PR #55, opened 2014-08-08
Improve cancellation support (#9).

davidmatson:
Here's how different cancellation events should trigger different actions:

| Event | CancellationToken for host.Call function | CancellationToken for auto-triggered function | Return from RunAndBlock |
| --- | --- | --- | --- |
| Cancellation of token passed to host.Call | Yes | No | No |
| host.Stop | No | Yes | Yes |
| Antares shutdown | Yes | Yes | Yes |
| host.Dispose | Yes | Yes | Yes |

I may ask for some assistance verifying these scenarios :)


advancedrei:
Can you guys please provide examples for how these cancellation options are supposed to work? Also, it would be nice in the WebJobsShutdownWatcher if you guys exposed events for when the shutdown was triggered, not just the cancellation token, so we can handle shutdown gracefully. For example, I need to flush my ApplicationInsights telemetry before the process shuts down. Thanks!

mathewc:
Can you explain your scenario a bit more and how you're using the shutdown watcher? You're just looking for an easy to use pattern for getting notified when the CancellationToken is actually being cancelled?

PR #500, opened 2015-02-12
Adding cancellation token parameter to RunAndBlock in JobHost

bwalks:
This allows cancellation token to be passed in through every public method that starts the JobHost.


mathewc:
Closing this, since it is very old, and the consensus seems to be no change. Brendan, if you still have passion on the matter, we can discuss further :) Thanks.

Issue #450, opened 2014-10-23
Add extensibility support


davidmatson:
Some dev notes on possible public surface area:
For runtime arguments:
====================
IValueProvider
IValueBinder
IWatchable
- IWatcher (return JObject rather than ParameterLog)

But not:
IOrderedValueBinder

For runtime attributes:
IBinding
- Return JObject rather than ParameterDescriptor
- ValueBindingContext
  - FunctionInstanceId
  - CancellationToken
  - Maybe some others (ConsoleOutput, FunctionCancellationToken, StorageAccount, etc.)
  - Probably keep constructor internal for now (not testable)
  - Avoid exposing properties that later would be DI constructor parameters instead
- BindingContext
  - Like ValueBindingContext, but add BindingData

For runtime trigger attributes:
ITriggerBinding
- Return JObject rather than ParameterDescriptor
- Refactor CreateFunctionDefinition to expose less? Maybe just IListenerFactory + trigger type?:
    IListenerFactory ListenerFactory { get; }
    Task<TriggerData> BindAsync(TTriggerValue value, ValueBindingContext context);
ITriggerData

Hopefully not:
ITriggeredFunctionBinding
ITriggeredFunctionInstanceFactory
FunctionDefinition
FunctionDescriptor
IInvoker

For runtime trigger listening:
IListener

For indexing time attributes:
IBindingProvider
- BindingProviderContext: NameResolver, ParameterInfo (abstract like MVC?), and CancellationToken, plus some kind of account access
- Expose BindingDataContract?
- Use DI instead for NameResolver?
- Probably keep constructor internal for now (not testable)

For indexing time trigger attributes:
ITriggerBindingProvider
- TriggerBindingProviderContext: Same as BindingProviderContext minus BindingDataContract

IListenerFactory
- ListenerFactoryContext - CancellationToken, account access, maybe SharedListeners?
- Refactor CreateAsync to expose less. Maybe just pass in an ITriggerExecutor that takes the trigger value + parent ID?
  - Move ParentId to ITriggerData: Guid? ParentId { get; }
    IListener CreateAsync(ITriggerExecutor<TTriggerValue> executor, ListenerFactoryContext context);
    public interface ITriggerExecutor<TTriggerValue>
    {
        bool TryExecute(TTriggerValue trigger);
    }

    
- Use DI to pass the trigger executor instead of an explicit parameter?

Hopefully not:
IFunctionInstance
IBindingSource
ExecutionReason
[...]

PR #508, opened 2015-05-06
Enabling 3rd Party Extensibility (Custom Bindings)

mathewc:
These changes introduce an extensibility model allowing 3rd party assemblies to integrate with the SDK and extend it with custom bindings, etc. To prove out the model, I've completely removed all of the previous internals visible allowances and specific assembly knowledge the Host assembly had for the ServiceBus extension, and refactored ServiceBus to use this extensibility model. This demonstrates that custom extensions can be written external to the core SDK, opening things up for the community to author their own.

This of course required making a bunch of previously internal interfaces public. I've reviewed this new public surface area carefully, and kept it as minimal as possible. I've also followed most of the recommendations in https://github.com/Azure/azure-webjobs-sdk/issues/450. As people write extensions, we'll likely find that we need to open more things up, and we'll do that as needed. For now the approach is to be conservative, and ensure that any new public surface area we add is as stable as possible.

This is the first stage of the extensibility work before the next SDK release. Next I'll be addressing specific asks, allowing more configurability of our built in bindings.

Notes:
======
* For DI injection, I've kept things simple and am just using a simple service container in JobHostConfiguration (IServiceProvider). This accomplishes our goals w/o locking into any specific DI solution (e.g. AutoFac)

* Previously the Dashboard required specific knowledge via type sharing between Host/Dashboard for formatting UI display strings (e.g. execution reason, parameter display details, etc.). Since this cannot be done for extensions, I've defined a small data contract for this (see ParameterDescriptor.DisplayHints). This of course is non-localized, but this is a trade off. We don't plan on localizing the Dashboard any time soon.

* For extensions, the ParameterDescriptor returned via IBinding.ToParameterDescriptor will of course be deserialized by the Dashboard as the base ParameterDescriptor type (since Dashboard doesn't have extension types). I've added ExtendedProperties to ParameterDescriptor to ensure that any custom extension properties are still deserialized by the Dashboard, and can be used for logging, etc.

* Since I've made some protocol additions, the plan is to do a updated Dashboard release BEFORE we release the updated SDK to the public. The changes I've made to the Dashboard are backwards compatible, but not forward compatible. For example, I've updated the ServiceBus extension to send its info in a different format to prove out the extensibility model, and that requires the updated Dashboard.

* The plan for the next SDK release is for there to be no breaking changes. I've Obsoleted one thing (JobHostConfiguration.ServiceBusConnectionString). It is true that to use ServiceBus now, you need to call the JobHostConfiguration.UseServiceBus method. That technically is a breaking change. We need to decide what to do there - allow this and tell people to add this single line of code, or perhaps for this release, put in a small shim that again loads the SB assembly dynamically and calls the UseServiceBus method (really don't want to regress back to that).


mathewc:
Adding a few reviewers/FYIs: @davidmatson @victorhurdugaci @alpaix @amitapl @rustd

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingContext.cs
```diff
@@ -10,7 +10,10 @@
 
 namespace Microsoft.Azure.WebJobs.Host.Bindings
 {
-    internal class BindingContext
+    /// <summary>
+    /// Context for parameter binding.
+    /// </summary>
+    public class BindingContext
```
mathewc: For all these BindingContext types that have been made public, before release I need to make a pass over all their public surface area and reduce what I can. This is an open TODO.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ServiceBusEndToEndTests.cs
```diff
@@ -125,9 +130,17 @@ private void ServiceBusEndToEndInternal()
                 TypeLocator = new FakeTypeLocator(this.GetType())
             };
 
-            _namespaceManager = NamespaceManager.CreateFromConnectionString(config.ServiceBusConnectionString);
+            ServiceBusConfiguration serviceBusConfig = new ServiceBusConfiguration();
+            config.UseServiceBus(serviceBusConfig);
```
mathewc: This line shows the new extensibility model in action. The user calls extension methods for the extensions they want to use.

src/Microsoft.Azure.WebJobs.Protocols/ParameterDisplayHints.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        /// <summary>
+        /// Gets or sets the default value.
+        /// </summary>
+        public virtual string DefaultValue { get; set; }
+
+        /// <summary>
+        /// Gets or sets the attribute text.
+        /// </summary>
+        public virtual string AttributeText { get; set; }
```
mathewc: Suggestions on whether this should be renamed? Currently it's only used by Dashboard LogAnalysis for outputting info on IBinder parameters. Perhaps this can be generalized/renamed?

src/Microsoft.Azure.WebJobs.Host/WebJobs.Host.csproj
```diff
@@ -367,7 +378,7 @@
     <Compile Include="Bindings\ValueBindingContext.cs" />
     <Compile Include="Bindings\IWatcher.cs" />
```
mathewc: I haven't made IWatcher functionality public yet (ServiceBus didn't use it). We can open that up as well when it is needed by other extensions. Likely there will be other instances of things we need to open up. We'll only discover these as we/others write extensions.

src/Microsoft.Azure.WebJobs.Host/Bindings/IOrderedValueBinder.cs
```diff
@@ -3,8 +3,14 @@
 
 namespace Microsoft.Azure.WebJobs.Host.Bindings
 {
-    internal interface IOrderedValueBinder : IValueBinder
+    /// <summary>
+    /// Defines an <see cref="IValueBinder"/> that provides an ordering hint./>
+    /// </summary>
+    public interface IOrderedValueBinder : IValueBinder
```
mathewc: We didn't want to make this public, but I don't see any other way. The ServiceBus bindings use it.

mathewc:
This PR has been merged into master.

mathewc:
I'm closing this, but I'll continue to address any feedback people have.

Issue #510, opened 2015-05-07
Support configurable retry policies


JacobSnyder:
We use the WebJobs SDK, specifically `QueueTriggers` in our service to process jobs and send messages to other services over HTTP.  

Currently, the behavior of the WebJobs SDK is that if an exception is thrown when processing a queue message, the visibility timeout window on the message will be updated such that the message is immediately visible in the queue:

```c#
private async Task ReleaseMessageAsync(CloudQueueMessage message, CancellationToken cancellationToken)
{
    try
    {
        // We couldn't process the message. Let someone else try.
        await _queue.UpdateMessageAsync(message, TimeSpan.Zero, MessageUpdateFields.Visibility, cancellationToken);
    }
    // snip...
```

What this means is the next time that the WebJobs SDK checks the queue, the message will be seen and reprocessed.

Due to the nature of sending requests to other services over HTTP, it's likely that an immediate retry might fail.  Consider the situation where the target receiving service is down.  In this case, immediately retrying likely won't work.  In this case, we would quickly go through the loop of "pull message from the queue > fail to process the message because the service is down > update the visibility of the message".  Eventually, we would hit the max dequeue count and the message would be moved to the Poison Queue.  This is a shame because if we had retried processing the message later (say in a few minutes or hours), the target receiving service might be up and processing the message would succeed.

It would be nice if the WebJobs SDK supported different backoff strategies out-of-the-box or allowed me to specify a function that would be called with the current dequeue count and then I could return the TimeSpan to set as the message's visibility timeout.  Either way, being able to support immediate retry, linear backoff, or exponential backoff would be nice.

Of course, we can "hack" this in by re-queueing the message in the queue and manually set the visibility timeout.  But this doesn't feel very "clean".  We would also have to use the storage APIs directly rather than use the WebJobs abstractions `ICollector<T>` or `IAsyncCollector<T>` because I don't think these allow us to set the visibility timeout.

mathewc:
Thanks for the detailed suggestion. Basically you're asking for a way to plug in your own IDelayStrategy (currently an internal interface) or some other interface, and have use allow that strategy to inspect/handle execution results and determine/return a delay for the next message batch read?

JacobSnyder:
[...]

Issue #505, opened 2015-04-27
Allow user configuration of CloudQueue/CloudQueueClient


mathewc:
Need to provide a way for users to configure the CloudQueue settings (and possibly CloudQueueClient settings) for all the queue objects we create internally. The customer scenario requiring this is that their message producer is using non default encoding settings, which requires the queue we create to use those same settings, otherwise messages won't deserialize properly (the customer is running into this).

This is related to the customer issue behind: https://github.com/Azure/azure-webjobs-sdk/issues/504.

PR #526, opened 2015-06-04
Queue Processing Extensibility

mathewc:
Adding adding a QueueProcessor strategy pattern to JobHostQueuesConfiguration, allowing users to override our default processing logic. This work addresses issues #510 and #505.

Features include:

* Ability to define a QueueProcessor per queue. This is done by providing your own IQueueProcessorFactory implementation
* Derive from default QueueProcessor and only override aspects you need to customize. E.g., override ReleaseMessage to implement your own message visibility backoff strategy for failed messages
* Ability to fully customize the CloudQueue and it's CloudQueueClient per queue


daniel-chambers:
I'd like to give a little feedback on this, if I may. :)

As someone who would want to inherit from `QueueProcessor` in order to do my own retry backoff policy, I'm looking at `CompleteProcessingMessageAsync` and I can see that it tells me the operation failed via the `succeeded` bool, but I can't see _why_ it failed (ie the exception).

In our particular scenario, the code that runs to process a queue message calls services that may either have transitive errors (in which case a fast retry with a backoff policy is desirable) or errors that indicate an unresolvable error (in which case putting the message in the poison queue immediately is desirable). Unfortunately, just a succeed/fail bool means I can't see what the cause of the problem was in order to react differently.

mathewc:
I agree with you Daniel. I had actually considered that, but decided to keep things simple - I wasn't sure it was needed. However, the fact that you immediately questioned the same thing made me realize it was important :)

I just pushed an update that flows a new FunctionResult type through. You can access any exception via functionResult.Exception.

mathewc:
Also, you can see an example of a custom QueueProcessorFactory/QueueProcessor in the end to end [functional tests here](https://github.com/Azure/azure-webjobs-sdk/blob/queues-extensibility/test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs#L221)

daniel-chambers:
Looks great! Thanks :)

You guys thinking of releasing any prerelease Nuget packages so we can play with the bits ahead of an official stable release?

mathewc:
Yes - actually I've already released these changes to the new App Service myget feed that I created recently :) https://www.myget.org/gallery/azure-appservice. This feed also contains the new WebJobs.Extensions package containing new out of band binding extensions that leverage all the new extensibility.

daniel-chambers:
Awesome. Cheers. :)

Issue #517, opened 2015-05-18
I'd like a table argument binding that supports a diversity of efficient operations


marklio:
Table bindings are really great, but the options available are either read or write, and mixing them (with multiple arguments) leads to observation of inconsistent state (because writes don't necessarily flush partitions).  I'd love to see a binding with the following abilities:
* Add (and I'd love Insert/Replace/Merge enforcement) (It would also be nice to have "bulk background adds" as well, with an associated WaitForBackgroundAddsAsync() or something)
* Flush (to force the flushing of batch operations so queries can see consistent state)
* AsQuery (that would return something IQueryable)
* Efficient single lookup by keys (Retrieve)
* Delete by keys or set of entities
(And async versions of all :) )

This, of course, can be accomplished with CloudTable directly, but requires far more code than the other bindings, and feels wrong in the world of WebJobs SDK.

Failing this, extensibility to add my own such binding would rock.

mathewc:
Thanks for the details. This is one of the target scenarios for the new extensibility model we'll be releasing soon. Rather than release a new table binding in the box, I'll consider adding this to the new [WebJobs.Extensions](https://github.com/Azure/azure-webjobs-sdk-extensions) repo/package. It is another great extensibility example for people to follow (customizing existing bindings) in addition to the authoring of entirely new bindings/triggers, which is what Extensions showcases thus far.

mathewc:
@marklio @rustd Guys, I've just pushed my proposed changes supporting Table binding extensions. Any feedback appreciated. https://github.com/Azure/azure-webjobs-sdk/pull/527

mathewc:
I've checked this extensibility in to master. @marklio So when you gonna write your fancy Table extension binding leveraging this? :) If you're really planning on writing one, it would be great to get it into the [WebJobs Extensions](https://github.com/azure/azure-webjobs-sdk-extensions) repo. What do you think? I'll update the latest alpha SDK version [on the myget feed](https://www.myget.org/gallery/azure-appservice) with these changes so they're available.

marklio:
I'll play with it this week. I'm on vacation next week, so it's likely to spill.

Sent from my Windows Phone
________________________________

rustd:
cool.

PR #527, opened 2015-06-09
Adding extensibility for Table bindings 

mathewc:
These changes allow users (or extension authors) to create their own custom Table bindings (e.g. a Table to IDictionary binding). This addresses the scenario discussed in #517. The way it works is, you author a ITableArgumentBindingExtensionProvider and register it with the JobHostConfiguration. We'll then pick up any extension bindings when binding functions.

```
JobHostConfiguration config = new JobHostConfiguration();
IExtensionRegistry extensions = config.GetService<IExtensionRegistry>();
extensions.RegisterExtension<ITableArgumentBindingExtensionProvider>(new MyTableArgumentBindingProvider());
```

### DIFF
--- a/test/Microsoft.Azure.WebJobs.Host.FunctionalTests/TableTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.FunctionalTests/TableTests.cs
@@ -3,12 +3,17 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
+using System.Reflection;
+using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Bindings;
 using Microsoft.Azure.WebJobs.Host.FunctionalTests.TestDoubles;
 using Microsoft.Azure.WebJobs.Host.Storage;
 using Microsoft.Azure.WebJobs.Host.Storage.Queue;
 using Microsoft.Azure.WebJobs.Host.Storage.Table;
+using Microsoft.Azure.WebJobs.Host.Tables;
 using Microsoft.WindowsAzure.Storage.Queue;
 using Microsoft.WindowsAzure.Storage.Table;
 using Newtonsoft.Json;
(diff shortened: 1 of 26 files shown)


test/Microsoft.Azure.WebJobs.Host.FunctionalTests/TableTests.cs
```diff
@@ -354,6 +379,17 @@ private class BindToIQueryableDynamicTableEntityProgram
             }
         }
 
+        private class CustomTableBindingExtensionProgram
+        {
+            public static void Run([QueueTrigger(TriggerQueueName)] CloudQueueMessage ignore,
+                [Table(TableName)] CustomTableBinding<Poco> table)
```
mathewc: Here's an example of the custom binding being used

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/TableTests.cs
```diff
@@ -403,5 +439,130 @@ private class PocoWithAllTypes
 
             public Poco PocoProperty { get; set; }
         }
+
+        /// <summary>
+        /// Binding type demonstrating how custom binding extensions can be used to bind to
+        /// arbitrary types
+        /// </summary>
+        /// <typeparam name="TEntity"></typeparam>
+        public class CustomTableBinding<TEntity>
```
mathewc: The following three types show what is involved in writing a binding extension. This test example doesn't actually perform any storage operations, it just demonstrates how you would.

mathewc:
One more thing to point out - I've only enabled extensions for Table bindings that bind to the Table, not bindings that bind to a single Entity/Poco. I don't think there are a lot of scenarios for that. We can open that up later if needed.

Likewise, I've only done this for Table bindings so far. TBD on whether we should do the same for the other bindings now (Blob, Queues), or wait until we get specific asks.

marklio:
This looks reasonable. However, is there a reason you wouldn't just enable the same extensibility mechanism across the board? I can think of several things I'd like to do with blobs, and even single entities.

mathewc:
Yes, I'm on the fence about it :) To do so will involve exposing new extensibility interfaces for each binding type. I was thinking of being conservative and only opening things up when we get specific asks or scenarios. Not sure yet...

What would you do with a single Table entity extension binding?

marklio:
It would be nice to have a wrapper that allowed me to control when changes are committed so I could handle optimistic concurrency issues.

Sent from my Windows Phone
________________________________

mathewc:
For that scenario, you likely don't want to go through all the work to implement and register your own binding for this. Probably what we should do for cases like this is allow a lighter weight way for you to specify OC or other strategies, and have the built in bindings look for and apply them.

Issue #275, opened 2014-09-11
No parallel execution for service bus triggers


victorhurdugaci:

HansOlavS:
Wow, this has been in the backlog since Sep 2014.

HansOlavS:
I've made a PR with a proposed fix for this issue at #533.

mathewc:
Related SO post here: http://stackoverflow.com/questions/31731733/intercept-azure-webjobs-queue-service-bus-triggers

This user needs to intercept all SB messages and run custom code BEFORE his job function gets called.

BrianVallelunga:
I'm the one who wrote the SO post. Specifically, I have a Tenant custom property set on my messages and want to be able to set Tenant-specific context data before the job is run. Ideally, I'd have access to this custom data from within my custom IJobActivator so I could inject tenant-specific services into my jobs. The job itself would then never need to know which tenant it was processing. That would all be moved down to the infrastructure level.

mathewc:
Guys, I've created an initial PR for this here: https://github.com/Azure/azure-webjobs-sdk/pull/558. @HansOlavS @BrianVallelunga please take a look and verify this will meet your needs. The E2E tests include an example of how custom processors can be specified [see this test](https://github.com/Azure/azure-webjobs-sdk/blob/sbconfig/test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ServiceBusEndToEndTests.cs#L99).

@HansOlavS This PR incorporates your PR #533 :) As I mentioned previously, we needed to also enable deeper customization, so I'm doing both here.

HansOlavS:
Looks great!

But I do feel that going concurrent is a pretty big breaking change. Isn't it better to opt-in and explicitly set the concurrency level yourself? Is there a default concurrency level on Azure Queues as well?

I feel the two should have the same default setup as far as possible...

mathewc:
Yes, I'm thinking you're right about the concurrency. I'll probably leave it at the existing default - at least we've opened it up now for people.

HansOlavS:
Another thing I've been thinking of: Wouldn't it be cool to adjust the concurrency-level per queue/method?
```c#
[ServiceBusTrigger("", OnMessageOptions = new OnMessageOptions()]
```

or just
```c#
[ServiceBusTrigger("", MaxConcurrentCalls = 5)]
```

I'd love to have those...

mathewc:
[...]

PR #533, opened 2015-06-21
Added OnMessageOptions to ServiceBusConfiguration.

HansOlavS:
I've added this to ServiceBusConfiguration and implemented using it throughout.
```c#
    public class ServiceBusConfiguration
    {
        ...
        public OnMessageOptions OnMessageOptions
        {
            get { return _onMessageOptions; }
            set { _onMessageOptions = value; }
        }
    }
```

There's a "tight coupling" between ServiceBusConfiguration and ServiceBus by using OnMessageOptions directly, but by doing this we'll support any changes that might later be introduced in the base ServiceBus library by using this instead of separately adding the properties of OnMessageOptions (like MaxConcurrentCalls, AutoComplete and so on).

As I'm just passing the OnMessageOptions through to the internally used MessageReceiver.OnMessageAsync() method, I haven't made any new tests to explicitly test setting different properties of OnMessageOptions.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.ServiceBus/Config/ServiceBusConfiguration.cs
+++ b/src/Microsoft.Azure.WebJobs.ServiceBus/Config/ServiceBusConfiguration.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using Microsoft.Azure.WebJobs.Host.Executors;
+using Microsoft.ServiceBus.Messaging;
 
 namespace Microsoft.Azure.WebJobs.ServiceBus
 {
@@ -12,6 +13,7 @@ namespace Microsoft.Azure.WebJobs.ServiceBus
     {
         private bool _connectionStringSet;
         private string _connectionString;
+        private OnMessageOptions _onMessageOptions = new OnMessageOptions();
 
         /// <summary>
         /// Gets or sets the Azure ServiceBus connection string.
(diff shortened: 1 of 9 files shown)


HansOlavS:
This PR is a proposed fix for issue #275.

mathewc:
I'm thinking for this we're going to want to follow the model I did recently when I opened up Azure Queue configurability (https://github.com/Azure/azure-webjobs-sdk/pull/526). That way, things can be customized more deeply, and can also vary *per queue*, rather than an option that applies globally to all queues. That's quite a bit more involved though, but does set us up better for the future I think. For example, I think I'd have a QueueProcessor that can customize more aspects of the processing pipeline for SB messages.

HansOlavS:
Rolling your own QueueProcessor is nice, but I also think that the out-of-the-box Trigger-attribute abstraction is really powerful and cool.

What about introducing a MaxConcurrentCalls on the QueueTrigger/ServiceBusTrigger itself? That way you get granular control per queue and still keep the high abstraction level of Trigger attributes.

```c#
[ServiceBusTrigger("my-queue", MaxConcurrentCalls = 16)]
```

or

```c#
[QueueTrigger("my-queue", MaxConcurrentCalls = 16)]
```

What do you think? Still no cigar?

mathewc:
The QueueProcessor strategy pattern doesn't change the attribute trigger model :) Please see the functional tests for an example of how QueueProcessor is used.

My main point is, before we add this, I need to decide on the approach we're going to take for SB queue extensibility, to ensure this plays well with that. It might be that what you had originally will be OK - that's the same model as we have now in JobHostQueuesConfiguration with the new QueueProcessor. There are some top level switches that you can set that affect ALL queues, and custom queue processors can choose to override those.

Here's what I'll do - I'll pull this issue into the current v1.1.0 release, and we'll discuss it in our next iteration meeting to decide which way we want to go. @rustd 

Thanks - I'll keep you posted :)

HansOlavS:
@mathewc, sounds great!

This might be a bit off-topic, but have you ever considered a trigger for batch-processing multiple messages? I find myself wanting this for performance in some cases (like sql server bulk-insert)

```c#
void Process([QueueBatchTrigger("my-queue")] string[] messages)
{
  // batch-process array of messages
}
```

mathewc:
FYI - I've created PR #558 which incorporates this change and goes a bit further. I'll close this PR and we can continue the discussion there. Thanks :)

Issue #546, opened 2015-07-20
Dashboard indexer fails when message contains line breaks


Benjimawoo:
If the output from a function invocation contains '\r\n' line breaks (the ones I've seen have all been in the message), the indexer fails to index that invocation, and retries repeatedly.

Overwriting the original output file with a duplicate with those line breaks removed resolves the issue.

Example input here: http://254shades.blob.core.windows.net/public/example.txt

Stack trace in the failed index logs:
```
Microsoft.WindowsAzure.Storage.StorageException: Specified value has invalid CRLF characters.
Parameter name: value ---> System.ArgumentException: Specified value has invalid CRLF characters.
Parameter name: value
   at System.Net.WebHeaderCollection.CheckBadChars(String name, Boolean isHeaderValue)
   at System.Net.WebHeaderCollection.Add(String name, String value)
   at Microsoft.WindowsAzure.Storage.Shared.Protocol.HttpWebRequestFactory.AddMetadata(HttpWebRequest request, IDictionary`2 metadata)
   at Microsoft.WindowsAzure.Storage.Blob.CloudBlockBlob.<PutBlobImpl>b__37(HttpWebRequest r, OperationContext ctx)
   at Microsoft.WindowsAzure.Storage.Core.Executor.Executor.ProcessStartOfRequest[T](ExecutionState`1 executionState, String startLogMessage)
   at Microsoft.WindowsAzure.Storage.Core.Executor.Executor.ExecuteSync[T](RESTCommand`1 cmd, IRetryPolicy policy, OperationContext operationContext)
   --- End of inner exception stack trace ---
   at Microsoft.WindowsAzure.Storage.Core.Executor.Executor.ExecuteSync[T](RESTCommand`1 cmd, IRetryPolicy policy, OperationContext operationContext)
   at Microsoft.WindowsAzure.Storage.Blob.CloudBlockBlob.UploadFromStreamHelper(Stream source, Nullable`1 length, AccessCondition accessCondition, BlobRequestOptions options, OperationContext operationContext)
   at Microsoft.WindowsAzure.Storage.Blob.CloudBlockBlob.UploadFromByteArray(Byte[] buffer, Int32 index, Int32 count, AccessCondition accessCondition, BlobRequestOptions options, OperationContext operationContext)
[...]

PR #547, opened 2015-07-22
Issue #546 - Dashboard indexer fails when message contains line breaks

Benjimawoo:


src/Dashboard/Data/FunctionInstanceSnapshot.cs
```diff
@@ -85,7 +85,11 @@ private string BuildFunctionDisplayTitle()
         {
             var name = new StringBuilder(FunctionShortName);
             IEnumerable<string> argumentValues = Arguments.Values.Select(v => v.Value);
-            string parametersDisplayText = String.Join(", ", argumentValues);
+
+
+            // Remove newlines to avoid 403/forbidden storage exceptions when saving display title to blob metadata
+            // for function indexes. Newlines may be present in JSON-formatted arguments.
+            string parametersDisplayText = String.Join(", ", argumentValues).Replace("\r\n", String.Empty);
```
mathewc: Do you think it would be better to replace newlines with a space rather than Empty? At least that preserves some separation.

test/Dashboard.UnitTests/Indexers/FunctionIndexerTests.cs
```diff
@@ -78,5 +78,24 @@ public void CreateSnapshot_CreatesExpectedSnapshot()
...
         }
+
+        [Fact]
+        public void FunctionInstanceSnapshot_CompletelyRemovesLineBreaks()
+        {
+            FunctionInstanceSnapshot message = new FunctionInstanceSnapshot
+            {
+                Arguments = new Dictionary<string, FunctionInstanceArgument> 
+                {
+                    //{"message", new FunctionInstanceArgument(){ Value = "{\r\n  \"QRPoint\": {\r\n    \"X\": 0,\r\n    \"Y\": 0\r\n  },\r\n  \"TimesheetId\": 0,\r\n  \"HasReadableQrCode\": false,\r\n  \"HasSignature\": false,\r\n  \"BlobAddress\": \"https://myaccount.blob.core.windows.net/container/file.png\",\r\n  \"$AzureWebJobsParentId\": \"511f605d-14bf-46ca-b321-96b59d9e81d6\"\r\n}"} }
```
mathewc: Please remove commented out code. Or, even better, make this test a [Theory] test, and include 4 or 5 [InlineData(...)] attributes to tests some permutations, including thsi commented out one (which looks like your original repro?)

test/Dashboard.UnitTests/Indexers/FunctionIndexerTests.cs
```diff
@@ -78,5 +78,24 @@ public void CreateSnapshot_CreatesExpectedSnapshot()
...
+                    {"message", new FunctionInstanceArgument(){ Value = "XXXXXX\r\nXXXXXXXXX\r\nYYYYYYY"} }
```
mathewc: If this special string was constructed with a specific length to trigger the bug, please add a comment to that effect

src/Dashboard/Data/FunctionInstanceSnapshot.cs
Benjimawoo: It would - I just left as Empty to keep existing behaviour

mathewc: Ok, leave as is then - existing behavior is good

PR #558, opened 2015-08-06
Support for ServiceBus parallel execution + custom MessageProcessors

mathewc:
This PR addresses the issues discussed in #275. Specifically, it includes the following:

- new ServiceBusConfiguration.OnMessageOptions global default (for simplicity)
- Default OnMessageOptions now has concurrency set to 5 (was left at default 1 previously)
- added MessageProcessor which allows custom message processors to be used per queue/topic
- MessageProcessors are created by ServiceBusConfiguration.MessageProcessorFactory - the default   factory can be replaced to take control of creation, return custom processors, etc.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.ServiceBus/Config/ServiceBusConfiguration.cs
+++ b/src/Microsoft.Azure.WebJobs.ServiceBus/Config/ServiceBusConfiguration.cs
@@ -2,6 +2,8 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using Microsoft.Azure.WebJobs.Host.Executors;
+using Microsoft.Azure.WebJobs.ServiceBus.Listeners;
+using Microsoft.ServiceBus.Messaging;
 
 namespace Microsoft.Azure.WebJobs.ServiceBus
 {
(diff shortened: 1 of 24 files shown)


mathewc:
Still need to add unit tests - only have an E2E test in place currently.

src/Microsoft.Azure.WebJobs.ServiceBus/Config/ServiceBusConfiguration.cs
```diff
@@ -14,6 +16,18 @@ public class ServiceBusConfiguration
         private string _connectionString;
 
         /// <summary>
+        /// Constructs a new instance.
+        /// </summary>
+        public ServiceBusConfiguration()
+        {
+            MessageOptions = new OnMessageOptions
+            {
+                MaxConcurrentCalls = 5
```
mathewc: Any comments on what this concurrency default should be?

rustd: Use what Azure Queues have?

mathewc: There isn't a corresponding setting for Queues. Concurrency is governed by Batch size and NewBatchThreshold.

There is a slight concern here that setting this to something other than 1 might regress people if they're relying on no concurrency. However, I think that is unlikely to be an issue and is outweighed by the benefit of giving people concurrency by default (which is what they would expect).

So we're looking for the right default number here.

PR #576, opened 2015-10-06
allows binding values to come from app settings

brendankowitz:


src/Microsoft.Azure.WebJobs.Host/Triggers/TriggeredFunctionBinding.cs
```diff
@@ -56,7 +58,7 @@ internal class TriggeredFunctionBinding<TTriggerValue> : ITriggeredFunctionBindi
             {
                 ITriggerData triggerData = await _triggerBinding.BindAsync(value, context);
                 triggerProvider = triggerData.ValueProvider;
-                bindingData = triggerData.BindingData;
+                bindingData = triggerData.BindingData.Concat(AppSettingsBinding.CreateBindingData()).ToDictionary(x => x.Key, x => x.Value);
```
mathewc: I think this would be a breaking change for existing applications. E.g. if they moved to the new release and had app settings that happened to coincide with binding data their app is already using (e.g. their own POCO types, etc.).

I'm also wondering if this is a feature that is really useful generally, or if we shouldn't constrain it to just Singleton right now. Singleton is a new feature, so we can't break anyone.

brendankowitz: Good point..I think in that case the AppSettings should supplement not override values that may already exist. Thoughts?

mathewc: Hey, I just realized that our existing JobHostConfiguration.INameResolver mechanism would satisfy your requirements. Basically, you provide your own INameResolver impl. that maps names to values (e.g. you can read from app settings, or wherever you want). Then in SingletonAttribute you can use an expression like [Singleton("%mysetting%")]. During the binding process, all our exising attributes that support binding expressions will use INameResolver to resolve any %% delimited values.

Issue was that Singleton didn't support this, but I just fixed that here: https://github.com/Azure/azure-webjobs-sdk/commit/1b9c21e593220b25945a371ab17733afb6f60e5f

:)

mathewc:
Hey, I just realized that our existing JobHostConfiguration.INameResolver mechanism would satisfy your requirements. Basically, you provide your own INameResolver impl. that maps names to values (e.g. you can read from app settings, or wherever you want). Then in SingletonAttribute you can use an expression like [Singleton("%mysetting%")]. During the binding process, all our exising attributes that support binding expressions will use INameResolver to resolve any %% delimited values.

Issue was that Singleton didn't support this, but I just fixed that here: 1b9c21e :) See my example TestNameResolver to see how to implement.

brendankowitz:
Great, thanks @mathewc

PR #606, opened 2015-12-09
Perf work for Dashboard logging and ServiceBus (PrefetchCount)

mathewc:
Adding a new **TraceLevelAttribute** that can be applied at the class/method level to control Console/Dashboard logging for a particular function. The idea is to allow very high scale functions to opt out of Dashboard logging (since this logging introduces a bit of overhead) to increase throughput. Notes on this feature:

* If the TraceLevel is set to Off for a function, no Console or Dashboard logs will be written for function invocations
* If the TraceLevel is set to Error for a function, normal invocations will not produce logs, but if a function errors it will produce logs (both Console and Dashboard)
* When functions are invoked manually via Dashboard Run/Replay, or via JobHost.Call, any TraceLevelAttribute annotation will be ignored, and normal full logs will be produced. This facilitates development/debugging, while retaining the benefits of actual steady state production behavior.

Adding a **ServiceBusConfiguration.PrefetchCount** setting to allow users to easily configure this for performance. The value will be applied to all MessageReceivers the runtime creates (`MessageReceiver.PrefetchCount`). Also adding a new virtual method **MessagingProvider.CreateMessageReceiver** to allow users to further customize MessageReceivers as needed. Users can override this to to return a receiver they have configured.

Using these two new features (setting PrefetchCount to 100), we are able to achieve throughput parity for a high scale ServiceBus job function (below). In my testing, this function could process the same number of messages per second as a non-WebJobs ServiceBus MessageReceiver configured with the same settings. In my testing, I was pushing this to ~800 messages per second on a single instance.

```csharp
[TraceLevel(TraceLevel.Error)]
public static async Task ProcessMessageAsync(
    [ServiceBusTrigger("perftest")] BrokeredMessage message)
{
    Program.MessageProcessed();
    await Task.Delay(10);
}
```

### DIFF
--- a/CustomDictionary.xml
+++ b/CustomDictionary.xml
@@ -69,7 +69,8 @@
       <Word>MessagingProvider</Word>
       <Word>CloudBlobContainer</Word>
       <Word>SingletonAttribute</Word>
-	  </Recognized>
+      <Word>Prefetch</Word>
+	</Recognized>
     <Deprecated/>
     <Compound>
       <Term CompoundAlternate="CancellationToken">cancellationToken</Term>
--- a/Settings.StyleCop
+++ b/Settings.StyleCop
@@ -22,6 +22,7 @@
       <Value>eTag</Value>
       <Value>bindable</Value>
 	  <Value>dequeue</Value>
+	  <Value>Prefetch</Value>
     </CollectionProperty>
   </GlobalSettings>
   <Parsers>
--- a/src/Dashboard/Properties/AssemblyInfo.cs
+++ b/src/Dashboard/Properties/AssemblyInfo.cs
@@ -4,7 +4,7 @@
 using System.Reflection;
 using System.Runtime.CompilerServices;
 
-[assembly: AssemblyVersion("1.2.0.0")]
-[assembly: AssemblyFileVersion("1.2.0.0")]
+[assembly: AssemblyVersion("1.1.1.0")]
+[assembly: AssemblyFileVersion("1.1.1.0")]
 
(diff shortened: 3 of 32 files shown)


softeering:
Hi Mathew, 

As discussed, I just tested this and it's far better. I managed to get a higher throughput (even better than the base ServiceBus implementation) when using a simple message processing callback.

Do you any ETA for the 1.1.1 stable release?

Thanks for your help

mathewc:
Maybe within a month. It's the Holidays soon, so probably looking at around the second week of January.

iremmats:
WOW! We are using WebJobs to put logs from Azure Service Bus into an Elasticsearch cluster. The bottleneck has always been the WebJobs throughput here. Im getting 500-1000 messages per second into the cluster from 1 small instance in Azure.
Using PrefetchCount = 100 and ConcurrentCalls = 5. What is the recommendation here? These numbers means I do 5 calls at the same time for 100 messages each right?

softeering:
Hi Mats,

PrefetchCount determines the size of the batch of messages grabbed from the queue in one call. Everytime the client gets messages from the bus, it will try to retrieve 100 messages in the same batch instead of getting 1 message only during each call (taking care about the max size of a batch and all that stuff)
MaxConcurrentCalls is just about how many "threads" are used to execute the message processing callback (i.e. the triggered function)
I don't have strong recommendations here: try modifying the MacConcurrentCalls depending on the time a message needs to be processed and the PrefetchCount depending on the size of the messages you're pulling from the bus
For example: the size of my messages is quite small and the I'm able to process ~ 5 messages per second in one callback - PrefetchCount is set to 100 and MaxConcurrentCalls is set to 50

Hope it helps

iremmats:
Thanks for input. I was mostly asking if there were any configuration you should really avoid. Ill do some testing on the parameters, still we are very happy with 500 message per second per small instance in Azure.

mathewc:
Thanks guys. What do you guys thing about the new **MessagingProvider.CreateMessageReceiver** overload I added to allow people to override the creation of the MessageReceiver used by the listener? I did that as an escape hatch in case any other additional advanced options might need to be configured for perf, like PrefetchCount. I promoted PrefetchCount to top level ServiceBusConfiguration because like OnMessageOptions it is likely a common thing to configure. Are there any other things that SB users might want to configure that I'm missing, or do you think these settings and methods allow most of the scenarios?

softeering:
Hi Mathew,

On my side, the override is enough. It allowed me to apply the same settings as what I was using with the base ServiceBus SDK. Maybe other guys will have more advanced use cases

Thanks

iremmats:
Hi,

Override works for me too. Its good to not confuse people with too many settings initially. :)

PR #630, opened 2016-02-12
Very very rough draft of EventHub extension.

MikeStall:
Missing xml comments, tests, and a ton of $$$ todo markers. Triggering binding is still rough.
but gives a basic view of the supported bindings and direction things are going in. 
This is a chance to comment on the high-level direction. 

This is exclusively in the ServiceBus extension (no core changes).
Added a \common directory for noneventhub specific support bindings. (The ServiceBus side should be able to use these too; and eventually we can push some of this into core)
\EventHub  directory is for eventhub specific.

Need to sort out the dashboard piece too. I tried to make this an extension-only change, but ServiceBus has support in the core for dashboard. So naively, to get service-bus caliber dashboard support, I'd need to make similar changes in core, which I'm hoping to be able to avoid.


src/Microsoft.Azure.WebJobs.ServiceBus/Common/CommonAsyncCollectorValueProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+using Microsoft.Azure.WebJobs.Host.Bindings;
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.ServiceBus
+{
+    // TUser is the parameter type from the user function we're binding to. 
+    // TMessage is from the underying IAsyncCollector<TMessage>
+    internal class CommonAsyncCollectorValueProvider<TUser, TMessage> : IOrderedValueBinder
```
MikeStall: test comment

MikeStall: test reply

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        public void AddSender(string eventHubName, string sendConnectionString)
+        {
+            var client = EventHubClient.CreateFromConnectionString(sendConnectionString, eventHubName);
+            AddSender(client);
+        }
+
+        
+        private void AddReceiver2(string eventHubName, EventProcessorHost listener)
```
mathewc: Nit: AddReceiver2? Shouldn't this just be an overload?

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
```diff
@@ -0,0 +1,129 @@
...
+        {
+            _listeners[eventHubName] = listener;
+        }
+
+        public void AddReceiver(string eventHubName, string receiverConnectionString)
+        {
+            // We can't get the storage string until we get a JobHostConfig. So defer this. 
+            _deferredWork.Add((jobHostConfig) =>
+           {
+               string storageConnectionString = jobHostConfig.DashboardConnectionString;
```
mathewc: Why does the EH receiver need the Dashboard connection string? Also you're assigning the dashboard connection to a storageConnection variable - they can be two different accounts.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
```diff
@@ -0,0 +1,129 @@
...
+using System.Threading.Tasks;
+using System.Threading;
+using Microsoft.ServiceBus.Messaging;
+using Microsoft.Azure.WebJobs.Host.Protocols;
+using System.Globalization;
+using Microsoft.Azure.WebJobs.Host.Triggers;
+
+namespace Microsoft.Azure.WebJobs.ServiceBus
+{
+    public class EventHubConfiguration : IExtensionConfigProvider, IEventHubProvider
```
mathewc: Is IEventHubProvider your interface, or EventHub's?

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
```diff
@@ -0,0 +1,129 @@
...
+
+            // Deferred list 
+            foreach (var action in _deferredWork)
+            {
+                action(context.Config);
+            }
+            _deferredWork.Clear();
+
+            // get the services we need to construct our binding providers
+            INameResolver nameResolver = context.Config.GetService<INameResolver>();
```
mathewc: fyi: context.Config.NameResolver

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubListener.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        public EventHubListener(ITriggeredFunctionExecutor executor, EventProcessorHost eventListener, EventProcessorOptions options, bool single)
+        {
+            this._executor = executor;
+            this._eventListener = eventListener;
+            this._singleDispatch = single;
+            this._options = options;
+        }
+
+        public void Cancel()
```
mathewc: Add $$$ so you don't forget to implement these :)

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubListener.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        async Task IEventProcessor.ProcessEventsAsync(PartitionContext context, IEnumerable<EventData> messages)
+        {
+            EventHubTriggerInput value = new EventHubTriggerInput
+            {
+                _events = messages.ToArray(),
+                _context = context
+            };
+
+            // Single dispatch
```
mathewc: What does this single dispatch mean for Dashboard? Is each function invocation then its own entry in the Dashboard?

MikeStall: Yes.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
MikeStall: IEventHubProvider is a private interface to the let the binders access the configuration.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubListener.cs
```diff
@@ -0,0 +1,122 @@
...
+                TriggeredFunctionData input = new TriggeredFunctionData
+                {
+                    ParentId = null,
+                    TriggerValue = value
+                };
+
+                FunctionResult result = await _executor.TryExecuteAsync(input, CancellationToken.None);
+            }
+
+            await context.CheckpointAsync();
```
mathewc: Probably should only checkpoint if the function succeeded? TryExecuteAsync won't throw if the function fails, it just returns a failed FunctionResult

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
MikeStall: EventHubs listener needs a storage account, which is uses to coordinate between multiple listener instances (creates a few leaseses, etc)
If you don't provide a storage account, we use the dashboard account to try and be nice.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/IEventHubProvider.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.ServiceBus.Messaging;
+
+namespace Microsoft.Azure.WebJobs.ServiceBus
+{
+    // Expose to binders so they can get the EventHub connections. 
+    interface IEventHubProvider
```
mathewc: If this is an interface users implement, you should consider making this an abstract class instead so we can add to it w/o that being a breaking change. That's the approach I've taken with all the other binding config extensibility points.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/IEventHubProvider.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        EventHubClient GetSender(string eventHubName);
```
mathewc: Nit: recommend having these method names match the things they return, rather than introducing our own taxonimy

src/Microsoft.Azure.WebJobs.ServiceBus/WebJobs.ServiceBus.csproj
```diff
@@ -28,6 +28,8 @@
     <WarningLevel>4</WarningLevel>
     <DocumentationFile>bin\Debug\Microsoft.Azure.WebJobs.ServiceBus.xml</DocumentationFile>
     <Prefer32Bit>false</Prefer32Bit>
+    <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
+    <RunCodeAnalysis>false</RunCodeAnalysis>
```
mathewc: Be sure to reenable this :) It'll make you change a lot of stuff I didn't comment on above :)

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
mathewc: Why the Dashboard account? Why not the AzureWebJobsStorage account which is required (Dashboard account is optional)

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/IEventHubProvider.cs
MikeStall: good point.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
MikeStall: I'll switch

PR #633, opened 2016-02-19
Use Microsoft.Build.Tasks.Core.dll instead of Microsoft.Build.Tasks.v4.0.dll

davidebbo:
This makes the build work out of the box on VS 2015.
http://stackoverflow.com/questions/31722877/cant-build-in-visual-studio-2015-because-microsoft-build-tasks-v14-0-dll-cann has more info.


mathewc:
I've already fixed this in master (https://github.com/Azure/azure-webjobs-sdk/commit/15cc73f887b09e32f34d48804aabdd396749dbe7)

PR #635, opened 2016-02-20
Add cases to ConverterManager to support additional bindings:

MikeStall:
1. automatic inheritence:
2. object --> TDest,   takes precedence

Remove parameter from GenericBinder


MikeStall:
changes where merged in from another request.

PR #637, opened 2016-02-26
introducing new hybrid strategy on logging blob changes based on the contianer size

shobak101:


src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -0,0 +1,162 @@
...
+            if (!_disposed)
+            {
+                _pollLogStrategy.Dispose();
+                _scanContainerStrategy.Dispose();
+                _cancellationTokenSource.Dispose();
+                _disposed = true;
+            }
+        }
+
+        private void ScnaContainersAndRegisterWithStrategy(CancellationToken cancellationToken)
```
davidebbo: Typo: 'Scna' :)

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/SharedBlobListener.cs
```diff
@@ -91,14 +91,8 @@ public void Dispose()
...
         {
-            if (!StorageClient.IsDevelopmentStorageAccount(account))
-            {
-                return new PollLogsStrategy();
-            }
-            else
-            {
-                return new ScanContainersStrategy();
-            }
+            account.ToString();
```
shobak101: that should be removed, the parameter should be removed all together, I hacked it quickly during testing forgot to remove it

davidebbo: Strange statement. Remove?

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/SharedBlobListener.cs
```diff
@@ -91,14 +91,8 @@ public void Dispose()
...
         private static IBlobListenerStrategy CreateStrategy(IStorageAccount account)
         {
-            if (!StorageClient.IsDevelopmentStorageAccount(account))
-            {
-                return new PollLogsStrategy();
-            }
-            else
-            {
-                return new ScanContainersStrategy();
-            }
```
davidebbo: Are there any code paths left that use either of these two existing strategies?

shobak101: only tests are left referencing it, I should add a unit test for the new strategy too, but no other references

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -0,0 +1,162 @@
...
+                {
+                    return;
+                }
+
+                IStorageBlobResultSegment results;
+
+                try
+                {
+                    // Non-async is correct here. ScanContainers occurs on a background thread. Unless it blocks, no one
+                    // else is around to observe the results.
```
davidebbo: No sure I understand why the background thread can't use async. Problem with blocking is that it increases the thread count, and thread usage can be an issue if there are too many.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -0,0 +1,162 @@
...
+                containerRegistrations = _registrations[container];
+            }
+            else
+            {
+                containerRegistrations = new List<ITriggerExecutor<IStorageBlob>>();
+                _registrations.Add(container, containerRegistrations);
+            }
+
+            containerRegistrations.Add(triggerExecutor);
+            return Task.FromResult(0);
```
davidebbo: What's does 0 mean?

MikeStall: it's just a convention for returning a completed task since this is not an 'async' method. Ideally, BCL would have had "Task.Completed" 

---
In reply to: [54306785](https://github.com/Azure/azure-webjobs-sdk/pull/637#discussion_r54306785) [](ancestors = 54306785)

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -0,0 +1,162 @@
...
+                new StorageBlobContainerComparer());
+            _scanContainerStrategy = new ScanContainersStrategy();
+            _pollLogStrategy = new PollLogsStrategy();
+            _cancellationTokenSource = new CancellationTokenSource();
+        }
+
+        public void Start()
+        {
+            ThrowIfDisposed();
+            // can not start the 2 strategies till all regiesterations are done
```
MikeStall: typo

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -0,0 +1,162 @@
...
+        {
+            ThrowIfDisposed();
+            // Register and Execute are not concurrency-safe.
+            // Avoiding calling Register while Execute is running is the caller's responsibility.
+
+            ICollection<ITriggerExecutor<IStorageBlob>> containerRegistrations;
+            if (_registrations.ContainsKey(container))
+            {
+                containerRegistrations = _registrations[container];
+            }
```
MikeStall: nit: can you use this form instead:
if (!_registrations.TryGetValue(container, out containerRegistrations))
{
// else clause here
}

That would be a single dictionary lookup instead of 2 lookups (ContainsKey and [] )

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
MikeStall: +1.  I'm assuming the problem is that Start(), which calls this, is synchronous and needs to block. If so, can you make this method async / await and push the blocking up up the stack? 

---
In reply to: [54306524](https://github.com/Azure/azure-webjobs-sdk/pull/637#discussion_r54306524) [](ancestors = 54306524)

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -0,0 +1,162 @@
...
+                _scanContainerStrategy.Dispose();
+                _cancellationTokenSource.Dispose();
+                _disposed = true;
+            }
+        }
+
+        private void ScnaContainersAndRegisterWithStrategy(CancellationToken cancellationToken)
+        {
+            foreach (IStorageBlobContainer container in _registrations.Keys)
+            {
```
MikeStall: nit:   You can eliminate redundant dictionary lookups (see 139, 147) with this pattern:
foreach(var kv in _registrations) {
  IStorageBlobContainer container = kv.Key;
  var registrations = kv.Value;


}

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -0,0 +1,162 @@
...
+                    // else is around to observe the results.
+                    results = container.ListBlobsSegmentedAsync(prefix: null, useFlatBlobListing: true, blobListingDetails: WindowsAzure.Storage.Blob.BlobListingDetails.None, maxResults: _scanBlobThreshold, currentToken: null, options: null, operationContext: null,
+                        cancellationToken: CancellationToken.None).GetAwaiter().GetResult();
+                }
+                catch (StorageException exception)
+                {
+                    if (exception.IsNotFound())
+                    {
+                        return;
+                    }
```
MikeStall: What if the container doesn't exist at startup?  We still want to listen on it .

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -0,0 +1,162 @@
...
+                    else
+                    {
+                        throw;
+                    }
+                }
+
+                // decide which strategy to use based on the existance of continuation token
```
MikeStall: This is a potential regression: 
Do we need to fault between strategies? 
Notably, if a container starts small (100 items) and we begin with the polling strategy, and then over the course of a week, it gets 100,000 items, we'd still be polling. Can the polling 'fault' over to a scan once the container gets large?

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
MikeStall: (Potential regression here) What if the container doesn't exist at startup?  We still want to listen on it .

MikeStall:
:-1:

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
mathewc: Yes, I mentioned that to Hosam on Friday evening as well. He is aware of that issue and is going to address it.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
davidebbo: Got it.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/SharedBlobListener.cs
mathewc: Your new strategy references the existing PollLogsStrategy - it can't be removed

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -0,0 +1,162 @@
...
+namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
+{
+    internal sealed class ScanBlobScanLogHybridPollingStrategy : IBlobListenerStrategy
+    {
+        private static readonly TimeSpan TwoSeconds = TimeSpan.FromSeconds(2);
+        private enum PollStrategy { ScanBlobs, ScanLogs, NotDetermined };
+        private int _scanBlobThreshold = 5000;
+        private readonly IDictionary<IStorageBlobContainer, PollStrategy> _pollStrategies;
+        private readonly IDictionary<IStorageBlobContainer, ICollection<ITriggerExecutor<IStorageBlob>>> _registrations;
+        private ScanContainersStrategy _scanContainerStrategy;
```
mathewc: This existing ScanContainersStrategy is basically untested code, since it effectively dead code since it only runs when dev storage is being used (which we don't support). So if we're now going to use it in the default strategy we'll need to test it thoroughly.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
mathewc: Recommend simply renaming this method to **Initialize** since it's job is to perform any pre-Start initialization required. I find the current name hard to understand.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -0,0 +1,162 @@
...
+                }
+
+                // decide which strategy to use based on the existance of continuation token
+                // if one exists it means there are more use PollLogStrategy otherwise use 
+                // ScanBlobStrategy
+
+                if (results.ContinuationToken != null)
+                {
+                    _pollStrategies.Add(container,PollStrategy.ScanLogs);
+                    foreach (var triggerExecuter in _registrations[container])
```
mathewc: This entire foreach loop is duplicated below. Hoist it out of the if/else and modify it to operate on an IBlobListenerStrategy local that you set in the if/else

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -0,0 +1,162 @@
...
+            return Task.FromResult(0);
+        }
+
+        public async Task<TaskSeriesCommandResult> ExecuteAsync(CancellationToken cancellationToken)
+        {
+            ThrowIfDisposed();
+            await Task.WhenAll(_scanContainerStrategy.ExecuteAsync(cancellationToken),
+                                _pollLogStrategy.ExecuteAsync(cancellationToken));
+
+            // Run subsequent iterations at 2 second intervals.
```
mathewc: For the scan algorithm, I think every 2 seconds is too often. For example: assume the app is monitoring 5 blob containers, each with ~3000 blobs. Then every 2 seconds you're reading 15,000 blobs, and often, for each of those blobs, the BlobTriggerExecutor will issue another storage query to check for the blob receipt. That's a lot of storage operations.

davidebbo: Agred, that's a little too aggressive. Though it does need to be fast enough to feel 'live'. Let's start with every 10 seconds.

Also, name the variable something like `PollingInterval` instead of `TwoSeconds`, so it's easier to change it without needing to change the code (and the comment too).

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/SharedBlobListener.cs
shobak101: yes it does, they are 2 members inside the new class

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
shobak101: That makes sense will do that

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
shobak101: there is already one test and I added a couple more will share in the next PR, but I think more than unit test and functional tests we need to come up with agreement whether we need to do some perf/stress test beforehand and what is it exactly we are testing for, during my investigations I did some tests to profile the time taken for requesting blob list and the data transferred

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -0,0 +1,162 @@
...
+                    }
+                }
+
+                // decide which strategy to use based on the existance of continuation token
+                // if one exists it means there are more use PollLogStrategy otherwise use 
+                // ScanBlobStrategy
+
+                if (results.ContinuationToken != null)
+                {
+                    _pollStrategies.Add(container,PollStrategy.ScanLogs);
```
MikeStall: Scanning logs is account wide. It doesn't make sense to set per-container.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
MikeStall: I thought the strategy was to go round-robbin. So if you listen on 5 containers, each container only gets scanned once every 2*5 = 10 seconds. 

---
In reply to: [54494106](https://github.com/Azure/azure-webjobs-sdk/pull/637#discussion_r54494106) [](ancestors = 54494106)

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -0,0 +1,261 @@
...
+        }
+
+        public async Task RegisterAsync(IStorageBlobContainer container, ITriggerExecutor<IStorageBlob> triggerExecutor, CancellationToken cancellationToken)
+        {
+            // Register and Execute are not concurrency-safe.
+            // Avoiding calling Register while Execute is running is the caller's responsibility.
+            ThrowIfDisposed();
+
+            // Register all in logPolling, there is no problem if we get 2 notifications of the new blob
+            Task registerLogPolling = _pollLogStrategy.RegisterAsync(container, triggerExecutor, cancellationToken);
```
MikeStall: from a simplicity perspective, can we just do the await here rather than run it in parallel with the rest of this function?  This is parallelize with in-memory dictionary lookups, which seems low value.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -0,0 +1,261 @@
...
+namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
+{
+    internal sealed class ScanBlobScanLogHybridPollingStrategy : IBlobListenerStrategy
+    {
+        private static readonly TimeSpan _pollintInterval = TimeSpan.FromSeconds(10);
+        private enum PollStrategy { ScanBlobs, ScanLogs, NotDetermined };
+        private int _scanBlobLimitPerPoll = 10000;
+        private readonly IDictionary<IStorageBlobContainer, ICollection<ITriggerExecutor<IStorageBlob>>> _registrations;
+        private readonly IDictionary<IStorageBlobContainer, DateTime> _lastSweepCycleStartTime;
+        private readonly IDictionary<IStorageBlobContainer, DateTime> _currentSweepCycleStartTime;
```
MikeStall: nit: these 4 dictionaries appear to be parallel. Could you declare a ContainerScanInfo class with the 4 members, and then have one
IDictionary<IStorageBlobContainer, ContainerScanInfo>

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -0,0 +1,261 @@
...
+                IStorageBlob blob;
+
+                if (!_blobsFoundFromScanOrNotification.TryDequeue(out blob))
+                {
+                    break;
+                }
+
+                notifications.Add(NotifyRegistrationsAsync(blob, failedNotifications, cancellationToken));
+            }
+            Task.WhenAll(notifications).Wait();
```
MikeStall: This is a blocking wait. Was that intended? if not, do:
async Task.WhenAll(notifications)

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -0,0 +1,261 @@
...
+                pollingTasks.Add(PollAndNotify(container, cancellationToken, failedNotifications));
+            }
+
+            // Re-add any failed notifications for the next iteration.
+            foreach (IStorageBlob failedNotification in failedNotifications)
+            {
+                _blobsFoundFromScanOrNotification.Enqueue(failedNotification);
+            }
+
+            await LogPollingTask;
```
MikeStall: This is currently serialized with the wait on the next line. 
You could do:
pollingTasks.Add(LogPollingTask) and then the waits will happen together.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
shobak101: what is happening in this approach is, every time I am polling a page out of each container, small containers will always be fully scanned every time, big ones will complete when the budget allows. this approach is the same complexity as the other one in terms of carrying around continuation tokens and paging, but it makes sure that small enough containers are polled every time even if the number of containers increases. while at the same time keeps the load (number of queried blobs) under control

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -0,0 +1,261 @@
...
+            foreach (ITriggerExecutor<IStorageBlob> registration in registrations)
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+
+                FunctionResult result = await registration.ExecuteAsync(blob, cancellationToken);
+                if (!result.Succeeded)
+                {
+                    // If notification failed, try again on the next iteration.
+                    failedNotifications.Add(blob);
+                }
```
MikeStall: what does this mean? 
If the user's function throws an exception, does this mean we'll re-execute it?

shobak101: that is the same approach used in the original ScanBlob strategy https://github.com/Azure/azure-webjobs-sdk/blob/master/src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanContainersStrategy.cs#L144 
I assumed that, this is what the SDK does, is that not correct?
Note: the registration here is not the customer code right away, it is another piece of our code that handles notifications! so what is happening here is if the other SDK piece of code failed we try again. I am not sure how that other piece of code handles customer code failures.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -0,0 +1,243 @@
...
+                cancellationToken.ThrowIfCancellationRequested();
+                await NotifyRegistrationsAsync(newBlob, failedNotifications, cancellationToken);
+            }
+        }
+
+        public void Notify(IStorageBlob blobWritten)
+        {
+            ThrowIfDisposed();
+            _blobsFoundFromScanOrNotification.Enqueue(blobWritten);
+            _pollLogStrategy.Notify(blobWritten);
```
mathewc: Why do you also have to notify _pollLogsStrategy? Isn't adding the blob to the queue sufficient?

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -0,0 +1,243 @@
...
+
+        private void ThrowIfDisposed()
+        {
+            if (_disposed)
+            {
+                throw new ObjectDisposedException(null);
+            }
+        }
+
+        public async Task<IEnumerable<IStorageBlob>> PollNewBlobsAsync(
```
mathewc: Please add some detailed comments in here about the algorithm you're using

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -0,0 +1,243 @@
...
+using Microsoft.WindowsAzure.Storage;
+using Microsoft.WindowsAzure.Storage.Blob;
+
+namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
+{
+    internal sealed class ScanBlobScanLogHybridPollingStrategy : IBlobListenerStrategy
+    {
+        private static readonly TimeSpan _pollintInterval = TimeSpan.FromSeconds(10);
+        private enum PollStrategy { ScanBlobs, ScanLogs, NotDetermined };
+        private int _scanBlobLimitPerPoll = 10000;
```
mathewc: Add detailed comments on what this limit is and how it is used

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -0,0 +1,243 @@
...
+
+        public async Task<IEnumerable<IStorageBlob>> PollNewBlobsAsync(
+            IStorageBlobContainer container, ContainerScanInfo containerScanInfo, CancellationToken cancellationToken)
+        {
+            IEnumerable<IStorageListBlobItem> currentBlobs;
+            IStorageBlobResultSegment blobSegment;
+            int blobPollLimitPerContainer = _scanBlobLimitPerPoll / _scanInfo.Count;
+            BlobContinuationToken continuationToken = containerScanInfo.ContinuationToken;
+
+            // if starting the cycle, keep the current time stamp to be used as curser
```
mathewc: nit: cursor

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
shobak101: good catch, that is a leftover from when I had the new strategy working as a delegate, no needed

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -0,0 +1,243 @@
...
+        private readonly IDictionary<IStorageBlobContainer, ContainerScanInfo> _scanInfo;
+        private readonly ConcurrentQueue<IStorageBlob> _blobsFoundFromScanOrNotification;
+        private PollLogsStrategy _pollLogStrategy;
+        private readonly CancellationTokenSource _cancellationTokenSource;
+        private bool _disposed;
+
+        public ScanBlobScanLogHybridPollingStrategy() : base()
+        {
+            _scanInfo = new Dictionary<IStorageBlobContainer, ContainerScanInfo>(new StorageBlobContainerComparer());
+            _pollLogStrategy = new PollLogsStrategy();
```
mathewc: The existing PollLogsStrategy does startup time background scans of all blobs in all containers. You're retaining that here, which means that your secondary scanning in this strategy will also be chugging through the exact same set of blobs?

mathewc:
In the latest revision, we've moved from the previous model of only doing the scan when containers were small (e.g. < 5000) to always running it behind the scenes, but managing the amount of work by spreading it out over containers/iterations, paging through progressively, Right? Just checking my understanding.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -0,0 +1,243 @@
...
+                DateTime lastModifiedTimestamp = properties.LastModified.Value.UtcDateTime;
+
+                if (lastModifiedTimestamp > containerScanInfo.LastSweepCycleStartTime)
+                {
+                    newBlobs.Add(currentBlob);
+                }
+            }
+
+            // record continuation token for next chunk retrieval
+            containerScanInfo.ContinuationToken = blobSegment.ContinuationToken;
```
mathewc: When running in a classic SDK AlwaysOn scenario, jobs are expected to be very long lived, so this continuation token is key. In cases where we're running in a free site that may get put to sleep after a short period of inactivity (functions), for large containers we'll likely **never get to the end** - we'll keep restarting at the beginning and chugging through to the point where we get idled out. Meaning blobs added at the end of the container might never get discovered. Note that this is the case even before your changes.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
mathewc: Yeah, the retries will continue until the blob poison queue handling kicks in after N failed attempts and moves the blob to the poison queue.

shobak101:
Yes that is exactly what we are doing

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
shobak101: I guess in this case here we have to tell customers that there is a budget constraining the number of blobs we poll, and in case of always on keeping that at a minimum enhances performance, and in case of free sites, keeping it at a minimum will ensure functionality

shobak101:
pushed to master f8d3ee9

PR #639, opened 2016-03-01
Allow running without any storage account 

MikeStall:
When we're not using Azure Storage listeners and don't want logging, allow explicitly disabling the storage accounts.

### DIFF
--- a/test/Microsoft.Azure.WebJobs.Host.UnitTests/JobHostConfigurationTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.UnitTests/JobHostConfigurationTests.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Diagnostics;
 using System.Linq;
+using System.Reflection;
 using Microsoft.Azure.WebJobs.Host.Executors;
 using Microsoft.Azure.WebJobs.Host.Indexers;
 using Microsoft.Azure.WebJobs.Host.TestCommon;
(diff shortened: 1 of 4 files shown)


mathewc:
The storage account is used for a other things that would also be broken. For example HostID generation/storage - See DynamicHostIdProvider. This is the default used, and it uses storage to store the ID.

Have you tried some e2e samples where you've completely disabled storage? I'd expect this to fail.

test/Microsoft.Azure.WebJobs.Host.UnitTests/JobHostConfigurationTests.cs
```diff
@@ -325,5 +326,52 @@ private static void TestHostIdThrows(string hostId)
...
+        // Test that we can explicitly disable storage and call through a function
+        [Fact]
+        public void JobHost_NoStorage_Succeeds()
+        {
+            string prevStorage  = Environment.GetEnvironmentVariable("AzureWebJobsStorage");
+            string prevDashboard = Environment.GetEnvironmentVariable("AzureWebJobsDashboard");
+            try
+            {
+                Environment.SetEnvironmentVariable("AzureWebJobsStorage", null);
+                Environment.SetEnvironmentVariable("AzureWebJobsDashboard", null);
```
mathewc: Do you need to modify machine settings like this if you're setting explicitly below? If this test fails mid way and the finally block isn't run, it will blow away my machine settings.

mathewc:
:shipit:

davidebbo:
Rebased and merged.

CarlosSardo:
Hi Guys,

Is this change already present in WebJobs SDK v1.1.2?

I am using it in a ConsoleApp like this:

`JobHostConfiguration config = new JobHostConfiguration();
            config.HostId = Guid.NewGuid().ToString("n");
            config.DashboardConnectionString = null;
            config.StorageConnectionString = null;


            JobHost host = new JobHost(config);
            host.RunAndBlock();`

But, I still get an exception in the RunAndBlock(), regarding the missing StorageConnectionString. Any clues?

Thanks!

PR #640, opened 2016-03-02
adding support for poco types to GenericBinder

brettsam:


src/Microsoft.Azure.WebJobs.ServiceBus/Common/GenericBinder.cs
```diff
@@ -90,7 +90,59 @@ internal class GenericBinder
...
+        /// to be generated for any POCO type.
+        /// </summary>
+        /// <param name="parameter">The ParameterInfo for the binding</param>
+        /// <param name="collectorGenericType">The generic type that must implement <see cref="IFlushCollector{T}"/> and have a public constructor with at most one parameter.</param>
+        /// <param name="collectorGenericArgumentType">The generic argument type for the <see cref="IFlushCollector{T}"/></param>
+        /// <param name="converterManager">A converter manager to pass along.</param>        
+        /// <param name="contextBinder">A Func that returns the TContext to be used by the <see cref="IFlushCollector{T}"/></param>
+        /// <returns></returns>
+        public static IBinding BindGenericCollector<TContext>(ParameterInfo parameter, Type collectorGenericType, Type collectorGenericArgumentType,
+            IConverterManager converterManager, Func<string, TContext> contextBinder)
```
MikeStall: Is this semantically the same as 'invokeStringBinder' parameter in the BindCollector instance below? If so, can we use the same name to stay consistent.

brettsam: Yep. Done.

src/Microsoft.Azure.WebJobs.ServiceBus/Common/GenericBinder.cs
```diff
@@ -90,7 +90,59 @@ internal class GenericBinder
...
+
+            // Create a delegate to pass as the builder func to BindCollector
+            Type funcType = typeof(Func<,,>).MakeGenericType(typeof(object), typeof(ValueBindingContext), flushCollectorInterfaceType);
+            MethodInfo getCollector = typeof(GenericBinder).GetMethod("GetCollector", BindingFlags.NonPublic | BindingFlags.Static)
+                .MakeGenericMethod(actualCollectorType, collectorGenericArgumentType);
+            var del = Delegate.CreateDelegate(funcType, getCollector);
+
+            // Create the method and parameters.
+            MethodInfo bindCollectorMethod = typeof(GenericBinder).GetMethod("BindCollector").MakeGenericMethod(collectorGenericArgumentType, typeof(TContext));
+            object[] parameters = new object[] { parameter, converterManager, del, null, invokeStringBinder };
```
MikeStall: I wouldn't block the change on this, but this parameter (invokeString) should be non-null.  We'll compute the initial TContext by calling invokeStringBinder(invokeString). See line 262.  
The idea is that the dashboard could replay this function and change the invokeString in the replayed instance.

brettsam: That makes sense. My current usage (which is restricted to only the EasyTable bindings) don't use an invokeString so I had no need for that. We can open this up to be more general in later updates.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
```diff
@@ -115,7 +115,63 @@ public static class BindingFactory
...
+        /// Creates a binding that binds to an <see cref="IAsyncCollector{T}"/> with the specified argument type. Allows for a binding
+        /// to be generated for any POCO type.
+        /// </summary>
+        /// <param name="parameter">The ParameterInfo for the binding</param>
+        /// <param name="collectorGenericType">The generic type that must implement <see cref="IAsyncCollector{T}"/> and have a public constructor with at most one parameter.</param>
+        /// <param name="collectorGenericArgumentType">The generic argument type for the <see cref="IAsyncCollector{T}"/></param>
+        /// <param name="converterManager">A converter manager to pass along.</param>        
+        /// <param name="invokeStringBinder">A <see cref="Func{T, TContext}"/> that returns the TContext to be used by the <see cref="IAsyncCollector{T}"/></param>
+        /// <returns></returns>
+        public static IBinding BindGenericCollector<TContext>(ParameterInfo parameter, Type collectorGenericType, Type collectorGenericArgumentType,
```
MikeStall: Does this type ever participate in a lookup to IConverterManager?

brettsam: It could -- but this method does nothing special other than pass the ConverterManager  along. 

The purpose of this is for when you want to support any kind of T in your IAsyncCollector<T> -- and you want the concrete Collector dynamically created based on the ParameterInfo. 

BindCollector requires you specify an builder Func that returns a concrete type. That requires a bit of reflection to handle -- so I moved all of that reflection in here. You tell this method the IAsyncCollector<T> type and the Type of T and it'll generate the builder Func for you (passing in a context object to the constructor). Everything else is passed along.

There may be a better way to do this internally in BindCollector if we refactor it a bit. We can do that in the future, I'd think.

---
In reply to: [55583043](https://github.com/Azure/azure-webjobs-sdk/pull/640#discussion_r55583043) [](ancestors = 55583043)

MikeStall:
:+1:

PR #649, opened 2016-03-07
fixing test cases and build warnings

shobak101:
3 fixes included
1- fixing warnings caused by coding style 
2- fixing the test case BlobTrigger_IfWritesToSecondBlobTrigger_TriggersOutputQuickly, the test case assumed that the polling interval was 2 seconds thus waited for 10 seconds only, now that the polling is 10 seconds I increased the wait time to 25, 20 is the worst case wait + extra 5 seconds for processing
3- in the hybrid stragegy, 2 initial scans were happening, disabled one and only the scan in logpolling takes place

### DIFF
--- a/WebJobs.sln
+++ b/WebJobs.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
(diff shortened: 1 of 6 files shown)


mathewc:
Looks good.

shobak101:
pushed to master 5442d2d

PR #655, opened 2016-03-12
Logging optimizations.

MikeStall:
Significant performance improvement to the SDK logging infrastructure. 
Create new WebJobs.Logging library with fast reader & writers based on table implementations.
Plumb Dashboard to accept the new reader if an appsetting is set.
SDK emits new events.
This is a preview of the changes. There are still some outanding things to address before the pull request is ready, noted by $$$ markers in the code.  Notably plumbing the function instance information through for Pass/Fail, arguments, and log output.


src/Microsoft.Azure.WebJobs.Logging/Entities/ContainerActiveEntity.cs
```diff
@@ -0,0 +1,104 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Logging
+{
+    // Entity to track active compute Containers during a timerange.
+    // A Container is a billable compute unit, like a virtual machine. 
+    // A container may run many functions simultaneously. 
+    public class ContainerActiveEntity : TableEntity
+    {
+        const string PartitionKeyFormat = TableScheme.ContainerActivePK;
```
mathewc: We need to make sure we design the table structure from the get go to be able to handle efficient log purge strategies. We know we'll want to be able to have retention policies, as well as the ability for customers to purge logs themselves by range. We have existing issues in the issues list tracking this.

Table Storage doesn't have a good way within a single table to be able to purge a range of items by query. E.g. see https://feedback.azure.com/forums/217298-storage/suggestions/2277650-ability-to-delete-from-table-storage-by-partition. You have to query each item for deletion one by one and delete it. In high throughput scenarios, that won't scale.

src/Dashboard/App_Start/AppModule.cs
```diff
@@ -35,46 +37,81 @@ public override void Load()
...
-            Bind<IFunctionIndexer>().To<FunctionIndexer>();
-            Bind<IIndexer>().To<UpgradeIndexer>();
-            Bind<IInvoker>().To<Invoker>();
-            Bind<IAbortRequestLogger>().To<AbortRequestLogger>();
-            Bind<IAborter>().To<Aborter>();
-
-            Bind<IIndexerLogWriter>().To<IndexerBlobLogWriter>();
-            Bind<IIndexerLogReader>().To<IndexerBlobLogReader>();
+
+            string tableLog = ConfigurationManager.AppSettings["AzureWebJobsLogTableName"];
```
mathewc: Since we're going to need to move to a more flexible multi-table approach, I don't think we should have a setting that indicates a table name. I think the table structure should be completely internal to the FastTableReader. If in the short term it only uses a single table, that should be an implementation detail, and we need to make sure all the new APIs will support the future.

If you don't have time now to implement the multi-table approach, we should have a bug to address that, so you can do it later.

src/Dashboard/App_Start/AppModule.cs
```diff
@@ -35,46 +37,81 @@ public override void Load()
...
+            if (tableLog != null)
+            {
+                // ASt table reader.
```
mathewc: spelling

src/Dashboard/App_Start/AppModule.cs
```diff
@@ -35,46 +37,81 @@ public override void Load()
...
-            Bind<IIndexerLogReader>().To<IndexerBlobLogReader>();
+
+            string tableLog = ConfigurationManager.AppSettings["AzureWebJobsLogTableName"];
+            if (tableLog != null)
+            {
+                // ASt table reader. 
+                var client = account.CreateCloudTableClient();
+                CloudTable logTable = client.GetTableReference(tableLog);
+
+                var s = new FastTableReader(logTable);
```
mathewc: Are there other interfaces that remain TODOs to also implement here? If so, please add a TODO comment

src/Dashboard/Dashboard.csproj
```diff
@@ -278,6 +278,16 @@
     <Compile Include="Data\TableParameterSnapshot.cs" />
     <Compile Include="Data\ConcurrentMetadataDocument.cs" />
     <Compile Include="Data\BlobConcurrentTextStore.cs" />
+    <Compile Include="FastTable\FastTableReader.cs" />
+    <Compile Include="FastTable\ExtensionHelpers.cs" />
+    <Compile Include="FastTable\NullAborter.cs" />
+    <Compile Include="FastTable\NullHeadbeatValidityMonitor.cs" />
```
mathewc: Not "Headbeat" :)

src/Dashboard/App_Start/AppModule.cs
MikeStall: I'd prefer to add a bug to track this. 

---
In reply to: [56221610](https://github.com/Azure/azure-webjobs-sdk/pull/655#discussion_r56221610) [](ancestors = 56221610)

src/Microsoft.Azure.WebJobs.Logging/Entities/ContainerActiveEntity.cs
MikeStall: This is tricky. I'll followup in email. 

---
In reply to: [56055211](https://github.com/Azure/azure-webjobs-sdk/pull/655#discussion_r56055211) [](ancestors = 56055211)

mathewc:
Closing since I believe this has all been merged.

PR #661, opened 2016-03-17
Making FunctionIndexingException public

mathewc:
And flowing it through TraceWriter pipeline. This supports scenarios where a user might want to register a TraceWriter that inspects/acts upon indexing exceptions.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
@@ -27,8 +27,9 @@ namespace Microsoft.Azure.WebJobs.Host.Indexers
         private readonly IFunctionExecutor _executor;
         private readonly HashSet<Assembly> _jobAttributeAssemblies;
         private readonly SingletonManager _singletonManager;
+        private readonly TraceWriter _trace;
 
-        public FunctionIndexer(ITriggerBindingProvider triggerBindingProvider, IBindingProvider bindingProvider, IJobActivator activator, IFunctionExecutor executor, IExtensionRegistry extensions, SingletonManager singletonManager)
+        public FunctionIndexer(ITriggerBindingProvider triggerBindingProvider, IBindingProvider bindingProvider, IJobActivator activator, IFunctionExecutor executor, IExtensionRegistry extensions, SingletonManager singletonManager, TraceWriter trace)
         {
             if (triggerBindingProvider == null)
             {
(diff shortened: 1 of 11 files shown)


src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
```diff
@@ -60,19 +61,45 @@ public FunctionIndexer(ITriggerBindingProvider triggerBindingProvider, IBindingP
...
+                {
+                    // Route the indexing exception through the TraceWriter pipeline
+                    _trace.Error(fex.Message, fex);
+
+                    // If the error has been marked as handled, ignore the indexing exception
+                    // and continue on to the rest of the methods. In this case the method in
+                    // error simply won't be running in the JobHost.
+                    if (fex.Handled)
+                    {
+                        continue;
```
davidebbo: This goes beyond the added trace support no?

So at the time exception is thrown, it may already be marked handled? I guess I don't have context to understand what that means.

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexingException.cs
```diff
@@ -3,17 +3,68 @@
...
+        /// Constructs a new instance.
+        /// </summary>
+        /// <param name="info">The <see cref="SerializationInfo"/>.</param>
+        /// <param name="context">The <see cref="StreamingContext"/>.</param>
+        protected FunctionIndexingException(SerializationInfo info, StreamingContext context) : base(info, context)
+        {
+            if (info == null)
+            {
+                throw new ArgumentNullException("info");
+            }
```
davidebbo: How easily can we just yank that rule that requires these useless checks? I'm guessing the amount of time we spend adding those checks for all new code adds up to much more than it would take to fix it. And I really don't like bloating the IL.

I'll keep bringing this one up! :)

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexingException.cs
```diff
@@ -3,17 +3,68 @@
...
+        /// <inheritdoc/>
+        public override void GetObjectData(SerializationInfo info, StreamingContext context)
+        {
+            if (info == null)
+            {
+                throw new ArgumentNullException("info");
+            }
+
+            info.AddValue("MethodName", this.MethodName);
+            info.AddValue("Handled", this.Handled);
```
davidebbo: `const` for those since more than one use

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
mathewc: See: https://github.com/Azure/azure-webjobs-sdk-script/pull/123 which requires these changes :)

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexingException.cs
mathewc: I'm open to relaxing rules where they make sense. In some repos where the APIs aren't a public SDK it makes sense, but this repo IS a public repo, so I'm more cautious. Let's discuss another time.

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexingException.cs
mathewc: Huh?

davidebbo: MethodName = info.GetString("MethodName");

and

    info.AddValue("MethodName", this.MethodName);

Rule of thumb is to not repeat literal strings in multiple places.

davidebbo:
:shipit: minor comments

PR #666, opened 2016-03-20
Adding new IBinderEx interface to support advanced IBinder scenarios

mathewc:
The base IBinder interface is limited - for example, it doesn't allow the specification of additional parameter attributes for the bind (e.g. StorageAccountAttribute). Introducing an additional IBinderEx interface that can be used to access these additional capabilities for advanced scenarios. IBinderEx is desiged to be more future proof, by having it take a single RuntimeBindingContext input which can be extended in the future.


src/Microsoft.Azure.WebJobs/StorageAccountAttribute.cs
```diff
@@ -18,10 +18,15 @@ public sealed class StorageAccountAttribute : Attribute
...
+        /// <param name="account">A string value indicating the Azure Storage connection string to use. This
+        /// string should be in one of the following formats. These checks will be applied in order and the
+        /// first match wins.
+        /// - The name of an "AzureWebJobs" prefixed app setting or connection string name. E.g., if your setting
+        ///   name is "AzureWebJobsMyStorage", you can specify "MyStorage" here.
+        /// - Can be a string containing %% values (e.g. %StagingStorage%). The value provided will be passed
+        ///   to any INameResolver registered on the JobHostConfiguration to resolve the actual setting name to use.
+        /// - Can be an app setting or connection string name of your choosing.
+        /// </param>
         public StorageAccountAttribute(string account)
```
brettsam: I didn't realize this existed until now. It seems generally useful. Can this be named something like ConnectionNameAttribute and be used by other bindings? I'm thinking we need to add this behavior throughout the extensions repo.

If it's too hard to generalize this one (I haven't looked through it all yet), maybe we could add it there and allow all those extensions to re-use it?

brettsam: I just saw your comment in the other repo about how you can apply this at the class level. So making a single attribute isn't the right answer -- but exposing the behavior so that other binding-specific attributes can use it would be useful.

src/Microsoft.Azure.WebJobs.Host/Bindings/Runtime/IBinderEx.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace Microsoft.Azure.WebJobs.Host.Bindings.Runtime
+{
+    /// <summary>
+    /// Interface defining functionality for dynamically binding to WebJobs SDK attributes
+    /// at runtime.
+    /// </summary>
+    /// <remarks>This interface enables imperative binding with attribute information specified at runtime.</remarks>
+    public interface IBinderEx
+    {
```
brettsam: Does anything implement this?

src/Microsoft.Azure.WebJobs/StorageAccountAttribute.cs
mathewc: Yeah, when you support class level attributes, you need a per connection type attribute. I chose this approach for all the storage bindings because there are many (and we'll add more soon, e.g. Azure Files). I didn't want to add a new duplicate property to each binding.

What behavior are you suggesting be exposed? Each binding type has different connection requirements and behavior.

src/Microsoft.Azure.WebJobs.Host/Bindings/Runtime/IBinderEx.cs
brettsam: n/m -- now I see RuntimeValueProvider :-)

src/Microsoft.Azure.WebJobs/StorageAccountAttribute.cs
brettsam: I'm suggesting that I'd like to support `[DocumentDbConnectionAttribute(string connection)]` and have the same behavior you have above -- looking for `AzureWebJobs`+connection, looking for %connection%, looking for an app setting named that specifically, looking for a connection string named that specifically.

Same thing for NH and EasyTables. I can see all of them using that same behavior. The BindingProviders would still need to call something to do the actual creation of the DocumentClient, MobileServiceClient, etc, but that logic of doing the lookups would be nice to share.

MikeStall:
:-1:

PR #709, opened 2016-05-04
updating nuget references

brettsam:
Moving all non-Dashboard nuget dependencies forward to their newest versions. There were two test changes that I'll call out in comments shortly. I also experienced a few test failures that passed upon re-runs. I've not run this repo's tests before so it'd be good if someone else could pull this change and run locally as well.

### DIFF
--- a/test/Microsoft.Azure.WebJobs.Host.FunctionalTests/TestDoubles/MemoryBlobStore.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.FunctionalTests/TestDoubles/MemoryBlobStore.cs
@@ -15,7 +15,7 @@ namespace Microsoft.Azure.WebJobs.Host.FunctionalTests.TestDoubles
     internal class MemoryBlobStore
     {
         private readonly ConcurrentDictionary<string, Container> _items = new ConcurrentDictionary<string, Container>();
-        private ServiceProperties _properties = new ServiceProperties();
+        private ServiceProperties _properties = new ServiceProperties(new LoggingProperties(), new MetricsProperties(), new MetricsProperties(), new CorsProperties());
 
         public string AcquireLease(string containerName, string blobName, TimeSpan? leaseTime)
         {
@@ -192,9 +192,6 @@ namespace Microsoft.Azure.WebJobs.Host.FunctionalTests.TestDoubles
                 DefaultServiceVersion = properties.DefaultServiceVersion,
                 HourMetrics = Clone(properties.HourMetrics),
                 Logging = Clone(properties.Logging),
-#pragma warning disable 0618
-                Metrics = Clone(properties.Metrics),
-#pragma warning restore 0618
                 MinuteMetrics = Clone(properties.MinuteMetrics)
             };
         }
--- a/test/Microsoft.Azure.WebJobs.ServiceBus.UnitTests/Listeners/ServiceBusListenerTests.cs
+++ b/test/Microsoft.Azure.WebJobs.ServiceBus.UnitTests/Listeners/ServiceBusListenerTests.cs
@@ -86,7 +86,7 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.UnitTests.Listeners
                 {
                     await _listener.StartAsync(CancellationToken.None);
                 });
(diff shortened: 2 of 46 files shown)


test/Microsoft.Azure.WebJobs.Host.FunctionalTests/TestDoubles/MemoryBlobStore.cs
```diff
@@ -192,9 +192,6 @@ private static ServiceProperties Clone(ServiceProperties properties)
                 DefaultServiceVersion = properties.DefaultServiceVersion,
                 HourMetrics = Clone(properties.HourMetrics),
                 Logging = Clone(properties.Logging),
-#pragma warning disable 0618
-                Metrics = Clone(properties.Metrics),
-#pragma warning restore 0618
```
brettsam: This was obsolete in the previous version and has now been removed.

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/TestDoubles/MemoryBlobStore.cs
```diff
@@ -15,7 +15,7 @@ namespace Microsoft.Azure.WebJobs.Host.FunctionalTests.TestDoubles
     internal class MemoryBlobStore
     {
         private readonly ConcurrentDictionary<string, Container> _items = new ConcurrentDictionary<string, Container>();
-        private ServiceProperties _properties = new ServiceProperties();
+        private ServiceProperties _properties = new ServiceProperties(new LoggingProperties(), new MetricsProperties(), new MetricsProperties(), new CorsProperties());
```
brettsam: In the previous Storage package, `ServiceProperties` initialized all of its properties. Now it does not. Instead they've added a second constructor that allows you to initialize them.

test/Microsoft.Azure.WebJobs.ServiceBus.UnitTests/Listeners/ServiceBusListenerTests.cs
```diff
@@ -86,7 +86,7 @@ public ServiceBusListenerTests()
                 {
                     await _listener.StartAsync(CancellationToken.None);
                 });
-            Assert.Equal("No such host is known", ex.Message);
+            Assert.Equal("Unable to connect to Service Bus using HTTP connectivity mode.", ex.Message);
```
brettsam: This error string has changed in the new package.

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/TestDoubles/MemoryBlobStore.cs
brettsam: @mathewc I wanted your opinion on this -- if you look at https://github.com/Azure/azure-webjobs-sdk/blob/b1fb20b0edf0530d62127d6dd9f50cc9345ca7b6/src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/BlobLogListener.cs#L184, you'll see that we do no null check on `serviceProperties.Logging`. In the previous Storage SDK, this was always initialized so we never hit it.

Should we change the behavior here to create a `new LoggingProperties()` object it if it's null on the `serviceProperties`? Or just add a null check and skip the logging setup? I'm not sure whether you'd ever actually get a null Logging property or what happens if that logging setup block is skipped.

davidebbo:
Other than the one outstanding comment for @mathewc, it looks like a straightforward (albeit large) package upgrade deal. As long as nothing appears broken, we should go with it.

mathewc:
Let me take a look

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/TestDoubles/MemoryBlobStore.cs
mathewc: Yeah, I'm not sure either - I haven't looked at this test code before. Since its just test code, I'd say if all the tests are passing with what you have, that's good.

mathewc:
Merged.

PR #711, opened 2016-05-04
Have [Queue] attribute run on top of new rule-pattern bindings.

MikeStall:
This adds one new file to describe the queue rules (see QueueBindingRules.cs) , and then removes the previous queue binding providers.


src/Microsoft.Azure.WebJobs.Host/Indexers/DefaultBindingProvider.cs
```diff
@@ -26,7 +25,11 @@ internal static class DefaultBindingProvider
             IExtensionRegistry extensions)            
         {
             List<IBindingProvider> innerProviders = new List<IBindingProvider>();
-            var ruleQueueOutput = new QueueAttributeBindingProvider(nameResolver, storageAccountProvider, messageEnqueuedWatcherGetter);
+
+            // Wire up new bindings 
+            IConverterManager converterManager = new ConverterManager();
+            var queueHelper = new QueueBindingRules(storageAccountProvider, messageEnqueuedWatcherGetter);
```
mathewc: I think it would look more natural to write:

var queueRules = new QueueBindingRules(...);
var queueBindingProvider = queueRules.Build(...);

Or maybe even:

var queueBindingProvider = QueueBindingRules.Build(...);

The point is to reduce the names/concepts in the mix.

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/AsyncCollectorArgumentBindingProvider.cs
```diff
@@ -1,84 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the MIT License. See License.txt in the project root for license information.
-
```
mathewc: Absolutely beautiful to see all these types removed. Are you **sure** the base helpers are handling all these scenarios? Seems too good to be true. I guess we have to rely on the existing tests, but lets make sure you've eyeballed all these deleted classes to look for any quirky things that might not have tests (maybe you already did this).

src/Microsoft.Azure.WebJobs.Host/Bindings/QueueBindingRules.cs
```diff
@@ -0,0 +1,217 @@
...
+            return new ResolvedQueueAttribute(queueName, client);
+        }
+
+        private ParameterDescriptor ToParameterDescriptorForCollector(QueueAttribute attr, ParameterInfo parameter, INameResolver nameResolver)
+        {
+            Task<IStorageAccount> t = Task.Run(() =>
+                _accountProvider.GetStorageAccountAsync(parameter, CancellationToken.None, nameResolver));
+            string accountName = t.GetAwaiter().GetResult().Credentials.AccountName;
+
+            return new QueueParameterDescriptor
```
mathewc: Does this parameter descriptor produce the exact same output as before? The reason I ask is that we have a bunch of Dashboard tests that cover the output. We'll need to run those.

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/ByteArrayArgumentBindingProvider.cs
```diff
@@ -1,83 +0,0 @@
...
-using Microsoft.Azure.WebJobs.Host.Bindings;
-using Microsoft.Azure.WebJobs.Host.Storage.Queue;
-
-namespace Microsoft.Azure.WebJobs.Host.Queues.Bindings
-{
-    internal class ByteArrayArgumentBindingProvider : IQueueArgumentBindingProvider
-    {
-        private readonly IContextGetter<IMessageEnqueuedWatcher> _messageEnqueuedWatcherGetter;
-
-        public ByteArrayArgumentBindingProvider(IContextGetter<IMessageEnqueuedWatcher> messageEnqueuedWatcherGetter)
```
mathewc: This IMessageEnqueuedWatcher functionality looks like it is being removed with no replacement? Isn't that breaking things?

MikeStall: It's  passed into the QueueBindingRules and flows through to the new AsyncCollector. 

---
In reply to: [62358535](https://github.com/Azure/azure-webjobs-sdk/pull/711#discussion_r62358535) [](ancestors = 62358535)

mathewc: Ah, I see, thanks. Just trying to be sure we're not losing anything unintentionally.

src/Microsoft.Azure.WebJobs.Host/Bindings/QueueBindingRules.cs
MikeStall: I think I'm missing a case... It's correct for the AsyncCollector rule, but missing a case for the exact type rule. 

---
In reply to: [62355693](https://github.com/Azure/azure-webjobs-sdk/pull/711#discussion_r62355693) [](ancestors = 62355693)

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/ByteArrayToStorageQueueMessageConverter.cs
```diff
@@ -1,34 +0,0 @@
...
-            }
-
-            _queue = queue;
-        }
-
-        public IStorageQueueMessage Convert(byte[] input)
-        {
-            if (input == null)
-            {
-                throw new InvalidOperationException("A queue message cannot contain a null byte array instance.");
```
mathewc: What do these custom conversion error strings get replaced in the new converter pipeline with out of curiosity?

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/MessageConverterFactory.cs
```diff
@@ -1,42 +0,0 @@
...
-            }
-            else if (typeof(TInput) == typeof(byte[]))
-            {
-                return (IMessageConverterFactory<TInput>)new ByteArrayMessageConverterFactory();
-            }
-            else
-            {
-                if (typeof(TInput).IsPrimitive)
-                {
-                    throw new NotSupportedException("Primitive types are not supported.");
```
mathewc: Are these same rules in place (no primitives, nested collections), or are they now supported?

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/MessageConverterFactory.cs
```diff
@@ -1,42 +0,0 @@
...
-                }
-
-                if (typeof(IEnumerable).IsAssignableFrom(typeof(TInput)))
-                {
-                    throw new InvalidOperationException("Nested collections are not supported.");
```
mathewc: Same question here: If I try to bind to an IAsyncCollector<IEnumerable<MyPoco>> (which is what I think this is trying to prevent) what kind of error message do I get?

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/ConverterValueBinder.cs
```diff
@@ -1,60 +0,0 @@
...
-
-        public ConverterValueBinder(IStorageQueue queue, IConverter<TInput, IStorageQueueMessage> converter,
-            IMessageEnqueuedWatcher messageEnqueuedWatcher)
-        {
-            _queue = queue;
-            _converter = converter;
-            _messageEnqueuedWatcher = messageEnqueuedWatcher;
-        }
-
-        public BindStepOrder StepOrder
```
mathewc: How is BindStepOrder handled in the new pipeline? How do users control this for their bindings - some like queues need to happen after all other bindings.

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/QueueAttributeBindingProvider.cs
```diff
@@ -1,92 +0,0 @@
...
-            QueueAttribute queueAttribute = parameter.GetCustomAttribute<QueueAttribute>(inherit: false);
-
-            if (queueAttribute == null)
-            {
-                return null;
-            }
-
-            string queueName = Resolve(queueAttribute.QueueName);
-            IBindableQueuePath path = BindableQueuePath.Create(queueName);
-            path.ValidateContractCompatibility(context.BindingDataContract);
```
mathewc: Is all this early contract validation still done in the new pipeline?

src/Microsoft.Azure.WebJobs.Host/Bindings/QueueBindingRules.cs
mathewc: Yeah, the Dashboard tests try all the binding permutations and verify the expected output. We want to make sure we don't break those. Tests are here: https://github.com/Azure/azure-webjobs-sdk-dashboard-tests/blob/master/test/Dashboard.EndToEndTests/Tests/ArgumentsDisplay/QueueArgumentsDisplayTests.cs

Are you able to run them locally? If not, it might be that since we've all moved to the latest VS the tests no longer run for us. I can update the test proj to try to get it running for VS2015.

src/Microsoft.Azure.WebJobs.Host/Bindings/QueueBindingRules.cs
```diff
@@ -0,0 +1,217 @@
...
+using Microsoft.Azure.WebJobs.Host.Queues;
+using Microsoft.Azure.WebJobs.Host.Storage;
+using Microsoft.Azure.WebJobs.Host.Storage.Queue;
+using Microsoft.WindowsAzure.Storage.Queue;
+
+namespace Microsoft.Azure.WebJobs.Host.Bindings
+{
+    // Write up bindinging rules for [Queue] attribute. 
+    // This is fundemantentally an IAsyncCollector<IStorageQueueMessage>
+    internal class QueueBindingRules
```
mathewc: This will be an example we point people to to follow, so I'd like to make sure we have good naming that we recommend people to follow.

What if we called this QueueBindingProvider, and it had a method Build? That makes sense to me. Maybe we steer people to a factory pattern, with a static QueueBindingProvider.Build?

Having a "build rules" method that returns an IBindingProvider seems off to me.

src/Microsoft.Azure.WebJobs.Host/Bindings/QueueBindingRules.cs
```diff
@@ -0,0 +1,217 @@
...
+            INameResolver nameResolver,
+            IConverterManager converterManager)
+        {
+            // IStorageQueueMessage is the core testing interface 
+            converterManager.AddConverter<byte[], IStorageQueueMessage, QueueAttribute>(ConvertByteArrayCloudQueueMessage);
+            converterManager.AddConverter<string, IStorageQueueMessage, QueueAttribute>(ConvertString2CloudQueueMessage);
+            converterManager.AddConverter<IStorageQueueMessage, string>(ConvertCloudQueueMessage2String);
+            converterManager.AddConverter<IStorageQueueMessage, byte[]>(ConvertCloudQueueMessage2ByteArray);
+
+            converterManager.AddConverter<CloudQueueMessage, IStorageQueueMessage>(ConvertReal2CloudQueueMessage);
```
mathewc: nit: move this next to the ones above (remove extraneous space)

src/Microsoft.Azure.WebJobs.Host/Bindings/QueueBindingRules.cs
```diff
@@ -0,0 +1,217 @@
...
+
+            var bf = new BindingFactory(nameResolver, converterManager);
```
mathewc: nit: rename bf to bindingFactory

src/Microsoft.Azure.WebJobs.Host/Bindings/QueueBindingRules.cs
```diff
@@ -0,0 +1,217 @@
...
+
+            var ruleQueueOutput = bf.BindToAsyncCollector<QueueAttribute, IStorageQueueMessage>(BuildFromQueueAttribute, ToParameterDescriptorForCollector, FixerUpper);
+            var ruleQueueClient = bf.BindToExactAsyncType<QueueAttribute, IStorageQueue>(BuildClientFromQueueAttributeAsync, ToParameterDescriptorForCollector, FixerUpper);
+            var ruleQueueClient2 = bf.BindToExactAsyncType<QueueAttribute, CloudQueue>(BuildRealClientFromQueueAttributeAsync, ToParameterDescriptorForCollector, FixerUpper);
+            var queueRules = new GenericCompositeBindingProvider<QueueAttribute>(
```
mathewc: rename queueRules to bindingProvider. I'd like us not to have name mismatches like this :)

src/Microsoft.Azure.WebJobs.Host/Bindings/QueueBindingRules.cs
```diff
@@ -0,0 +1,217 @@
...
+            converterManager.AddConverter<byte[], IStorageQueueMessage, QueueAttribute>(ConvertByteArrayCloudQueueMessage);
+            converterManager.AddConverter<string, IStorageQueueMessage, QueueAttribute>(ConvertString2CloudQueueMessage);
+            converterManager.AddConverter<IStorageQueueMessage, string>(ConvertCloudQueueMessage2String);
+            converterManager.AddConverter<IStorageQueueMessage, byte[]>(ConvertCloudQueueMessage2ByteArray);
+
+            converterManager.AddConverter<CloudQueueMessage, IStorageQueueMessage>(ConvertReal2CloudQueueMessage);
+
+            var bf = new BindingFactory(nameResolver, converterManager);
+
+            var ruleQueueOutput = bf.BindToAsyncCollector<QueueAttribute, IStorageQueueMessage>(BuildFromQueueAttribute, ToParameterDescriptorForCollector, FixerUpper);
```
mathewc: nit: can we rename these to something like: collectorProvider, mockQueueClientProvider, queueClientProvider? They're returning IBindingProviders not "rules"

src/Microsoft.Azure.WebJobs.Host/Bindings/QueueBindingRules.cs
```diff
@@ -0,0 +1,217 @@
...
+            var ruleQueueClient = bf.BindToExactAsyncType<QueueAttribute, IStorageQueue>(BuildClientFromQueueAttributeAsync, ToParameterDescriptorForCollector, FixerUpper);
+            var ruleQueueClient2 = bf.BindToExactAsyncType<QueueAttribute, CloudQueue>(BuildRealClientFromQueueAttributeAsync, ToParameterDescriptorForCollector, FixerUpper);
+            var queueRules = new GenericCompositeBindingProvider<QueueAttribute>(
+                (attr) => ValidateQueueAttribute(attr, nameResolver),
+                ruleQueueClient, ruleQueueClient2, ruleQueueOutput);
+
+            return queueRules;
```
mathewc: rename queueRules to bindingProvider

src/Microsoft.Azure.WebJobs.Host/Bindings/QueueBindingRules.cs
```diff
@@ -0,0 +1,217 @@
...
+        }
+
+        // Hook to apply hacky rules all at once.
```
mathewc: Can we clean thus up? You mention "hacky" rules, and FixerUpper seems like temp code naming :) People are going to look at this as an example to follow, so I'd like to have it as clean as possible :)

src/Microsoft.Azure.WebJobs.Host/Bindings/QueueBindingRules.cs
```diff
@@ -0,0 +1,217 @@
...
+                (attr) => ValidateQueueAttribute(attr, nameResolver),
+                ruleQueueClient, ruleQueueClient2, ruleQueueOutput);
+
+            return queueRules;
+        }
+
+        // Hook to apply hacky rules all at once. 
+        private async Task<QueueAttribute> FixerUpper(QueueAttribute attrResolved, ParameterInfo parameter, INameResolver nameResolver)
+        {
+            // Look for [Storage] attribute and squerrel aover
```
mathewc: nit: spelling errors

src/Microsoft.Azure.WebJobs.Host/Bindings/QueueBindingRules.cs
```diff
@@ -0,0 +1,217 @@
...
+        {
+            // Look for [Storage] attribute and squerrel aover 
+            IStorageAccount account = await _accountProvider.GetStorageAccountAsync(parameter, CancellationToken.None, nameResolver);
+            StorageClientFactoryContext clientFactoryContext = new StorageClientFactoryContext
+            {
+                Parameter = parameter
+            };
+            IStorageQueueClient client = account.CreateQueueClient(clientFactoryContext);
+
+            // Create the queue if needed?
```
mathewc: Previously the queue was created on demand the first time a message was added. It looks like you're doing this eagerly now at bind/startup time?

I think that's the more desirable behavior - i.e. if I have 20 queue functions pointing to different queues but I only ever use 1, I don't want all 20 queues created. I think that will be true of a lot of other bindings - e.g. for expensive resources like DocumentDB, I don't want to create tables until they're actually used.

src/Microsoft.Azure.WebJobs.Host/Bindings/QueueBindingRules.cs
```diff
@@ -0,0 +1,217 @@
...
+
+        // This is a static validation (so only %% are resolved; not {} ) 
+        // For runtime validation, the regular builder functions can do the resolution.
+        private static void ValidateQueueAttribute(QueueAttribute attribute, INameResolver nameResolver)
+        {
+            string queueName = NormalizeQueueName(attribute, nameResolver);
+            QueueClient.ValidateQueueName(queueName);
+        }
+
+        private IStorageQueueMessage ConvertReal2CloudQueueMessage(CloudQueueMessage arg)
```
mathewc: nit: ConvertToStorageQueueMessage

src/Microsoft.Azure.WebJobs.Host/Bindings/QueueBindingRules.cs
```diff
@@ -0,0 +1,217 @@
...
+        {
+            return new StorageQueueMessage(arg);
+        }
+
+        private byte[] ConvertCloudQueueMessage2ByteArray(IStorageQueueMessage arg)
```
mathewc: Please replace all "2"s with "To"

src/Microsoft.Azure.WebJobs.Host/Bindings/QueueBindingRules.cs
```diff
@@ -0,0 +1,217 @@
...
+        private IAsyncCollector<IStorageQueueMessage> BuildFromQueueAttribute(QueueAttribute attrResolved)
+        {
+            var attr = (ResolvedQueueAttribute)attrResolved;
+            var queue = attr.GetQueue();
+            return new QueueAsyncCollector(queue, _messageEnqueuedWatcherGetter.Value);
+        }
+
+        // Queue attributes are paired with a separate [StorageAccount]. 
+        // Consolidate the information from both attributes into a single attribute.
+        internal sealed class ResolvedQueueAttribute : QueueAttribute
```
mathewc: We're going to need to find a better pattern for this. While you were able to work around this specific instance with some hacks (I think you'll agree they're hacks :)) other instances like this are going to arise. I'm wondering if the way out isn't to introduce a context object that gets flowed around through your new pipeline, along with the attribute. I think we're going to need something like that - otherwise people are going to just start hanging all kinds of crap of their attributes, abusing the .NET attribute model and treating the attributes as context bags.

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/MessageConverterFactory.cs
MikeStall: restrictions and exception type came from [Queue]. 

---
In reply to: [62360839](https://github.com/Azure/azure-webjobs-sdk/pull/711#discussion_r62360839) [](ancestors = 62360839)

src/Microsoft.Azure.WebJobs.Host/Bindings/QueueBindingRules.cs
```diff
@@ -0,0 +1,217 @@
...
+            converterManager.AddConverter<string, IStorageQueueMessage, QueueAttribute>(ConvertString2CloudQueueMessage);
+            converterManager.AddConverter<IStorageQueueMessage, string>(ConvertCloudQueueMessage2String);
+            converterManager.AddConverter<IStorageQueueMessage, byte[]>(ConvertCloudQueueMessage2ByteArray);
+
+            converterManager.AddConverter<CloudQueueMessage, IStorageQueueMessage>(ConvertReal2CloudQueueMessage);
+
+            var bf = new BindingFactory(nameResolver, converterManager);
+
+            var ruleQueueOutput = bf.BindToAsyncCollector<QueueAttribute, IStorageQueueMessage>(BuildFromQueueAttribute, ToParameterDescriptorForCollector, FixerUpper);
+            var ruleQueueClient = bf.BindToExactAsyncType<QueueAttribute, IStorageQueue>(BuildClientFromQueueAttributeAsync, ToParameterDescriptorForCollector, FixerUpper);
```
brettsam: BindToExactType doesn't have these last two parameters -- should it?

src/Microsoft.Azure.WebJobs.Host/Bindings/QueueBindingRules.cs
```diff
@@ -0,0 +1,217 @@
...
+            var ruleQueueClient2 = bf.BindToExactAsyncType<QueueAttribute, CloudQueue>(BuildRealClientFromQueueAttributeAsync, ToParameterDescriptorForCollector, FixerUpper);
```
brettsam: When is FixerUpper called?

mathewc:
What is the status of this review? There are outstanding comments from 11 days ago.

src/Microsoft.Azure.WebJobs.Host/Bindings/QueueBindingRules.cs
MikeStall: I'm preserving the original behavior.  (we have tests for it). 
This is actually dead code and could be removed (note the if on line 67).

The original rules are:
- if the parameter type is CloudQueue, then eagerly create. Which we do on in BuildClientFromQueueAttributeAsync
- else, defer creation until a message is added: which happens in QueueAsyncCollector 


---
In reply to: [62366656](https://github.com/Azure/azure-webjobs-sdk/pull/711#discussion_r62366656) [](ancestors = 62366656)

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/ConverterValueBinder.cs
MikeStall: This class gets replaced with AsyncCollectorValueProvider, which implements StepOrder. 



---
In reply to: [62361334](https://github.com/Azure/azure-webjobs-sdk/pull/711#discussion_r62361334) [](ancestors = 62361334)

src/Microsoft.Azure.WebJobs.Host/Bindings/QueueBindingRules.cs
MikeStall: I agree this isn't ideal. However, I only end up doing this to cope with the state split between [Queue] and [Storage] attributes. Most extensions shouldn't have that split since Functions will naturally push us to a single attribute. I really want to avoid another context object (ala IBinding, IValueProvider) if we can. 


---
In reply to: [62368731](https://github.com/Azure/azure-webjobs-sdk/pull/711#discussion_r62368731) [](ancestors = 62368731)

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/ByteArrayToStorageQueueMessageConverter.cs
MikeStall: In this specific case, this is an inconsistency and we should allow null and map it to "don't queue an object". 
I believe the errors are handled in both cases. In general, the declarative system lets us give rich errors. 

---
In reply to: [62360641](https://github.com/Azure/azure-webjobs-sdk/pull/711#discussion_r62360641) [](ancestors = 62360641)

MikeStall: in other words, we should still get the same (or comparable) errors at the same stage. 

---
In reply to: [63594695](https://github.com/Azure/azure-webjobs-sdk/pull/711#discussion_r63594695) [](ancestors = 63594695,62360641)

src/Microsoft.Azure.WebJobs.Host/Bindings/QueueBindingProvider.cs
```diff
@@ -0,0 +1,262 @@
...
+            var bindClient = bindingFactory.BindToExactAsyncType<QueueAttribute, IStorageQueue>(BuildClientFromQueueAttributeAsync, ToReadWriteParameterDescriptorForCollector, CollectAttributeInfo);
+            var bindSdkClient = bindingFactory.BindToExactAsyncType<QueueAttribute, CloudQueue>(BuildRealClientFromQueueAttributeAsync, ToReadWriteParameterDescriptorForCollector, CollectAttributeInfo);
+                        
+            var bindingProvider = new GenericCompositeBindingProvider<QueueAttribute>(
+                ValidateQueueAttribute, nameResolver, bindClient, bindSdkClient, bindAsyncCollector);
+
+            return bindingProvider;
+        }
+
+        // [Queue] has some legacy behavior where the storage account can be specified outside of the [Queue] attribute.
```
mathewc: It's not legacy behavior, and it's also not specific to Queue (it applies to all Storage bindings).

src/Microsoft.Azure.WebJobs.Host/Bindings/QueueBindingProvider.cs
```diff
@@ -0,0 +1,262 @@
...
+            _messageEnqueuedWatcherGetter = messageEnqueuedWatcherGetter;
+        }
+
+        public static IBindingProvider Build(
+            IStorageAccountProvider accountProvider,
+            IContextGetter<IMessageEnqueuedWatcher> messageEnqueuedWatcherGetter,
+            INameResolver nameResolver,
+            IConverterManager converterManager)
+        {
+            var closure = new QueueBindingProvider(accountProvider, messageEnqueuedWatcherGetter);
```
mathewc: As an outsider, this closure naming for a new binding provider is somewhat confusing :)

src/Microsoft.Azure.WebJobs.Host/Bindings/QueueBindingProvider.cs
```diff
@@ -0,0 +1,262 @@
...
+            StorageClientFactoryContext clientFactoryContext = new StorageClientFactoryContext
+            {
+                Parameter = parameter
+            };
+            IStorageQueueClient client = account.CreateQueueClient(clientFactoryContext);
+                    
+            return new ResolvedQueueAttribute(attrResolved.QueueName, client);
+        }
+
+        // Cloudqueue version. Can technically read from it. Preserves compat with older SDK.
```
mathewc: This comment doesn't seem to apply to this method? (at least I can't understand the comment at all :))

src/Microsoft.Azure.WebJobs.Host/Bindings/QueueBindingProvider.cs
```diff
@@ -0,0 +1,262 @@
...
+        }
+
+        private static IStorageQueue GetQueue(QueueAttribute attrResolved)
+        {
+            var attr = (ResolvedQueueAttribute)attrResolved;
+            IStorageQueue queue = attr.GetQueue();
+            return queue;
+        }
+
+        // Queue attributes are paired with a separate [StorageAccount].
```
mathewc: Queue attributes _can_ be paired ...

mathewc:
:ship:

src/Microsoft.Azure.WebJobs.Host/Bindings/QueueBindingProvider.cs
MikeStall: I'll reword.

---
In reply to: [64935628](https://github.com/Azure/azure-webjobs-sdk/pull/711#discussion_r64935628) [](ancestors = 64935628)

src/Microsoft.Azure.WebJobs.Host/Bindings/QueueBindingProvider.cs
MikeStall: I'll rename "legacy" to "pre-existing".  My point was that new extensions shouldn't follow this pattern because it doesn't jive well with Functions.  So new extensions won't need to do this extra hook. 

---
In reply to: [64934730](https://github.com/Azure/azure-webjobs-sdk/pull/711#discussion_r64934730) [](ancestors = 64934730)

MikeStall:
Commited to Master.

PR #717, opened 2016-05-09
Improve error handling for %% cases.

MikeStall:
Improvement to AttributeCloner. Resolve %% once upfront rather than on every invocation.
Add more tests.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/DefaultAttributeInvokerDescriptor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/DefaultAttributeInvokerDescriptor.cs
@@ -9,7 +9,10 @@ using Newtonsoft.Json;
 
 namespace Microsoft.Azure.WebJobs.Host.Bindings
 {
-    // Helpers for providing default implementations of IAttributeInvokeDescriptor
+    // Helpers for providing default behavior for an IAttributeInvokeDescriptor that 
+    // convert between a TAttribute and a string representation (invoke string). 
+    // Properties with [AutoResolve] are the interesting ones to serialize and deserialize. 
+    // Assume any property without a [AutoResolve] attribute is read-only and so doesn't need to be included in the invoke string. 
     internal static class DefaultAttributeInvokerDescriptor<TAttribute>
         where TAttribute : Attribute
     {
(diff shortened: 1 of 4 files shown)


src/Microsoft.Azure.WebJobs.Host/Bindings/DefaultAttributeInvokerDescriptor.cs
```diff
@@ -47,15 +37,19 @@ public TAttribute FromInvokeString(string invokeString)
             }
         }
 
-        public string ToInvokeString()
+        // Default implementation of IAttributeInvokeDescriptor which can be operate on any attribute and 
+        // follows basic conventions with a [ResolveProperty].
```
mathewc: Shouldn't [ResolveProperty] be [AutoResolve]?

src/Microsoft.Azure.WebJobs.Host/Bindings/DefaultAttributeInvokerDescriptor.cs
```diff
@@ -47,15 +37,19 @@ public TAttribute FromInvokeString(string invokeString)
             }
         }
 
-        public string ToInvokeString()
+        // Default implementation of IAttributeInvokeDescriptor which can be operate on any attribute and
```
mathewc: nit: this isn't an implementation of IAttributeInvokeDescriptor

src/Microsoft.Azure.WebJobs.Host/Bindings/DefaultAttributeInvokerDescriptor.cs
```diff
@@ -9,26 +9,16 @@
 
 namespace Microsoft.Azure.WebJobs.Host.Bindings
 {
-    // Default implementation of IAttributeInvokeDescriptor which can be operate on any attribute and 
-    // follows basic conventions with a [ResolveProperty]. 
-    // Assume any property without a [ResolveProperty] attribute is read-only. 
-    internal class DefaultAttributeInvokerDescriptor<TAttribute> : IAttributeInvokeDescriptor<TAttribute>
+    // Helpers for providing default implementations of IAttributeInvokeDescriptor
+    internal static class DefaultAttributeInvokerDescriptor<TAttribute>
```
mathewc: Seems like this type should be renamed since it isn't an InvokerDescriptor anymore

src/Microsoft.Azure.WebJobs.Host/Bindings/DefaultAttributeInvokerDescriptor.cs
MikeStall: good catch

---
In reply to: [62568238](https://github.com/Azure/azure-webjobs-sdk/pull/717#discussion_r62568238) [](ancestors = 62568238)

PR #723, opened 2016-05-13
Add validate hook for index-type validate.

MikeStall:
Validator hook is dispatched after the %% are resolved, but before any runtime invocation (so no { } resolved yet).

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Config/ExtensionConfigContextExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Config/ExtensionConfigContextExtensions.cs
@@ -1,0 +1,72 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Azure.WebJobs.Host.Bindings;
+
+namespace Microsoft.Azure.WebJobs.Host.Config
+{
+    /// <summary>
+    /// Helper Extension methods for extension configuration. 
+    /// </summary>
+    public static class ExtensionConfigContextExtensions
+    {
+        /// <summary>
+        /// Register a set of rules to handle binding to a given attribute. 
+        /// </summary>
+        /// <typeparam name="TAttribute">The attribute to bind to</typeparam>
+        /// <param name="context"></param>
+        /// <param name="validator">a validator function to invoke on the attribute before runtime. </param>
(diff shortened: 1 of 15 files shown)


mathewc:
Can you provide more context? What is the scenario here? I don't understand the above description.

MikeStall:
This is a hook to validate static things in the attribute and get early errors at index time (not waiting until runtime). For example, it may validate that the properties  conforms to some regex or are in some range.
________________________________

src/Microsoft.Azure.WebJobs.Host/Config/ExtensionConfigContextExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        /// <summary>
+        /// Register a set of rules to handle binding to a given attribute. 
+        /// </summary>
+        /// <typeparam name="TAttribute">The attribute to bind to</typeparam>
+        /// <param name="context"></param>
+        /// <param name="validator">a validator function to invoke on the attribute before runtime. </param>
+        /// <param name="bindingProviders">list of binding providers to handle this attribute. 
+        /// If none of these handle the attribute, an error is thrown at indexing. </param>
+        public static void RegisterBindingRules<TAttribute>(
```
mathewc: Related to my comments in the outstanding PR https://github.com/Azure/azure-webjobs-sdk/pull/711 I think we shouldn't use this "Rules" terminology, and instead call these what they are. E.g. RegisterBindingProviders.

mathewc: Also, I don't see any tests or usages of these new extension methods in this review, so it's hard to see usage context. I suggest either adding tests, or moving this into your other Queue PR and having one of the bindings use these extensions, so we can see the usage.

MikeStall: There are some tests in ValidationTests.cs

---
In reply to: [63563653](https://github.com/Azure/azure-webjobs-sdk/pull/723#discussion_r63563653) [](ancestors = 63563653)

MikeStall:
This is a hook to validate static things in the attribute and get early errors at index time (not waiting until runtime). For example, it may validate that the properties  conforms to some regex or are in some range.

---
In reply to: [219741200](https://github.com/Azure/azure-webjobs-sdk/pull/723#issuecomment-219741200) [](ancestors = 219741200)

mathewc:
closing, since I believe this is merged already?

PR #726, opened 2016-05-17
Add new binding rules.

MikeStall:
Add 2 new binding rules to help Bret. 
1. merge Type filter & Validator into 1 fule
2. new generic item so that we can skip IValueProvider.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
@@ -12,7 +12,7 @@ using Microsoft.Azure.WebJobs.Host.Triggers;
 namespace Microsoft.Azure.WebJobs.Host.Bindings
 {
     /// <summary>
-    /// Helper class for producing binding rules. 
+    /// Helper class for creating some generally useful BindingProviders
     /// </summary>
     public class BindingFactory
     {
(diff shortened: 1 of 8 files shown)


src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
```diff
@@ -47,6 +47,19 @@ public IConverterManager ConverterManager
         }
 
         /// <summary>
+        /// Creating a type filter predicate around another rule. Filter doubles as a validator if it throws.  
+        /// </summary>
+        /// <param name="predicate">type predicate. Only apply inner rule if this predicate as applied to the user parameter type is true. </param>
+        /// <param name="innerRule">Inner rule. </param>
+        /// <returns></returns>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Performance", "CA1822:MarkMembersAsStatic")]
```
mathewc: nit: can you add exclusions to the global file rather than inline in source? That's the pattern we generally practice across these repos.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
```diff
@@ -47,6 +47,19 @@ public IConverterManager ConverterManager
         }
 
         /// <summary>
+        /// Creating a type filter predicate around another rule. Filter doubles as a validator if it throws.  
+        /// </summary>
+        /// <param name="predicate">type predicate. Only apply inner rule if this predicate as applied to the user parameter type is true. </param>
```
mathewc: nit: Type

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
```diff
@@ -47,6 +47,19 @@ public IConverterManager ConverterManager
...
+        /// <param name="innerRule">Inner rule. </param>
```
mathewc: nit: extra space

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
```diff
@@ -55,7 +68,7 @@ public IConverterManager ConverterManager
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Performance", "CA1822:MarkMembersAsStatic")]
         public IBindingProvider AddTypeFilter(Func<Type, bool> predicate, IBindingProvider innerRule)
```
mathewc: I've commented on this in your other reviews - an IBindingProvider isn't a "rule" really. Hopefully we can change the naming across all this new stuff to match what the types/interfaces really are.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
```diff
@@ -79,7 +92,7 @@ public IBindingProvider AddValidator<TAttribute>(Action<TAttribute, Type> valida
         /// <typeparam name="TAttribute"></typeparam>
```
mathewc: nit: fill out all the xml comment sections here and throughout :)

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/FilteringBindingProvider.cs
```diff
@@ -26,7 +33,16 @@ public Task<IBinding> TryCreateAsync(BindingProviderContext context)
...
             Type parametertype = context.Parameter.ParameterType;
-            if (!_predicate(parametertype))
+
+            var attr = context.Parameter.GetCustomAttribute<TAttribute>();
+
+            var cloner = new AttributeCloner<TAttribute>(attr, _nameResolver);
+            var attrNameResolved = cloner.GetNameResolvedAttribute();
+
+            // This may do validation and throw too. 
+            bool useThisRule = _predicate(attrNameResolved, parametertype);
```
mathewc: nit: bool canBind?

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/GenericItemBindingProvider.cs
```diff
@@ -0,0 +1,70 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Reflection;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Protocols;
+
+namespace Microsoft.Azure.WebJobs.Host.Bindings
+{
+    internal class GenericItemBindingProvider<TAttribute> : IBindingProvider
```
mathewc: Can you add a descriptive xml block to this?

test/Microsoft.Azure.WebJobs.Host.UnitTests/Common/BindToGenericItemTests.cs
```diff
@@ -0,0 +1,81 @@
...
+                var bf = context.Config.BindingFactory;
+
+                // Add [Test] support                
+                var rule = bf.BindToGenericItem<TestAttribute>(Builder);
+                context.RegisterBindingRules<TestAttribute>(rule);
+            }
+
+            private Task<object> Builder(TestAttribute attrResolved, Type parameterType)
+            {
+                var method = parameterType.GetMethod("New", BindingFlags.Static | BindingFlags.Public);
```
mathewc: Why are you using reflection to create this instance? Why not just call a ctor on the type?

test/Microsoft.Azure.WebJobs.Host.UnitTests/Common/BindingProviderFilterTests.cs
```diff
@@ -0,0 +1,93 @@
...
+
+
+        public class FakeExtClient : IExtensionConfigProvider
+        {
+            public void Initialize(ExtensionConfigContext context)
+            {
+                var bf = context.Config.BindingFactory;
+
+                // Add [Test] support                
+                var rule = bf.BindToExactType<TestAttribute, string>(attr => attr.Path);
```
mathewc: I wonder if generally for people who have several filters/validations they need to do, it if won't be easier for them to just implement an IBindingProvider themselves, since in that case they can put whatever arbitrarily complex logic they want in their TryCreateAsync, w/o having to grok all these composition rules and extension methods that take Funcs?

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
brettsam: Do we still need AddTypeFilter and AddValidator? I was hoping we could go from 3 'filter' methods to 1 and reduce the concept count.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Common/BindToGenericItemTests.cs
MikeStall: Technically, I don't know the type.  I wanted the test to simulate the same situation that production could would have.  

---
In reply to: [63630342](https://github.com/Azure/azure-webjobs-sdk/pull/726#discussion_r63630342) [](ancestors = 63630342)

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
MikeStall: I have a strong preference for inlining them -  a) their ugly presence is a motivator to fix it; b) the global file easily gets out of sync with stale entries. 


---
In reply to: [63629010](https://github.com/Azure/azure-webjobs-sdk/pull/726#discussion_r63629010) [](ancestors = 63629010)

mathewc: My assumption is that you have no intention of fixing this. If you do, then make it static in this PR. Otherwise this is just code litter. That's why we generally shove all these nitpicky errors to the supression file. For things you plan on fixing fine, but generally all the violations are things we have no intention of ever fixing, so we don't need the code litter.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Common/BindingProviderFilterTests.cs
MikeStall: Why would they have multiple filters? They're passing a delegate so it can do whatever they need. 

---
In reply to: [63631087](https://github.com/Azure/azure-webjobs-sdk/pull/726#discussion_r63631087) [](ancestors = 63631087)

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/FilteringBindingProvider.cs
```diff
@@ -26,7 +33,16 @@ public Task<IBinding> TryCreateAsync(BindingProviderContext context)
...
+
+            var attr = context.Parameter.GetCustomAttribute<TAttribute>();
+
+            var cloner = new AttributeCloner<TAttribute>(attr, _nameResolver);
+            var attrNameResolved = cloner.GetNameResolvedAttribute();
+
+            // This may do validation and throw too. 
+            bool canBind = _predicate(attrNameResolved, parametertype);
+
+            if (!canBind)
```
brettsam: If I add several filters in a row -- shouldn't it try all the filters and only return null if none of them returned true?

i.e. -- if i want to support TypeA, TypeB, TypeC, I would think that I could do:

AddFilter(IsTypeA)
AddFilter(IsTypeB)
AddFilter(IsTypeC)

If the type is TypeB, I'd expect it to match

MikeStall: Do AddFilter( t => IsTypeA(t) || IsTypeB(t) || IsTypeC(t), ...) 

---
In reply to: [63770416](https://github.com/Azure/azure-webjobs-sdk/pull/726#discussion_r63770416) [](ancestors = 63770416)

brettsam: The method name 'AddFilter' does not really convey that to me. I'm guessing most people will read this as 'there's a list of filters that get run, and I'm adding to it'.

---
In reply to: [63771789](https://github.com/Azure/azure-webjobs-sdk/pull/726#discussion_r63771789) [](ancestors = 63771789,63770416)

PR #728, opened 2016-05-19
Fixing SingletonListener lock release bug

mathewc:
Fixing an issue reported by Hamid in his ApiHub trigger binding. The issue was that due to this bug, on error paths (e.g. listener fails to start due to invalid path), when the host is disposed, the singleton lock that was aquired as part of startup was not released.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/JobHostContext.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/JobHostContext.cs
@@ -78,7 +78,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
 
         public void Dispose()
         {
-            if (_disposed)
+            if (!_disposed)
             {
                 _listener.Dispose();
 
--- a/src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonListener.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonListener.cs
@@ -72,10 +72,7 @@ namespace Microsoft.Azure.WebJobs.Host.Listeners
                 LockTimer.Stop();
             }
 
-            if (_lockHandle != null)
-            {
-                await _singletonManager.ReleaseLockAsync(_lockHandle, cancellationToken);
-            }
+            await ReleaseLockAsync(cancellationToken);
 
             if (_isListening)
             {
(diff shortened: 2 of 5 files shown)


src/Microsoft.Azure.WebJobs.Host/Executors/JobHostContext.cs
```diff
@@ -78,7 +78,7 @@ public IAsyncCollector<FunctionInstanceLogEntry> FastLogger
 
         public void Dispose()
         {
-            if (_disposed)
```
mathewc: Doh! This error has been in the SDK since the early days (2014 at least: https://github.com/Azure/azure-webjobs-sdk/blob/95c5f4547749a89f69ee3a12c9c252fd3af51144/src/Microsoft.Azure.WebJobs.Host/Executors/JobHostContext.cs)

mathewc: I think this hasn't been a big issue noticed by users because generally when their Console app host is shutting down, they don't care about proper disposal since the process is going down anyways.

However in Azure Functions scenarios, we're starting/stopping hosts in memory and proper cleanup is very important.

src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonListener.cs
```diff
@@ -101,6 +98,10 @@ public void Dispose()
                 LockTimer.Dispose();
             }
 
+            // When we Dispose, it's important that we release the lock if we
+            // have it.
+            ReleaseLockAsync().Wait();
```
MikeStall: Could this result in deadlocks? Sync context, etc. 

We may need to do this:
  var theTask = Task.Run( ()=> ReleaseLockAsync());
  theTask.GetAwaiter().GetResult();

mathewc: Can you elaborate on how you think this might cause a deadlock? I don't see it.

MikeStall: In general, any call to .Wait() in can deadlock. This is a particularly common issue in ASP.Net due to the sync context.  :|  
The GetAwaiter() goo compensates for it. 
A good explanation is at: [http://blog.stephencleary.com/2012/07/dont-block-on-async-code.html](http://blog.stephencleary.com/2012/07/dont-block-on-async-code.html)

---
In reply to: [63921309](https://github.com/Azure/azure-webjobs-sdk/pull/728#discussion_r63921309) [](ancestors = 63921309)

safihamid:
🚢

src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonListener.cs
mathewc: Thanks. Now using GetAwaiter

PR #729, opened 2016-05-20
Don't invoke filter for AsyncCollector if type mismatches and we know…

MikeStall:
… we won't use the rule.


src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactoryHelpers.cs
```diff
@@ -73,6 +73,45 @@ internal static class BindingFactoryHelpers
...
+                else if (genericType == typeof(ICollector<>))
+                {
+                    return elementType;
+                }
+
+                return null;
+            }
+            else
+            {
+                if (parameterType.IsByRef)
```
brettsam: Should you check IsByRef first? I'd expect something like 'out GenericType<SomeType>' to return GenericType<SomeType> as the Type.

I *think* this would actually return null for that case.

MikeStall: It turns out this is correct. IsGenericType returns false for ByRef types. I added  a unit test to verify

brettsam:
:shipit:

PR #730, opened 2016-05-23
Add some regression tests for [Queue] attribute.

MikeStall:
No product changes. Added an internal method on JobHostConfiguration to let the test set the storage account.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
@@ -73,7 +73,7 @@ namespace Microsoft.Azure.WebJobs
             string value = ConfigurationUtility.GetSettingFromConfigOrEnvironment(Constants.EnvironmentSettingName);
             IsDevelopment = string.Compare(Constants.DevelopmentEnvironmentValue, value, StringComparison.OrdinalIgnoreCase) == 0;
         }
-
+   
         /// <summary>
         /// Gets a value indicating whether the <see cref="JobHost"/> is running in a Development environment.
         /// You can use this property in conjunction with <see cref="UseDevelopmentSettings"/> to default
@@ -268,7 +268,7 @@ namespace Microsoft.Azure.WebJobs
             }
         }
 
-        private IJobHostContextFactory ContextFactory
+        internal IJobHostContextFactory ContextFactory
         {
             get
             {
@@ -278,6 +278,11 @@ namespace Microsoft.Azure.WebJobs
                 }
 
                 return _contextFactory;
+            }
+            set
+            {
+                // Expose this for unit tests to override. 
+                _contextFactory = value;
             }
         }
 
--- a/src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/QueueAttributeBindingProvider.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/QueueAttributeBindingProvider.cs
@@ -70,6 +70,10 @@ namespace Microsoft.Azure.WebJobs.Host.Queues.Bindings
             }
 
             IStorageAccount account = await _accountProvider.GetStorageAccountAsync(context.Parameter, context.CancellationToken, _nameResolver);
+            if (account == null)
+            {
+                throw new InvalidOperationException("Can't bind Queue since no storage account is set.");
+            }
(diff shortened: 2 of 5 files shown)


src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
```diff
@@ -302,6 +302,12 @@ public StorageClientFactory StorageClientFactory
             }
         }
 
+        // Test hook to set storage account. 
+        internal void SetAccountProvider(IStorageAccountProvider provider)
```
mathewc: I don't think this method should be added. There is already a mechanism to do this, by setting the ContextFactory property. That's what other tests do. This doesn't feel like a pattern we want to promote - adding a new setter for every new thing a test wants to mock.

MikeStall: How do I set the _contextFactory? The setter is private.  Existing tests end up creating a whole new IServiceProvider (and skip JobHostConfiguration altogether) and go down a different path than prod. 
Alternatively - I could make the ContextFactory property internal?

mathewc: That sounds better




On Mon, May 23, 2016 at 1:50 PM -0700, "Mike Stall" <notifications@github.com<mailto:notifications@github.com>> wrote:


In src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs<https://github.com/Azure/azure-webjobs-sdk/pull/730#discussion_r64286416>:

> @@ -302,6 +302,12 @@ public StorageClientFactory StorageClientFactory
>              }
>          }
>
> +        // Test hook to set storage account.
> +        internal void SetAccountProvider(IStorageAccountProvider provider)


How do I set the _contextFactory? The setter is private. Existing tests end up creating a whole new IServiceProvider (and skip JobHostConfiguration altogether) and go down a different path than prod.
Alternatively - I could make the ContextFactory property internal?

-

MikeStall:
cr feedback applied and merged.

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/QueueAttributeBindingProvider.cs
```diff
@@ -70,6 +70,10 @@ internal class QueueAttributeBindingProvider : IBindingProvider
             }
 
             IStorageAccount account = await _accountProvider.GetStorageAccountAsync(context.Parameter, context.CancellationToken, _nameResolver);
+            if (account == null)
```
mathewc: If we're adding this validation, it would be good to do it consistently across all the other bindings.

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/QueueAttributeBindingProvider.cs
```diff
@@ -70,6 +70,10 @@ internal class QueueAttributeBindingProvider : IBindingProvider
...
+            {
+                throw new InvalidOperationException("Can't bind Queue since no storage account is set.");
```
mathewc: nit: "Unable to bind Queue because no storage account has been configured."

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Queues/QueueTests.cs
```diff
@@ -18,6 +21,158 @@ public class QueueTests
         private const string TriggerQueueName = "input";
         private const string QueueName = "output";
 
+        // Test binding to generics.
```
mathewc: Thanks for adding this :)

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Queues/QueueTests.cs
```diff
@@ -18,6 +21,158 @@ public class QueueTests
...
+
+            // Now peek at messages. 
+            var queue = account.CreateQueueClient().GetQueueReference(QueueName);
+            var msgs = queue.GetMessages(10).ToArray();
+
+            Assert.Equal(1, msgs.Length);
+            Assert.Equal("123", msgs[0].AsString);
+        }
+
+        public class Program2
```
mathewc: Can you please give these programs descriptive names rather than Program1/2/3 to make it clear what they're doing?

PR #731, opened 2016-05-24
Validate

MikeStall:
Have AttributeCloner enforce binding contract compatibility the same way that [Queue] does. Add unit test to catch it.

### DIFF
--- a/test/Microsoft.Azure.WebJobs.Host.UnitTests/AttributeClonerTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.UnitTests/AttributeClonerTests.cs
@@ -38,6 +38,34 @@ namespace Microsoft.Azure.WebJobs.Host.UnitTests
             public string ConstantProp { get; private set; }
         }
 
+        // Helper to easily generate a fixed binding contract. 
+        private static IReadOnlyDictionary<string, Type> GetBindingContract(params string[] names)
+        {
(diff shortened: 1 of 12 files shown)


mathewc:
Yes, this index time validation is very important

test/Microsoft.Azure.WebJobs.Host.UnitTests/AttributeClonerTests.cs
```diff
@@ -38,6 +38,34 @@ public Attr2(string resolvedProp2, string constantProp)
...
+
+        // Enforce binding contracts statically. 
+        [Fact]
+        public void BindingContractMismatch()
+        {
+            Attr1 a1 = new Attr1 { Path = "{name}" };
+
+            try
+            {
+                var cloner = new AttributeCloner<Attr1>(a1, EmptyContract);
```
mathewc: nit: The xunit Assert.Throws<InvalidOperationException> helper facilitates tests like these

mathewc:
One small comment. :ship:

test/Microsoft.Azure.WebJobs.Host.UnitTests/AttributeClonerTests.cs
MikeStall: I didn't use it because the helper won't let me verify the message contents.  

---
In reply to: [64429034](https://github.com/Azure/azure-webjobs-sdk/pull/731#discussion_r64429034) [](ancestors = 64429034)

PR #744, opened 2016-06-08
New logging API to track the "volume" of function throughput.  

MikeStall:
PR is ready - Math is tweaked and I added a bunch of unit tests for it.
Volume is the for getting "# of functions" x "size"
API is in WebJobs.Logging  and exposed through Dashboard APIs.
Since there can be millions of function instances, requires a background poller to aggregate.
"Size" comes from an env var.

### DIFF
--- a/src/Dashboard/ApiControllers/FunctionsController.cs
+++ b/src/Dashboard/ApiControllers/FunctionsController.cs
@@ -213,8 +213,9 @@ namespace Dashboard.ApiControllers
             // Return as parallel arrays since that's more efficient than an array of structs.
             var result = new
             {
-                Times = Array.ConvertAll(data, x => x.Item1),
-                Counts = Array.ConvertAll(data, x => x.Item2)
+                Times = Array.ConvertAll(data, x => x.Time),
+                Counts = Array.ConvertAll(data, x => x.Volume),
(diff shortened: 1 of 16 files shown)


src/Microsoft.Azure.WebJobs.Logging/Internal/LogReader.cs
```diff
@@ -26,6 +27,37 @@ public LogReader(CloudTable table)
...
+
+            IEnumerable<InstanceCountEntity> results = _instanceTable.ExecuteQuery(query);
+                        
+            int N = numberBuckets;
+            long startTicks = startTime.Ticks;
+            double bucketWidth = ((double)(endTime.Ticks - startTicks)) / N;
+
+            int[] values = new int[N];
+
+            foreach (var entity in results)
```
MikeStall: This math needs to be reworked to properly handle numberBuckets.

src/Dashboard/ApiControllers/FunctionsController.cs
```diff
@@ -202,6 +202,24 @@ public IHttpActionResult GetInvocationsForFunction(string functionId, [FromUri]P
             return Invocations(indexSegment);
         }
 
+        [Route("api/functions/volume")]
```
MikeStall: This is the new API that Sisira would consume.

Issue #750, opened 2016-06-20
BindingTemplate should support case insensitive bindings 


safihamid:
Hit the issue when trying to bind the 'Path' attribute of the file based SaaS providers. These paths similar to Windows file system are case insensitive but WebJobs SDK template binding currently only supports case sensitive paths.

Here is an example:

In Dropbox create a folder called 'test'
Then in WebJobs, create an ApiHubFileTrigger as:(with capital 'T' for 'Test' )
            [ApiHubFileTrigger("dropbox", @"Test/{name}", PollIntervalInSeconds = 5)] Stream input,

When a file is added to that Dropbox folder, we get an exception trying to call BindAsync.

PR #751, opened 2016-06-21
Support Case Insensitive Binding Templates (#750)

mathewc:


src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplate.cs
```diff
@@ -78,13 +78,25 @@ public static BindingTemplate FromString(string input)
...
         /// </exception>
-        public string Bind(IReadOnlyDictionary<string, string> parameters)
+        public string Bind(IReadOnlyDictionary<string, string> parameters, bool ignoreCase = false)
         {
             StringBuilder builder = new StringBuilder();
 
+            if (ignoreCase && parameters != null)
+            {
+                // convert to a case insensitive dictionary
+                var caseInsensitive = new Dictionary<string, string>(parameters.Count, StringComparer.OrdinalIgnoreCase);
```
davidebbo: Simpler:

    var caseInsensitive = new Dictionary<string, string>(parameters, StringComparer.OrdinalIgnoreCase);

mathewc: Yeah, I tried that first but it doesn't work because parameters is an IReadOnlyDictionary :)

davidebbo: Damn it

davidebbo:
:shipit:

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplate.cs
safihamid: Can we add the ignoreCase param to public static BindingTemplate FromString(string input) as well?
That's what we are using mostly in  extentions

mathewc: Done. See update.

safihamid:
🚢

Issue #653, opened 2016-03-11
Improve AppSetting/EnvironmentVar resolution


brettsam:
Several configurations are using nearly identical code to look up replacements in app settings first, and then environment variables. This should be provided automatically -- either through INameResolver or another mechanism.

brettsam:
PR #760 addresses this by removing the need to perform lookups yourself. Instead, use the NameResolver.

PR #760, opened 2016-06-30
Adding support for Settings in AttributeCloner; Changing default INam…

brettsam:
…eResolver

Some discussion here: https://github.com/Azure/azure-webjobs-sdk-script/issues/454

This enables you to identify a 'Setting' (which is a value that we look up directly from the INameResolver -- without expecting %%) by using `[AutoResolve(IsSetting = true)]` on any property of your extension attribute. This will call INameResolver.Resolve with this value and throw if it is not found -- similar to how we throw if a %% replacement is not found.

I've also removed the DefaultNameResolver (which simply threw) with one that looks at AppConfig, then environment variables. This is technically a breaking change -- although if anyone used our default one they would have been seeing exceptions whenever they called it.


src/Microsoft.Azure.WebJobs.Host/Indexers/AppSettingOrEnvironmentNameResolver.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Host.Indexers
+{
+    /// <summary>
+    /// An implementation of <see cref="INameResolver"/> that resolves tokens by looking first
+    /// in App Settings and then in environment variables.
+    /// </summary>
+    internal class AppSettingOrEnvironmentNameResolver : INameResolver
```
mathewc: nit: Why not just keep the default resolver named DefaultNameResolver?

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
```diff
@@ -65,15 +66,15 @@ public JobHostConfiguration(string dashboardAndStorageConnectionString)
 
             AddService<IExtensionRegistry>(extensions);
             AddService<StorageClientFactory>(new StorageClientFactory());
-            AddService<INameResolver>(new DefaultNameResolver());
+            AddService<INameResolver>(new AppSettingOrEnvironmentNameResolver());
```
mathewc: So this is the default, but if someone sets JobHostConfiguration.NameResolver themselves, they'd override this, right?

mathewc:
A few comments, otherwise looks good. :ship:

mathewc:
When you absorb this in Functions, you'll remove our default name resolver? I'n the case of Functions, we actually want to derive from the SDK DefaultNameResolver to extend. Customers will want to do that as well. Recommend making DefaultNameResolver public.

src/Microsoft.Azure.WebJobs.Host/Indexers/AppSettingOrEnvironmentNameResolver.cs
brettsam: Sure, that makes sense. I was trying to be descriptive, but the comments should take care of that. Will also make public as you mentioned elsewhere.

---
In reply to: [69617436](https://github.com/Azure/azure-webjobs-sdk/pull/760#discussion_r69617436) [](ancestors = 69617436)

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
brettsam: Correct.

---
In reply to: [69617747](https://github.com/Azure/azure-webjobs-sdk/pull/760#discussion_r69617747) [](ancestors = 69617747)

brettsam:
Yeah, we should derive from this. I'll make it public as well.

---
In reply to: [230574600](https://github.com/Azure/azure-webjobs-sdk/pull/760#issuecomment-230574600) [](ancestors = 230574600)

brettsam:
This addresses #653.

Issue #685, opened 2016-03-30
Shared Blob listener needs to be Singleton


mathewc:
The new shared blob listener strategy polls containers and enqueues messages for blobs that need to be processed. This listener should only be running on one scaled out instance to reduce storage account load. Desired scale out for blob processing will still be achieved because each instance is monitoring the queue.

mathewc:
Fixed by https://github.com/Azure/azure-webjobs-sdk/pull/762

PR #762, opened 2016-07-06
Making the shared blob listener a Singleton (#685)

mathewc:
Done in collaboration with @shobak101  (his work was in branch https://github.com/Azure/azure-webjobs-sdk/tree/singletonListener)

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/BlobListenerFactory.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/BlobListenerFactory.cs
@@ -30,6 +30,7 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
         private readonly IStorageBlobContainer _container;
         private readonly IBlobPathSource _input;
         private readonly ITriggeredFunctionExecutor _executor;
+        private readonly SingletonManager _singletonManager;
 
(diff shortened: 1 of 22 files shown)


mathewc:
I'm still planning on looking to see if any new tests can/should be added

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/BlobListenerFactory.cs
```diff
@@ -159,13 +167,19 @@ internal class BlobListenerFactory : IListenerFactory
...
                 BlobETagReader.Instance, new BlobReceiptManager(blobClient),
                 new BlobTriggerQueueWriter(hostBlobTriggerQueue, messageEnqueuedWatcher));
+
             await sharedBlobListener.RegisterAsync(_container, triggerExecutor, cancellationToken);
-            return new BlobListener(sharedBlobListener);
+
+            // we only want the blob discovery/scan listener to be running on a single instance,
+            // so we make it a singleton
+            var blobListener = new BlobListener(sharedBlobListener);
+            return singletonManager.CreateHostSingletonListener(blobListener, "Blobs");
```
mathewc: This will result in a host level lock blob of the form **4496ba9d19da4040aaf65fdf92690e8a/Blob.Listener**

Issue #680, opened 2016-03-30
add unique user agent to all storage requests to be able to get logs server side


shobak101:
to monitor and get insights about customer trends in using storage related operations (listening on blobs/ queues - blob outputs) and basically all storage requests that we control, we need to send a unique user agent header. This way we can collect metrics and logs from Storage server side.

mathewc:
Can you please provide a full description of this work item?

mathewc:
Regarding the PR for this issue: Since this is a global static setting, rather than doing it explicitly via StorageBlobClient/StorageQueueClient, I think you should instead do it in a single place on startup, e.g. via a JobHost static ctor or something. Note that we also want this to happen for Table storage as well, so doing it once centrally makes sense, rather than doing for each storage service.

mathewc:
Also, I think you should remove the hyphen from the agent string "-AzureWebJobs" and use a space delimiter instead.

mathewc:
This work is done.

PR #763, opened 2016-07-08
add user agent to storage requests listing blobs and queue messages f…

shobak101:
…ixes #680 
sending this to start a conversation on where this duplicated code should go!

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Storage/Blob/StorageBlobClient.cs
+++ b/src/Microsoft.Azure.WebJobs.Storage/Blob/StorageBlobClient.cs
@@ -58,6 +58,16 @@ namespace Microsoft.Azure.WebJobs.Host.Storage.Blob
             BlobListingDetails blobListingDetails, int? maxResults, BlobContinuationToken currentToken,
             BlobRequestOptions options, OperationContext operationContext, CancellationToken cancellationToken)
         {
+            // Add WebJobs to user agent to make it easier to collect usage information from Storage telemetry
+            if (operationContext == null)
+            {
+                operationContext = new OperationContext();
+            }
+            operationContext.SendingRequest += (sender, e) =>
+            {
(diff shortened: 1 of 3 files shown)


src/Microsoft.Azure.WebJobs.Storage/Blob/StorageBlobClient.cs
```diff
@@ -58,6 +58,16 @@ public Task<ServiceProperties> GetServicePropertiesAsync(CancellationToken cance
             BlobListingDetails blobListingDetails, int? maxResults, BlobContinuationToken currentToken,
             BlobRequestOptions options, OperationContext operationContext, CancellationToken cancellationToken)
         {
+            // Add WebJobs to user agent to make it easier to collect usage information from Storage telemetry
+            if (operationContext == null)
```
mathewc: Is there a way we can specify this globally on the clients themselves (CloudBlobClient/CloudQueueClient/etc.). E.g some client level event, or perhaps via https://msdn.microsoft.com/en-us/library/microsoft.windowsazure.storage.blob.cloudblobclient.setserviceproperties(v=azure.10).aspx?

src/Microsoft.Azure.WebJobs.Storage/Blob/StorageBlobClient.cs
```diff
@@ -58,6 +58,16 @@ public Task<ServiceProperties> GetServicePropertiesAsync(CancellationToken cance
...
+            {
+                operationContext = new OperationContext();
+            }
+            operationContext.SendingRequest += (sender, e) =>
+            {
+                e.Request.UserAgent += "-WebJobsBlob";
```
mathewc: Can we use a single agent identifier like AzureWebJobs? Not sure we need to have Blob/Queue variants.

PR #764, opened 2016-07-09
revisiting storage blobs and tables naming validation 

shobak101:


src/Microsoft.Azure.WebJobs.Host/Blobs/BlobClient.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs
     internal static class BlobClient
     {
         // Tested against storage service on Jan 2014. All other unsafe and reserved characters work fine.
-        private static readonly char[] UnsafeBlobNameCharacters = { '\\', '[', ']' };
```
mathewc: Please update the outdated comment.

src/Microsoft.Azure.WebJobs.Host/Tables/TableClient.cs
```diff
@@ -110,7 +110,26 @@ public static void ValidateAzureTableName(string tableName)
 
         public static bool IsValidAzureTableName(string tableName)
         {
-            return Regex.IsMatch(tableName, "^[A-Za-z][A-Za-z0-9]{2,62}$");
+            // Table Name Rules are summerized in this page https://msdn.microsoft.com/en-us/library/azure/dd179338.aspx
+            // Table names may contain only alphanumeric characters.
+            if (!Regex.IsMatch(tableName, "^[A-Za-z][A-Za-z0-9]{2,62}$"))
```
mathewc: The above regex seems to already capture the rules you added below.

mathewc:
Fixes: https://github.com/Azure/azure-webjobs-sdk-script/issues/410

src/Microsoft.Azure.WebJobs.Host/Tables/TableClient.cs
shobak101: actually you are right I misinterpreted it, reverting that change and will only keep the comment referencing the naming rules

mathewc: you can remove the if block, but keep your new comment

mathewc:
:ship:

PR #767, opened 2016-07-13
Adding built-in rand-guid binding parameter support

mathewc:
Moving the `rand-guid` support we added to Functions into the core SDK. This serves two goals:

1) Make it available to all users
2) Makes it a per-invocation resolution, so it works for early bound functions as well. In Functions currently, it only works for late bound functions (i.e. non-C#)

### DIFF
--- a/test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
@@ -46,9 +46,11 @@ namespace Microsoft.Azure.WebJobs.Host.EndToEndTests
         private static TimeSpan _timeoutJobDelay;
 
         private readonly CloudQueue _testQueue;
+        private readonly TestFixture _fixture;
 
         public AsyncChainEndToEndTests(TestFixture fixture)
         {
+            _fixture = fixture;
             _resolver = new RandomNameResolver();
             _hostConfig = new JobHostConfiguration()
             {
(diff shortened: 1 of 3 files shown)


test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
```diff
@@ -222,6 +224,42 @@ public void FunctionFailures_LogsExpectedTraceEvent()
...
+                host.Call(methodInfo, arguments);
+            }
+
+            // We expect 3 separate blobs to have been written
+            var blobs = container.ListBlobs().Cast<CloudBlockBlob>().ToArray();
+            Assert.Equal(3, blobs.Length);
+            foreach (var blob in blobs)
+            {
+                string content = blob.DownloadText(Encoding.UTF8);
+                int blobInt = int.Parse(content.Trim(new char[] { '\uFEFF', '\u200B' }));
```
brettsam: nit: could you also use Guid.TryParse() to make sure the blob name is a valid Guid? -- just to make sure nothing's getting mangled

brettsam:
:shipit:

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
mathewc: It's an int, not a guid. The Trim is me removing the BOM

brettsam: I'm suggesting you check the *name* of the blob. Isn't that what you're using {rand-guid} to replace? The content of the blob is an int, but the name should be a guid.

---
In reply to: [70699572](https://github.com/Azure/azure-webjobs-sdk/pull/767#discussion_r70699572) [](ancestors = 70699572)

mathewc: got it. Yep, I'll do that.

Issue #768, opened 2016-07-13
Invalid Queue message should not bring down the host


brettsam:
A customer reported an issue in the Script repo: https://github.com/Azure/azure-webjobs-sdk-script/issues/434. The issue there was that the message they were putting into the queue was bad. However, we shouldn't bring down the host when this happens.

An easy way to repro this 'bad' message with a python script is in the issue linked above.

**Notes from our mail discussion:**

The underlying issue in the message is that it’s been inserted with a type of Base64Encoded, but the raw string is not Base64. So when the Storage SDK tries to decode it, it throws the FormatException. 

Also, I think I mentioned it before, but the message is never put in the poison queue – I’m guessing b/c the host crashes and we never do the CompleteProcessingMessageAsync step, which checks the dequeue count and moves it. I see some messages in there with 90+ dequeue counts (I inserted them yesterday during my testing).

Exception Details:
The input is not a valid Base-64 string as it contains a non-base 64 character, more than two padding characters, or an illegal character among the padding characters.
```
   at System.Convert.FromBase64_Decode(Char* startInputPtr, Int32 inputLength, Byte* startDestPtr, Int32 destLength)
   at System.Convert.FromBase64CharPtr(Char* inputPtr, Int32 inputLength)
   at System.Convert.FromBase64String(String s)
   at Microsoft.WindowsAzure.Storage.Queue.CloudQueueMessage.get_AsString() in c:\Program Files (x86)\Jenkins\workspace\release_dotnet_master\Lib\Common\Queue\CloudQueueMessage.Common.cs:line 208
   at Microsoft.Azure.WebJobs.Host.Storage.Queue.StorageQueueMessage.get_AsString() in C:\git\azure-webjobs-sdk\src\Microsoft.Azure.WebJobs.Storage\Queue\StorageQueueMessage.cs:line 44
```
The original exception is thrown here: https://github.com/Azure/azure-webjobs-sdk/blob/b1fb20b0edf0530d62127d6dd9f50cc9345ca7b6/src/Microsoft.Azure.WebJobs.Storage/Queue/StorageQueueMessage.cs#L44.

The origination of the call (not shown above) is here: https://github.com/Azure/azure-webjobs-sdk/blob/master/src/Microsoft.Azure.WebJobs.Host/Queues/Listeners/QueueListener.cs#L254

[...]

PR #769, opened 2016-07-14
improving queue trigger handling of invalid messages

brettsam:
Addresses #768

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Queues/StorageQueueMessageExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Queues/StorageQueueMessageExtensions.cs
@@ -27,6 +27,10 @@ namespace Microsoft.Azure.WebJobs.Host.Queues
                 // when the message type is binary, AsString will throw if the bytes aren't valid UTF-8.
                 value = null;
             }
+            catch (FormatException)
+            {
+                value = null;
+            }
 
             return value;
         }
--- a/test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AzureStorageEndToEndTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AzureStorageEndToEndTests.cs
@@ -3,8 +3,11 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
+using System.Reflection;
 using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host.TestCommon;
 using Microsoft.WindowsAzure.Storage;
 using Microsoft.WindowsAzure.Storage.Blob;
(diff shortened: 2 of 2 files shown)


mathewc:
I don't think we want to do the change in QueueProcessor. We don't want to automatically move these messages to poison queue. E.g. We want to support binding to CloudQueueMessage rather than string - in this case that should work. I.e. if the user isn't trying to bind as a string, we shouldn't be enforcing that its a string, or doing an AsString.

I think we need to simply make our checks in [StorageQueueMessageExtensions](https://github.com/Azure/azure-webjobs-sdk/blob/master/src/Microsoft.Azure.WebJobs.Host/Queues/StorageQueueMessageExtensions.cs#L25) more robust. We should handle the FormatException there as well. That will let the message go through, and it will get normal poison queue handling.

brettsam:
Good point about letting CloudQueueMessage get through; I hadn't tested that. I think you're right. Handling it there seems to throw an exception within the function invocation (likely in a converter), which then triggers the proper error and poison-queue flow. Let me play with it a bit more and I'll submit a revised PR.

mathewc:
I do like your idea of having a way of identifying a message like this that will **never** succeed - in such cases, doing retries will never work. However I think that is a potential improvement that would require more thought on the right way to do it across the board (e.g. blob trigger has poison queue handling as well). So lets keep this one scoped.

brettsam:
The code change is simpler now. I ended up writing an e2e test since I wanted to ensure that this worked through the whole pipeline -- that CloudQueueMessage params got the message and that string params threw and eventually went to the poison queue.

src/Microsoft.Azure.WebJobs.Host/Queues/StorageQueueMessageExtensions.cs
```diff
@@ -27,6 +27,10 @@ public static string TryGetAsString(this IStorageQueueMessage message)
                 // when the message type is binary, AsString will throw if the bytes aren't valid UTF-8.
                 value = null;
             }
+            catch (FormatException)
```
mathewc: nit: 

```catch (Exception ex)
{
    if (ex is DecoderFallbackException || ex is FormatException)
    {
        value = null;
        return;
    }
    else
    {
        throw;
    }
}
```

mathewc:
:ship:

src/Microsoft.Azure.WebJobs.Host/Queues/StorageQueueMessageExtensions.cs
fabiocav: Exception filters clean the proposed code, but we'd need to be able to use C# 6 in the project :)

mathewc: Yeah, we can't do that yet until we move the SDK and Dashboard projects from TeamCity to AppVeyor. If somebody magically made that happen while I'm away on vacation, I'd buy them a case of beer :)

PR #777, opened 2016-07-27
Add user agent to storage requests for telemetry fixes #680

### LINKED ISSUE #680: add unique user agent to all storage requests to be able to get logs server side

shobak101:
checked with Storage client team and there is no way to set this on the whole client, so added the call for list blobs and get queue messages

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Storage/Blob/StorageBlobClient.cs
+++ b/src/Microsoft.Azure.WebJobs.Storage/Blob/StorageBlobClient.cs
@@ -27,6 +27,12 @@ namespace Microsoft.Azure.WebJobs.Host.Storage.Blob
     {
         private readonly CloudBlobClient _sdk;
 
+        static StorageBlobClient()
+        {
+            // Add WebJobs to user agent to make it easier to collect usage information from Storage telemetry
+            OperationContextInitializer.Initialize();
+        }
+
         /// <summary>Initializes a new instance of the <see cref="StorageBlobClient"/> class.</summary>
         /// <param name="sdk">The SDK client to wrap.</param>
         public StorageBlobClient(CloudBlobClient sdk)
--- a/src/Microsoft.Azure.WebJobs.Host/WebJobs.Host.csproj
+++ b/src/Microsoft.Azure.WebJobs.Host/WebJobs.Host.csproj
@@ -332,6 +332,9 @@
     <Compile Include="..\Microsoft.Azure.WebJobs.Storage\StorageExceptionExtensions.cs">
       <Link>Storage\StorageExceptionExtensions.cs</Link>
     </Compile>
+    <Compile Include="..\Microsoft.Azure.WebJobs.Storage\OperationContextInitializer.cs">
+      <Link>Storage\OperationContextInitializer.cs</Link>
+    </Compile>
     <Compile Include="..\Microsoft.Azure.WebJobs.Storage\Table\IEntityResolver.cs">
       <Link>Storage\Table\IEntityResolver.cs</Link>
     </Compile>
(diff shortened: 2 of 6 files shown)


fabiocav:
@brettsam I'll take a look at this, but could you help and provide some feedback?

src/Microsoft.Azure.WebJobs.Storage/OperationContextHelper.cs
```diff
@@ -0,0 +1,37 @@
...
+            context.SendingRequest += (sender, e) =>
+            {
+                e.Request.UserAgent += userAgent;
+            };
+
+            return context;
+        }
+
+        /// <summary>Returns a default operation context.</summary>
+        public static OperationContext GetDefaultOperationContext()
```
fabiocav: Doesn't look like this method is used anywhere. Is it needed?

fabiocav:
Added a minor comment about an unused method, but looks good.

Have you considered using the `GlobalSendingRequest` event? If we want to have that header sent with every request made to storage, would that be a better option (and also, potentially prevent issues down the road where we may forget to call the helper to add the event handler)?

src/Microsoft.Azure.WebJobs.Storage/OperationContextHelper.cs
```diff
@@ -0,0 +1,37 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.WindowsAzure.Storage;
+
+#if PUBLICSTORAGE
+namespace Microsoft.Azure.WebJobs.Storage
+#else
+namespace Microsoft.Azure.WebJobs.Host.Storage
+#endif
```
davidebbo: Why the conditional namespace?

src/Microsoft.Azure.WebJobs.Storage/OperationContextHelper.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    internal class OperationContextHelper
+    {
+        /// <summary>Adds handler to manipulate User Agent for Requests</summary>
+        public static OperationContext AddWebJobsUserAgentToOperationContext(OperationContext context)
+        {
+            var userAgent = "-AzureWebJobs";
```
davidebbo: I'd just inline this where it's used.

src/Microsoft.Azure.WebJobs.Storage/Blob/StorageBlobClient.cs
```diff
@@ -58,6 +58,9 @@ public Task<ServiceProperties> GetServicePropertiesAsync(CancellationToken cance
             BlobListingDetails blobListingDetails, int? maxResults, BlobContinuationToken currentToken,
             BlobRequestOptions options, OperationContext operationContext, CancellationToken cancellationToken)
         {
+            // Add WebJobs to user agent to make it easier to collect usage information from Storage telemetry
+            operationContext = OperationContextHelper.AddWebJobsUserAgentToOperationContext(operationContext);
```
brettsam: >operationContext = OperationContextHelper.AddWebJobsUserAgentToOperationContext(operationContext); [](start = 12, length = 98)

Could you make this an extension method so the call was something like 'operationContext.AddWebJobsUserAgent();'?

brettsam:
I was just about to ask whether we'd need to do this in Extensions as well (Timer uses storage, for example). If GlobalSendingRequest would handle that for every storage operation coming from the function app, that would be nice.

---
In reply to: [235451342](https://github.com/Azure/azure-webjobs-sdk/pull/777#issuecomment-235451342) [](ancestors = 235451342)

shobak101:
@brettsam I just tested Fabio's suggestion and it works, all calls to storage have this -AzureWebJobs added to the useragent, will send a new PR now with the latest changes

fabiocav:
Looks good. A minor thing that might be helpful is to add a comment to the `Initialize` method body to explain that it is a noop and exists to expose a convenient way to trigger the static initialization. That might give others some context and help avoid confusion (or the temptation to remove it :) ) 
Aside from that, good to :shipit: !

shobak101:
added the comment and pushed to master

PR #799, opened 2016-08-23
Optionally raise FunctionTimeoutExceptions during a timeout

brettsam:
A couple of ideas that allow Scripting to handle Timeouts more nicely...


brettsam:
Ready for review....

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -161,6 +165,12 @@ public HostOutputMessage HostOutputMessage
                 await _functionInstanceLogger.DeleteLogFunctionStartedAsync(functionStartedMessageId, cancellationToken);
             }
 
+            // We want to immediately raise any fatal exceptions
+            if (exceptionInfo != null && exceptionInfo.SourceException.IsFatal())
```
mathewc: Can SourceException ever be null? Might guard against that

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -499,11 +521,22 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
-            var timer = StartFunctionTimeout(instance, globalFunctionTimeout, functionCancellationTokenSource, traceWriter);
+            var timer = StartFunctionTimeout(instance, globalFunctionTimeout, timeoutWhileDebugging, functionCancellationTokenSource, traceWriter);
             try
             {
                 await invoker.InvokeAsync(invokeParameters);
             }
+            catch (OperationCanceledException)
+            {
+                // if cancellationToken is not signaled, it means this is a timeout
+                if (throwOnFunctionTimeout && !cancellationToken.IsCancellationRequested)
```
mathewc: This feels a little fragile. Is the issue that you can't throw the exception from the background timer thread? How about using another mechanism for determining whether the timer has expired, e.g. subscribing to the Elapsed with a local delegate that sets a flag you can check?

brettsam: Right, we need to throw here so the rest of the logging, etc, happens before we rethrow the 'fatal' exception. Having the timer delegate throw means the exception is coming from some other thread.

Good idea on the additional delegate. I'll do that.

---
In reply to: [76653248](https://github.com/Azure/azure-webjobs-sdk/pull/799#discussion_r76653248) [](ancestors = 76653248)

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
```diff
@@ -164,6 +167,17 @@ public string StorageConnectionString
         /// </summary>
         public TimeSpan? FunctionTimeout { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value indicating whether function invocations will timeout when
+        /// a <see cref="FunctionTimeout"/> is specified and a debugger is attached. False by default.
+        /// </summary>
+        public bool TimeoutWhileDebugging { get; set; }
```
mathewc: How about moving all 3 of these related settings into a new FunctionTimeoutConfiguration class, and exposing it via JobHostConfiguration.FunctionTimeout property?

Then I'd have the new config class have properties: DefaultTimeout, ThrowOnTimeout and IgnoreWhileDebugging

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
```diff
@@ -164,6 +167,17 @@ public string StorageConnectionString
...
+
+        /// <summary>
+        /// When true, an exception is thrown when a function timeout expires.
+        /// </summary>
+        public bool ThrowExceptionOnFunctionTimeout { get; set; }
```
mathewc: When using this with the cooperative model as well (CancellationToken) how long after we signal cancellation will we throw? I.e., in this model what is the graceful shutdown period?

src/Microsoft.Azure.WebJobs.Host/Timers/BackgroundExceptionDispatcher.cs
```diff
@@ -7,20 +7,16 @@
 
 namespace Microsoft.Azure.WebJobs.Host.Timers
 {
-    internal class BackgroundExceptionDispatcher : IBackgroundExceptionDispatcher
+    /// <summary>
+    /// Default implementation for <see cref="IWebJobExceptionHandler"/>. 
+    /// </summary>
+    public class WebJobExceptionHandler : IWebJobExceptionHandler
```
mathewc: nit: I'd use WebJobs (plural) in these names

src/Microsoft.Azure.WebJobs.Host/Timers/TaskSeriesTimer.cs
```diff
@@ -21,7 +21,7 @@ internal sealed class TaskSeriesTimer : ITaskSeriesTimer
         private Task _run;
         private bool _disposed;
 
-        public TaskSeriesTimer(ITaskSeriesCommand command, IBackgroundExceptionDispatcher backgroundExceptionDispatcher,
+        public TaskSeriesTimer(ITaskSeriesCommand command, IWebJobExceptionHandler backgroundExceptionDispatcher,
```
mathewc: Once the interface name is finalized, can you change the parameter names to match throughout all these files?

brettsam: Yeah, there's a bunch of these. Even though it's mostly search/replace, I didn't want to have to re-do it, so I figured I'd wait until we finalized it.

---
In reply to: [76664040](https://github.com/Azure/azure-webjobs-sdk/pull/799#discussion_r76664040) [](ancestors = 76664040)

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
brettsam: At the moment there's no grace period by default (but we also don't throw by default). In the Script repo I do this waiting in the ExceptionHandler -- and it's 2 seconds -- which is just an arbitrary starting point that I picked.

I could do something similar in the default ExceptionHandler. We could also have that grace period be configurable.

---
In reply to: [76662449](https://github.com/Azure/azure-webjobs-sdk/pull/799#discussion_r76662449) [](ancestors = 76662449)

mathewc: Yeah, I think we need to do something, otherwise mixing the two models doesn't work well. If I've implemented cancellation tokens, I'd want those to be cancelled giving me a chance to clean up before the process goes down.

src/Microsoft.Azure.WebJobs.Host/Extensions/ExceptionExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace System
+{
+    internal static class ExceptionExtensions
+    {
+        public static bool IsFatal(this Exception exception)
+        {
+            while (exception != null)
+            {
+                if ((exception as OutOfMemoryException != null && exception as InsufficientMemoryException == null) ||
```
fabiocav: The first check/cast is not needed. With the current logic, you'd only need to check for `InsufficientMemoryException`. I'm assuming what we really want is to interpret `OutOfMemoryException` as fatal, in which case, the second check is unnecessary.

fabiocav: We should also be able to replace simplify this by using `is` instead.

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
brettsam: Check out my latest commit. I've given it some quick testing from Script and looks good. I pushed the grace period into the core SDK and made it configurable.

---
In reply to: [76665848](https://github.com/Azure/azure-webjobs-sdk/pull/799#discussion_r76665848) [](ancestors = 76665848)

src/Microsoft.Azure.WebJobs.Host/Extensions/ExceptionExtensions.cs
brettsam: Yeah, I thought the same. I didn't touch it (other than formatting) after copying from the link that Mathew sent -- [http://vasters.com/archive/Are-You-Catching-Falling-Knives.html](http://vasters.com/archive/Are-You-Catching-Falling-Knives.html)

I'll swap all of this to use 'is'.

---
In reply to: [76674918](https://github.com/Azure/azure-webjobs-sdk/pull/799#discussion_r76674918) [](ancestors = 76674918)

brettsam: InsufficientMemoryException is an OutOfMemoryException

So we want to throw on OOM but not IME.

---
In reply to: [76674693](https://github.com/Azure/azure-webjobs-sdk/pull/799#discussion_r76674693) [](ancestors = 76674693)

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -499,11 +546,23 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
             try
             {
                 await invoker.InvokeAsync(invokeParameters);
             }
+            catch (OperationCanceledException)
+            {
+                // if cancellationToken is not signaled, it means this is a timeout
+                if (timeoutConfig.ThrowExceptionOnFunctionTimeout && hasTimerElapsed)
+                {
+                    string errorMessage = string.Format("Timeout while executing function: {0}", instance.FunctionDescriptor.ShortName);
```
MikeStall: nit: Could this message include a) how long we ran; b) what the timeout threshold is?  In lots of cases, we only have an exception msg, and that could be critical info.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/BlobListenerFactory.cs
```diff
@@ -35,7 +35,7 @@ internal class BlobListenerFactory : IListenerFactory
 
         public BlobListenerFactory(IHostIdProvider hostIdProvider,
             IQueueConfiguration queueConfiguration,
-            IBackgroundExceptionDispatcher backgroundExceptionDispatcher,
+            IWebJobsExceptionHandler backgroundExceptionDispatcher,
```
mathewc: nit: fix name

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/SharedBlobListener.cs
```diff
@@ -22,7 +22,7 @@ internal sealed class SharedBlobListener : ISharedListener
         private bool _disposed;
 
         public SharedBlobListener(IStorageAccount storageAccount,
-            IBackgroundExceptionDispatcher backgroundExceptionDispatcher)
+            IWebJobsExceptionHandler backgroundExceptionDispatcher)
```
mathewc: nit: fix name all the other places as well

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -161,9 +161,38 @@ public HostOutputMessage HostOutputMessage
                 await _functionInstanceLogger.DeleteLogFunctionStartedAsync(functionStartedMessageId, cancellationToken);
             }
 
+            await ThrowIfTimeoutOrFatal(exceptionInfo, _timeoutConfig.TimeoutGracePeriod, _backgroundExceptionDispatcher);
+
             return exceptionInfo != null ? new ExceptionDispatchInfoDelayedException(exceptionInfo) : null;
         }
 
+        internal static async Task ThrowIfTimeoutOrFatal(ExceptionDispatchInfo exceptionInfo, TimeSpan timeoutGracePeriod, IWebJobsExceptionHandler exceptionHandler)
```
mathewc: nit: I'd rename this HandleExceptionAsync, and only call it above if exceptionInfo != null

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -475,9 +521,10 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
 
         internal static async Task ExecuteWithWatchersAsync(IFunctionInstance instance,
             IReadOnlyDictionary<string, IValueProvider> parameters,
-            TimeSpan? globalFunctionTimeout,
+            JobHostFunctionTimeoutConfiguration timeoutConfig,
```
mathewc: Your call on whether you want to name this with the "JobHost" prefix or remove that, as I've done with SingletonConfiguration

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -499,11 +546,23 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
-            var timer = StartFunctionTimeout(instance, globalFunctionTimeout, functionCancellationTokenSource, traceWriter);
+            bool hasTimerElapsed = false;
+            var timer = StartFunctionTimeout(instance, timeoutConfig.Timeout, timeoutConfig.TimeoutWhileDebugging, functionCancellationTokenSource, traceWriter, () => hasTimerElapsed = true);
             try
             {
                 await invoker.InvokeAsync(invokeParameters);
             }
+            catch (OperationCanceledException)
+            {
+                // if cancellationToken is not signaled, it means this is a timeout
```
mathewc: remove obsolete comment

src/Microsoft.Azure.WebJobs.Host/JobHostFunctionTimeoutConfiguration.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// <summary>
+        /// Gets or sets a value indicating whether function invocations will timeout when
+        /// a <see cref="Timeout"/> is specified and a debugger is attached. False by default.
+        /// </summary>
+        public bool TimeoutWhileDebugging { get; set; }
+
+        /// <summary>
+        /// When true, an exception is thrown when a function timeout expires.
+        /// </summary>
+        public bool ThrowExceptionOnFunctionTimeout { get; set; }
```
mathewc: nit: ThrowExceptionOnTimeout?

src/Microsoft.Azure.WebJobs.Host/Timers/BackgroundExceptionDispatcher.cs
```diff
@@ -7,20 +7,16 @@
 
 namespace Microsoft.Azure.WebJobs.Host.Timers
 {
-    internal class BackgroundExceptionDispatcher : IBackgroundExceptionDispatcher
```
mathewc: Did you rename the files to match the new names? I can't tell

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
brettsam: I'll remove it JobHost; it's a mouthful

---
In reply to: [76691087](https://github.com/Azure/azure-webjobs-sdk/pull/799#discussion_r76691087) [](ancestors = 76691087)

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
brettsam: The timeout value is already logged elsewhere, but I've added the Timeout to both the message and a TImeSpan value to the exception itself. It may be useful later, so might as well do it now.

---
In reply to: [76689057](https://github.com/Azure/azure-webjobs-sdk/pull/799#discussion_r76689057) [](ancestors = 76689057)

src/Microsoft.Azure.WebJobs.Host/Timers/IWebJobsExceptionHandler.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// </summary>
+        /// <returns>A Task.</returns>
+        Task OnTimeoutExceptionAsync(ExceptionDispatchInfo exceptionInfo, TimeSpan timeoutGracePeriod);
+
+        /// <summary>
+        /// Implement this method to handle an unhandled exception.
+        /// </summary>
+        /// <param name="exceptionInfo">The <see cref="ExceptionDispatchInfo"/>.</param>
+        Task OnUnhandledExceptionAsync(ExceptionDispatchInfo exceptionInfo);
+    }
```
brettsam: I want to call out this change.

I realized that simply re-throwing a FunctionTimeoutException wasn't very useful if we weren't able to cancel the shutdown token as well -- that will alert all *other* running functions that they should clean up because we're about to shut down.

So I decided to change this interface to look like this -- it allows you to grab the host during initialization, and then call StopAsync/Dispose on it to inform everyone else they should shut down. This would also stop all Listeners from trying to pick up anything new during the grace period.

You can see the default implementation in WebJobsExceptionHandler -- I'm writing tests for that now but wanted some feedback ASAP.

It feels wrong to have something that the JobHost created be responsible for killing it, but it seems to work. I'm open to other suggestions, too.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -347,22 +372,42 @@ internal static System.Timers.Timer StartFunctionTimeout(IFunctionInstance insta
...
+
             trace.Error(message, null, TraceSource.Execution);
 
             trace.Flush();
 
-            // only cancel the token AFTER we've logged our error, since
-            // the Dashboard function output is also tied to this cancellation
-            // token and we don't want to dispose the logger prematurely.
-            cancellationTokenSource.Cancel();
+            // Only cancel the timeout if not debugging
```
mathewc: nit: Only cancel the token

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -303,42 +327,43 @@ internal static TraceLevel GetFunctionTraceLevel(IFunctionInstance functionInsta
...
-                // up the cancellation timer
-                return null;
-            }
-
-            // first see if there is a Timeout applied to the method or class
-            TimeSpan? timeout = globalTimeout;
-            TimeoutAttribute timeoutAttribute = TypeUtility.GetHierarchicalAttributeOrNull<TimeoutAttribute>(method);
-            if (timeoutAttribute != null)
+            TimeSpan? timeout = null;
+            if (attribute != null)
```
mathewc: Just to make things really clear, I'd put an if (attribute == null) { return null } first thing in the method.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInvoker.cs
```diff
@@ -48,6 +48,9 @@ public IReadOnlyList<string> ParameterNames
 
         public async Task InvokeAsync(object[] arguments)
         {
+            // Return a task immediately in case the method is not async.
+            await Task.Yield();
```
mathewc: I don't understand this

src/Microsoft.Azure.WebJobs.Host/Timers/IWebJobsExceptionHandler.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// <summary>
+    /// Represents an unhandled exception handler.
+    /// </summary>
+    public interface IWebJobsExceptionHandler
+    {
+        /// <summary>
+        /// Called during host creation. Allows a reference to the host.
+        /// </summary>
+        /// <param name="host">The JobHost.</param>
+        void Initialize(JobHost host);
```
mathewc: Why does this need to be part of the interface? I think instead a concrete implementation like WebJobsExceptionHandler  should just take the host as a ctor param, since its an implementation detail.

src/Microsoft.Azure.WebJobs.Host/Timers/IWebJobsExceptionHandler.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        /// <summary>
+        /// Called when a timeout occurs.
+        /// </summary>
+        /// <returns>A Task.</returns>
+        Task OnTimeoutExceptionAsync(ExceptionDispatchInfo exceptionInfo, TimeSpan timeoutGracePeriod);
```
mathewc: Why is the explicit method needed - isn't the fact that its a timeout discoverable via the ExceptionDispatchInfo? I.e., don't you have everything you need in a signature like the one below to handle all the scenarios?

src/Microsoft.Azure.WebJobs.Host/Timers/IWebJobsExceptionHandler.cs
brettsam: That's what I'd done originally -- but then I realized that other implementors may want access to this as well and there's no way for them to get it.

Maybe I was overthinking it?

---
In reply to: [77049998](https://github.com/Azure/azure-webjobs-sdk/pull/799#discussion_r77049998) [](ancestors = 77049998)

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInvoker.cs
brettsam: We don't get a Task returned to us until the first await -- in this case it's  _methodInvoker.InvokeAsync() below.

If the function ends up being synchronous and has an infinite loop, we'll never actually return a Task. That means we can't cancel because we're effectively blocked here.

---
In reply to: [77048332](https://github.com/Azure/azure-webjobs-sdk/pull/799#discussion_r77048332) [](ancestors = 77048332)

mathewc: I see. Yeah, since the config needs to be created before the host, other implementors can't really wire it up in a clean way. Ok for now - might improve later.

src/Microsoft.Azure.WebJobs.Host/Timers/IWebJobsExceptionHandler.cs
mathewc: I see. Yeah, since the config needs to be created before the host, other implementors can't really wire it up in a clean way. Ok for now - might improve later.

src/Microsoft.Azure.WebJobs/TimeoutAttribute.cs
```diff
@@ -26,11 +26,30 @@ public sealed class TimeoutAttribute : Attribute
...
         /// <summary>
         /// Gets the timeout value.
         /// </summary>
         public TimeSpan Timeout { get; private set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether function invocations will timeout when
+        /// a <see cref="Timeout"/> is specified and a debugger is attached. False by default.
+        /// </summary>
+        public bool TimeoutWhileDebugging { get; set; }
```
mathewc: some naming suggestions, take 'em or leave 'em:

ThrowExceptionOnTimeout => ThrowOnTimeout
TimeoutGracePeriod => GracePeriod

src/Microsoft.Azure.WebJobs.Host/Timers/IWebJobsExceptionHandler.cs
brettsam: Fabio also suggested a factory for scenarios like this. I'm going to leave it as-is for the moment and revisit for improvements later.

---
In reply to: [77075564](https://github.com/Azure/azure-webjobs-sdk/pull/799#discussion_r77075564) [](ancestors = 77075564)

src/Microsoft.Azure.WebJobs/TimeoutAttribute.cs
brettsam: I'll take 'em!

---
In reply to: [77077864](https://github.com/Azure/azure-webjobs-sdk/pull/799#discussion_r77077864) [](ancestors = 77077864)

mathewc:
:ship"

Issue #800, opened 2016-08-24
EventHubTrigger error: "Controller has already started"


mathewc:
We've gotten a few reports on this issue now. We need to investigate to see how we allow the JobHost to get into this state, and prevent it.

https://social.msdn.microsoft.com/Forums/en-US/bb2f3726-82ec-43d2-b7fa-4b964b969da2/systeminvalidoperationexception-controller-has-already-started?forum=AzureFunctions

https://social.msdn.microsoft.com/Forums/azure/en-US/3bbdef39-58eb-4085-a336-c1088f6f6a73/issue-while-binding-multiple-functions-to-the-same-event-hub-trigger?forum=AzureFunctions

aesalon:
Thanks for opening this issue; adding my Function's function.json as per request. Any other files that would be helpful?

Of note: I've also been experiencing binding issues with this same function which casue an exception like this: mscorlib: Object of type 'Microsoft.WindowsAzure.Storage.Table.TableQuery`1[Submission#0+AggregatedSettings]' cannot be converted to type 'System.Linq.IQueryable`1[Submission#0+AggregatedSettings]'.

When I removed the bindings and accessed table storage output and event hub input explicitly from within the function (going through the CloudStorageAccount.Parse, CloudTable, TableQuery<T> etc dance), I stopped getting the exception covered by this issue.

[function.zip](https://github.com/Azure/azure-webjobs-sdk/files/439811/function.zip)

mathewc:
@MikeStall Is looking into this one. He can provide more details

MikeStall:
notes: I can't repro the error (either in a Function or even in a standalone EventHub where I intentionally make some bad calls) The log in the original appears to be coming from: https://github.com/Azure/azure-webjobs-sdk-script/blob/dev/src/WebJobs.Script/Host/ScriptHostManager.cs#L135

christopheranderson:
This might be related to some of the consumer group issues we've seen?

MikeStall:
The exception happens when multiples *threads* call EventProcessorHost.RegisterEventProcessorAsync on the same instance. (there's some TLS going on there) 
EventHubConfiguration has a global map of EventProcessorHosts. 
So I'm theorizing one way this could happen is if there's a refresh (file update), and ScriptHostManager orphans the old files (but the EPH is still active) and starts listening on the new files (same EPH), and so we end up with a double listen.

MikeStall:
[...]

PR #807, opened 2016-08-31
EventHub fixes

MikeStall:
- add ConsumerGroup as 1st class property on EventHubTrigger. This is to allow users to read from a consumer group without having to create their own EventProcessorHost.
- avoid reusing  the same EventProcessorHost across multiple JobHosts. EventHubConfiguration will now create a new host instance per time.
- call EventData.Dispose() on both events dispatched and events sent.
- include support for PartitionManagerOptions
- better usage of EventProcessorHost. Stopping listening will eagerly abort ProcessEventsAsync.
- more commenting around PartitionContext.CheckpointAsync

There's another companion change to Scripts to take advantage of this to solve #800.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
+++ b/src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
@@ -20,21 +20,37 @@ namespace Microsoft.Azure.WebJobs.ServiceBus
     {
         // Event Hub Names are case-insensitive
         private readonly Dictionary<string, EventHubClient> _senders = new Dictionary<string, EventHubClient>(StringComparer.OrdinalIgnoreCase);
-        private readonly Dictionary<string, EventProcessorHost> _listeners  = new Dictionary<string, EventProcessorHost>(StringComparer.OrdinalIgnoreCase);
-        private readonly EventProcessorOptions _options;
-        private readonly List<Action<JobHostConfiguration>> _deferredWork = new List<Action<JobHostConfiguration>>();
+        
+        private readonly EventProcessorOptions _options; 
+        private readonly PartitionManagerOptions _partitionOptions; // optional, used to create EventProcessorHost
+        private readonly Dictionary<string, ReceiverCreds> _receiverCreds = new Dictionary<string, ReceiverCreds>();
+        private readonly Dictionary<string, EventProcessorHost> _explicitlyProvidedHosts = new Dictionary<string, EventProcessorHost>(StringComparer.OrdinalIgnoreCase);
+
+        private string _defaultStorageString; // set to JobHostConfig.StorageConnectionString
+
+        /// <summary>
+        /// default constructor. Callers can reference this without having any assembly references to service bus assemblies. 
+        /// </summary>
+        public EventHubConfiguration()
+            : this(null, null)
+        {
+        }
 
         /// <summary>
         /// Constructs a new instance.
         /// </summary>
         /// <param name="options">The optional <see cref="EventProcessorOptions"/> to use when receiving events.</param>
-        public EventHubConfiguration(EventProcessorOptions options = null)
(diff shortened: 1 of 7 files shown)


src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubListener.cs
```diff
@@ -38,77 +38,115 @@ void IListener.Cancel()
...
         {
-            // Begin listener 
-            return Task.FromResult(0);
+            return new Listenter(this);
         }
 
-        async Task IEventProcessor.ProcessEventsAsync(PartitionContext context, IEnumerable<EventData> messages)
+        // We get a new instance each time Start() is called. 
+        // We'll get a listener per partition - so they can potentialy run in parallel even on a single machine.
+        private class Listenter : IEventProcessor
```
mathewc: spelling (Listenter)

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubListener.cs
```diff
@@ -38,77 +38,115 @@ void IListener.Cancel()
...
         {
-            EventHubTriggerInput value = new EventHubTriggerInput
+            private readonly EventHubListener _parent;
+            private readonly CancellationTokenSource _cts = new CancellationTokenSource();
+            private static Task _done = Task.FromResult(0);
```
mathewc: nit: might as well use Task.Completed for this instead of your own static

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubListener.cs
```diff
@@ -38,77 +38,115 @@ void IListener.Cancel()
...
-            {
+                    Task[] dispatches = new Task[len];
+                    for (int i = 0; i < len; i++)
+                    {
+                        Task task;
+                        if (_cts.IsCancellationRequested)
+                        {
+                            // If we stopped the listener, then we may lose the lease and be unable to checkpoint. 
+                            // So skip running the rest of the batch. The new listener will pick it up. 
+                            task = _done;
```
mathewc: What is the benefit of adding a bunch of completed Tasks? Why can't you just exit the loop if the token is cancelled (using a collection instead of an array then)?

MikeStall: I need to avoid passing null tasks to Task.When. And I preallocate the dispatches array upfront so I need to put a task in it. 
I could switch to using a List<Task>.

mathewc: Yeah, a collection or list might be better

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubListener.cs
MikeStall: Where exactly is that defined?  I don't see a completed property on System.Threading.Tasks.Task and I don't see any other reference to that in the SDK.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
```diff
@@ -20,21 +20,37 @@ public class EventHubConfiguration : IExtensionConfigProvider, IEventHubProvider
...
 
         /// <summary>
         /// Constructs a new instance.
         /// </summary>
         /// <param name="options">The optional <see cref="EventProcessorOptions"/> to use when receiving events.</param>
-        public EventHubConfiguration(EventProcessorOptions options = null)
+        /// <param name="partitionOptions">Optional <see cref="PartitionManagerOptions"/> to use to configure any EventProcessorHosts. </param>
+        public EventHubConfiguration(
+            EventProcessorOptions options, 
+            PartitionManagerOptions partitionOptions = null)
```
mathewc: parameter is mispelled

PR #813, opened 2016-09-02
fixing executor logic

brettsam:
Behavior for timeouts and cancellations (such as host stopping) are now:

If a timeout:
-	if not `ThrowOnTimeout`, we wait forever for the task to exit after cancelling the token that we passed to the function. This is how it behaved before my change.
-	if `ThrowOnTimeout`, we throw right away, which will bubble up and provide a grace period before calling the host.Stop().

If a cancellation (Stop/Shutdown):
-	 We continue to wait for the original invocation along with the timeout. If a timeout fires, we behave as above.

I had previously added logic that prevented functions from running forever -- we'd exit the Executor and report failure if the host was stopping or there was a timeout. This restores the original behavior and the only difference arises when you set TimeoutAttribute.ThrowOnTimeout = true.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
@@ -538,20 +538,11 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                 TimeoutAttribute timeoutAttribute = TypeUtility.GetHierarchicalAttributeOrNull<TimeoutAttribute>(method);
                 bool throwOnTimeout = timeoutAttribute == null ? false : timeoutAttribute.ThrowOnTimeout;
                 var timer = StartFunctionTimeout(instance, timeoutAttribute, timeoutTokenSource, traceWriter);
+                TimeSpan timerInterval = timer == null ? TimeSpan.MinValue : TimeSpan.FromMilliseconds(timer.Interval);
                 try
                 {
-                    await InvokeAsync(invoker, invokeParameters, timeoutTokenSource, functionCancellationTokenSource, throwOnTimeout);
-                }
-                catch (OperationCanceledException)
-                {
-                    if (timeoutAttribute != null && timeoutAttribute.ThrowOnTimeout && timeoutTokenSource.IsCancellationRequested)
-                    {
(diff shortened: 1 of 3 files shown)


src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -633,6 +592,75 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
             }
         }
 
+        internal static async Task InvokeAsync(IFunctionInvoker invoker, object[] invokeParameters, CancellationTokenSource timeoutTokenSource,
+            CancellationTokenSource functionCancellationTokenSource, bool throwOnTimeout, TimeSpan timerInterval, IFunctionInstance instance)
+        {
+            // There are three ways the function can complete:
+            //   1. The invokeTask itself completes first.
+            //   2. A cancellation is requested (by host.Stop(), fore example).
```
mathewc: nit: fore

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -633,6 +592,75 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+            if (completedTask == timeoutTask)
+            {
+                if (onTimeout != null)
+                {
+                    onTimeout();
+                }
+
+                if (throwOnTimeout)
+                {
+                    // If we need to throw, await the task to throw now. This will bubble up and eventually
```
mathewc: You say "await the task to throw now" but you're not awaiting anything?

mathewc:
:ship:

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
brettsam: I changed the behavior (I used to await here, then catch a TaskCancledException upstream and throw the FunctionTimeoutException there) -- but I never updated the comment.

Fixed.

---
In reply to: [77361102](https://github.com/Azure/azure-webjobs-sdk/pull/813#discussion_r77361102) [](ancestors = 77361102)

Issue #785, opened 2016-08-09
Blob scanning strategy can have gaps due to clock skew


brettsam:
`ScanBlobScanLogHybridPollingStrategy` determines whether a blob is 'new' or not by comparing its timestamp to the local machine's time. I've seen instances where the clocks differ by a couple of seconds, causing a new blob to be missed in the poll. The blob is eventually seen by the log-scanning strategy that's also running, but it'd be nice if we can tighten this up to prevent the gap. I think this could be done by storing the newest `LastModifiedTime` of the blobs we've looked at rather than the `UtcNow` time of the local machine.

https://github.com/Azure/azure-webjobs-sdk/blob/master/src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs#L181

christopheranderson:
Same story as #786, Brett

PR #821, opened 2016-09-12
addressing blob scan clock skew

brettsam:
Addresses #785.

It is possible for blob LastModified timestamps to be off by a few seconds when compared to the local machine. This would result in us missing new blobs after our initial scans. They'd later be picked up by the log-reading strategy, but that could be a delay of several minutes. This change reduces the chances of missing due to clock skew by basing our last scan-time on the latest-read blob, rather than on the local machine's time.

An example of where we could miss a blob before this change -- assume our local machine's clock is ahead of Azure by 2 seconds:

1. BlobA is added at 12:00:00. 
2. This is our first scan, so we pull it and see that it is new immediately after it has been added. We mark the 'last scan time' as 12:00:02 (because we're skewed ahead of Azure).
3. BlobB is added at 12:00:01.
4. We pull it and compare it to our 'last scan time' of 12:00:02. Because we think this is old, we never process it.

With this change, in step 2, we'd put our 'last scan time' as 12:00:00 (the timestamp of BlobA), and thus pick up BlobB with the next scan.

Note that this does not completely eliminate the chances of missing a blob with this scan. I've noticed that LastModified timestamps are all rounded to the nearest second. So it's possible that we pull a blob at exactly 12:00:00 and mark 12:00:00 as our 'last timestamp seen'. Then a new blob can come in at 12:00:00.5, but also get a timestamp of 12:00:00 when stored in Azure. On our next scan, we'd miss that because the timestamp is equal to our last scan. There may be ways to account for this but it also may not be worth the effort. I gave it some thought but all of my solutions had drawbacks. I'd rather not move to use `>=` as that would mean we'd always have at least one 'new blob' that we'd later reject when we found its receipt. Any suggestions welcome :-)

### DIFF
--- a/test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategyTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategyTests.cs
@@ -14,6 +14,8 @@ using Microsoft.Azure.WebJobs.Host.FunctionalTests.TestDoubles;
 using Microsoft.Azure.WebJobs.Host.Listeners;
 using Microsoft.Azure.WebJobs.Host.Storage;
 using Microsoft.Azure.WebJobs.Host.Storage.Blob;
+using Microsoft.WindowsAzure.Storage;
+using Microsoft.WindowsAzure.Storage.Blob;
 using Xunit;
 
 namespace Microsoft.Azure.WebJobs.Host.FunctionalTests.Blobs.Listeners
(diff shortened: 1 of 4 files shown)


test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategyTests.cs
```diff
@@ -99,10 +101,12 @@ public void TestBlobListenerWithMultipleContainers()
 
             RunExecuteWithMultiPollingInterval(firstContainerExpectedNames, product, executor, testScanBlobLimitPerPoll / containerCount);
 
+            Thread.Sleep(10);
+
             List<string> secondContainerExpectedNames = new List<string>();
             for (int i = 0; i < 2; i++)
             {
-                firstContainerExpectedNames.Add(CreateAblobAndUploadToContainer(secondContainer));
+                secondContainerExpectedNames.Add(CreateAblobAndUploadToContainer(secondContainer));
```
brettsam: Note this test was wrong before I made my change. It actually failed until I added a Sleep to make sure the second scan saw these blobs as later. I made sure this passed both before and after my change.

mathewc:
:ship:

PR #823, opened 2016-09-12
allow listening on event hubs with arbitrary names. 

MikeStall:
Stop calling the EventProcessorHost ctor that the Event hub team wants to deprecate.
Generate our own storage container name; and fix 

Fix https://github.com/Azure/azure-webjobs-sdk-script/issues/625

### DIFF
--- a/src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
+++ b/src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Text;
 using Microsoft.Azure.WebJobs.Host;
 using Microsoft.Azure.WebJobs.Host.Bindings;
@@ -230,12 +231,16 @@ namespace Microsoft.Azure.WebJobs.ServiceBus
                     sb.EntityPath = null; // need to remove to use with EventProcessorHost
                 }
 
-                EventProcessorHost host = new EventProcessorHost(
-                   eventProcessorHostName,
-                   actualPath,
-                   consumerGroup,
-                   sb.ToString(),
-                   storageConnectionString);
+                string @namespace = GetServiceBusNamespace(creds.EventHubConnectionString);
+                string containerName = EventHubLeaseFactory.GetContainerName(eventHubName, @namespace, storageConnectionString);
+
+                EventProcessorHost host = new EventProcessorHost( 
+                    hostName: eventProcessorHostName,
+                    eventHubPath: actualPath,
+                    consumerGroupName: consumerGroup,
+                    eventHubConnectionString: sb.ToString(), 
+                    storageConnectionString: storageConnectionString, 
+                    leaseContainerName: containerName);
 
                 if (_partitionOptions != null)
                 {
@@ -313,6 +318,13 @@ namespace Microsoft.Azure.WebJobs.ServiceBus
             var eventData = new EventData(Encoding.UTF8.GetBytes(input));
             return eventData;
         }
+
+        // Given an eventhub connection string, get the service bus namespace. 
+        private static string GetServiceBusNamespace(string connectionString)
+        {
+            var cx = new ServiceBusConnectionStringBuilder(connectionString);
+            return cx.Endpoints.First().Host.ToString();
+        }       
 
         // Hold credentials for a given eventHub name. 
         // Multiple consumer groups (and multiple listeners) on the same hub can share the same credentials. 
--- a/src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubLeaseFactory.cs
+++ b/src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubLeaseFactory.cs
@@ -1,0 +1,133 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Globalization;
+using System.Text;
+using Microsoft.WindowsAzure.Storage;
+using Microsoft.WindowsAzure.Storage.Table;
+
+namespace Microsoft.Azure.WebJobs.ServiceBus
+{
+    // EventProcessorHost requires a container name that the multiple host instances use to coordinate with each other and create leases. 
+    // An event hub is described by an (namespace name, event hub path name). The mapping from the event hub to the container name
+    // must be a one-to-one injective function such that:
+    // 1. it's deterministic: when we scale out ot N separate instances, they must all compute the same container name for the same event hub. 
+    // 2. it's collision free: different event hubs shouldn't share the same container. So we can't just use a hash. 
+    //  Note that the event hub name is much more expressive than the storage container name, so we can't do a stateless collision-free transform. 
+    // Hence we use a lookaside storage table to save the transform. 
+    internal class EventHubLeaseFactory
+    {
+        // Name of Azure table that maps "eventHub" --> container name. 
+        private const string TableName = "AzureFunctionsEventHubTable";
+             
+        // Get container name. Lease name that multiple eventhub instances use to coordinate with each other. 
+        public static string GetContainerName(string eventHubName, string eventHubNamespace, string storageConnectionString)
+        {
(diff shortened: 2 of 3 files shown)


src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubLeaseFactory.cs
```diff
@@ -0,0 +1,133 @@
...
+    internal class EventHubLeaseFactory
+    {
+        // Name of Azure table that maps "eventHub" --> container name. 
+        private const string TableName = "AzureFunctionsEventHubTable";
+             
+        // Get container name. Lease name that multiple eventhub instances use to coordinate with each other. 
+        public static string GetContainerName(string eventHubName, string eventHubNamespace, string storageConnectionString)
+        {
+            CloudStorageAccount account = CloudStorageAccount.Parse(storageConnectionString);
+            var client = account.CreateCloudTableClient();
```
mathewc: I don't think we should create any table resources in the users storage account. Note that nowhere else in the SDK do we do this currently. Instead, for things we need to store we use blob storage. We also try to minimize creation of new containers. For small storage items like this, I recommend you create a blob in our azure-webjobs-host container. Other extensions are already using this container for that purpose.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubLeaseFactory.cs
```diff
@@ -0,0 +1,133 @@
...
+                        return entity.ContainerName;
+                    }
+                    else
+                    {
+                        // Storage container, restrict to: [a-z0-9\-]{3,63}, no consecutive dashes, start&end with alpha
+                        // Guid is 32 chars long and ToString() yields lowercase.  
+                        string newName = "eventhubs-" + Guid.NewGuid().ToString("N");
+
+                        // Not found. We'll create one. 
+                        var entity = new EventHubTableEntity { PartitionKey = pk, RowKey = rk, ContainerName = newName };
```
mathewc: Do we really need a separate unique container per event hub? I'd expect that within the container, they're managing the uniqueness of their lease IDs. Why can't we just specify a single simple top level container like "eventhub-leases" and let them manage it? I can't imagine that they expect all of their users to have to replicate logic like you've done here - if that is required, it should be centralized in their SDK.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubLeaseFactory.cs
MikeStall: We do create table for the Functions (albeit not SDK) logging, and I followed that naming convention to try and be consistent.  (Although it'd be great to switch SDK over to that faster table logging). 

I preferred Table because it's better than Blob for small independent updates like this. For Blob, I didn't want a blob per entity (inefficient); and I didn't want a single blob that all entities shared (since the entities are independent from each other; I wanted to minimize concurrency cases) 

That said, if you'd rather this be a single blob in the  azure-webjobs-host ,  I can change it.

mathewc: Yeah, I think we should minimize the number of visible storage entities we create in the users account. Definitely for low level plumbing like this. Logs are different I think, since there are valid scenarios for users wanting access to that log storage.

The mapping you're doing here is the same kind of problem that our DynamicHostIdProvider solves via blob storage. In that case it is mapping full assembly name => Guid. Why do you think having a unique blob per entity is inefficient? You just create it once if it doesn't exist - that's fast.

MikeStall: My opinion is that 1 table or 1 container was the same "footprint" to the owner of the storage account. 

This problem is definitely solvable with blob.  I had gotten feedback from storage team previously that entities in a table was a better pattern than individual blobs  for dealing with small amounts of information, so I defaulted to that out of habit. For example, you can query multiple table entities in a single REST call, so this would let you do something like "tell me all event hubs in a given namespace". 

Arguably, that's matters much less when we're only dealing with a handful of entries.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubLeaseFactory.cs
MikeStall: Unfortunately, yes. This is a limitation of EventProcessorHost.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubLeaseFactory.cs
```diff
@@ -0,0 +1,133 @@
...
+    // must be a one-to-one injective function such that:
+    // 1. it's deterministic: when we scale out ot N separate instances, they must all compute the same container name for the same event hub. 
+    // 2. it's collision free: different event hubs shouldn't share the same container. So we can't just use a hash. 
+    //  Note that the event hub name is much more expressive than the storage container name, so we can't do a stateless collision-free transform. 
+    // Hence we use a lookaside storage to save the transform. 
+    internal class EventHubLeaseFactory
+    {
+        // Name of Azure blob that maps "eventHub" --> container name. 
+        private const string InternalContainerName = "azure-webjobs-hosts";
+        private const string InternalBlobPath = "eventhub/leasenames.json";
```
mathewc: These aren't lease names, right? They're container names.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubLeaseFactory.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        // Concurrent safe mutation of a blob. 
+        // Apply mutator to update the blob (as a deserialized JSON contents). 
+        // If blob doesn't exist yet, then this allocates a new TBlob() object and calls the mutator on that. 
+        // Mutator also extracts a value and returns it. 
+        private static TResult MutateBlob<TBlob, TResult>(CloudBlockBlob blob, Func<TBlob, TResult> mutator) where TBlob : new()
```
mathewc: Mutating a single blob seems a lot more complicated and error prone than simply doing:

- GET azure-webjobs-hosts/eventhub/lease-containers/<namespace>-<hubname>
- If blob exists, read it to get the container name, else PUT a new blob to that path with the container name

That's basically the same logic you had before for table storage, except instead of table queries you're doing direct blob references.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
```diff
@@ -230,12 +231,16 @@ EventProcessorHost IEventHubProvider.GetEventProcessorHost(string eventHubName,
...
                 }
 
-                EventProcessorHost host = new EventProcessorHost(
-                   eventProcessorHostName,
-                   actualPath,
-                   consumerGroup,
-                   sb.ToString(),
-                   storageConnectionString);
+                string @namespace = GetServiceBusNamespace(creds.EventHubConnectionString);
+                string containerName = EventHubLeaseFactory.GetContainerName(eventHubName, @namespace, storageConnectionString);
```
mathewc: We need tests for this code - tests that exercise this blob code and verify that the correct blob is created, etc. Also a test that verifies concurrency is handled correctly - e.g. three Tasks that run concurrently, and verify that each gets the same container name.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubLeaseFactory.cs
```diff
@@ -0,0 +1,133 @@
...
+                {
+                    return containerName;
+                }
+
+               // Storage container, restrict to: [a-z0-9\-]{3,63}, no consecutive dashes, start&end with alpha
+               // Guid is 32 chars long and ToString() yields lowercase.  
+               string newName = "eventhubs-" + Guid.NewGuid().ToString("N");
+
+               // Create a new container name for the event hub. 
+               map[key] = newName;
```
mathewc: nit: formatting is off

MikeStall:
I discussed with the EventProcessorHost team and they will make EPH allow multiple eventhubs listening through a single container. That's goodness for all clients. This means the clients can pass in a fixed container name and that greatly simplifies things here. So this PR would be unnecessary.

Issue #786, opened 2016-08-09
Blob scanning a large container at startup can take several minutes


brettsam:
In my tests on a local machine, I'm seeing the initial blob scan of a container take roughly 160 milliseconds per blob (to check receipts, etc). As you get more and more blobs, the time that it takes to perform the initial scan on startup gets longer and longer. We've seen our Script tests timeout after a minute of waiting for ~500 blobs to be scanned on startup.

The impact here is that if you have container with a large number of blobs, any new blobs may take several minutes before they are processed after the host starts.

We may be able to take an approach similar to that of Timers and store the 'latest processed time' in its own blob and read from there on startup, skipping all of the ones we know we've already looked at.

https://github.com/Azure/azure-webjobs-sdk/blob/master/src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs

christopheranderson:
@brettsam - could you look at this? @shobak101 isn't likely to have time in the near time to address this.

brettsam:
Baseline -- Running locally with ~1100 blobs in a container, it takes right around 3.5 minutes for my machine to complete a full scan. This is because it takes roughly 150ms to process each blob and we do this sequentially.

### **Option 1: Store a 'Latest Blob Processed' timestamp in Blob storage.** 
If we know where we left off, we don't need to re-scan old, already-processed blobs during every restart. This is what is currently implemented in #824.

**Pros:** Doesn't matter how many blobs you have in the container, it'll never re-scan old blobs.
**Cons:** The fact that we scan all blobs on startup is a bit of a 'failsafe'. It means that even if something fell through the cracks, it'll get re-evaluated when the host starts.

I've mitigated the con by doing a couple of things:
- I only write the 'Latest Blob Processed' timestamp after we've successfully created Queue messages.
- If there were any errors in creating the Queue messages, I make sure to keep the timestamp rolled back to the earliest failure so that if the host were to crash, we'd pick it back up.
- I roll the timestamp back by 1ms so that we re-process the last item again upon startup. This is to account for the chance of timestamp rounding that blobs do. You could process a blob at 12:00:00 and have a new blob come in at 12:00:00.1, but it's timestamp would be rounded down to 12:00:00. By rounding the timestamp back slightly, you guarantee this will be picked up on a restart.

### **Option 2: Improve the ~150ms-per-blob scan time** 
[...]

PR #824, opened 2016-09-12
serializing ScanInfo to improve blob startup processing

brettsam:


mamaso:
Clarify a bit for me:

Currently on host start we have no record of which blobs containers have been scanned.  The goal of this is to save the previous scan info in the storage account so that it persists between restarts, correct?

brettsam:
Correct. Maybe it'd be easier if I linked to the issue -- #786

brettsam:
This is now ready for a proper review.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -223,10 +228,13 @@ private void ThrowIfDisposed()
...
 
-            // if ending a cycle then copy currentSweepCycleStartTime to lastSweepCycleStartTime
-            if (blobSegment.ContinuationToken == null)
+            // if ending a cycle then copy currentSweepCycleStartTime to lastSweepCycleStartTime, if changed
+            if (blobSegment.ContinuationToken == null &&
+                containerScanInfo.CurrentSweepCycleLatestModified > containerScanInfo.LastSweepCycleLatestModified)
             {
                 containerScanInfo.LastSweepCycleLatestModified = containerScanInfo.CurrentSweepCycleLatestModified;
+                await _blobScanInfoManager.UpdateLatestScanAsync(container.ServiceClient.Credentials.AccountName,
+                    container.Name, containerScanInfo.LastSweepCycleLatestModified);
```
brettsam: This is currently serializing the last scan timestamp immediately after we've determined the 'new blobs' but before we've executed any functions. With a large number of blobs, this means we may update this, then spend several minutes running functions. If something crashes, we'd skip re-running all of those other blobs and the log scanning would have to pick them up.

It may make more sense to do this after we're done looping through and running functions against the 'new blobs'. Then any restart would just pick back up from the beginning.

I can't think of a downside to that approach but I'll see what others say.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/IBlobScanInfoManager.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using System;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
+{
+    internal interface IBlobScanInfoManager
+    {
+        Task<DateTime?> LoadLatestScanAsync(string storageAccountName, string containerName);
+        Task UpdateLatestScanAsync(string storageAccoutnName, string containerName, DateTime scanInfo);
```
mamaso: storageAccoutnName misspelled in UpdateLatestScanAsync

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
mamaso: Does log scanning only run on new storage logs, or does it keep track of the last logs it saw?  If it only runs on new logs we could have unprocessed blobs if container scan finds blob -> storage logs with those blobs created -> crash during blob processing -> restart (the blob logs have already occurred, so the blobs will never process).

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/SharedBlobListenerFactory.cs
```diff
@@ -13,8 +13,9 @@ internal class SharedBlobListenerFactory : IFactory<SharedBlobListener>
         private readonly IStorageAccount _account;
         private readonly IWebJobsExceptionHandler _exceptionHandler;
         private readonly IContextSetter<IBlobWrittenWatcher> _blobWrittenWatcherSetter;
+        private readonly string _hostId;
 
-        public SharedBlobListenerFactory(IStorageAccount account,
+        public SharedBlobListenerFactory(string hostId, IStorageAccount account,
```
mamaso: my ignorance here, is the hostId web app specific? function specific? instance specific?

Is there only one instance which listens to blobs, or do we parallelize somehow?

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/StorageBlobScanInfoManager.cs
```diff
@@ -0,0 +1,113 @@
...
+                {
+                    latestScan = scanInfo.LatestScan;
+                }
+
+                return latestScan;
+            }
+            catch (StorageException exception)
+            {
+                if (exception.RequestInformation != null &&
+                    exception.RequestInformation.HttpStatusCode == 404)
```
mamaso: nit: could probably use one of the StorageExceptionExtension methods, this also fine

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/StorageBlobScanInfoManager.cs
```diff
@@ -0,0 +1,113 @@
...
+                string scanInfoLine = await scanInfoBlob.DownloadTextAsync(CancellationToken.None);
+                ScanInfo scanInfo;
+                using (StringReader stringReader = new StringReader(scanInfoLine))
+                {
+                    scanInfo = (ScanInfo)_serializer.Deserialize(stringReader, typeof(ScanInfo));
+                }
+
+                if (scanInfo != null)
+                {
+                    latestScan = scanInfo.LatestScan;
```
mamaso: nit: do we have access to the `?.` accessor?  Could shorten a fair bit of null checking

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
mathewc: The log scan goes back in time [2 hours](https://github.com/Azure/azure-webjobs-sdk/blob/master/src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/BlobLogListener.cs#L52). So if the container scan finds a blob modified at time T and persists a scan info T2 > T (e.g. another blob was modified a little after), if blob processing fails, container scan would never find that blob again. But presumably log scan will find it because it goes back 2 hours?

brettsam: @mathewc  -- is the full scan at startup considered our failsafe? This persistence may not be a good idea, then, if it's removing that. I had assumed that the log scanning was fullproof (within 2 hours) and thus was the failsafe here. But it sounds like that may not be the case if logs aren't guaranteed.

I was even thinking that, on startup, we could check the last updated time of the scanInfo blob and delete it if it were older than 2 hours so we'd only have to do a full scan if we think we may have blobs added outside of that 2 hour window.

If we do, indeed, need that full scan at startup, maybe I can see if there's a way to trim down that 140ms per-blob hit we take while we scan. Parallelize it, try to reduce network calls, etc.

brettsam:
I did some thinking here and moved the save to after all notifications have been sent (i.e. after all queue messages are added). I also took failures into account -- if we previously had failures, we want to make sure we pick them up on a restart (they're currently only held in memory). I'm trying to figure out the best way to test the failure thing now but wanted to get it out.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/StorageBlobScanInfoManager.cs
brettsam: Not until we move to AppVeyor.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/SharedBlobListenerFactory.cs
brettsam: For your other comment -- I think Mathew answered it above. Log scanning goes back 2 hours.

brettsam:
Tests are now ready -- @mathewc I'd like to get your eyes on this before it goes in. I've thought through a bunch of scenarios and have them covered, but you may have others that I've not considered.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/SharedBlobListenerFactory.cs
mathewc: To clarify a bit more - only one instance does the log scanning and population of the blob trigger queue. However ALL instances listen to that queue, so blob processing itself is parallelized. I did that because we don't want this relatively expensive scanning to be duplicated on all instances.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
mathewc: Correct, blob logs are NOT guaranteed. See their documentation [here](https://blogs.msdn.microsoft.com/windowsazurestorage/2011/08/02/windows-azure-storage-logging-using-logs-to-track-storage-requests/). I excerpt: _"During normal operation all requests are logged; but it is important to note that  logging is provided on a best effort basis.  This means we do not guarantee that every message will be logged due to the fact that the log data is buffered in memory at the storage front-ends before being written out, and if a role is restarted then its buffer of logs would be lost."_

So the log scanning helps by improving the possibility that we get faster notifications when a scan process is chugging through a large container. Scanning for new blobs is the only way we can be SURE we get everything.

brettsam: So what's your take on the approach I've taken here? Yay/nay? Scrap it? I can fill in some other things I looked into in the issue and we can discuss it there.

mathewc:
:ship:

Issue #819, opened 2016-09-09
Make logging more resilient in the event of host crash + usability long running events


brettsam:
Discussed on an internal thread. From @MikeStall -- 

What I’d like to do is:
-	Don’t send a “start” event – that would be too slow since it would guarantee double the log messages in the host path case. (say the function runs for 1 sec; we don’t want to emit 2 log msgs) 
-	Instead, send the heart  beat message and defer it for 10 seconds, and send every 30 sec after that.  That means quick 1s functions won’t be penalized.

lindydonna:
From triage discussion: this causes a problem because it causes some logging events to be missed entirely. The reason for the change in behavior relative to Web Jobs is to improve logging performance.

@brettsam Will your "graceful shutdown" improvements make this issue less troublesome?

lindydonna:
@MikeStall Please come up with a design that handles these failure cases and we'll try to make it fit for RC.

brettsam:
If anything, the graceful shutdowns makes matters worse because now we'll deliberately shut the host down for timeouts. We don't have a way to grab metadata from all in-flight functions and log that they've been killed. 

I'm not sure what "potentially optimize failure cases" means here. We'll log if there's a failure. But if there's a host shutdown and you're in the middle of a 1-minute function run, we have no success or failure -- the function simply isn't complete when the host exits. And we don't forcefully kill any functions on a shutdown. If they don't exit quickly when we cancel the shutdown token (only available to C#), we'll never get any details about them.

Mike's design above is meant to let quick-running functions get by without paying any logging penalty and long-running functions save a heartbeat so we have a record of them.

morsh:
Hi,
I'm hoping this is the right place to add those comments, but I'm experiencing the same problems with the following scenarios:
1. exceptions with requiring the function file like missing npm peckages exceptions
2. unhandled async exceptions like having an unhandled exception while expecting a callback

Basically I would expect monitoring the concurrent active node js requests and adding some process level exception handling.

MikeStall:
#826 and #708 should fix this.

Issue #708, opened 2016-05-04
FastLogger Should handle Table deletions


mathewc:
While the host us up and running, if the FastLogger table is deleted from Table storage, it doesn't get recreated, and no logs are written.

The logger should optimistically write to the table and handle 404s by creating the table if it no longer exists.

mathewc:
Also, we should change the name of the Default table created from "AzureFunctionsLogTable". The core SDK is independent of Azure Functions, and people will want to use FastLogging outside of AzureFunctions.

MikeStall:
(commit comment from https://github.com/Azure/azure-webjobs-sdk/commit/dd176eed31aecd6d46e1df6bec1490cd6975f340 had typo referring to this bug. That commit was actually for #819 )

mathewc:
I opened https://github.com/Azure/azure-webjobs-sdk/issues/832 for the table rename suggestion which was part of this issue above.

PR #826, opened 2016-09-14
Fix 819 - have "function started" log event 

MikeStall:
Include a "Function started" logging event.  This is useful for longer running functions and lets us show "In-progress" vs "completed". It also means that if we crash in the middle of a long running function, we have some evidence of execution. 

Log events are coalesced so that a completed row overwrites a started row - this avoids having to do an 
indexing operation.

Fixes https://github.com/Azure/azure-webjobs-sdk/issues/819

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging/Internal/LogWriter.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging/Internal/LogWriter.cs
@@ -9,6 +9,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Logging.Internal;
 using Microsoft.WindowsAzure.Storage.Table;
+using System.Collections;
 
 namespace Microsoft.Azure.WebJobs.Logging
 {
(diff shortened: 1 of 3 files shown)


src/Microsoft.Azure.WebJobs.Logging/Internal/LogWriter.cs
```diff
@@ -161,33 +162,33 @@ private static int GetContainerSize()
...
             }
 
-            // Time aggregate is flushed later. 
-            // Don't flush until we've moved onto the next interval. 
+            if (item.IsCompleted())
             {
-                var rowKey = TimelineAggregateEntity.RowKeyTimeInterval(item.FunctionName, item.StartTime, _uniqueId);
-
-                lock(_lock)
+                // For completed items, aggregate total passed and failed withing a time bucket.
```
mathewc: nit: withing

src/Microsoft.Azure.WebJobs.Logging/Internal/LogWriter.cs
```diff
@@ -161,33 +162,33 @@ private static int GetContainerSize()
...
+
+                    lock (_lock)
                     {
-                        // Can we flush the old counters?
-                        x = TimelineAggregateEntity.New(_containerName, item.FunctionName, item.StartTime, _uniqueId);
-                        _timespan[rowKey] = x;
+                        TimelineAggregateEntity x;
+                        if (!_timespan.TryGetValue(rowKey, out x))
+                        {
+                            // Can we flush the old counters?
```
mathewc: it's not clear what this question refers to

src/Microsoft.Azure.WebJobs.Logging/Internal/LogWriter.cs
```diff
@@ -161,33 +162,33 @@ private static int GetContainerSize()
...
+                            x = TimelineAggregateEntity.New(_containerName, item.FunctionName, item.StartTime, _uniqueId);
```
mathewc: can you give this "x" a better name for readability?

src/Microsoft.Azure.WebJobs.Logging/Internal/LogWriter.cs
```diff
@@ -329,6 +330,35 @@ private static void Increment(FunctionInstanceLogItem item, TimelineAggregateEnt
 
 
             await Task.WhenAll(t);
-        }    
+        }
+
+        // Collection where adding in the same RowKey replaces a previous entry with that keyl.
```
mathewc: nit: keyl

src/Microsoft.Azure.WebJobs.Logging/Internal/LogWriter.cs
```diff
@@ -329,6 +330,35 @@ private static void Increment(FunctionInstanceLogItem item, TimelineAggregateEnt
 
 
             await Task.WhenAll(t);
-        }    
+        }
+
+        // Collection where adding in the same RowKey replaces a previous entry with that keyl. 
+        // This is single-threaded. Caller must lock. 
+        private class RowKeyCollection<T> where T : TableEntity
```
mathewc: This type confuses me. It's not really a collection of row keys, its a collection of TableEntities. Also, it doesn't really seem to add much behavior above Dictionary itself. I.e., you could just use a Dictionary instead of this type and use the indexer for assignment rather than Add.

mathewc:
:ship:

MikeStall:
Committed.
https://github.com/Azure/azure-webjobs-sdk/commit/dd176eed31aecd6d46e1df6bec1490cd6975f340 
(The commit comments there have a typo and said #708 , but it was actually this one)

PR #827, opened 2016-09-15
Throw error if blob-only storage account

### LINKED ISSUE #797: Blob Only Storage Accounts not supported as primary storage account
They aren't supported as the "primary" storage account (since we need access to queues, etc.). We need to document this, and ideally find a way to throw an error if the wrong type of account is used.

mamaso:
resolves #797

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageAccountProvider.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageAccountProvider.cs
@@ -153,6 +153,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
         public async Task<IStorageAccount> GetAccountAsync(string connectionStringName, CancellationToken cancellationToken)
         {
             IStorageAccount account = null;
+            var isPrimary = true;
             if (connectionStringName == ConnectionStringNames.Dashboard)
             {
                 account = DashboardAccount;
@@ -168,13 +169,14 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                 if (!string.IsNullOrEmpty(connectionString))
                 {
                     account = ParseAccount(connectionStringName, connectionString);
+                    isPrimary = false;
                 }
             }
 
             if (account != null)
             {
                 // On the first attempt, this will make a network call to verify the credentials work.
-                await _storageCredentialsValidator.ValidateCredentialsAsync(account, cancellationToken);
+                await _storageCredentialsValidator.ValidateCredentialsAsync(account, isPrimary, cancellationToken);
             }
 
             return account;
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
@@ -4,10 +4,12 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host.Storage;
 using Microsoft.Azure.WebJobs.Host.Storage.Blob;
+using Microsoft.Azure.WebJobs.Host.Storage.Queue;
 using Microsoft.WindowsAzure.Storage;
 using Microsoft.WindowsAzure.Storage.Auth;
 
@@ -16,8 +18,9 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
     internal class DefaultStorageCredentialsValidator : IStorageCredentialsValidator
     {
         private readonly HashSet<StorageCredentials> _validatedCredentials = new HashSet<StorageCredentials>();
+        private StorageCredentials _primaryCredentials = null;
 
-        public async Task ValidateCredentialsAsync(IStorageAccount account, CancellationToken cancellationToken)
+        public async Task ValidateCredentialsAsync(IStorageAccount account, bool isPrimaryAccount, CancellationToken cancellationToken)
         {
             if (account == null)
             {
@@ -27,17 +30,17 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
             StorageCredentials credentials = account.Credentials;
 
             // Avoid double-validating the same account and credentials.
-            if (_validatedCredentials.Contains(credentials))
+            if ((!isPrimaryAccount && _validatedCredentials.Contains(credentials)) || _primaryCredentials == credentials)
             {
                 return;
             }
 
-            await ValidateCredentialsAsyncCore(account, cancellationToken);
+            await ValidateCredentialsAsyncCore(account, isPrimaryAccount, cancellationToken);
             _validatedCredentials.Add(credentials);
         }
 
-        private static async Task ValidateCredentialsAsyncCore(IStorageAccount account,
-            CancellationToken cancellationToken)
+        private async Task ValidateCredentialsAsyncCore(IStorageAccount account,
+            bool isPrimaryAccount, CancellationToken cancellationToken)
         {
             // Verify the credentials are correct.
             // Have to actually ping a storage operation.
@@ -56,8 +59,37 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
             catch
             {
                 string message = String.Format(CultureInfo.CurrentCulture,
-                    "The account credentials for '{0}' are incorrect.", account.Credentials.AccountName);
+                    "Invalid storage account '{0}'. Please make sure your credentials are correct.",
+                    account.Credentials.AccountName);
                 throw new InvalidOperationException(message);
+            }
+
+            if (isPrimaryAccount)
+            {
+                // Primary storage accounts require Queues
+                IStorageQueueClient queueClient = account.CreateQueueClient();
+                IStorageQueue queue = queueClient.GetQueueReference("name");
+                try
+                {
+                    await queue.ExistsAsync(cancellationToken);
+                    _primaryCredentials = account.Credentials;
+                }
+                catch (OperationCanceledException)
+                {
+                    throw;
+                }
+                catch (StorageException exception)
+                {
+                    WebException webException = exception.GetBaseException() as WebException;
+                    if (webException != null && webException.Status == WebExceptionStatus.NameResolutionFailure)
+                    {
+                        string message = String.Format(CultureInfo.CurrentCulture,
+                            "Invalid storage account '{0}'. Primary storage accounts must be general "
+                            + "purpose accounts and not restricted blob storage accounts.", account.Credentials.AccountName);
+                        throw new InvalidOperationException(message);
+                    }
+                    throw;
+                }
             }
         }
     }
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/IStorageCredentialsValidator.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/IStorageCredentialsValidator.cs
@@ -9,6 +9,6 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
 {
     internal interface IStorageCredentialsValidator
     {
-        Task ValidateCredentialsAsync(IStorageAccount account, CancellationToken cancellationToken);
+        Task ValidateCredentialsAsync(IStorageAccount account, bool isPrimaryAccount, CancellationToken cancellationToken);
     }
 }
--- a/test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
@@ -151,7 +151,7 @@ namespace Microsoft.Azure.WebJobs.Host.EndToEndTests
                 Assert.True(queueProcessorFactory.CustomQueueProcessors.Sum(p => p.CompleteProcessingCount) >= 2);
 
                 Assert.Equal(16, storageClientFactory.TotalBlobClientCount);
-                Assert.Equal(9, storageClientFactory.TotalQueueClientCount);
+                Assert.Equal(11, storageClientFactory.TotalQueueClientCount);
                 Assert.Equal(0, storageClientFactory.TotalTableClientCount);
 
                 Assert.Equal(6, storageClientFactory.ParameterBlobClientCount);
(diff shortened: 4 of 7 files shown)


src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
```diff
@@ -41,24 +41,32 @@ internal class DefaultStorageCredentialsValidator : IStorageCredentialsValidator
...
             // Have to actually ping a storage operation.
-            IStorageBlobClient client = account.CreateBlobClient();
+            IStorageQueueClient client = account.CreateQueueClient();
 
             try
             {
                 // This can hang for a long time if the account name is wrong. 
                 // If will fail fast if the password is incorrect.
-                await client.GetServicePropertiesAsync(cancellationToken);
+                await client.SdkObject.GetServicePropertiesAsync(cancellationToken);
```
mathewc: Is this because our StorageQueueClient impl. doesn't implement GetServicePropertiesAsync? How hard would that be to add, so you don't have to access SdkObject here?

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
```diff
@@ -41,24 +41,32 @@ internal class DefaultStorageCredentialsValidator : IStorageCredentialsValidator
...
             {
                 throw;
             }
-            catch
+            catch (StorageException exception)
             {
                 string message = String.Format(CultureInfo.CurrentCulture,
-                    "The account credentials for '{0}' are incorrect.", account.Credentials.AccountName);
+                                        "The account credentials for '{0}' are incorrect.", account.Credentials.AccountName);
+                if (exception.IsRemoteNameNotResolved())
```
mathewc: Rather than try to figure out via IsRemoteNameNotResolved what type of error this is, how about just changing the error message to something that handles all cases, like: "Invalid storage account '{0}'. Please make sure your credentials are correct, and that the storage account is a general purpose account and not a restricted blob storage account"? You get the idea - we just need to get the right verbiage for referencing these account types from official Azure Storage sources.

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
mamaso: Ended up being a significant amount of change required in the test project, so I changed the methodology a bit.

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
mamaso: Validator is used for all storage accounts yep.  @lindydonna had mentioned that we don't want to support blob-only, but I've lifted that restriction for secondary storage accounts with this new iteration (since they work at the moment, no need for a breaking change).

mathewc: We definitely want to support blob-only for secondary storage accounts.

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageAccountProvider.cs
```diff
@@ -174,7 +174,7 @@ private IStorageAccount StorageAccount
             if (account != null)
             {
                 // On the first attempt, this will make a network call to verify the credentials work.
-                await _storageCredentialsValidator.ValidateCredentialsAsync(account, cancellationToken);
+                await _storageCredentialsValidator.ValidateCredentialsAsync(account, cancellationToken, account == StorageAccount);
```
mathewc: I think this validator is also used for Dashboard accounts? If so, it would be primary if either StorageAccount or DashboardAccount (both require queues access).

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
```diff
@@ -27,21 +27,21 @@ internal class DefaultStorageCredentialsValidator : IStorageCredentialsValidator
             StorageCredentials credentials = account.Credentials;
 
             // Avoid double-validating the same account and credentials.
-            if (_validatedCredentials.Contains(credentials))
+            if ((!isPrimaryAccount && _validatedCredentials.Contains(credentials)) || _primaryCredentials == credentials)
```
mathewc: I don't understand why these changes are needed.

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
```diff
@@ -56,9 +56,36 @@ internal class DefaultStorageCredentialsValidator : IStorageCredentialsValidator
...
+                    await queue.ExistsAsync(cancellationToken);
+                    _primaryCredentials = account.Credentials;
+                }
+                catch (OperationCanceledException)
+                {
+                    throw;
+                }
+                catch (StorageException exception)
+                {
+                    if (!exception.IsNotFoundQueueNotFound())
```
mathewc: Could this check catch more errors than we want it to? Can you modify it so that it also verifies you're getting the expected 4xx status code so you know the failure is not a 5xx Service Unavailable or something? Looks like currently, if you got a 503, you'd complain about the account.

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
```diff
@@ -56,9 +56,36 @@ internal class DefaultStorageCredentialsValidator : IStorageCredentialsValidator
...
+                    {
+                        string message = String.Format(CultureInfo.CurrentCulture,
+                            "Invalid storage account '{0}'. The primary storage account (AzureWebJobsStorage) must be a general "
```
mathewc: If it turns out this also applies to Dashboard accounts, then we can generalize the wording to:  "... Primary storage accounts must be general purpose accounts, not restricted blob storage accounts."

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageAccountProvider.cs
mamaso: Validator is used for all storage accounts, yep.  I'll look into the dashboard usage

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
mamaso: It's a corner case if a secondary account and a primary account use the same connection string.

1. Blob only storage account as secondary storage is validated (does not do queue check)
2. The same account is also used as primary
3. Because it already exists in validatedCredentials, the queue check is skipped

Might not be worth the more complex code, I doubt this will get hit and if it does we'll throw an error anyway (though likely more cryptic)

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
mamaso: Yep, it will.  Maybe I'll add the extension method to test the exception back in of 'queue not resolved' to be specific

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageAccountProvider.cs
```diff
@@ -153,6 +153,7 @@ private IStorageAccount StorageAccount
         public async Task<IStorageAccount> GetAccountAsync(string connectionStringName, CancellationToken cancellationToken)
         {
             IStorageAccount account = null;
+            var isPrimary = true;
```
mamaso: avoiding the StorageAccount / DashboardAccount getters and setters as they invoke parsing

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
```diff
@@ -56,9 +56,37 @@ internal class DefaultStorageCredentialsValidator : IStorageCredentialsValidator
...
+                    await queue.ExistsAsync(cancellationToken);
+                    _primaryCredentials = account.Credentials;
+                }
+                catch (OperationCanceledException)
+                {
+                    throw;
+                }
+                catch (StorageException exception)
+                {
+                    if (exception.Message.Contains("The remote name could not be resolved"))
```
mamaso: @mathewc Responding to your previous comments about exception filtering, there is no status code to check as this is thrown from client side.  Inspecting the message is probably the easiest way to do this

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
```diff
@@ -16,8 +15,9 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
     internal class DefaultStorageCredentialsValidator : IStorageCredentialsValidator
     {
         private readonly HashSet<StorageCredentials> _validatedCredentials = new HashSet<StorageCredentials>();
+        private StorageCredentials _primaryCredentials = null;
 
-        public async Task ValidateCredentialsAsync(IStorageAccount account, CancellationToken cancellationToken)
+        public async Task ValidateCredentialsAsync(IStorageAccount account, CancellationToken cancellationToken, bool isPrimaryAccount)
```
fabiocav: nit: move `CancellationToken` to last parameter

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
```diff
@@ -27,21 +27,21 @@ internal class DefaultStorageCredentialsValidator : IStorageCredentialsValidator
...
 
-        private static async Task ValidateCredentialsAsyncCore(IStorageAccount account,
-            CancellationToken cancellationToken)
+        private async Task ValidateCredentialsAsyncCore(IStorageAccount account,
+            CancellationToken cancellationToken, bool isPrimaryAccount)
         {
             // Verify the credentials are correct.
             // Have to actually ping a storage operation.
-            IStorageBlobClient client = account.CreateBlobClient();
+            var client = account.CreateBlobClient();
```
fabiocav: Revert this change. As a rule of thumb, if the type is not clear (as in the result of a method call), you want to be explicit about the type.

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
fabiocav: This is pretty fragile. Is there an inner exception in those cases (potentially a `WebException`) we can inspect?

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
```diff
@@ -56,9 +56,37 @@ internal class DefaultStorageCredentialsValidator : IStorageCredentialsValidator
...
+                {
+                    throw;
+                }
+                catch (StorageException exception)
+                {
+                    if (exception.Message.Contains("The remote name could not be resolved"))
+                    {
+                        string message = String.Format(CultureInfo.CurrentCulture,
+                            "Invalid storage account '{0}'. Primary storage accounts must be general "
+                            + "purpose accounts and not restricted blob storage accounts.", account.Credentials.AccountName);
```
fabiocav: Have you tested that an incorrect host/account in the connection string won't end up down this path? I expect that to be the case (as it will lead to the same failure) and if so, this message may be a bit misleading/confusing. We may need to update this verbiage, as without a reliable check against some account metadata, we can't assume that we failed because of a blob-only account.

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
fabiocav: Is the notion of a "primary account" something that is being introduced with this changes?

mamaso: From a user perspective you're right, it's an implementation detail that we haven't had to surface.  A 'primary account' has existed already:  it's the account in which logs / queues live, which is AzureWebJobsStorage/Dashboard.

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
mamaso: Yep, good feedback.  Changing to:
```c#
var webException = exception.GetBaseException() as WebException;
if (webException != null && webException.Status == WebExceptionStatus.NameResolutionFailure)
...
```

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
mamaso: Invalid hosts are caught with this call to the blob endpoint: https://github.com/Azure/azure-webjobs-sdk/pull/827/files#diff-d66d3cb7ae4b38012579fd160bda4f73R50

We only reach this point if the blob endpoint call is successful, indicating that credentials work, but there is no queue endpoint.

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
mamaso: yep, I was setting up isPrimaryAccount as a parameter with default = false, but as it turns out I still needed to update every call so it isn't worth it.  I'll make this switch

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
mamaso: Snuck in with earlier changes as I'd switched this from Blob -> Queue initially.

Interestingly that's the opposite advice that c++ gives, it's recommended to always use `auto`.  However, there aren't issues with implicit casting leading to unintended copies in c#, so I guess it makes sense :)

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
fabiocav: This is great!

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
fabiocav: At the end, in C#, the compiled result will be the same, but when using `var` for a variable being set as a result of a method call, the type isn't clear, which would require others to inspect the return type (hover, etc.), but it's worse when looking at code outside of an IDE since you wouldn't have that option. Just makes things clear. If you're instantiating a type, though, `var` all day long :)

PR #829, opened 2016-09-16
resilient to table getting deleted.

MikeStall:
Logging will recreate the table if it's deleted. 
Fix https://github.com/Azure/azure-webjobs-sdk/issues/708

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging/Internal/Utility.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging/Internal/Utility.cs
@@ -10,6 +10,7 @@ using System.IO;
 using System.Linq;
 using System.Net;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 using System.Xml;
 
@@ -76,11 +77,8 @@ namespace Microsoft.Azure.WebJobs.Logging
         }
 
 
-        private static async Task SafeCreateAsync(this CloudTable table)
+        private static async Task SafeCreateAsync(this CloudTable table, int intervalMilliseconds = 5000, int totalMilliseconds = 120*1000)
         {
-            var interval = TimeSpan.FromSeconds(5);
-            var total = TimeSpan.FromSeconds(120);
-
             while(true)
             {
                 try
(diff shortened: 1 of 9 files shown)


src/Microsoft.Azure.WebJobs.Logging/Internal/Utility.cs
```diff
@@ -72,5 +74,80 @@ public static string DecodeBase64(string base64)
...
+                    }
+                }
+                await Task.Delay(interval);
+                total = total.Subtract(interval);
+            }
+        }
+
+        // Write table entry. 
+        // If table doesn't exist (such as if it was deleted), then recreate it. 
+        public static async Task SafeWriteAsync(this CloudTable table, TableBatchOperation batch)
```
mathewc: I'd rename this to SafeExecuteAsync since it isn't write specific - it's just executing an arbitrary operation

src/Microsoft.Azure.WebJobs.Logging/Internal/Utility.cs
```diff
@@ -72,5 +74,80 @@ public static string DecodeBase64(string base64)
             string str = Encoding.UTF8.GetString(bytes);
             return str;
         }
+
+
+        private static async Task SafeCreateAsync(this CloudTable table)
```
mathewc: Please add tests verifying this new code works. E.g. execute a write to a table that doesn't exist, etc.

src/Microsoft.Azure.WebJobs.Logging/Internal/Utility.cs
MikeStall: I'd rather just keep this restricted to Write operations than generalize it and sort out the Delete semantics.

src/Microsoft.Azure.WebJobs.Logging/Internal/Utility.cs
MikeStall: I'll change it to have the logging stop eagerly creating the tables.  That will then mean all the tests will implicitly test the auto-create path.

src/Microsoft.Azure.WebJobs.Logging/Internal/Utility.cs
mathewc: Sure

src/Microsoft.Azure.WebJobs.Logging/Internal/Utility.cs
mathewc: That would be good. It won't be testing your while(true) retry stuff though, ensuring that we don't go in to an infinite loop, etc.

MikeStall: I have an idea to unit test the while(true) part. I'll include in next iteration.

mathewc:
:ship:

PR #830, opened 2016-09-19
Correct connection string name in exception if none is given

FPar:
If no connection string in the config is set, an exception is thrown. The message of this exception says, that there is no connection string with the name "AzureWebJobs". This is misleading, because the expected connection string name is "AzureWebJobsServiceBus".


mathewc:
Merged, thanks for fixing this :) FYI - I also added a test for this. Thanks.

Issue #811, opened 2016-09-01
Function Monitor APIs requires the Function Name to be modified to lower case & remove dash (-) character if any


panchagnula:
When using the API <scmsite>/azurejobs/api/functions/definitions/" + funcName + "/invocations?limit=20";
the funcName above needs to be modified to all lowercase & remove any "-" characters if exist other wise we get a 404.

MikeStall:
This may be by-design.  
The funcName argument should be the FunctionId, returned by other APIs like 
api/functions/invocations/recent.  

Where are you getting the argument from?  If the other APIs are giving you a function id that when passed to this function gives you a 404, that's a bug.

panchagnula:
Copying the notes from email conversation with Mike here 

API is called from here I am calling the API from FunctionsPortal https://github.com/projectkudu/AzureFunctionsPortal/blob/dev/AzureFunctions.Client/app/components/function-monitor.component.ts

The API <scmsite>/azurejobs/api/functions/definitions?Limit=100 HTTP/1.1 returns a full list of functions
and we filter out the selected function from this list using the functionName(==name of the folder), functionId is not something exposed to be used. 
See the method ‘getAggregateErrorsAndInvocationsForSelectedFunction’ here https://github.com/projectkudu/AzureFunctionsPortal/blob/dev/AzureFunctions.Client/app/services/function-monitor.service.ts

MikeStall:
We agreed on some naming restrictions: https://github.com/Azure/azure-webjobs-sdk-script/issues/15 
Notably, function names are case-insensitive case-preserving.  
So callers to the dashboard APIs should not need to call ToLower(). 
But dash is a legal character and shouldn't require any mangling.

PR #831, opened 2016-09-19
Make logger use new casing rules

MikeStall:
Make Logging  comply with Function naming rules. Notably, case insensitive and case preserving.

Fixes https://github.com/Azure/azure-webjobs-sdk/issues/811


mathewc:
:ship:

MikeStall:
Change has been Pushed.

PR #836, opened 2016-09-20
refactoring FunctionExecutor so we can pass the invokeTask to FunctionTimeoutException

brettsam:
Required to fix https://github.com/Azure/azure-webjobs-sdk-script/issues/655. With the Task included in the exception, an ExceptionHandler can wait and see if the task exits on its own after the token is canceled. This means we won't have to restart. 

The PR for script is also ready -- https://github.com/Azure/azure-webjobs-sdk-script/pull/708

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/FunctionTimeoutException.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/FunctionTimeoutException.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Runtime.Serialization;
+using System.Threading.Tasks;
 
 namespace Microsoft.Azure.WebJobs.Host
 {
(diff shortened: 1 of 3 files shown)


src/Microsoft.Azure.WebJobs.Host/FunctionTimeoutException.cs
```diff
@@ -33,17 +34,23 @@ protected FunctionTimeoutException(SerializationInfo info, StreamingContext cont
...
 
         /// <summary>
         /// The function timeout value that expired.
         /// </summary>
         public TimeSpan Timeout { get; set; }
 
+        /// <summary>
+        /// The task that did not complete due to a timeout.
+        /// </summary>
+        public Task FunctionTask { get; set; }
```
mathewc: nit: I wonder if you shouldn't just name this "Task", like you use "Timeout" above, rather than "FunctionTimeout"

mathewc:
One rename suggestion, otherwise looks good. :ship:

src/Microsoft.Azure.WebJobs.Host/FunctionTimeoutException.cs
brettsam: Seems reasonable. I'll do that.

src/Microsoft.Azure.WebJobs.Host/FunctionTimeoutException.cs
```diff
@@ -33,17 +34,23 @@ protected FunctionTimeoutException(SerializationInfo info, StreamingContext cont
         }
 
         /// <inheritdoc/>
-        public FunctionTimeoutException(string message, Guid instanceId, string methodName, TimeSpan timeout, Exception innerException)
+        public FunctionTimeoutException(string message, Guid instanceId, string methodName, TimeSpan timeout, Task functionTask, Exception innerException)
```
MikeStall: >Task functionTask [](start = 110, length = 17)

Is this just diagnostics? Or do we use it somewhere at runtime?

brettsam: We'll use it in Script. We don't use it anywhere here (but we could). In Script, we have a custom IWebJobsExceptionHandler that will waita few seconds to see if the Task exits gracefully. If it does, we can continue on without restarting the site. See here for usage: https://github.com/Azure/azure-webjobs-sdk-script/pull/708

MikeStall: Ok, cool. :shipit:

PR #838, opened 2016-09-22
Fix 625, allow double dash in name 

MikeStall:
Use new EventProcessorHost sdk to pass a blob path. This lets us handle arbitrary event hub names.

Fix https://github.com/Azure/azure-webjobs-sdk-script/issues/625

### DIFF
--- a/src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
+++ b/src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
@@ -38,7 +38,7 @@ namespace Microsoft.Azure.WebJobs.ServiceBus
         /// Name of the blob container that the EventHostProcessor instances uses to coordinate load balancing listening on an event hub. 
         /// Each event hub gets its own blob prefix within the container. 
         /// </summary>
-        public const string LeaseContainerName = "AzureWebJobsEventHub";
+        public const string LeaseContainerName = "azure-webjobs-eventhub";
 
         /// <summary>
         /// default constructor. Callers can reference this without having any assembly references to service bus assemblies. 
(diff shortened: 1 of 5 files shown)


src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
```diff
@@ -256,6 +270,90 @@ EventProcessorHost IEventHubProvider.GetEventProcessorHost(string eventHubName,
...
+                return string.Format(CultureInfo.InvariantCulture, "::{0:X4}", ordinalValue);
+            }
+        }
+                
+        // Escape a blob path.  
+        // For diagnostics, we want human-readble strings that resemble the input. 
+        // Inputs are most commonly alphanumeric with a fex extra chars (dash, underscore, dot). 
+        // Escape character is a ':', which is also escaped. 
+        // Blob names are case sensitive; whereas input is case insensitive, so normalize to lower.  
+        private static string EscapeBlobPath(string path)
```
mathewc: Need to add tests for all these new helper methods. That would also help us better visualize the transformations.

src/Microsoft.Azure.WebJobs.ServiceBus/packages.config
```diff
@@ -1,15 +1,15 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="Microsoft.Azure.KeyVault.Core" version="1.0.0" targetFramework="net45" />
-  <package id="Microsoft.Azure.ServiceBus.EventProcessorHost" version="2.2.5" targetFramework="net45" />
+  <package id="Microsoft.Azure.ServiceBus.EventProcessorHost" version="2.2.6" targetFramework="net45" />
```
mathewc: You also need to update the WebJobs.ServiceBus.nuspec. We have a build task that should be complaining about the mismatched version (or maybe we only added that to the other repos?)

MikeStall: Good catch. No complaints from build.

mathewc:
:ship:

MikeStall:
Merged.

PR #839, opened 2016-09-22
Fixing bug in ServiceBusTriggerBinding

mathewc:
Fix for issue reported here: https://github.com/projectkudu/AzureFunctionsPortal/issues/585

### DIFF
--- a/src/Microsoft.Azure.WebJobs.ServiceBus/Triggers/ServiceBusTriggerBinding.cs
+++ b/src/Microsoft.Azure.WebJobs.ServiceBus/Triggers/ServiceBusTriggerBinding.cs
@@ -30,30 +30,31 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Triggers
         private readonly AccessRights _accessRights;
         private readonly ServiceBusConfiguration _config;
 
-        public ServiceBusTriggerBinding(string parameterName, Type parameterType, 
-            ITriggerDataArgumentBinding<BrokeredMessage> argumentBinding, ServiceBusAccount account, string queueName, AccessRights accessRights, ServiceBusConfiguration config)
+        public ServiceBusTriggerBinding(string parameterName, Type parameterType, ITriggerDataArgumentBinding<BrokeredMessage> argumentBinding, ServiceBusAccount account,
+            AccessRights accessRights, ServiceBusConfiguration config, string queueName)
+            : this(parameterName, parameterType, argumentBinding, account, accessRights, config)
+        {
+            _queueName = queueName;
+            _entityPath = queueName;
+        }
+
+        public ServiceBusTriggerBinding(string parameterName, Type parameterType, ITriggerDataArgumentBinding<BrokeredMessage> argumentBinding, ServiceBusAccount account,
+            AccessRights accessRights, ServiceBusConfiguration config, string topicName, string subscriptionName)
+            : this(parameterName, parameterType, argumentBinding, account, accessRights, config)
+        {
+            _topicName = topicName;
+            _subscriptionName = subscriptionName;
+            _entityPath = SubscriptionClient.FormatSubscriptionPath(topicName, subscriptionName);
+        }
+
+        private ServiceBusTriggerBinding(string parameterName, Type parameterType, ITriggerDataArgumentBinding<BrokeredMessage> argumentBinding, 
+            ServiceBusAccount account, AccessRights accessRights, ServiceBusConfiguration config) 
         {
             _parameterName = parameterName;
             _converter = CreateConverter(parameterType);
             _argumentBinding = argumentBinding;
             _account = account;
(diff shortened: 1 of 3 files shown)


src/Microsoft.Azure.WebJobs.ServiceBus/Triggers/ServiceBusTriggerBinding.cs
```diff
@@ -30,30 +30,31 @@ internal class ServiceBusTriggerBinding : ITriggerBinding
...
         private readonly ServiceBusConfiguration _config;
 
-        public ServiceBusTriggerBinding(string parameterName, Type parameterType, 
-            ITriggerDataArgumentBinding<BrokeredMessage> argumentBinding, ServiceBusAccount account, string queueName, AccessRights accessRights, ServiceBusConfiguration config)
+        public ServiceBusTriggerBinding(string parameterName, Type parameterType, ITriggerDataArgumentBinding<BrokeredMessage> argumentBinding, ServiceBusAccount account,
+            AccessRights accessRights, ServiceBusConfiguration config, string queueName)
+            : this(parameterName, parameterType, argumentBinding, account, accessRights, config)
         {
-            _parameterName = parameterName;
-            _converter = CreateConverter(parameterType);
```
mathewc: Here I'm removing duplicated constructor code to make this more maintainable. The actual bug was that we weren't setting _converter in the topic ctor, only the queue ctor. That caused the null ref exception that was reported.

test/Microsoft.Azure.WebJobs.ServiceBus.UnitTests/Triggers/ServiceBusTriggerBindingIntegrationTests.cs
```diff
@@ -19,14 +19,16 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.UnitTests.Triggers
...
+        private ITriggerBinding _topicBinding;
 
         public ServiceBusTriggerBindingIntegrationTests()
         {
             IQueueTriggerArgumentBindingProvider provider = new UserTypeArgumentBindingProvider();
             ParameterInfo pi = new StubParameterInfo("parameterName", typeof(UserDataType));
             var argumentBinding = provider.TryCreate(pi);
-            _binding = new ServiceBusTriggerBinding("parameterName", typeof(UserDataType), argumentBinding, null, "queueName", AccessRights.Manage, new ServiceBusConfiguration());
+            _queueBinding = new ServiceBusTriggerBinding("parameterName", typeof(UserDataType), argumentBinding, null, AccessRights.Manage, new ServiceBusConfiguration(), "queueName");
+            _topicBinding = new ServiceBusTriggerBinding("parameterName", typeof(UserDataType), argumentBinding, null, AccessRights.Manage, new ServiceBusConfiguration(), "subscriptionName", "topicName");
```
mathewc: Regression test - this failed before.

src/Microsoft.Azure.WebJobs.ServiceBus/Triggers/ServiceBusTriggerBinding.cs
```diff
@@ -94,7 +95,8 @@ public string EntityPath
         public async Task<ITriggerData> BindAsync(object value, ValueBindingContext context)
         {
             BrokeredMessage message = value as BrokeredMessage;
-            if (message == null && !_converter.TryConvert(value, out message))
```
mathewc: For topics, _converted was null here

fabiocav:
ship-it-squirrel

src/Microsoft.Azure.WebJobs.ServiceBus/Triggers/ServiceBusTriggerBinding.cs
```diff
@@ -94,7 +95,8 @@ public string EntityPath
...
+            if (message == null && _converter != null && 
+                !_converter.TryConvert(value, out message))
```
brettsam: I know `_converter` shouldn't be null anymore -- but if it is, won't this just skip over the throw here now? I may be reading it wrong, too :-)

mathewc: This if block is about trying to do a conversion. We only want to do that if the message isn't already a BrokeredMessage, and we actually have a converter. I just put this here to be defensive, since this was the source of the null ref. converter can't actually be null anymore

mathewc: It'll throw elsewhere you're right though, but with a better error message hopefully

mathewc: I'll just remove the check I added though to remove confusion

Issue #825, opened 2016-09-14
Allow EventHub triggers to bind to byte[] 


MikeStall:
[EventHubTrigger("hub")] byte[] one --> binds to one event as a byte[]

[EventHubTrigger("hub")] byte[][] many--> binds to multiple events. 

This is technically a breaking change since byte[] today would bind to a multiple of type 'byte'; not a single of type byte[].

christopheranderson:
@MikeStall - do you want to pick this up soon?

mathewc:
Yeah, hopefully we can get this out in the next release.

MikeStall:
If we register a Item --> byte[] with the converter manager, this should work.

PR #841, opened 2016-09-23
Allow EventData to bind to byte[].

MikeStall:
The heavy lifting is in the common code; and the extension just provides the converter.

This really means:
1. EventHub's config can register a EventData-->Byte[] converter
2. The trigger support leverages the converter manager to find that converter.
If no such converter is found, then we have the same behavior as today.

Fixes https://github.com/Azure/azure-webjobs-sdk/issues/825

### DIFF
--- a/src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
+++ b/src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
@@ -278,6 +278,7 @@ namespace Microsoft.Azure.WebJobs.ServiceBus
             cm.AddConverter<string, EventData>(ConvertString2EventData);
             cm.AddConverter<EventData, string>(ConvertEventData2String);
             cm.AddConverter<byte[], EventData>(ConvertBytes2EventData); // direct, handles non-string representations
+            cm.AddConverter<EventData, byte[]>(ConvertEventData2Bytes); // direct, handles non-string representations
 
             var bf = new BindingFactory(nameResolver, cm);
 
@@ -306,6 +307,12 @@ namespace Microsoft.Azure.WebJobs.ServiceBus
         {
             var eventData = new EventData(input);
             return eventData;
+        }
+
+        private static byte[] ConvertEventData2Bytes(EventData input)
+        {
+            var bytes = input.GetBytes();
+            return bytes;
         }
 
         private static EventData ConvertString2EventData(string input)
--- a/test/Microsoft.Azure.WebJobs.Host.UnitTests/Common/TriggerTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.UnitTests/Common/TriggerTests.cs
@@ -43,6 +43,25 @@ namespace Microsoft.Azure.WebJobs.Host.UnitTests.Common
                 this.Finished();
             }
         }
+                
+        public class FunctionsByteArray : FunctionsBase
+        {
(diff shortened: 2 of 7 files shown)


test/Microsoft.Azure.WebJobs.Host.UnitTests/Common/TriggerTests.cs
```diff
@@ -214,15 +233,83 @@ public void TestSingleStringDispatch()
...
+            Assert.Equal(1, items.Length);
+            
+            // This uses the Item --> byte[] converter. Dispatch as a single item.
+            // Received as 1 object, a byte[]. 
+            var bytes = System.Text.Encoding.UTF8.GetBytes(e0.Message);
+            Assert.Equal(bytes, items[0]);
+        }
+                
+        [Fact]
+        public void TestByteArrayDispatch2()
```
mathewc: can you please give this test and the one below it more descriptive names so it is clear what they're verifying?

test/Microsoft.Azure.WebJobs.Host.UnitTests/Common/TriggerTests.cs
```diff
@@ -43,6 +43,25 @@ public void MultipleTrigger([FakeQueueTrigger] FakeQueueData[] batch)
...
+        public class FunctionsByteArray : FunctionsBase
+        {
+            public void Trigger([FakeQueueTrigger] byte[] single)
+            {
+                _collected.Add(single);
+                this.Finished();
+            }
+        }
+
+        // Exact semantics depend on how this is registered.
```
mathewc: On how what is registered?

test/Microsoft.Azure.WebJobs.Host.UnitTests/FakeQueue/FakeQueueData.cs
```diff
@@ -7,10 +7,13 @@ namespace Microsoft.Azure.WebJobs.Host.UnitTests
     // Fake queue message. Equivalent of CloudQueueMessage or EventData
     public class FakeQueueData
     {
-        // This correpsonds to string & poco conversion. 
+        // This corresonds to string & poco conversion.
```
mathewc: Try again :)

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
```diff
@@ -278,6 +278,7 @@ void IExtensionConfigProvider.Initialize(ExtensionConfigContext context)
             cm.AddConverter<string, EventData>(ConvertString2EventData);
             cm.AddConverter<EventData, string>(ConvertEventData2String);
             cm.AddConverter<byte[], EventData>(ConvertBytes2EventData); // direct, handles non-string representations
+            cm.AddConverter<EventData, byte[]>(ConvertEventData2Bytes); // direct, handles non-string representations
```
mathewc: Just to make sure - we'll now handle both byte[] and byte[][] automatically, w/o the user having to register any other converters?

mathewc:
Just a few minor comments. Once those are addressed 🚢

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
MikeStall: Correct! They just register a single Item-->Byte[] converter, and that's enough for us to figure everything out.

MikeStall:
CR feedback addressed and pushed.

Issue #840, opened 2016-09-22
In new Dashboard, Search button leads to yellow screen of death


davidebbo:
Steps:

- In a function app, go to dashboard
- Click on Search Blobs

Result: ASP.NET Yellow Screen of Death

Note: it works in the non-beta dashboard

MikeStall:
This is a dup of https://github.com/Azure/azure-webjobs-sdk/issues/677 
They're crashing due to unregistered (Ninject) components. 

We should disable both buttons:  
1. Search blobs doesn't work in the Azure Functions; 
2. 'About' just shows the indexer logs; but Functions logging doesn't use indexing.

davidebbo:
Note that the same issue repros if you're using the new Dashboard with a legacy site. So we can't just disable the buttons as that would regress existing functionality. But potentially we could disable them only for Function apps, and fix the issue for legacy sites.

MikeStall:
I meant disable when used for Functions.  New dashboard on a legacy site should work.

davidebbo:
Understood. So it sounds like there are two distinct work items:

1. Disable those buttons for functions
2. Fix the error for legacy scenarios

MikeStall:
Actually, I think I can make SearchBlobs just work outright in Functions.  I'll go for that. 
For legacy - I just tested that and it works. Are you sure it's  a legacy site?

davidebbo:
The legacy site I had tried on may have been dirty (might have been switched back and forth between old/new before), so let’s ignore if it doesn’t repro for you.

lindydonna:
@MikeStall Did the disable button issue get fixed in your recent commits?

MikeStall:
@lindydonna  - It's fixed in this PR: https://github.com/Azure/azure-webjobs-sdk/pull/846 
The PR is outstanding and hasn't been committed yet.

MikeStall:
PR was commited. This is fixed.

Issue #677, opened 2016-03-28
WebJobs dashboard has broken links


davidebbo:
*Copying from https://github.com/projectkudu/WebJobsPortal/issues/240. Opened by @BilalAlam*

![image](https://cloud.githubusercontent.com/assets/17733334/14085844/0a092c4e-f4d7-11e5-8481-fa297b352b7c.png)

The "search blobs" link and the "about" link both don't seem to work at all.  We should fix that up or avoid showing the links.

MikeStall:
Fixed.

PR #846, opened 2016-09-26
Fix "Search for blobs" link in dashboard.

MikeStall:
Fix "About link" - it was showing indexing errors.

https://github.com/Azure/azure-webjobs-sdk/issues/840
https://github.com/Azure/azure-webjobs-sdk/issues/677

### DIFF
--- a/src/Dashboard/App_Start/AppModule.cs
+++ b/src/Dashboard/App_Start/AppModule.cs
@@ -77,6 +77,7 @@ namespace Dashboard
                 Bind<IHeartbeatValidityMonitor>().To<NullHeartbeatValidityMonitor>();
 
                 Bind<IAborter>().To<NullAborter>();
+                Bind<IInvoker>().To<NullInvoker>();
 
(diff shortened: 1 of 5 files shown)


mathewc:
How can we add regression tests that ensure all the configurations we're making on the "fast logger" app init path? We either need some unit tests (less expensive) that verify this, or perhaps some Dashboard UI tests (more expensive) that verify this.

mathewc:
Is this review still active? I was wondering if there are any tests in place for this stuff. That would be ideal. I'm OK with just getting this fix in if you think adding tests will be too hard, but I really think we need them.

MikeStall:
It's still active. It got pushed to the back burner while I tried to think how to test it. I agree we need some strong testing.

MikeStall:
I've filed https://github.com/Azure/azure-webjobs-sdk/issues/857 and assigned to me to track the testing issues. I'll go ahead and push this fix since it's a small delta but fixes some obvious things.

PR #849, opened 2016-09-27
fixing threading bug in logging

### LINKED ISSUE #675: keep getting Unhandled Exception: System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less
if I have multiple jobs running at the same time and using log.writeline to write log every a few seconds, I get the following error occasionally. This causes our jobs fail sometime and we can't catch this error to report the issue. it looks like StringBuilder in UpdateOutputLogCommand.UpdateOutputBlob has concurrency issue.

Error:
[03/25/2016 14:22:29 > bdae78: ERR ] 
[03/25/2016 14:22:29 > bdae78: ERR ] Unhandled Exception: System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection.
[03/25/2016 14:22:29 > bdae78: ERR ] Parameter name: chunkLength
[03/25/2016 14:22:29 > bdae78: ERR ]    at System.Text.StringBuilder.ToString()
[03/25/2016 14:22:29 > bdae78: ERR ]    at System.IO.StringWriter.ToString()
[03/25/2016 14:22:29 > bdae78: ERR ]    at Microsoft.Azure.WebJobs.Host.Loggers.UpdateOutputLogCommand.<UpdateOutputBlob>d__10.MoveNext()
[03/25/2016 14:22:29 > bdae78: ERR ] --- End of stack trace from previous location where exception was thrown ---
[03/25/2016 14:22:29 > bdae78: ERR ]    at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[03/25/2016 14:22:29 > bdae78: ERR ]    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[03/25/2016 14:22:29 > bdae78: ERR ]    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[03/25/2016 14:22:29 > bdae78: ERR ]    at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd(Task task)
[03/25/2016 14:22:29 > bdae78: ERR ]    at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
[03/25/2016 14:22:29 > bdae78: ERR ]    at Microsoft.Azure.WebJobs.Host.Loggers.UpdateOutputLogCommand.<TryExecuteAsync>d__3.MoveNext()
[03/25/2016 14:22:29 > bdae78: ERR ] --- End of stack trace from previous location where exception was thrown ---
[03/25/2016 14:22:29 > bdae78: ERR ]    at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[03/25/2016 14:22:30 > bdae78: ERR ]    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[03/25/2016 14:22:30 > bdae78: ERR ]    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[03/25/2016 14:22:30 > bdae78: ERR ]    at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd(Task task)
[03/25/2016 14:22:30 > bdae78: ERR ]    at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
[03/25/2016 14:22:30 > bdae78: ERR ]    at Microsoft.Azure.WebJobs.Host.Timers.RecurrentTaskSeriesCommand.<ExecuteAsync>d__0.MoveNext()
[03/25/2016 14:22:30 > bdae78: ERR ] --- End of stack trace from previous location where exception was thrown ---
[03/25/2016 14:22:30 > bdae78: ERR ]    at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[03/25/2016 14:22:30 > bdae78: ERR ]    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[03/25/2016 14:22:30 > bdae78: ERR ]    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[03/25/2016 14:22:30 > bdae78: ERR ]    at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd(Task task)
[03/25/2016 14:22:30 > bdae78: ERR ]    at System.Runtime.CompilerServices.TaskAwaiter`1.GetResult()
[03/25/2016 14:22:30 > bdae78: ERR ]    at Microsoft.Azure.WebJobs.Host.Timers.TaskSeriesTimer.<RunAsync>d__d.MoveNext()
[03/25/2016 14:22:30 > bdae78: ERR ] --- End of stack trace from previous location where exception was thrown ---
[03/25/2016 14:22:30 > bdae78: ERR ]    at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
[03/25/2016 14:22:30 > bdae78: ERR ]    at Microsoft.Azure.WebJobs.Host.Timers.BackgroundExceptionDispatcher.<>c__DisplayClass1.<Throw>b__0()
[03/25/2016 14:22:30 > bdae78: ERR ]    at System.Threading.ThreadHelper.ThreadStart_Context(Object state)
[03/25/2016 14:22:30 > bdae78: ERR ]    at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
[03/25/2016 14:22:30 > bdae78: ERR ]    at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
[03/25/2016 14:22:30 > bdae78: ERR ]    at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
[03/25/2016 14:22:30 > bdae78: ERR ]    at System.Threading.ThreadHelper.ThreadStart()
[03/25/2016 14:22:30 > bdae78: SYS ERR ] Job failed due to exit code -532462766
[03/25/2016 14:22:30 > bdae78: SYS INFO] Process went down, waiting for 0 seconds

brettsam:
Fixes #675. 

`StringBuilder.ToString()` first allocates the string, then starts copying the buffered data to it until it reaches the end of its data. If you `Append()` more strings after the output string has been allocated, it will pass the end of the allocated string and you'll end up with the Exception we were seeing. 

By snapshotting the Length, we don't have that problem. Now the copying stops at the appropriate place, even if the `Append()` is called during `ToString()`. Thanks for brainstorming the solution @fabiocav!

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Loggers/UpdateOutputLogCommand.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Loggers/UpdateOutputLogCommand.cs
@@ -139,7 +139,10 @@ namespace Microsoft.Azure.WebJobs.Host.Loggers
                     _synchronizedWriter.Flush();
                 }
 
-                snapshot = _innerWriter.ToString();
+                // Explicitly specify the length. Without this, any writes occurring 
+                // during the call to ToString() could throw.
+                StringBuilder innerBuilder = _innerWriter.GetStringBuilder();
+                snapshot = innerBuilder.ToString(0, innerBuilder.Length);
 
                 if (flushAndClose)
                 {


src/Microsoft.Azure.WebJobs.Host/Loggers/UpdateOutputLogCommand.cs
```diff
@@ -139,7 +139,10 @@ public void Dispose()
                     _synchronizedWriter.Flush();
                 }
 
-                snapshot = _innerWriter.ToString();
+                // Explicitly specify the length. Without this, any writes occurring
```
MikeStall: >// Explicitly specify the length. Without this, any writes occurring  [](start = 16, length = 69)

This is already under a lock; and we have a synchronized writer.    Do we still need those?

brettsam: The synchronized writer -- yes. The underlying `StringBuilder` used by `StringWriter` isn't thread-safe so that guarantees we don't have issues with calling `Append()` from multiple threads.
The lock -- I'm assuming it is still needed. This is the only place it's used, but I'm guessing the original writer wanted to make sure this block was locked. I'd be wary of removing it unless we were sure it wasn't needed.

MikeStall:
:shipit:

mathewc:
Were you able to reproduce this consistently before the fix, and not after? I wish we had a multi-threaded test for regressions in this area.

brettsam:
Yeah, it repro'ed very reliably. I turned down the flush timeout from 1 minute to 5 seconds and had a function spinning up a bunch of tasks and logging. I always hit it within a few flushes.

After my fix I had it running for several minutes without any exception. Looking through the `StringBuilder.ToString()` source, it was pretty clear what was happening.

I don't like fixing something without adding a test either. I thought about a reliable test that would prove this is fixed and couldn't really come up with one -- it's such a timing-based bug. But I'll give it a little more thought tomorrow and see if I can add something.

fabiocav:
👍

mathewc:
Can't you just make the repro you mentioned above a test? A higher level test like that might help us catch other issues in the future as well. Might require making the flush timeout test configurable.

If there isn't an easy way to get a regression test in place, then don't worry about it. I'm just glad to have this fixed :)

brettsam:
I've added a specific regression test that failed every time I ran it before the fix: https://github.com/Azure/azure-webjobs-sdk/commit/cabe07dac13f8f7afc16ab26aeb479aa6fa924dd

I agree that we need some kind of dedicated stress testing, but that's a larger issue that we can follow up on separately.

PR #851, opened 2016-09-27
MAke eventhub sending fit in 256k size limit

MikeStall:
EventHub can't send more than 256k of data at once. So the EventHub async collector needs to batch it. 
This is generally useful, so make the collector be public. 

Fix https://github.com/Azure/azure-webjobs-sdk/issues/834

Need to think of a good way to test this - none of the EventHub methods are virtual.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubAsyncCollector.cs
+++ b/src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubAsyncCollector.cs
@@ -26,7 +26,7 @@ namespace Microsoft.Azure.WebJobs.ServiceBus
         private const int BatchSize = 100;
 
         // Suggested to use 240k instead of 256k to leave padding room for headers.
-        private const int MaxByteSize = 240 * 1000; 
+        private const int MaxByteSize = 240 * 1024; 
 
         /// <summary>
         /// Create a sender around the given client. 
@@ -34,6 +34,10 @@ namespace Microsoft.Azure.WebJobs.ServiceBus
         /// <param name="client"></param>
         public EventHubAsyncCollector(EventHubClient client)
         {
+            if (client == null)
+            {
+                throw new ArgumentNullException("client");
+            }
             _client = client;
         }
 
@@ -45,6 +49,11 @@ namespace Microsoft.Azure.WebJobs.ServiceBus
         /// <returns></returns>
         public async Task AddAsync(EventData item, CancellationToken cancellationToken = default(CancellationToken))
         {
+            if (item == null)
+            {
+                throw new ArgumentNullException("item");
+            }
+
             while (true)
             {
                 lock (_list)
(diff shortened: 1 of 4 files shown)


src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubAsyncCollector.cs
```diff
@@ -1,50 +1,93 @@
...
+                {
+                    var size = (int)item.SerializedSizeInBytes;
 
-            if (flush)
-            {
-                await this.FlushAsync(cancellationToken);
+                    if (size > MaxByteSize)
+                    {
+                        // Single event is too large to add.
+                        string msg = string.Format("Event is too large. Event is approximately {0}b and max size is {0}b", size, MaxByteSize);
```
mathewc: I think your substitution indices are off - both are {0}

mathewc: Let's add tests for this

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubAsyncCollector.cs
```diff
@@ -1,50 +1,93 @@
...
+                        throw new InvalidOperationException(msg);
+                    }
+
+                    if ((_currentByteSize + size > MaxByteSize) || (_list.Count >= BatchSize))
```
mathewc: Can we just get rid of this empty if by applying Demorgans to it?

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubAsyncCollector.cs
```diff
@@ -1,50 +1,93 @@
...
-    // Core object to send events to EventHub. 
-    // Any user parameter that sends events will eventually get bound to this object. 
-    // This gets wrappers with various adapters and passed to the user function. 
-    internal class EventHubAsyncCollector : IAsyncCollector<EventData>
+    /// <summary>
+    /// Core object to send events to EventHub. 
+    /// Any user parameter that sends EventHub events will eventually get bound to this object. 
+    /// This will queue events and send in batches, also keeping under the 256kb event hub limit per batch. 
+    /// </summary>
+    public class EventHubAsyncCollector : IAsyncCollector<EventData>
```
MikeStall: It doesn't *need* to be public. But I made it public since it's useful in standalone cases. It's also useful for testing a WebJobs function where I want to directly invoke the C# method outside of the model binder.  

The public surface tests don't appear to include the service bus extension.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubAsyncCollector.cs
MikeStall: Yes.  I thought it'd be easier to follow the logic like this, but I can flip it.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubAsyncCollector.cs
mathewc: I just pushed public surface tests for ServiceBus and your new Logging. Please pull and rebase. You should also review the public types in the Logging assembly, now that this test exposes the list. See commit https://github.com/Azure/azure-webjobs-sdk/commit/af8406f9faeebbd5649fa9038ace3c257cf41227

MikeStall: Awesome - I'll rebase on those.  I'd still like to make it public.

mathewc:
:ship:

Issue #784, opened 2016-08-08
Enrich TraceEvent to include invoked function name


abatishchev:
Here's how we're currently writing function invocation exceptions to Application Insights:
```csharp
var logger = container.GetInstance<IExceptionLogger>();
var traceMonitor = new TraceMonitor()
	.Filter(e => e.Exception != null)
	.Subscribe(f =>
		{
			foreach (var e in f.Events.ToArray())
			{
				logger.LogException(e.Source, e.Exception.InnerException ?? e.Exception);
			}
		});

config.Tracing.Tracers.Add(traceMonitor);
```
The problem is that `e.Source` always equals to `WebJobs.Execution`. Is it possible to replace it with respectful function name, or add a new property if this would be a breaking change otherwise?

mathewc:
Yes, the Source property is used for other purposes. We could add more details like this not as top level properties of the general `TraceEvent` type (since that event is also used in cases where a function isn't even being invoked) but as additional data in the `TraceEvent.Properties` property bag which was added for that purpose.

The place in the code where we would add these properties would be in `TraceWriterFunctionInstanceLogger`, e.g. [here](https://github.com/Azure/azure-webjobs-sdk/blob/master/src/Microsoft.Azure.WebJobs.Host/Loggers/TraceWriterFunctionInstanceLogger.cs#L27).

abatishchev:
Yes, this sounds exactly what we're looking for!
Is this class extendable/pluggable by users, or the changes should be done on the SDK side first? If the former than is it available in current stable release, or we need wait version 2.0 to go live?

mathewc:
The change needs to be made in the SDK itself, since we're emitting the event.

abatishchev:
I see. How can we make it extensible? I'd rather spend more time on this than on what to put there by default. Every project likely will have their very own requirements here.

btw what is current schedule, when to wait version 2.0?

mathewc:
We should just add the function name to the Properties collection. We might be able to leverage that in Functions. Currently in Functions we have our own custom trace writer that injects the function name.

abatishchev:
Please add! This will significantly improve the readability of the telemetry (we write logs and exceptions to AI).

brettsam:
`TraceEvent.Properties` will now include function information -- available properties found here: https://github.com/Azure/azure-webjobs-sdk/blob/master/src/Microsoft.Azure.WebJobs.Host/Loggers/TraceEventExtensions.cs

Issue #774, opened 2016-07-21
Expose WebJob Context Information To Custom TraceWriters


aaronhoffman:
In an attempt to make reviewing/querying WebJob logs easier, I took advantage of the `JobHostConfiguration.Tracing.Tracers` collection and added a custom `SqlTraceWriter` to my WebJob project ([sample code here](https://gist.github.com/aaronhoffman/3e319cf519eb8bf76c8f3e4fa6f1b4ae)).

This works pretty well but can get somewhat confusing due to WebJobs being executed in parallel.

Would it be possible to expose some of the [WebJob Context](https://blog.kloud.com.au/2016/03/14/azure-webjob-logs-demystified/) information to the TraceWriter?

Either via a static class or by passing it into the `TraceWriter.Trace()` method perhaps?

The various *`InstanceId`s could be added to the SQL Table to help identify logs from the same job.

christopheranderson:
Are there specific information on the context you'd like?

aaronhoffman:
From: https://blog.kloud.com.au/2016/03/14/azure-webjob-logs-demystified/

`HostInstanceId`, `FunctionInstanceId`, `Function.FullName`, and the original message payload would be nice (for message queue trigger jobs).

And I'm not sure if it would be possible, but if there was a way to get a unique value identifying the Web Site instance and the `Deployment Slot` the webjob is running in, that would be great! I'm not sure if it's appropriate to pull a couple values from an environment variable or not... `WEBSITE_INSTANCE_ID` maybe?

mathewc:
Related to #784 which asks for FunctionName.

brettsam:
`TraceEvent.Properties` will now include function information -- available properties found here: https://github.com/Azure/azure-webjobs-sdk/blob/master/src/Microsoft.Azure.WebJobs.Host/Loggers/TraceEventExtensions.cs

aaronhoffman:
I'd love to include this feature in a talk I'm giving in a couple weeks. What is the release schedule like for the related nuget package?

Looks like the last release was a 2.0.0 beta in July? https://www.nuget.org/packages/Microsoft.Azure.WebJobs/2.0.0-beta1 

49 commits behind current master https://github.com/Azure/azure-webjobs-sdk/compare/977072a...master

mathewc:
@brettsam can get it up on our myget feed for you to access. We won't be doing nuget.org update for a while.

PR #854, opened 2016-09-29
injecting function details into TraceEvent.Properties

brettsam:
For #774 and #784.

@mathewc -- The tests are not ready yet but wanted a quick review of the approach here. This adds the hostId, instanceId, and functionDescriptor values to the Properties collection in both our `TraceWriterFunctionInstanceLogger` and the `TraceWriter` we pass to the function itself.

I investigated including the trigger object as well. I could inject the *object* fairly easily with the `TraceWriter` that we pass as a parameter to functions. And I can include the *string* from `ToInvokeString()` with the `TraceWriterFunctionInstanceLogger`, but that felt odd to have one form in one message and another elsewhere. I'll keep looking into that.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInstanceTraceWriter.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInstanceTraceWriter.cs
@@ -1,0 +1,39 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+
+namespace Microsoft.Azure.WebJobs.Host.Executors
+{
+    /// <summary>
+    /// The <see cref="TraceWriter"/> passed to function instances. This class adds function instance details
+    /// to all trace events.
+    /// </summary>
+    internal class FunctionInstanceTraceWriter : TraceWriter
+    {
+        private TraceWriter _innerWriter;
+        private IFunctionInstance _instance;
+        private Guid _hostInstanceId;
+
+        public FunctionInstanceTraceWriter(IFunctionInstance instance, Guid hostInstanceId, TraceWriter innerWriter, TraceLevel level)
+            : base(level)
+        {
+            _innerWriter = innerWriter;
+            _instance = instance;
+            _hostInstanceId = hostInstanceId;
+        }
+
+        public override void Trace(TraceEvent traceEvent)
+        {
+            if (traceEvent == null)
+            {
+                throw new ArgumentNullException("traceEvent");
+            }
+
+            traceEvent.AddFunctionInstanceDetails(_hostInstanceId, _instance.FunctionDescriptor, _instance.Id);
+
+            _innerWriter.Trace(traceEvent);
+        }
+    }
+}
--- a/src/Microsoft.Azure.WebJobs.Host/Loggers/TraceEventExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Loggers/TraceEventExtensions.cs
@@ -1,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Azure.WebJobs.Host.Protocols;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    internal static class TraceEventExtensions
+    {
+        private const string FunctionDescriptorKey = "MS_FunctionDescriptor";
+        private const string FunctionInvocationIdKey = "MS_FunctionInvocationId";
+        private const string HostInstanceIdKey = "MS_HostInstanceId";
+
+        public static void AddFunctionInstanceDetails(this TraceEvent traceEvent, Guid hostInstanceId, FunctionDescriptor descriptor, Guid functionId)
+        {
+            traceEvent.Properties[HostInstanceIdKey] = hostInstanceId;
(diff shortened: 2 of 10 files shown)


brettsam:
Ready for review.

src/Microsoft.Azure.WebJobs.Host/Loggers/TraceEventExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+    internal static class TraceEventExtensions
+    {
+        private const string FunctionDescriptorKey = "MS_FunctionDescriptor";
+        private const string FunctionInvocationIdKey = "MS_FunctionInvocationId";
+        private const string HostIdKey = "MS_HostId";
+
+        public static void AddFunctionInstanceDetails(this TraceEvent traceEvent, Guid hostId, FunctionDescriptor descriptor, Guid functionId)
+        {
+            traceEvent.Properties[HostIdKey] = hostId;
+            traceEvent.Properties[FunctionDescriptorKey] = descriptor;
```
mathewc: FunctionDescriptor is public, right? So you're doing this as a way to include the function Name (which was the original ask) as well as additional details?

src/Microsoft.Azure.WebJobs.Host/Loggers/TraceEventExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    internal static class TraceEventExtensions
+    {
+        private const string FunctionDescriptorKey = "MS_FunctionDescriptor";
+        private const string FunctionInvocationIdKey = "MS_FunctionInvocationId";
+        private const string HostIdKey = "MS_HostId";
+
+        public static void AddFunctionInstanceDetails(this TraceEvent traceEvent, Guid hostId, FunctionDescriptor descriptor, Guid functionId)
+        {
+            traceEvent.Properties[HostIdKey] = hostId;
```
mathewc: Wondering what the scenario is for HostId? This will be the same for all logs, right? So I'm wondering how this would be useful for someones custom trace writer. If they wanted HostId they could access that themselves by passing it into their custom tracer on construction.

src/Microsoft.Azure.WebJobs.Host/Loggers/TraceEventExtensions.cs
brettsam: Yeah it's public. It includes a few other things that someone may want -- I didn't have a good reason to restrict them: https://github.com/Azure/azure-webjobs-sdk/blob/master/src/Microsoft.Azure.WebJobs.Protocols/FunctionDescriptor.cs

src/Microsoft.Azure.WebJobs.Host/Loggers/TraceWriterFunctionInstanceLogger.cs
```diff
@@ -24,28 +25,38 @@ public TraceWriterFunctionInstanceLogger(TraceWriter trace)
 
         public Task<string> LogFunctionStartedAsync(FunctionStartedMessage message, CancellationToken cancellationToken)
         {
-            _trace.Info(string.Format(CultureInfo.InvariantCulture, "Executing: '{0}' - Reason: '{1}'", message.Function.ShortName, message.FormatReason()), TraceSource.Execution);
+            string traceMessage = string.Format(CultureInfo.InvariantCulture, "Executing: '{0}' - Reason: '{1}'", message.Function.ShortName, message.FormatReason());
+            Trace(message.HostInstanceId, message.Function, message.FunctionInstanceId, traceMessage, TraceLevel.Info, TraceSource.Execution);
```
mathewc: nit: move the TraceLevel param to the front for clarity

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInstanceTraceWriter.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+
+namespace Microsoft.Azure.WebJobs.Host.Executors
+{
+    internal class FunctionInstanceTraceWriter : TraceWriter
```
mathewc: nit: add a comment block to this guy describing what he's for

src/Microsoft.Azure.WebJobs.Host/Loggers/TraceEventExtensions.cs
brettsam: HostID was explicitly asked for in #774. Don't you have to construct and register your tracer with the Config before you have a Host id? That's what I thought -- but I can double-check if there's any way to do that.

mathewc: They were asking for HostInstanceId not HostId. HostId is the same across all scaled out instances, HostInstanceId is unique per instance.

mathewc:
Agree with our general approach. I'd forget about adding the trigger object or anything like that for now.

src/Microsoft.Azure.WebJobs.Host/Loggers/TraceEventExtensions.cs
brettsam: Well, I learned something today :-)

brettsam:
Well, turns out I was using `HostInstanceId` anyway. I just was using the wrong parameter names -- I've updated all of that. Addressed other comments as well. Anything else?

mathewc:
:ship:

Issue #847, opened 2016-09-26
Execution logging can hit table storage limits


paulbatum:
A customer found their function app to be running very unreliably. On investigation we found that the host was crashing constantly with the below exception. We believe the table storage limit was being hit because of a very large parameter payload. This logging code path needs to be more defensive - for example the parameters should be truncated if they exceed some max length. Exceptions thrown from this code path should also not crash the host process.

```
Microsoft.WindowsAzure.Storage.StorageException: Element 10 in the batch returned an unexpected response code.
   at Microsoft.WindowsAzure.Storage.Core.Executor.Executor.EndExecuteAsync[T](IAsyncResult result)
   at Microsoft.WindowsAzure.Storage.Table.CloudTable.EndExecuteBatch(IAsyncResult asyncResult)
   at Microsoft.WindowsAzure.Storage.Core.Util.AsyncExtensions.<>c__DisplayClass1`1.<CreateCallback>b__0(IAsyncResult ar)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.Azure.WebJobs.Logging.Utility.<SafeExecuteAsync>d__8.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.Azure.WebJobs.Logging.LogWriter.<WriteBatchAsync>d__2c`1.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.Azure.WebJobs.Logging.LogWriter.<FlushIntancesAsync>d__1c.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[...]

PR #856, opened 2016-09-30
Truncate log entries with silly  long string values.

MikeStall:
Fix https://github.com/Azure/azure-webjobs-sdk/issues/847

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging/Entities/InstanceTableEntity.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging/Entities/InstanceTableEntity.cs
@@ -19,6 +19,8 @@ namespace Microsoft.Azure.WebJobs.Logging
 
         internal static InstanceTableEntity New(FunctionInstanceLogItem item)
         {
+            item.Truncate();
+
             string argumentJson = null;
             if (item.Arguments != null)
             {
--- a/src/Microsoft.Azure.WebJobs.Logging/FunctionInstanceLogItem.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging/FunctionInstanceLogItem.cs
@@ -14,6 +14,14 @@ namespace Microsoft.Azure.WebJobs.Logging
     /// </summary>
     public class FunctionInstanceLogItem : IFunctionInstanceBaseEntry
     {
+        // Max lengths of various fields.
+        // Except for output logging, these shouldn't get hit in practice.
+        // Truncate if they're exceeded. 
+        private const int MaxTriggerReasonLength = 200;
+        private const int MaxErrorLength = 500;
+        private const int MaxLogOutputLength = 2000 + 3;
+        private const int MaxParameterPayloadLength = 1000;
+
         /// <summary>Gets or sets the function instance ID.</summary>
         public Guid FunctionInstanceId { get; set; }
 
@@ -49,8 +57,7 @@ namespace Microsoft.Azure.WebJobs.Logging
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Usage", "CA2227:CollectionPropertiesShouldBeReadOnly")]
         public IDictionary<string, string> Arguments { get; set; }
 
-        // Direct inline capture for small log outputs. For large log outputs, this is faulted over to a blob. 
-        /// <summary></summary>
+        /// <summary>Direct inline capture for small log outputs. For large log outputs, this is faulted over to a blob. </summary>
         public string LogOutput { get; set; }
 
         /// <summary>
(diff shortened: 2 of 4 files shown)


src/Microsoft.Azure.WebJobs.Logging/FunctionInstanceLogItem.cs
```diff
@@ -49,8 +57,7 @@ public class FunctionInstanceLogItem : IFunctionInstanceBaseEntry
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Usage", "CA2227:CollectionPropertiesShouldBeReadOnly")]
         public IDictionary<string, string> Arguments { get; set; }
 
-        // Direct inline capture for small log outputs. For large log outputs, this is faulted over to a blob. 
-        /// <summary></summary>
+        /// <summary>Direct inline capture for small log outputs. For large log outputs, this is faulted over to a blob. </summary>
```
mathewc: How/where is this faulted over to blob?

src/Microsoft.Azure.WebJobs.Logging/FunctionInstanceLogItem.cs
```diff
@@ -153,6 +160,57 @@ public void Validate()
...
+            // This is fundamentally driven by performance. We need to fit log entries into table rows.
+            // Truncate to ensure that we're under table's maximum request payload size (4mb).
+            // None of these limits (except for output log)  should actually get hit in normal scenarios. 
+            this.TriggerReason = Truncate(this.TriggerReason, MaxTriggerReasonLength);
+            this.ErrorDetails = Truncate(this.ErrorDetails, MaxErrorLength);
+
+            // Logger should already have truncated this, but just in case. 
+            this.LogOutput = Truncate(this.LogOutput, MaxLogOutputLength);
+
+            // Arguments may have 1 larger argument for the trigger.
```
mathewc: Why do you have to do the double pass here? Can't you just do a single pass and truncate inline?

mathewc: Also, functions can have an abitrary set of parameters, so theoretically a function can still go over the limit if it has too many parameters, even with this truncation, right? Perhaps you should allocate a maximum size to the entire arguments list, and compute the max size for each based on that?

src/Microsoft.Azure.WebJobs.Logging/FunctionInstanceLogItem.cs
```diff
@@ -153,6 +160,57 @@ public void Validate()
...
+        /// </summary>
+        public void Truncate()
+        {
+            // This is fundamentally driven by performance. We need to fit log entries into table rows.
+            // Truncate to ensure that we're under table's maximum request payload size (4mb).
+            // None of these limits (except for output log)  should actually get hit in normal scenarios. 
+            this.TriggerReason = Truncate(this.TriggerReason, MaxTriggerReasonLength);
+            this.ErrorDetails = Truncate(this.ErrorDetails, MaxErrorLength);
+
+            // Logger should already have truncated this, but just in case.
```
mathewc: Where does logger do truncation?

src/Microsoft.Azure.WebJobs.Logging/FunctionInstanceLogItem.cs
MikeStall: It's not yet. That was an initial goal, but nyi.

src/Microsoft.Azure.WebJobs.Logging/FunctionInstanceLogItem.cs
MikeStall: Double pass because you can't modify a dictionary while enumerating it.

src/Microsoft.Azure.WebJobs.Logging/Entities/InstanceTableEntity.cs
```diff
@@ -19,6 +19,8 @@ internal class InstanceTableEntity : TableEntity
 
         internal static InstanceTableEntity New(FunctionInstanceLogItem item)
         {
+            item.Truncate();
```
MikeStall: This is where the truncation happens (when we're converting it to an actual table entity)

src/Microsoft.Azure.WebJobs.Logging/FunctionInstanceLogItem.cs
MikeStall: I added a code-review comment to call that out. It's in the InstanceTableEntity ctor.

src/Microsoft.Azure.WebJobs.Logging/FunctionInstanceLogItem.cs
MikeStall: I think for a large number of arguments, I'll just cut the whole thing.  This shouldn't happen in practice. 
I think the common case is 1 (potentially large) argument from the trigger; and several smaller ones (like output bindings).

src/Microsoft.Azure.WebJobs.Logging/FunctionInstanceLogItem.cs
mathewc: Then please remove the confusing comment :)

mathewc:
:ship:

src/Microsoft.Azure.WebJobs.Logging/FunctionInstanceLogItem.cs
MikeStall: yes.

Issue #684, opened 2016-03-30
WebJobs.Logging needs to support log purge / retention policies


mathewc:
The current design is based on a single table which won't allow us to efficiently delete / purge logs. We need to move to a rolling multi-table design which will allow us to delete old logs with a single table delete statement. This is how all other logging systems built on Table Storage are designed.

sandeep-at-airlift:
Is there a date for this being fixed? I have to decide whether to invest in a homegrown logging solution or not. this is killing our ability to debug issues.

mathewc:
We're not making any more investments in the Dashboard - the recommendation for logging going forward is to use Application Insights https://github.com/Azure/azure-webjobs-sdk/wiki/Application-Insights-Integration.

PR #862, opened 2016-10-04
switch logging over to shard per-table. 

MikeStall:
Allows deleting a table at a time.  Each month's worth of data gets its own table. 
Fix https://github.com/Azure/azure-webjobs-sdk/issues/684 

A few outstanding items here before :
1. The existing 30+ logging tests pass as is (The changes to the test file are cosmetic). But add more tests for the new corner cases (like range queries across tables). 
2. Optimize the LogReader.SafeExecuteQuerySegmentedAsync() method to use continuation token.


src/Microsoft.Azure.WebJobs.Logging/Internal/LogReader.cs
```diff
@@ -132,6 +156,93 @@ public LogReader(CloudTable table)
             return new Segment<ActivationEvent>(l.ToArray(), null);
         }
 
+        // Epochs may be sparse. 
+        // We may frequently start an enumeration at DateTime.MaxValue
+        class EpochTableIterator
```
MikeStall: This class needs to be cleaned up. And the SafeExecuteQuerySegmentedAsync specifically needs to be optimized to use continuation tokens. But that's just a contained implementation detail and I wanted to share the PR out to give a chance for feedback on the rest of this change.

test/Microsoft.Azure.WebJobs.Logging.FunctionalTests/LoggerTest.cs
```diff
@@ -9,51 +9,60 @@
...
         static string CommonFuncName1 = "gamma";
 
+        private List<CloudTable> _tables = new List<CloudTable>();
+
+        private string _tableNamePrefix = "logtestYY" + Guid.NewGuid().ToString("n");
+
+        private CloudTableClient _tableClient;
+
+        // Delete any tables we created. 
+        public void Dispose()
```
MikeStall: Dispose() gets called at the end of every test. I should have written the tests using this before hand rather than the try/finally. 
The changes in the test file here are primarily cosmeitc - there's no 'functional" change.

test/Microsoft.Azure.WebJobs.Logging.FunctionalTests/LoggerTest.cs
```diff
@@ -101,56 +110,124 @@ public void NormalizeFunctionName(string name, string expected)
...
+
+            // Various combinations of straddling an epoch boundary 
+            await Verify(reader, Before(times[1]), After(times[2]), logs[2], logs[1]); 
+            await Verify(reader, Before(times[1]), Before(times[2]), logs[1]); 
+            await Verify(reader, After(times[1]), Before(times[2]));
+
+            // Skipping over an empty epoch 
+            await Verify(reader, Before(times[1]), Before(times[3]), logs[2], logs[1]);
+
+            // Now... delete the middle table; and verify the other data is still there.
```
MikeStall: Here's where we test that table deletion actually works! This would fail before.

src/Microsoft.Azure.WebJobs.Logging/Internal/LogReader.cs
```diff
@@ -155,20 +184,186 @@ public LogReader(CloudTable table)
...
+                    if (epoch == physicalEpochs[i])
+                    {
+                        return epoch;
+                    }
+                }
+
+                return -1; // done 
+            }
+
+            // Helper workaround for this bug:
```
panchagnula: demo

src/Microsoft.Azure.WebJobs.Logging/Internal/LogReader.cs
```diff
@@ -155,20 +184,186 @@ public LogReader(CloudTable table)
...
+                for (var i = physicalEpochs.Length - 1; i >= 0; i--)
+                {
+
+                    if (epoch > physicalEpochs[i])
+                    {
+                        epoch = physicalEpochs[i];
+                        return epoch;
+                    }
+
+                    if (epoch == physicalEpochs[i])
```
mamaso: nit: combine these conditions with >= as it's essentially the same logic (unless if saving the assignment is critical)

src/Microsoft.Azure.WebJobs.Logging/Internal/LogReader.cs
```diff
@@ -155,20 +184,186 @@ public LogReader(CloudTable table)
...
+                    d[epoch] = table;
+                }
+                return d;
+            }
+
+            // Given an epoch, find the next one in physicalEpochs that it would fall into, using reverse-chronological order.              
+            // physicalEpochs - epochs for existing tables, sorted in ascending order.  This is generally short (expected under 50 elements)
+            // Returns a physical epoch (from the list passed in) that the requested epoch falls into 
+            // or -1 if there's no physical epoch below this.
+            private static long FindNext(long epoch, long[] physicalEpochs)
```
mamaso: what if we reversed the sorting order, can simplify this whole function to something like
`physicalEpochs.First((physical) => physical <= epoch)`

src/Microsoft.Azure.WebJobs.Logging/Internal/LogReader.cs
mamaso: see above comment as well

src/Microsoft.Azure.WebJobs.Logging/Internal/LogReader.cs
```diff
@@ -155,20 +184,186 @@ public LogReader(CloudTable table)
...
+
+            // Date range queries start with most recent (endTime) and then return entities in descending chronological order. 
+            public async Task<Segment<TElement>> SafeExecuteQuerySegmentedAsync<TElement>(
+                TableQuery<TElement> rangeQuery,
+                DateTime startTime, // Range for query 
+                DateTime endTime,
+                string continuationToken
+                ) where TElement : ITableEntity, new()
+            {
+                _tables = await InitAsync(_tableLookup);
```
mamaso: do we want to reload the tables every time we loop through?

src/Microsoft.Azure.WebJobs.Logging/Internal/LogReader.cs
```diff
@@ -155,20 +184,186 @@ public LogReader(CloudTable table)
...
-                realContinuationToken, 
-                cancellationToken);
+            var iter = new EpochTableIterator(_tableLookup);
+            var results = await iter.SafeExecuteQuerySegmentedAsync<RecentPerFuncEntity>(rangeQuery, queryParams.Start, queryParams.End);
+
+            return results.As<IRecentFunctionEntry>();
+        }
+
+        // Helper to run queries which can span multiple tables.         
+        private class EpochTableIterator
```
mamaso: nit: feel like EpochTableIterator is kind of a misnomer as it's api is not that of an iterator

src/Microsoft.Azure.WebJobs.Logging/Internal/TimeBucket.cs
```diff
@@ -27,5 +29,63 @@ public static long ConvertToBucket(DateTime dt)
             }
             return min;
         }
+
+        // Used to compute epoch. 
+        const long ticksBaseline = 633663648000000000L;
+        const long ticksPerMonth = 26785144663987L;
+
```
mamaso: nit: capitalize these constants

src/Dashboard/App_Start/AppModule.cs
```diff
@@ -134,46 +133,37 @@ public static void Load(ContainerBuilder builder)
             }
         }
 
-        // Get a fast log table name 
+        // Get a fast log table name. Returns a table name prefix if
```
mathewc: incomplete sentence?

src/Microsoft.Azure.WebJobs.Logging/ILogTableProvider.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+using Microsoft.WindowsAzure.Storage.Table;
+
+namespace Microsoft.Azure.WebJobs.Logging
+{
+    /// <summary>
+    /// Logs are stored across multiple tables to aide in purging old logs.
```
mathewc: Please update the xml comments on these public types to be what you'd expect to see as a 3rd party outsider. E.g., for this interface, I'd expect something like "Defines an interface for managing log tables.".

src/Microsoft.Azure.WebJobs.Logging/ILogTableProvider.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace Microsoft.Azure.WebJobs.Logging
+{
+    /// <summary>
+    /// Logs are stored across multiple tables to aide in purging old logs.  
+    /// </summary>
+    public interface ILogTableProvider
+    {
+        /// <summary>
+        /// Return a table with the given suffix. The logging will create this table if it doesn't exist.
```
mathewc: Please clarify these comments. "The logging will create" doesn't make it clear whose responsibility it is. Is it the implementor of this interface?

src/Microsoft.Azure.WebJobs.Logging/ILogTableProvider.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// </summary>
+        /// <param name="suffix">Suffix will be legal table name characters. </param>
```
mathewc: "The suffix to use for the log table name. Should only include legal table name characters."

src/Microsoft.Azure.WebJobs.Logging/Internal/EpochTableProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+            var table = _tableClient.GetTableReference(tableName);
+            return table;
+        }
+
+        // List all tables that we may have handed out. 
+        public async Task<CloudTable[]> ListTablesAsync()
+        {
+            List<CloudTable> list = new List<CloudTable>();
+            TableContinuationToken continuationToken = null;
+            while (true)
```
mathewc: nit: `do {} while(continuationToken != null)`?

src/Microsoft.Azure.WebJobs.Logging/Internal/TimeBucket.cs
```diff
@@ -27,5 +29,63 @@ public static long ConvertToBucket(DateTime dt)
...
+        private static long GetEpochNumber(DateTime epoch)
+        {
+            var year = epoch.Year; // 4 digit
+            var month = epoch.Month; // 1..12
+
+            int i = year * 100 + month;
+            return i;
+        }
+
+        public static CloudTable GetTableForEpoch(this ILogTableProvider tableLookup, DateTime epoch)
```
mathewc: Why are all these helper methods extension methods rather than part of the ILogTableProvider interface? E.g. I'd expect a function `ILogTableProvider.GetLogTable(DateTime))` that encapsulates all this logic. That way all details of how much data (week, month, etc.) and how to map a time to a table is encapsulated in the implementation.

src/Microsoft.Azure.WebJobs.Logging/Internal/TimeBucket.cs
```diff
@@ -27,5 +29,63 @@ public static long ConvertToBucket(DateTime dt)
             }
             return min;
         }
+
+        // Used to compute epoch. 
+        const long ticksBaseline = 633663648000000000L;
+        const long ticksPerMonth = 26785144663987L;
+        
+        public static DateTime CommonEpoch = new DateTime(0, DateTimeKind.Utc); // maps to epoch 0
```
mathewc: What is this "common epoch" used for? I feel like I'm back in geology class :)

src/Microsoft.Azure.WebJobs.Logging/Internal/TimeBucket.cs
```diff
@@ -27,5 +29,63 @@ public static long ConvertToBucket(DateTime dt)
...
         }
+
+        // Used to compute epoch. 
+        const long ticksBaseline = 633663648000000000L;
+        const long ticksPerMonth = 26785144663987L;
+        
+        public static DateTime CommonEpoch = new DateTime(0, DateTimeKind.Utc); // maps to epoch 0 
+        public static string CommonEpochSuffix = "common";
+
+        public static long GetEpochNumberFromTable(CloudTable table)
```
mathewc: I thought you moved from opaque epoch numbering to datetime strings YYYYMM? If so, this function won't parse correctly, right?

src/Microsoft.Azure.WebJobs.Logging/Internal/TimeBucket.cs
MikeStall: My intention was to make the interface as simple as possible (and hence do as much possible before making the call). The only thing the interface absolutely *has* to do is the suffix--> table mapping.   
But as you're alluding to, that means the interface can't control the policy (group by month, etc).  I can shift this around. 

---
In reply to: [82906498](https://github.com/Azure/azure-webjobs-sdk/pull/862#discussion_r82906498) [](ancestors = 82906498)

src/Microsoft.Azure.WebJobs.Logging/Internal/TimeBucket.cs
MikeStall: Used for the list of FunctionDefinitions, which are timeless (they don't have any time stamp information). 

---
In reply to: [82906813](https://github.com/Azure/azure-webjobs-sdk/pull/862#discussion_r82906813) [](ancestors = 82906813)

src/Microsoft.Azure.WebJobs.Logging/Internal/TimeBucket.cs
MikeStall: It will return -1 for the common case. 

---
In reply to: [82907220](https://github.com/Azure/azure-webjobs-sdk/pull/862#discussion_r82907220) [](ancestors = 82907220)

src/Microsoft.Azure.WebJobs.Logging/Internal/TimeBucket.cs
```diff
@@ -27,5 +29,63 @@ public static long ConvertToBucket(DateTime dt)
...
+            if (long.TryParse(suffix, out epoch))
+            {
+                return epoch;
+            }
+            return -1;
+        }
+
+        // Epoch must be positive, orderd integers.
+        // Use YYYYMM 
+        private static long GetEpochNumber(DateTime epoch)
```
mathewc: This mapping from DateTime to opaque epoch numbers seems complicated. Why is it needed at all? Given a set of tables each table holding a month's worth of data, you can find out directly which table a given DateTime maps to, right? The suffix of each table defines a min/max date/time.

src/Microsoft.Azure.WebJobs.Logging/LogFactory.cs
```diff
@@ -15,23 +13,35 @@ public static class LogFactory
...
         /// Multiple compute instances can write to the same table simultaneously without interference. 
         /// </summary>
         /// <param name="computerContainerName">name of the compute container. Likley %COMPUTERNAME%. </param>
-        /// <param name="table">underlying azure storage table to write to.</param>
+        /// <param name="tableLookup">underlying azure storage table to write to. 
+        /// Passed a string suffix (which will consist of valid Azure table characters). 
+        /// This must be a deterministic (replayable) function.</param>
         /// <returns></returns>
-        public static ILogWriter NewWriter(string computerContainerName, CloudTable table)
+        public static ILogWriter NewWriter(string computerContainerName, ILogTableProvider tableLookup)
```
mathewc: nit: containerName. You use "containerName" elsewhere

src/Microsoft.Azure.WebJobs.Logging/Internal/TimeBucket.cs
MikeStall: Per your comment below, we could indeed push some of this into the TableProvider interface.   But here are some forces we still need to account for: 
1. We need to be able to convert from Date --> Table and Table --> Date. 
2. Also, tables may not be consecutive (hence the need for ListTables).  We may have data for Jan 1st, nothing for 2 months, and then April 1st.  
3. minor issue: Table names are a very restricted character set (basically alpah numeric) and can't represent datetime.ToString(). so using integer also solved this problem. 
4. minor issue: I wanted the tables sorted lexographically in table viewing tools. DateTime doesn't sort like that. 



---
In reply to: [82908776](https://github.com/Azure/azure-webjobs-sdk/pull/862#discussion_r82908776) [](ancestors = 82908776)

src/Microsoft.Azure.WebJobs.Logging/LogFactory.cs
```diff
@@ -15,23 +13,35 @@ public static class LogFactory
...
         }
 
         /// <summary>
         /// Create a new writer for the given compute container name that writes to the given table.
         /// Multiple compute instances can write to the same table simultaneously without interference. 
         /// </summary>
         /// <param name="computerContainerName">name of the compute container. Likley %COMPUTERNAME%. </param>
-        /// <param name="table">underlying azure storage table to write to.</param>
+        /// <param name="tableLookup">underlying azure storage table to write to. 
+        /// Passed a string suffix (which will consist of valid Azure table characters).
```
mathewc: I don't understand what "passed a string suffix" refers to. Please clean up the xml comments for these public types

src/Microsoft.Azure.WebJobs.Logging/Internal/LogWriter.cs
```diff
@@ -47,18 +47,18 @@ internal class LogWriter : ILogWriter
         static ContainerActiveLogger _container;
         CloudTableInstanceCountLogger _instanceLogger;
 
-        public LogWriter(string computerContainerName, CloudTable table)
+        public LogWriter(string computerContainerName, ILogTableProvider tableLookup)
```
mathewc: nit: containerName

src/Microsoft.Azure.WebJobs.Logging/LogFactory.cs
```diff
@@ -15,23 +13,35 @@ public static class LogFactory
...
+
+        /// <summary>
+        /// Get a default table provider for the given tableClient. This will generate table names with the given prefix.
+        /// </summary>
+        /// <param name="tableClient">storage client for where to generate tables</param>
+        /// <param name="tableNamePrefix">prefix for tables to generate. This should be a valid azure table name.</param>
+        public static ILogTableProvider NewTableProvider(CloudTableClient tableClient, string tableNamePrefix = LogFactory.DefaultLogTableName)
         {
-            return new LogWriter(computerContainerName, table);
+            return new EpochTableProvider(tableClient, tableNamePrefix);
```
mathewc: ILogTableProvider is public so people can implement instances of it, but they can't plug in their implementations into our infrastructure, right? I'm not saying that is a goal, just trying to understand the provider interface and your intent for it.

test/Dashboard.UnitTests/RestApiTests.cs
```diff
@@ -134,14 +136,22 @@ public Fixture()
 
             public void Dispose()
             {
-                _logTable.DeleteIfExists();                
+                DisposeAsync().Wait();
+            }
+            private async Task DisposeAsync()
+            {
+                var tables = await _provider.ListTablesAsync();
```
mathewc: nit: might consider deleting these in parallel

test/Dashboard.UnitTests/RestApiTests.cs
```diff
@@ -166,23 +176,18 @@ public void Dispose()
                 return list.ToArray();
             }
 
-            CloudTable GetNewLoggingTable()
+            CloudTableClient GetNewLoggingTable()
```
mathewc: This method doesn't seem to be named correctly

src/Microsoft.Azure.WebJobs.Logging/LogFactory.cs
MikeStall: The dashboard will work with the default settings. That was part of the rational to keep the provider interface simple and not make it attractive to customize. But if a highly determined individual created their own arbitrary table provider here, it would log to it, but it wouldn't work with the dashboard (although they could deploy their own dashboard with their own custom logger) 

---
In reply to: [82910137](https://github.com/Azure/azure-webjobs-sdk/pull/862#discussion_r82910137) [](ancestors = 82910137)

src/Microsoft.Azure.WebJobs.Logging/Internal/TimeBucket.cs
mathewc: You're using a format like YYYYMM now for table name suffices, right? So can't you do away with the epoch number conversions now?

src/Dashboard/App_Start/AppModule.cs
```diff
@@ -134,46 +133,37 @@ public static void Load(ContainerBuilder builder)
...
                 {
-                    return table;
+                    // Explicitly set to old mode. 
+                    return null;
                 }
+
+                // This is the common case for Azure Functions. 
+                // No prefix, so use the default. 
+                var provider = LogFactory.NewTableProvider(tableClient);
+                return provider;
```
MikeStall: I'm actually a fan of the additional variables. It generates the same underlying code, but I find it easier to follow. It's also helpful in the debugger. 

---
In reply to: [82874457](https://github.com/Azure/azure-webjobs-sdk/pull/862#discussion_r82874457) [](ancestors = 82874457)

src/Microsoft.Azure.WebJobs.Logging/Internal/TimeBucket.cs
MikeStall: actually - we don't need them anymore in the latest iteration. So I'll just deleted them. 

---
In reply to: [82881677](https://github.com/Azure/azure-webjobs-sdk/pull/862#discussion_r82881677) [](ancestors = 82881677)

src/Microsoft.Azure.WebJobs.Logging/Internal/TimeBucket.cs
MikeStall: Also, this is all internal. So we could do a massive rename if somebody gets better inspiration. 

---
In reply to: [82907667](https://github.com/Azure/azure-webjobs-sdk/pull/862#discussion_r82907667) [](ancestors = 82907667,82906813)

src/Microsoft.Azure.WebJobs.Logging/Internal/TimeBucket.cs
MikeStall: I think it amounts to some renames - not an underlying logic change. 

---
In reply to: [82912142](https://github.com/Azure/azure-webjobs-sdk/pull/862#discussion_r82912142) [](ancestors = 82912142)

src/Dashboard/App_Start/AppModule.cs
```diff
@@ -134,46 +133,37 @@ public static void Load(ContainerBuilder builder)
...
-                        // Appsetting specifically opts us in. Use fast-tables.
-                        table.CreateIfNotExists();
-                        return table;
-                    }
+                    // No Func appsetting, this is using sdk-style older logging. 
+                    return null;
                 }
 
-                if (table.Exists())
+                if (string.Equals(ver, FunctionExtensionVersionDisabled, StringComparison.OrdinalIgnoreCase))
```
MikeStall: I prefer to separate the cases out - and then it gives me a place to comment each. 

---
In reply to: [82874256](https://github.com/Azure/azure-webjobs-sdk/pull/862#discussion_r82874256) [](ancestors = 82874256)

src/Microsoft.Azure.WebJobs.Logging/LogFactory.cs
MikeStall: I'll switch the other places to computerContainerName. "ContainerName" is too overloaded and easily confused with Azure Storage. 

---
In reply to: [82909392](https://github.com/Azure/azure-webjobs-sdk/pull/862#discussion_r82909392) [](ancestors = 82909392)

src/Microsoft.Azure.WebJobs.Logging/Internal/LogReader.cs
MikeStall: Fair - it's not IEnumerable. But it does take in a continuation token and support iteration via continuation tokens. . 
it's internal, so we can change later. 

---
In reply to: [82881404](https://github.com/Azure/azure-webjobs-sdk/pull/862#discussion_r82881404) [](ancestors = 82881404)

src/Microsoft.Azure.WebJobs.Logging/Internal/LogReader.cs
MikeStall: we do anyways; but you're right - this could be cleaner.  see next iteration

---
In reply to: [82880453](https://github.com/Azure/azure-webjobs-sdk/pull/862#discussion_r82880453) [](ancestors = 82880453)

src/Microsoft.Azure.WebJobs.Logging/Internal/LogReader.cs
```diff
@@ -78,33 +83,53 @@ public LogReader(CloudTable table)
...
-            TableResult retrievedResult = await _instanceTable.SafeExecuteAsync(retrieveOperation);
+            Task<FunctionInstanceLogItem>[] taskLookups = Array.ConvertAll(tables, async instanceTable =>
+            {
+                // Create a retrieve operation that takes a customer entity.
+                TableOperation retrieveOperation = InstanceTableEntity.GetRetrieveOperation(id);
+
+                // Execute the retrieve operation.
+                TableResult retrievedResult = await instanceTable.SafeExecuteAsync(retrieveOperation);
+
+                var x = (InstanceTableEntity)retrievedResult.Result;
```
mathewc: nit: make "x" a more descriptive name

mathewc:
I'd like to see some unit tests for all the nitty gritty helpers functions / logic that you use for various calculations. It looks like you only had some higher level end to end tests (which are needed as well).

Other than that, no further comments. :ship:

MikeStall:
Added some awesome unit tests for it. I'll go ahead and push but leave this open for lingering comments.

---
In reply to: [253626205](https://github.com/Azure/azure-webjobs-sdk/pull/862#issuecomment-253626205) [](ancestors = 253626205)

MikeStall:
Also, I'll send out a Scripts PR to take advantage of these changes. 

---
In reply to: [253639744](https://github.com/Azure/azure-webjobs-sdk/pull/862#issuecomment-253639744) [](ancestors = 253639744,253626205)

MikeStall:
Committed.

PR #864, opened 2016-10-10
Add unit tests for Dashboard.

MikeStall:
Test registration,
No functional changes to the product. Some minor changes in startup to aide in test factoring. 

Switch over the DI from AutoFac to Ninject.
- Ninject uses global statics and is harder to test.
- the cool kids recommend AutoFac.


MikeStall:
I've gone ahead and pushed this commit; but I'll leave this pull request open for a bit longer for comment.

src/Dashboard/App_Start/NinjectWebCommon.cs
```diff
@@ -1,86 +0,0 @@
...
-using System.Web.Http;
-using System.Web.Mvc;
-using Dashboard.AppStart;
-using Dashboard.Infrastructure;
-using Microsoft.Web.Infrastructure.DynamicModuleHelper;
-using Ninject;
-using Ninject.Modules;
-using Ninject.Web.Common;
-
-[assembly: WebActivatorEx.PreApplicationStartMethod(typeof(NinjectWebCommon), "Start")]
```
mathewc: After removing this, we're no longer using any of the WebActivator/WebActivatorEx packages right? Can we remove those?

src/Dashboard/App_Start/NinjectWebCommon.cs
```diff
@@ -1,86 +0,0 @@
...
-        }
-
-        /// <summary>
-        /// Starts the application
-        /// </summary>
-        [SuppressMessage("Microsoft.Reliability", "CA2000:Dispose objects before losing scope")]
-        public static void Start()
-        {
-            // Registration must occur before application start. So use: WebActivator.PreApplicationStartMethod 
-            DynamicModuleUtility.RegisterModule(typeof(OnePerRequestHttpModule));
```
mathewc: So none of this startup time logic is needed for AutoFac? It's not clear from looking at the code that all the same behaviors are maintained.

MikeStall: Correct. That's part of why Autofac won over Ninject.  There's no intrinsic need for any of this.

src/Dashboard/App_Start/NinjectWebCommon.cs
MikeStall: I'll try...

mathewc:
I'm assuming all the Dashboard tests pass with these changes? :)

src/Dashboard/App_Start/FilterConfig.cs
```diff
@@ -15,7 +15,7 @@ public static void RegisterGlobalFilters(GlobalFilterCollection filters)
                 throw new ArgumentNullException("filters");
             }
 
-            filters.Add(new HandleErrorAttribute());
+            //filters.Add(new HandleErrorAttribute());
```
MikeStall: we can delete the whole file. This is now handled by AutoFac.

src/Dashboard/App_Start/NinjectWebCommon.cs
MikeStall: It looks like we can!

MikeStall:
applied Cr feedback.

Issue #850, opened 2016-09-27
Missing causality  on queue output messages


MikeStall:
Output queue messages got stamped with a causality ID matching the function id that sent the message. This could be used to determine parent-child relationships. 

It was via a call to:
https://github.com/Azure/azure-webjobs-sdk/blob/master/src/Microsoft.Azure.WebJobs.Host/Queues/QueueCausalityManager.cs#L30 

That call got deleted when we switched everything over in the big AsyncCollector<> refactoring.  

We should find a place to make that call in the new pipeline so that the listener can set the parent id.

MikeStall:
This used to be called from UserTypeToStorageQueueMessageConverter, which got deleted and replaced with the BindingHelpers.

MikeStall:
Fix checked in.

PR #868, opened 2016-10-12
Convert

MikeStall:
Fix for:
https://github.com/Azure/azure-webjobs-sdk/issues/850
Need to let converter manager include the Function instance guid so that it can stamp it.

Use a delegate FuncArgumentBuilder<TAttribute> instead of a Func<>,
Let ConverterManager take a Object-->JObject hook that can override JSON serialization. 
Queue bindings then use this to stamp the function instance id; as they did before. 

Some cleanups to the IConverterManager interface. These are binary breaking changes, but not source breaking changes.


MikeStall:
Review has been open for 2 weeks, and I need these changes for some other RC-blockers in table; so I've gone ahead and committed this.

PR #869, opened 2016-10-13
Extending struct data binding to support nullable types

mathewc:
Currently our binding pipeline doesn't support binding method parameters to nullable types. All our current binding contract types have been non-nullable I guess.

The need for this arose in Azure Functions recently, where we have an HttpTrigger binding whose binding data contract includes route parameters, which can be **optional** (i.e. nullable). See the example below where the `int?` parameter comes from an optional route parameter `products/{category:alpha}/{id:int?}`. These changes allow this to work, whereas before we'd get casting errors / type mismatch errors.

```csharp
public class ProductInfo
{
    public string Category { get; set; }
    public int? Id { get; set; }
}

public static ProductInfo Run(ProductInfo info, string category, int? id, TraceWriter log)
{
    log.Info($"ProductInfo: Category={category} Id={id}");
    log.Info($"Parameters: category={category} id={id}");

    return info;
}
```

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/Data/StructDataBinding.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/Data/StructDataBinding.cs
@@ -3,23 +3,28 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host.Converters;
 using Microsoft.Azure.WebJobs.Host.Protocols;
 
 namespace Microsoft.Azure.WebJobs.Host.Bindings.Data
 {
+    /// <summary>
+    /// Handles value types (structs) as well as nullable types.
+    /// </summary>
     internal class StructDataBinding<TBindingData> : IBinding
-        where TBindingData : struct
     {
         private static readonly IObjectToTypeConverter<TBindingData> Converter =
             ObjectToTypeConverterFactory.CreateForStruct<TBindingData>();
 
+        private readonly bool _isNullable;
         private readonly string _parameterName;
         private readonly IArgumentBinding<TBindingData> _argumentBinding;
 
         public StructDataBinding(string parameterName, IArgumentBinding<TBindingData> argumentBinding)
         {
+            _isNullable = TypeUtility.IsNullable(typeof(TBindingData));
             _parameterName = parameterName;
             _argumentBinding = argumentBinding;
         }
@@ -40,7 +45,7 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings.Data
 
             if (!Converter.TryConvert(value, out typedValue))
             {
-                throw new InvalidOperationException("Unable to convert value to " + typeof(TBindingData).Name + ".");
+                throw new InvalidOperationException("Unable to convert value to " + TypeUtility.GetFriendlyName(typeof(TBindingData)) + ".");
             }
 
             return BindAsync(typedValue, context);
@@ -63,10 +68,10 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings.Data
 
             object untypedValue = bindingData[_parameterName];
 
-            if (!(untypedValue is TBindingData))
+            if (!(untypedValue is TBindingData) && !(untypedValue == null && _isNullable))
             {
-                throw new InvalidOperationException("Binding data for '" + _parameterName +
-                    "' is not of expected type " + typeof(TBindingData).Name + ".");
+                throw new InvalidOperationException(
+                    string.Format(CultureInfo.InvariantCulture, "Binding data for '{0}' is not of expected type {1}.", _parameterName, TypeUtility.GetFriendlyName(typeof(TBindingData))));
             }
 
             TBindingData typedValue = (TBindingData)untypedValue;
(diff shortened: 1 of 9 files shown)


src/Microsoft.Azure.WebJobs.Host/Bindings/Data/StructDataBindingProvider.cs
```diff
@@ -8,8 +8,10 @@
 
 namespace Microsoft.Azure.WebJobs.Host.Bindings.Data
 {
+    /// <summary>
+    /// Handles value types (structs) as well as nullable types.
+    /// </summary>
     internal class StructDataBindingProvider<TBindingData> : IBindingProvider
```
mathewc: The switch where we create this provider type is in `DataBindingProvider.CreateTypedBindingProvider`. If for the target type `Type.IsValueType` is true, we create this provider. The problem is a nullable type IS a value type, but is not a struct.

MikeStall:
>     internal static class ObjectToTypeConverterFactory

Note for the future: I'd be great if we could find some way to replace these with the new IConverterManager and have a single converter story. 

---
Refers to: src/Microsoft.Azure.WebJobs.Host/Bindings/ObjectToTypeConverterFactory.cs:8 in c7604b7. [](commit_id = c7604b7c1ea26abf539b7abe86aece0f07a4557d, deletion_comment = False)

test/Microsoft.Azure.WebJobs.Host.UnitTests/Bindings/Data/DataBindingProviderTests.cs
```diff
@@ -14,6 +15,40 @@ namespace Microsoft.Azure.WebJobs.Host.UnitTests.Bindings.Data
...
+        {
+            // Arrange
+            IBindingProvider product = new DataBindingProvider();
+
+            string parameterName = "p";
+            Type parameterType = typeof(int?);
+            BindingProviderContext context = CreateBindingContext(parameterName, parameterType);
+
+            // Act
+            IBinding binding = product.TryCreateAsync(context).GetAwaiter().GetResult();
```
MikeStall: >).GetAwaiter().GetResult(); [](start = 61, length = 27)

nit: can you use "await" here? It's an Async method.

src/Microsoft.Azure.WebJobs.Host/Bindings/Data/StructDataBinding.cs
```diff
@@ -9,17 +9,21 @@
...
         private static readonly IObjectToTypeConverter<TBindingData> Converter =
             ObjectToTypeConverterFactory.CreateForStruct<TBindingData>();
 
+        private readonly bool _isNullable;
         private readonly string _parameterName;
         private readonly IArgumentBinding<TBindingData> _argumentBinding;
 
         public StructDataBinding(string parameterName, IArgumentBinding<TBindingData> argumentBinding)
         {
+            _isNullable = typeof(TBindingData).IsGenericType && typeof(TBindingData).GetGenericTypeDefinition() == typeof(Nullable<>);
```
MikeStall: >typeof(TBindingData).IsGenericType && typeof(TBindingData).GetGenericTypeDefinition() == typeof(Nullable<>); [](start = 26, length = 108)

nit: move to the existing static TypeUtility class: 
  bool TypeUtility.IsNullable(Type)

MikeStall:
>                 throw new InvalidOperationException("Unable to convert value to " + typeof(TBindingData).Name + ".");

see comment below. TypeUtility.GetFriendlyName()? 

---
Refers to: src/Microsoft.Azure.WebJobs.Host/Bindings/Data/StructDataBinding.cs:47 in c7604b7. [](commit_id = c7604b7c1ea26abf539b7abe86aece0f07a4557d, deletion_comment = False)

src/Microsoft.Azure.WebJobs.Host/Bindings/StructOutputConverter.cs
```diff
@@ -1,23 +1,25 @@
...
     internal class StructOutputConverter<TInput, TOutput> : IObjectToTypeConverter<TOutput>
-        where TInput : struct
     {
+        private readonly bool _isNullable;
         private readonly IConverter<TInput, TOutput> _innerConverter;
 
         public StructOutputConverter(IConverter<TInput, TOutput> innerConverter)
         {
+            _isNullable = typeof(TInput).IsGenericType && typeof(TInput).GetGenericTypeDefinition() == typeof(Nullable<>);
             _innerConverter = innerConverter;
```
MikeStall: see other comment. TypeUtility.IsNullable()

MikeStall:
:shipit:

mathewc:
I'll make all your suggested changes and update the PR with the changes before checking in.

src/Microsoft.Azure.WebJobs.Host/Bindings/Data/StructDataBinding.cs
```diff
@@ -63,7 +67,7 @@ public Task<IValueProvider> BindAsync(BindingContext context)
 
             object untypedValue = bindingData[_parameterName];
 
-            if (!(untypedValue is TBindingData))
+            if (!(untypedValue is TBindingData) && !(untypedValue == null && _isNullable))
             {
                 throw new InvalidOperationException("Binding data for '" + _parameterName +
                     "' is not of expected type " + typeof(TBindingData).Name + ".");
```
mathewc: Added a test for this as well

PR #875, opened 2016-10-18
Let reader read old (pre-sharded) logs. Needed so latest reader (in d…

MikeStall:
…ashboard) can still consume older writers.

### DIFF
--- a/src/Dashboard/app/views/AboutHome.html
+++ b/src/Dashboard/app/views/AboutHome.html
@@ -4,3 +4,4 @@
         <ng-include src="'app/views/IndexerLogTable.html'"></ng-include>
     </div>
 </div>
+1.2
--- a/src/Microsoft.Azure.WebJobs.Logging/Entities/FunctionDefinitionEntity.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging/Entities/FunctionDefinitionEntity.cs
@@ -49,5 +49,10 @@ namespace Microsoft.Azure.WebJobs.Logging
                 OriginalName = functionName
             };
         }
+
+        public override string ToString()
+        {
+            return this.OriginalName;
+        }
     }
 }
--- a/src/Microsoft.Azure.WebJobs.Logging/Internal/DefaultLogTableProvider.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging/Internal/DefaultLogTableProvider.cs
@@ -32,8 +32,14 @@ namespace Microsoft.Azure.WebJobs.Logging
 
         public CloudTable GetTable(string suffix)
         {
+            var table = LegacyTableReader.TryGetLegacy(_tableClient, suffix);
+            if (table != null)
+            {
+                return table;
+            }
+
             var tableName = _tableNamePrefix + suffix;
-            var table = _tableClient.GetTableReference(tableName);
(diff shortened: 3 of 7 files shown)


src/Microsoft.Azure.WebJobs.Logging/Internal/LegacyHack.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.WindowsAzure.Storage.Table;
+
+namespace Microsoft.Azure.WebJobs.Logging
+{
+    // LogReader needs to handle older format.
+    // DO this in one central place so that we can remove this class when deprecate the old format.
```
mathewc: Let's create a bug to remove this, after the old versions are removed from the feed post GA

src/Microsoft.Azure.WebJobs.Logging/Internal/LegacyHack.cs
```diff
@@ -0,0 +1,76 @@
...
+            }
+            return null;
+        }
+
+        // -1 if not matched
+        public static long GetEpochFromTable(CloudTable table)
+        {
+            if (table.Name == OldTableName)
+            {
+                return 201609;
```
mathewc: What does this mean?

src/Microsoft.Azure.WebJobs.Logging/Internal/LegacyHack.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.WindowsAzure.Storage.Table;
+
+namespace Microsoft.Azure.WebJobs.Logging
+{
+    // LogReader needs to handle older format.
+    // DO this in one central place so that we can remove this class when deprecate the old format. 
+    // One central place 
+    internal class LegacyHack
```
mathewc: nit: I'd rename this to LegacyTableHelper - if this ever were to surface in a stack trace, that name would make people feel better. If we create a bug as I suggested above to remove this, we'll remember w/o the name :)

src/Microsoft.Azure.WebJobs.Logging/Internal/LegacyHack.cs
MikeStall: Filed: https://github.com/Azure/azure-webjobs-sdk/issues/876

src/Microsoft.Azure.WebJobs.Logging/Internal/LegacyHack.cs
MikeStall: September 2016 - suffix number for before version .7; used to keep things in order.

mathewc:
:ship:

src/Dashboard/app/views/AboutHome.html
```diff
@@ -3,4 +3,5 @@
         <h3>About</h3>
         <ng-include src="'app/views/IndexerLogTable.html'"></ng-include>
     </div>
-</div>
\ No newline at end of file
+</div>
+1.2
```
mathewc: what is this?

PR #877, opened 2016-10-20
Service Bus: re-using MessagingFactory based on connection string

brettsam:


mamaso:
Looks good, though I think I like option 2 in #881 better, as I can see this TCP connection issue surfacing in other bindings.

brettsam:
We don't have a solid plan for this. Closing it for now but will keep branch around.

PR #878, opened 2016-10-21
Allowing SingletonListener to fall back to 'secondary' mode

brettsam:
This needs some more tests and perhaps some design tweaks, but want to get some feedback.


brettsam:
We're not going to do this -- closing.

Issue #883, opened 2016-10-24
Support multiple log writers to the same storage account 


MikeStall:
Multiple hosts can write to a single storage account. This is a common scenario in Functions. 
FastLogging APIs need to handle this. FunctionId is HostName + FunctionName.  Use FunctionIds similar to how the classic SDK does.

PR #884, opened 2016-10-24
Support multiple loggers to the same storage account

MikeStall:
Fix  https://github.com/Azure/azure-webjobs-sdk/issues/883


src/Dashboard/Data/FunctionIndexReader.cs
```diff
@@ -70,7 +70,8 @@ public DateTimeOffset GetCurrentVersion()
             return _versionMapper.GetVersion(blob.Metadata);
         }
 
-        public IResultSegment<FunctionIndexEntry> Read(int maximumResults, string continuationToken)
+        // Classic logging ignores hostName filter.
```
mathewc: Is this for the top level Dashboard view showing all functions across all hosts? I.e., the same function name might be shown twice in the table, although the actual invocation logs ARE separated by host id.

src/Microsoft.Azure.WebJobs.Logging/FunctionId.cs
```diff
@@ -0,0 +1,108 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Build directly from a string. This skips escaping and should only be used by the reader for interop. 
+        /// This is the direct inverse of ToString().
+        /// </summary>
+        /// <param name="value">raw value from a previos ToString() call. </param>
+        /// <returns></returns>
+        public static FunctionId DangerousFromString(string value)
```
mathewc: Can you remove this "Dangerous" terminology from the public API - it's strange? E.g., since this is the inverse of ToString, How about a Parse or TryParse method, which is a common BCL pattern. Or perhaps just a Build overload that takes a string?

src/Microsoft.Azure.WebJobs.Logging/FunctionId.cs
```diff
@@ -0,0 +1,108 @@
...
+            return new FunctionId { Value = value };
+        }
+
+        /// <summary>
+        /// Build a FunctionId from the host and function name. 
+        /// </summary>
+        /// <param name="hostName"></param>
+        /// <param name="functionName"></param>
+        /// <returns></returns>
+        public static FunctionId Build(string hostName, string functionName)
```
mathewc: Any reason why you can't just use a constructor for creation?

src/Microsoft.Azure.WebJobs.Logging/FunctionInstanceLogItem.cs
```diff
@@ -29,9 +28,13 @@ public class FunctionInstanceLogItem : IFunctionInstanceBaseEntry
         /// <summary>Gets or sets the Function ID of the ancestor function instance.</summary>
         public Guid? ParentId { get; set; }
 
-        /// <summary>Name of this function. This can be used in querying other instances of this function. </summary>
+        /// <summary>Short Name of this function.</summary>
         public string FunctionName { get; set; }
```
mathewc: Doesn't FunctionId already contain the FunctionName as a component? Can it expose it?

src/Microsoft.Azure.WebJobs.Logging/Internal/TableScheme.cs
```diff
@@ -15,12 +15,36 @@ internal static class TableScheme
     {
         // List all partition keys in once place to ensure they're disjoint. 
         internal const string InstancePK = "I"; // InstanceTableEntity
-        internal const string TimelineAggregatePK = "T"; // TimelineAggregateEntity
-        internal const string RecentFuncIndexPK = "R"; // RecentPerFuncEntity
+        internal const string TimelineAggregatePK = "T2"; // TimelineAggregateEntity
+        internal const string RecentFuncIndexPK = "R2"; // RecentPerFuncEntity
         internal const string ContainerActivePK = "C"; // ContainerActiveEntity
-        internal const string FuncDefIndexPK = "FD"; // FunctionDefinitionEntity
+        internal const string FuncDefIndexPK = "FD2"; // FunctionDefinitionEntity x
```
mathewc: Why the rename? Also, what is the "x" in your comment?

test/Dashboard.UnitTests/RestApiTests.cs
```diff
@@ -36,21 +36,33 @@ public RestApiTests(Fixture fixture)
...
+            var x = response.Entries.ToArray();
+            Assert.Equal(1, x.Length);
+            Assert.Equal("alpha", x[0].functionName);
+        }
+
+        // Pass ?host=  parameter when getting definitions to filter to a specific host. 
+        [Fact]
+        public async Task GetDefinitionSingleHost()
+        {
+            var response = await _client.GetJsonAsync<FunctionStatisticsSegment>(_endpoint + "/api/functions/definitions?limit=100&host=missing");
```
mathewc: What is the case sensitivity of host id when used in queries?

src/Microsoft.Azure.WebJobs.Logging/FunctionId.cs
MikeStall: I can change to Parse. 
I don't want to overload Build since the semantics are different (escape vs. unescape). Only the reader should call this. Consumers shouldn't call this. 

---
In reply to: [84855861](https://github.com/Azure/azure-webjobs-sdk/pull/884#discussion_r84855861) [](ancestors = 84855861)

src/Microsoft.Azure.WebJobs.Logging/FunctionId.cs
MikeStall: It's technically possible; but I avoided it because it loses semantics. Specifically, whether it escapes (Build) or doesn't escape (Parse).

---
In reply to: [84856006](https://github.com/Azure/azure-webjobs-sdk/pull/884#discussion_r84856006) [](ancestors = 84856006)

src/Microsoft.Azure.WebJobs.Logging/FunctionInstanceLogItem.cs
MikeStall: FunctionId is an escaped normalized form.  
FunctionId may be "host-foo:41bar", whereas functionName may be "foo-BAR".


---
In reply to: [84856180](https://github.com/Azure/azure-webjobs-sdk/pull/884#discussion_r84856180) [](ancestors = 84856180)

src/Microsoft.Azure.WebJobs.Logging/Internal/TableScheme.cs
MikeStall: Rename because the schema is breaking. This gaurantees we won't read old schemas. 
I removed the x. 


---
In reply to: [84856799](https://github.com/Azure/azure-webjobs-sdk/pull/884#discussion_r84856799) [](ancestors = 84856799)

test/Dashboard.UnitTests/RestApiTests.cs
MikeStall: host id is case insensitive.

---
In reply to: [84857185](https://github.com/Azure/azure-webjobs-sdk/pull/884#discussion_r84857185) [](ancestors = 84857185)

test/Microsoft.Azure.WebJobs.Logging.FunctionalTests/FunctionIdTests.cs
```diff
@@ -0,0 +1,46 @@
...
+namespace Microsoft.Azure.WebJobs.Logging.FunctionalTests
+{
+    public class FunctionIdTests
+    {
+        // Test normalization 
+        [Fact]
+        public void Operators()
+        {
+            var f1 = FunctionId.Build("h1", "FFF1");
+            var f2 = FunctionId.Build("h1", "fff1");
```
MikeStall: >h1 [](start = 39, length = 2)

I'll update this test to verify hosts too.

src/Dashboard/Data/FunctionIndexReader.cs
MikeStall: Yes.

---
In reply to: [84854250](https://github.com/Azure/azure-webjobs-sdk/pull/884#discussion_r84854250) [](ancestors = 84854250)

MikeStall:
This has been merged. I'll leave the PR open for a bit longer to facilitate lingering comments.

Issue #810, opened 2016-09-01
Test against Blob Only storage accounts for blob trigger/input/output


christopheranderson:
We need to decide if we support blob only storage accounts for blob trigger/input/output.

mathewc:
I think this is a duplicate of https://github.com/Azure/azure-webjobs-sdk/issues/797.

lindydonna:
@mathewc I think you're right. This issue was to track whether or not we wanted to support the scenario at all, especially because there's probably a perf impact. Are you saying that we should decide that we are disallowing them and update the portal pickers appropriately?  We do have issues to track the portal work FYI.

mathewc:
Blob only storage accounts can't be used for our "primary" storage account, since we use queues internally. They should be supported in general though, however there have been some core SDK bugs logged for issues recently we may need to investigate there.

lindydonna:
In that case, this is not a duplicate. The other issue is about showing an error in the host (if I understood it correctly), whereas this is about determine whether we care to support the scenario. (I.e., let's firm up the "should work" statement). The easiest option would be to just decide not to support for RC and revisit later.

lindydonna:
This scenario probably works, but we don't want to support it for RC. There is a related issue for filtering out blob-only accounts from the portal picker.

paulbatum:
@mamaso Confirm blob only works at the SDK level. If it does, then reactivate https://github.com/Azure/azure-webjobs-sdk-script/issues/773

mamaso:
Investigated, blob-only storage accounts work without a hitch.

The premium storage account issues are due to a bug with GetServicePropertiesAsync in the storage sdk: https://github.com/Azure/azure-storage-net/issues/354

We use GetServicePropertiesAsync to do account validation & enable blob logging.  
Workarounds: I could handle this particular exception, assuming that it's caused by premium storage, or we can wait for storage to fix the bug.

Also put in a request for an api to determine storage account type (general, blob only, premium)
https://github.com/Azure/azure-storage-net/issues/355

lindydonna:
@davidebbo and @paulbatum: The Storage team closed both of the SDK issues that @mamaso opened, since the Storage REST API doesn't actually support these features.

My 2 cents: catching the exception on GetServicePropertiesAsync and assuming it's a premium account seems hacky and fragile. I vote to keep the exception in place. 

Unfortunately, it looks like we can't filter these accounts from the portal either, due to Azure/azure-storage-net#355.

[...]

PR #886, opened 2016-10-26
Validate storage account type against attribute bindings

mamaso:
related to #810 and [script/#773](https://github.com/Azure/azure-webjobs-sdk-script/issues/773)

Looking for initial feedback before I add tests & polish.

Added `StorageAccountType` to storage accounts, which is set to `premium` or `blobonly` in `validateCredentialsAsync`.  ~~In each binding provider I test if the storage account type is compatible with that attribute.~~

When you request a StorageAccount you supply the supported storage account types.  If the account is not of the supported type, we throw an error indicating the invalid type & what types are supported.

 Added `SupportedTypes` to StorageAccountAttribute with default `All`, so that from functions we can disabled Premium storage for BlobTriggerAttributes.

``` c#
// to throw if 'somename' is a Premium storage account, you can write this:
StorageAccount('somename', ~StorageAccountTypes.Premium)
```

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Blobs/Triggers/BlobTriggerAttributeBindingProvider.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Blobs/Triggers/BlobTriggerAttributeBindingProvider.cs
@@ -159,6 +159,8 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs.Triggers
 
             IStorageAccount hostAccount = await _accountProvider.GetStorageAccountAsync(context.CancellationToken);
             IStorageAccount dataAccount = await _accountProvider.GetStorageAccountAsync(context.Parameter, context.CancellationToken, _nameResolver);
+            // premium does not support blob logs, so disallow for blob triggers
+            dataAccount.AssertTypeOneOf(StorageAccountType.GeneralPurpose, StorageAccountType.BlobOnly);
 
             ITriggerBinding binding = new BlobTriggerBinding(parameter, argumentBinding, hostAccount, dataAccount, path,
                 _hostIdProvider, _queueConfiguration, _exceptionHandler, _blobWrittenWatcherSetter,
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageAccountProvider.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageAccountProvider.cs
@@ -2,7 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
-using System.Reflection;
+using System.Globalization;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host.Storage;
@@ -176,7 +176,12 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
             if (account != null)
             {
                 // On the first attempt, this will make a network call to verify the credentials work.
-                await _storageCredentialsValidator.ValidateCredentialsAsync(account, isPrimary, cancellationToken);
+                await _storageCredentialsValidator.ValidateCredentialsAsync(account, cancellationToken);
+
+                if (isPrimary)
+                {
+                    account.AssertTypeOneOf(StorageAccountType.GeneralPurpose);
+                }
             }
 
             return account;
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
@@ -18,9 +18,8 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
     internal class DefaultStorageCredentialsValidator : IStorageCredentialsValidator
     {
         private readonly HashSet<StorageCredentials> _validatedCredentials = new HashSet<StorageCredentials>();
-        private StorageCredentials _primaryCredentials = null;
 
-        public async Task ValidateCredentialsAsync(IStorageAccount account, bool isPrimaryAccount, CancellationToken cancellationToken)
+        public async Task ValidateCredentialsAsync(IStorageAccount account, CancellationToken cancellationToken)
         {
             if (account == null)
             {
@@ -30,17 +29,17 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
             StorageCredentials credentials = account.Credentials;
 
             // Avoid double-validating the same account and credentials.
-            if ((!isPrimaryAccount && _validatedCredentials.Contains(credentials)) || _primaryCredentials == credentials)
+            if (_validatedCredentials.Contains(credentials))
             {
                 return;
             }
 
-            await ValidateCredentialsAsyncCore(account, isPrimaryAccount, cancellationToken);
+            await ValidateCredentialsAsyncCore(account, cancellationToken);
             _validatedCredentials.Add(credentials);
         }
 
-        private async Task ValidateCredentialsAsyncCore(IStorageAccount account,
-            bool isPrimaryAccount, CancellationToken cancellationToken)
+        // Test that the credentials are valid and classify the account.Type as one of StorageAccountTypes
+        private async Task ValidateCredentialsAsyncCore(IStorageAccount account, CancellationToken cancellationToken)
         {
             // Verify the credentials are correct.
             // Have to actually ping a storage operation.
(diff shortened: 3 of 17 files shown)


src/Microsoft.Azure.WebJobs.Host/Blobs/Triggers/BlobTriggerAttributeBindingProvider.cs
```diff
@@ -160,6 +160,11 @@ private static IBlobArgumentBindingProvider CreateProvider(IEnumerable<Type> clo
             IStorageAccount hostAccount = await _accountProvider.GetStorageAccountAsync(context.CancellationToken);
             IStorageAccount dataAccount = await _accountProvider.GetStorageAccountAsync(context.Parameter, context.CancellationToken, _nameResolver);
 
+            if (dataAccount.Type == StorageAccountType.Premium)
```
paulbatum: It makes more sense for this logic to be applied at the ScriptHost level. At the webjobs level there is no problem with the trigger.

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
```diff
@@ -56,40 +54,42 @@ internal class DefaultStorageCredentialsValidator : IStorageCredentialsValidator
...
-            }
-
-            if (isPrimaryAccount)
-            {
-                // Primary storage accounts require Queues
-                IStorageQueueClient queueClient = account.CreateQueueClient();
-                IStorageQueue queue = queueClient.GetQueueReference("name");
-                try
+                var storageException = e as StorageException;
+                if (storageException != null && storageException.RequestInformation.HttpStatusCode == 400 &&
```
paulbatum: This needs some explanatory comments :)

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
```diff
@@ -56,40 +54,42 @@ internal class DefaultStorageCredentialsValidator : IStorageCredentialsValidator
...
+                await queue.ExistsAsync(cancellationToken);
+            }
+            catch (OperationCanceledException)
+            {
+                throw;
+            }
+            catch (StorageException exception)
+            {
+                WebException webException = exception.GetBaseException() as WebException;
+                if (webException != null && webException.Status == WebExceptionStatus.NameResolutionFailure)
```
paulbatum: Same as above, needs some //WARNING: HACK INCOMING comments.

src/Microsoft.Azure.WebJobs/StorageAccountAttribute.cs
```diff
@@ -27,14 +27,22 @@ public sealed class StorageAccountAttribute : Attribute
...
 
         /// <summary>
         /// Gets the name of the Azure Storage connection string to use.
         /// </summary>
         public string Account { get; private set; }
+
+        /// <summary>
+        /// Gets the disabled storage account types.
+        /// </summary>
+        public StorageAccountTypes Supported { get; private set; }
```
mathewc: Let's not complicate the public UI for people with this - they shouldn't have to tell us what the account type is. I don't want our inability to determine a storage account type to leak through, since that will be temporary - we need to get a way to do this from storage (which we should/will be able to do eventually). We should push instead for that.

mamaso: The motivation is to allow functions to disable blobtriggers on premium.  I agree with keeping the public api simple (this probably doesn't have much value to an average joe user).  

Is it preferable to have a way of detecting whether the web jobs sdk is running in a functions environment?  @fabiocav argued against this, hence the attribute.

mamaso: @mathewc Any preference on how to proceed?  Look for existence of FUNCTIONS_EXTENSION_VERSION and disable BlobTrigger on premium accounts?

I think it's either that or surface a method (like the storage account attribute) so that script code can control whether these are enabled, unless if you see another option.  Is there a global configuration for the core sdk?

mamaso:
@mathewc @fabiocav implemented the stuff discussed yesterday.

- Removed all public api stuff: StorageAccountType enum, StorageAccountAttribute 'SupportedTypes' property
- Disabled BlobTrigger for premium storage accounts at the webjobs level
- Added / fixed tests

Behavior:  Throws an exception like `Storage account 'name' is of unsupported type 'Premium'. Storage account must be of type(s) GeneralPurpose, BlobOnly` upon job host start

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
```diff
@@ -56,40 +55,52 @@ internal class DefaultStorageCredentialsValidator : IStorageCredentialsValidator
...
-                IStorageQueue queue = queueClient.GetQueueReference("name");
-                try
+                var storageException = e as StorageException;
+                if (storageException != null && storageException.RequestInformation != null && 
+                    storageException.RequestInformation.HttpStatusCode == 400 &&
+                    storageException.RequestInformation.ExtendedErrorInformation.ErrorCode == "InvalidQueryParameterValue")
                 {
-                    await queue.ExistsAsync(cancellationToken);
-                    _primaryCredentials = account.Credentials;
+                    // WARNING: HACK INCOMING
```
mathewc: Can you change this warning to a TODO pointing to a new bug number, tracking getting the storage team to give us a better way of identifying this? I believe you mentioned there was an issue in their repo related to this.

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
```diff
@@ -56,40 +55,52 @@ internal class DefaultStorageCredentialsValidator : IStorageCredentialsValidator
...
-                        string message = String.Format(CultureInfo.CurrentCulture,
-                            "Invalid storage account '{0}'. Primary storage accounts must be general "
-                            + "purpose accounts and not restricted blob storage accounts.", account.Credentials.AccountName);
-                        throw new InvalidOperationException(message);
-                    }
-                    throw;
+                    // WARNING: HACK INCOMING
+                    // Blob-only storage accounts do not support services other than Blob.  
+                    // If we see a name resolution failure on the queue endpoint classify as a blob-only account
+                    account.Type = StorageAccountTypes.BlobOnly;
```
mathewc: Should this method default the Type to GeneralPurpose at the top of the method?

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
```diff
@@ -56,40 +55,52 @@ internal class DefaultStorageCredentialsValidator : IStorageCredentialsValidator
...
-                    WebException webException = exception.GetBaseException() as WebException;
-                    if (webException != null && webException.Status == WebExceptionStatus.NameResolutionFailure)
-                    {
-                        string message = String.Format(CultureInfo.CurrentCulture,
-                            "Invalid storage account '{0}'. Primary storage accounts must be general "
-                            + "purpose accounts and not restricted blob storage accounts.", account.Credentials.AccountName);
-                        throw new InvalidOperationException(message);
-                    }
-                    throw;
+                    // WARNING: HACK INCOMING
```
mathewc: Why do you need this HACK comment here? I'd just remove it. We're not planning on having a better way here, right?

src/Microsoft.Azure.WebJobs.Host/Executors/StorageAccountProviderExtensions.cs
```diff
@@ -19,7 +19,7 @@ public static Task<IStorageAccount> GetDashboardAccountAsync(this IStorageAccoun
                 throw new ArgumentNullException("provider");
             }
 
-            return provider.GetAccountAsync(ConnectionStringNames.Dashboard, cancellationToken);
+            return provider.GetAccountAsync(ConnectionStringNames.Dashboard, StorageAccountTypes.GeneralPurpose, cancellationToken);
```
mathewc: I think you should have a default parameter for the account type, so for the mainline case you don't have to modify the code everywhere. We should only have to pass in this parameter in cases where you need to restrict.

src/Microsoft.Azure.WebJobs.Host/Executors/StorageAccountProviderExtensions.cs
```diff
@@ -54,7 +50,7 @@ public static Task<IStorageAccount> GetStorageAccountAsync(this IStorageAccountP
                 }
             }
 
-            return provider.GetAccountAsync(connectionStringName, cancellationToken);
+            return provider.GetAccountAsync(connectionStringName, bindingSupported, cancellationToken);
```
mathewc: I'd rename this parameter to accountTypeSupported?

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageAccountProvider.cs
```diff
@@ -169,14 +168,21 @@ private IStorageAccount StorageAccount
...
                 }
             }
 
             if (account != null)
             {
                 // On the first attempt, this will make a network call to verify the credentials work.
-                await _storageCredentialsValidator.ValidateCredentialsAsync(account, isPrimary, cancellationToken);
+                await _storageCredentialsValidator.ValidateCredentialsAsync(account, cancellationToken);
+
+                if (!supportedTypes.HasFlag(account.Type))
```
mathewc: How about we move this out into its own static helper method, since the logic is separate from validating/classifying the account. E.g. perhaps an extension method to IStorageAccount.

That way, you don't have to modify all of the existing function signatures to pass in the type. I think there are only 2 places where you're actually doing an account restriction check anyways - so you can call to the new method in those places only.

src/Microsoft.Azure.WebJobs.Host/Storage/StorageAccountType.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+namespace Microsoft.Azure.WebJobs.Host.Storage
+{
+    /// <summary>Storage account types.</summary>
+    [Flags]
+    internal enum StorageAccountTypes
```
mathewc: I don't think this enum needs to be a flags. Why can't it be named StorageAccountType, with GeneralPurpose, BlobOnly, Premium? For the checks you need to do, I don't think you need flags.

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
mamaso: sure thing

src/Microsoft.Azure.WebJobs.Host/Executors/StorageAccountProviderExtensions.cs
mamaso: good point, left over from when I was doing attributeSupported vs bindingSupported stuff

src/Microsoft.Azure.WebJobs.Host/Executors/StorageAccountProviderExtensions.cs
mamaso: I agree, in the past I've gotten flak for not having cancellation token be the last parameter so I did it this way this time.  I'm all for cleaner code so I'll use a default.

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
mamaso: the better way is storage api to determine account type, yeah hack comment is unnecessary

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
mamaso: Perhaps for clarity.  I just had every storage account constructed as GeneralPurpose by default, but it probably is useful to show that here

src/Microsoft.Azure.WebJobs.Host/Storage/StorageAccountType.cs
mamaso: Arose from the storage account attribute stuff.  I'd still say it's useful because you need to pass a list of enums around to do the below matrix, unless if we have extension methods for 'onlyAllow' and 'doNotAllow' (which would not scale if they added another account type with weird restrictions that didnt' fall in these buckets).

- Storage/Dashboard accounts / table bindings / queue bindings == GeneralPurpose
- Blob Triggers / SingletonManager == General Purpose || BlobOnly
- Blob input / output == all types

mathewc:
Much cleaner now. :ship:

PR #896, opened 2016-11-03
improving singleton logging

brettsam:
This removes our very verbose singleton logging (half of https://github.com/Azure/azure-webjobs-sdk-script/issues/728) and adds detailed debugging information should a renewal fail. Also adds function invocation ids to some of our other logging as it's currently tough to correlate events when looking through the logs.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Loggers/TraceWriterFunctionInstanceLogger.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Loggers/TraceWriterFunctionInstanceLogger.cs
@@ -25,7 +25,7 @@ namespace Microsoft.Azure.WebJobs.Host.Loggers
 
         public Task<string> LogFunctionStartedAsync(FunctionStartedMessage message, CancellationToken cancellationToken)
         {
-            string traceMessage = string.Format(CultureInfo.InvariantCulture, "Executing: '{0}' - Reason: '{1}'", message.Function.ShortName, message.FormatReason());
+            string traceMessage = string.Format(CultureInfo.InvariantCulture, "Executing '{0}' (Reason='{1}', Id={2})", message.Function.ShortName, message.FormatReason(), message.FunctionInstanceId);
             Trace(TraceLevel.Info, message.HostInstanceId, message.Function, message.FunctionInstanceId, traceMessage, TraceSource.Execution);
             return Task.FromResult<string>(null);
         }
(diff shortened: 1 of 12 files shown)


src/Microsoft.Azure.WebJobs.Host/Loggers/TraceWriterFunctionInstanceLogger.cs
```diff
@@ -25,7 +25,7 @@ public TraceWriterFunctionInstanceLogger(TraceWriter trace)
 
         public Task<string> LogFunctionStartedAsync(FunctionStartedMessage message, CancellationToken cancellationToken)
         {
-            string traceMessage = string.Format(CultureInfo.InvariantCulture, "Executing: '{0}' - Reason: '{1}'", message.Function.ShortName, message.FormatReason());
+            string traceMessage = string.Format(CultureInfo.InvariantCulture, "Executing '{0}' (Reason='{1}', Id={2})", message.Function.ShortName, message.FormatReason(), message.FunctionInstanceId);
```
mathewc: The FunctionStarted/FunctionCompleted logs already contain this Id. Not opposed to adding this here as well, just wondering how you think this will help?

src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonManager.cs
```diff
@@ -103,8 +104,6 @@ internal TimeSpan MinimumLeaseRenewalInterval
 
         public async virtual Task<object> TryLockAsync(string lockId, string functionInstanceId, SingletonAttribute attribute, CancellationToken cancellationToken, bool retry = true)
         {
-            _trace.Verbose(string.Format(CultureInfo.InvariantCulture, "Waiting for Singleton lock ({0})", lockId), source: TraceSource.Execution);
```
mathewc: What do you guys (@fabiocav @MikeStall) think about defining a set of ExtendedTraceLevel constants so that we can have a higher Diagnostic level (level 5). While not part of the System.Diagnostic.TraceLevel enum, we can still assign such an int. That would allow us to add TraceWriter.Diagnostic(). We could then retain these log statements and switch them on if needed.

src/Microsoft.Azure.WebJobs.Host/Loggers/TraceWriterFunctionInstanceLogger.cs
brettsam: I was trying to find the 'Reason' (specifically, the blob name) that triggered a failing function. I had the invocation id that I was looking for, but when you're combing through logs of hundreds of functions/minute, all the logs are interleaved and it's really hard to figure out which 'Reason' goes with the 'Id'. This just removes that issue.

src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonManager.cs
brettsam: I agree. I think we do need some higher level of debug logs that go beyond 'Verbose'. I've really wanted to get super-detailed with what we do in the blob trigger, for example, so it's easier to track down tough scenarios. Today you almost need to attach a debugger to get a good sense of what's happening.

src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonManager.cs
```diff
@@ -584,15 +587,29 @@ public RenewLeaseCommand(IStorageBlockBlob leaseBlob, string leaseId, string loc
...
+                        // Log the details we've been accumulating to help with debugging this scenario
+                        DateTimeOffset now = DateTimeOffset.UtcNow;
+                        StringBuilder sb = new StringBuilder();
+                        sb.AppendFormat("Singleton lock renewal failed ({0})", _lockId).AppendLine();
+                        sb.AppendFormat("  Last successful request:         {0} ({1} ago)", _lastSuccessRequest.ToString("o"), now - _lastSuccessRequest).AppendLine();
+                        sb.AppendFormat("  Last successful response:        {0} ({1} ago)", _lastSuccessResponse.ToString("o"), now - _lastSuccessResponse).AppendLine();
+                        sb.AppendFormat("  Last successful request latency: {0}", _lastSuccessResponse - _lastSuccessRequest).AppendLine();
+                        sb.AppendFormat("  Last failed request:             {0} ({1} ago)", _lastFailedRequest.ToString("o"), now - _lastFailedRequest).AppendLine();
+                        sb.AppendFormat("  Last delay:            {0}", _lastDelay);
+
```
brettsam: I came up with a better format for these logs -- I'll update this shortly.

src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonManager.cs
brettsam: @mathewc -- I know there's an outstanding question above -- should we add that as an issue to address in the future? Cleaning up these logs now would still be valuable, I think.

mathewc:
:ship:

PR #910, opened 2016-11-16
Expanding system binding parameter support (adding datetime) (#731)

### LINKED ISSUE #731: Add support for a built-in datetime binding
Similar to our existing {rand-guid} binding which allows people to easily generate unique blob paths declaratively (e.g. `path: "images/{rand-guid}")` we should also support a datetime binding that allows people to bind to date/time strings.

This might look like **{datetime:mm-dd-yyyy}** allowing standard date/time format strings to be specified. So in your binding you can simply do:

```
{
  "type": "blob",
  "name": "myOutputBlob",
  "path": "outcontainer/{datetime:yyyy-mm-dd}",
  "connection": "function43dc41c29824_STORAGE",
  "direction": "out"
}
```

We should generalize this so the pattern is `{name:format}` allowing format strings to be specified. E.g. allow `{datetime:yyyy-mm-dd}` to be specified, or `{rand-guid:G}`.

mathewc:
Resolves https://github.com/Azure/azure-webjobs-sdk-script/issues/731

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingParameterResolver.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingParameterResolver.cs
@@ -1,0 +1,123 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.ObjectModel;
+using System.Globalization;
+using System.Linq;
+
+namespace Microsoft.Azure.WebJobs.Host.Bindings.Path
+{
+    internal abstract class BindingParameterResolver
+    {
+        private static Collection<BindingParameterResolver> _resolvers;
+
+        static BindingParameterResolver()
+        {
+            // create the static set of built in system resolvers
+            _resolvers = new Collection<BindingParameterResolver>();
+            _resolvers.Add(new RandGuidResolver());
+            _resolvers.Add(new DateTimeResolver());
+        }
+
+        public abstract string Name { get; }
+
+        public static bool IsSystemParameter(string value)
+        {
+            if (string.IsNullOrEmpty(value))
+            {
+                throw new ArgumentNullException("value");
+            }
+
+            BindingParameterResolver resolver = null;
+            return TryGetResolver(value, out resolver);
+        }
+
+        public static bool TryGetResolver(string value, out BindingParameterResolver resolver)
+        {
+            if (string.IsNullOrEmpty(value))
+            {
+                throw new ArgumentNullException("value");
+            }
+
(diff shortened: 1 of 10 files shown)


src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingParameterResolver.cs
```diff
@@ -0,0 +1,123 @@
...
+namespace Microsoft.Azure.WebJobs.Host.Bindings.Path
+{
+    internal abstract class BindingParameterResolver
+    {
+        private static Collection<BindingParameterResolver> _resolvers;
+
+        static BindingParameterResolver()
+        {
+            // create the static set of built in system resolvers
+            _resolvers = new Collection<BindingParameterResolver>();
```
mathewc: We can easily open this up in the future for extensibility, but for now just keeping these interfaces internal.

test/Microsoft.Azure.WebJobs.Host.UnitTests/TypeUtilityTests.cs
```diff
@@ -13,7 +13,6 @@ public class TypeUtilityTests
         [InlineData(typeof(string), false)]
         [InlineData(typeof(int), false)]
         [InlineData(typeof(int?), true)]
-        [InlineData(typeof(Nullable<int>), true)]
```
mathewc: We were getting warnings from xunit for duplicate test cases here and below

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingParameterResolver.cs
```diff
@@ -0,0 +1,123 @@
...
+                }
+            }
+
+            public override string Resolve(string value)
+            {
+                string format = GetFormatOrNull(value);
+
+                if (!string.IsNullOrEmpty(format))
+                {
+                    return DateTime.Now.ToString(format, CultureInfo.InvariantCulture);
```
mathewc: Using local rather than utc so that users can control this via the WEBSITE_TIME_ZONE WebApp setting. By default, VMs will already be using UTC, but if people want to change their timezone they can via this setting, and using Local time will allow that to be picked up properly. See https://github.com/Azure/azure-webjobs-sdk-extensions/issues/32 for details.

fabiocav (approved):

alexgman:
Anyone know where can we find a list of the currently-available parameter bindings, such as `{rand.guid}` or `{dateTime.Now}`?

PAVaillancourt-IvadoLabs:
@alexgman here's the documentation page about bindings, there is a "list" of parameter bindings at the bottom.
https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-expressions-patterns

Issue #905, opened 2016-11-13
ServiceBus: Improve JSON conversions with null ContentType


brettsam:
Our converters special-case a few `BrokeredMessage` `ContentTypes` (links below). If you supply a message with a JSON string but do not set the appropriate `ContentType` you end up with a vague message (especially in Scripts): `There was an error deserializing the object of type System.String. The input source is not correctly formatted. System.Runtime.Serialization: The input source is not correctly formatted.` 

Theory -- it looks like ServiceBus will assume that a `null` ContentType is XML. Since the JSON is not XML, it throws. This may apply to the byte[] converter as well.

We should remove these checks and just read from the Stream directly and pass the string/byte[] along. We'll need to verify that this won't break any scenarios.

Links:
- https://github.com/Azure/azure-webjobs-sdk/blob/master/src/Microsoft.Azure.WebJobs.ServiceBus/Triggers/BrokeredMessageToStringConverter.cs#L22-L24
- https://github.com/Azure/azure-webjobs-sdk/blob/master/src/Microsoft.Azure.WebJobs.ServiceBus/Triggers/BrokeredMessageToByteArrayConverter.cs#L16-L18

PR #912, opened 2016-11-17
ServiceBus: improving BrokeredMessage string and byte[] conversions

brettsam:
Improving `BrokeredMessageToString` by not special-casing any `ContentType`s. We'll first try to read from the Stream and return that string. If that fails (because it was serialized differently), we'll then try the default serializer used by Service Bus. If that fails, we'll provide a slightly better error message than the default. Previously, sending a plain string with a null ContentType would fail as we'd attempt to deserialize it using the default serializer.

Improving `BrokeredMessageToByteArray` by throwing a slightly better error message if we're unable to deserialize. We can't implement the fallback here because deserializing a byte[] from a Stream will always work -- so we use the ContentTypes as a hint. The behavior here hasn't changed but I wrap the exception message (which was previously confusing, especially when seen in Functions) and added a bunch more tests to make the behavior clear.

Addresses #905

### DIFF
--- a/src/Microsoft.Azure.WebJobs.ServiceBus/Triggers/BrokeredMessageToStringConverter.cs
+++ b/src/Microsoft.Azure.WebJobs.ServiceBus/Triggers/BrokeredMessageToStringConverter.cs
@@ -2,7 +2,9 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Globalization;
 using System.IO;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host.Converters;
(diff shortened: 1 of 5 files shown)


brettsam:
@mathewc @MikeStall -- pinging for review.

src/Microsoft.Azure.WebJobs.ServiceBus/Triggers/BrokeredMessageToStringConverter.cs
```diff
@@ -2,54 +2,76 @@
...
-                if (stream == null)
+                cancellationToken.ThrowIfCancellationRequested();
+                try
                 {
-                    return Task.FromResult<string>(null);
+                    return await reader.ReadToEndAsync();
+                }
+                catch (DecoderFallbackException)
+                {
+                    // we'll try again below
```
MikeStall: This is a bit confusing:
1. If the body is read once, then can we safely do a 2nd attempted read if line 37 throws? Or do we need to buffer the contents?
2. TextReader.ReadToEnd  isn't documented to throw DecoderFallbackException. (https://msdn.microsoft.com/en-us/library/system.io.textreader.readtoendasync(v=vs.110).aspx ) Where does that come from?

mathewc: If you search for DecoderFallbackException you'll find we have handling for this across many of our other bindings as well. I know I've seen this thrown and have put some of these in as well.

src/Microsoft.Azure.WebJobs.ServiceBus/Triggers/BrokeredMessageToStringConverter.cs
```diff
@@ -2,54 +2,76 @@
...
             {
                 throw new ArgumentNullException("input");
             }
+            Stream stream = input.GetBody<Stream>();
+            if (stream == null)
+            {
+                return null;
+            }
 
-            if (input.ContentType == ContentTypes.TextPlain ||
```
mathewc: You're removing all of these in this converter, but leaving them in the byte[] converter. Can we be consistent across all?

src/Microsoft.Azure.WebJobs.ServiceBus/Triggers/BrokeredMessageToStringConverter.cs
```diff
@@ -2,54 +2,76 @@
...
+                {
+                    // we'll try again below
                 }
 
-                try
+                // We may get here if the message is a string yet was DataContract-serialized when created. We'll
+                // try to deserialize it here. This may fail as well, in which case we'll try to give a decent error. We cannot read the 
+                // body twice so we need to use the DataContractSerializer directly rather than use input.GetBody<string>().
+                stream.Position = 0;
+                DataContractSerializer serializer = new DataContractSerializer(typeof(string));
```
mathewc: Can you just create a new local BrokeredMessage passing in the stream to the ctor and use the SB APIs for reading? That way we're still getting their serialization logic.

brettsam: Good points. Let me give those a try...

src/Microsoft.Azure.WebJobs.ServiceBus/Triggers/BrokeredMessageToStringConverter.cs
brettsam: It happens if the Reader can't deal with the bytes in the stream -- the message is: `Unable to translate bytes [99] at index 61 from specified code page to Unicode.` Because we don't own both ends of the pipeline, we need to deal with messages that are just plain strings (in which case this works) and those that are DataContract-serialized (and you need to go back to using `GetBody<string>()`).

src/Microsoft.Azure.WebJobs.ServiceBus/Triggers/BrokeredMessageToStringConverter.cs
brettsam: Looks like this works with a `Clone()`.

src/Microsoft.Azure.WebJobs.ServiceBus/Triggers/BrokeredMessageToStringConverter.cs
brettsam: Unfortunately no -- there's really no failure for the byte[] converter so there's nothing to fallback to. Since we're just straight copying the bytes, it won't ever hit any Decoder exception (or similar) -- it doesn't care, it'll copy the bytes happily either way. So we need to use the ContentType as a hint as to how things were encoded.

mathewc:
Looks good to me :ship:

src/Microsoft.Azure.WebJobs.ServiceBus/Triggers/BrokeredMessageToStringConverter.cs
brettsam: I agree it's not obvious. I had to write some tests to make sure I understood (there's combinations of all this in the tests I added). There's (at least) two ways you can create a BrokeredMessage and pass it bytes:
1. You pass bytes in and let the BrokeredMessage deal with the serialization:
     
        byte[] bytes = Encoding.UTF8.GetBytes("test");
        BrokeredMessage message = new BrokeredMessage(bytes);
2. You create a Stream yourself and pass that in: 

        MemoryStream ms = new MemoryStream();
        StreamWriter sw = new StreamWriter(ms);
        sw.Write("test");
        sw.Flush();
        ms.Seek(0, SeekOrigin.Begin);
        BrokeredMessage message = new BrokeredMessage(ms);

And then there are two ways to try to get the bytes out:
A. `message.GetBody<byte[]>()`
B. `message.GetBody<Stream>()` and then read as a `MemoryStream`

Let me run through what happens if you mix these and then read the bytes back into a string:
1A: `test`
1B: `@base64Binary3http://schemas.microsoft.com/2003/10/Serialization/�test`
2A: `SerializationException: The input source is not correctly formatted.`
2B: `test`

So if you're writing in as a Stream, you need to tell us via the `ContentType` so we can read it out correctly and give you 2B -- otherwise you end up with the exception in 2A.
I imagine the most common path is 1A, but the other support was always there so I didn't do much to touch it, other than provide a nicer message should you hit 2A.

brettsam:
Made a tiny change to use an interpolated string in the ServiceBus assembly. We'll see if it works...

Issue #837, opened 2016-09-21
SingletonManager Lease Improvement


mathewc:
We should investigate using a proposed lease ID when acquiring our blob leases in SingletonManager.TryAcquireLeaseAsync (code [here](https://github.com/Azure/azure-webjobs-sdk/blob/master/src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonManager.cs#L366)). Currently when we loose a lease and try to reacquire it, we have to wait for the lease to expire. If we use the same lease ID, we'll be able to acquire it directly. This will improve startup times for the host.

E.g. we may just be able to use JobHost.HostId.

It's easy to see the existing issue. If you start/start a single instance of the Samples app (which uses Timers) back to back, you'll see the second time you start it up, there will be a delay before the timer listener can start up, because it is waiting to acquire the lease. If running with AZURE_WEBJOBS_ENV = Development, the default lease length is 15 seconds, so the delay will be a max of 7 seconds. However, when running in production, the delay can be up to 30 seconds. If the above fix works as intended, we should be able to reacquire our lost lease immediately on restart.

lindydonna:
@mathewc Is talking to himself. He will fill out the customer scenario later OR will fix the bug first.

brettsam:
One thing that's not mentioned above is that Functions puts the Host into Development mode when you are connected via the portal (and that setting remains until the host restarts, which could be a while). This means that your Singleton leases are reduced down to 15 seconds, greatly increasing the chance of a 409 under load. This change removes the need to shorten that lease time, which means they will always be 60 seconds.

I'm hoping it'll slightly help with #822 for those using Functions.

brettsam:
We had a discussion a while back on this (see #913 for some of it) and came to the conclusion that the risk here was far greater than the reward. This was mostly to allow for faster turnaround during development, but it brought up a whole mess of issues that could potentially break. The major one being that only one host could be running on a machine at a time. In the Script world, we often have overlapping hosts -- so that would cause all kinds of issues. Closing this out as I don't think there's a great solution.

Issue #770, opened 2016-07-15
Consider Redesign of Control Queue monitoring


mathewc:
The JobHost monitors host instance specific control queues for Dashboard communication (Abort messages, etc.). Since each _instance_ of the host has a new ID, these means a new queue for each instance. Now these queues are only created if the Dashboard actually adds a message (e.g. the user clicks the "Abort" button). However, it does mean that the host is constantly **polling non existent queues** waiting for a message. This results in storage 404s.

While these aren't errors per say, we have had lots of customers complain about this, since it fills their storage logs, App Insights logs, etc. with tons of 404s, producing a lot of noise.

See these related issues:
- https://github.com/Azure/azure-webjobs-sdk/issues/765
- https://github.com/Azure/azure-webjobs-sdk/issues/1215
- https://github.com/Azure/azure-webjobs-sdk/issues/1483

hrboyceiii:
:+1:  
It took me a while to figure out what was going on here.  Is there a reason the SDK doesn't create the Queue -- either on startup or another point -- similar to the behavior for poison queues?

What would be the preferred approach here?

mathewc:
Moving this back to triage, since people are using App Insights more and more, causing these 404s to become more visible. See https://github.com/Azure/azure-webjobs-sdk/issues/770

mathewc:
@hrboyceiii Note that these queues are function invocation instance specific - so they can't be created on startup. The instance ID is only know when an invocation has started.

christopheranderson:
@christopheranderson - this will impact monitoring improvements you're looking at

snboisen:
I see some serious problems with all this 404 garbage being generated by WebJobs SDK:
* User confusion - a working program is not supposed to generate errors.
* Noise - the failures generated by WebJobs SDK may mask real program errors.
* Quota expenditure - we are on the free plan for Application Insights and are constantly hitting the cap.
* Frustrates WebJobs debugging - we have had an issue with the WebJobs dashboard since the beginning, but were unable to ever solve it. These constant 404 errors makes it hard to debug problems with the WebJobs SDK.

christopheranderson:
@mathewc - since we're not planning on investing in the dashboard going forward, would it make sense to give folks the option of disabling these control queues, thus avoiding the 404 noise?

darewreck54:
@christopheranderson Does that mean that the dashboard will be treated as is or would it be depreciated in favor of other logging systems?

[...]

PR #913, opened 2016-11-17
adding HostInstanceId and using it for proposed singleton listener locks

brettsam:
Addresses #837.


src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
```diff
@@ -123,6 +125,31 @@ public string HostId
...
+                {
+                    _hostInstanceId = Environment.GetEnvironmentVariable("WEBSITE_INSTANCE_ID")
+                        ?? Environment.MachineName.GetHashCode().ToString("X", CultureInfo.InvariantCulture).PadLeft(32, '0');
+
+                    _hostInstanceId = _hostInstanceId.Substring(0, 32).ToLowerInvariant();
+                }
+
+                return _hostInstanceId;
+            }
+
```
brettsam: This logic was all taken from Script -- we should be able to remove that and use this property instead.

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
```diff
@@ -282,7 +309,6 @@ public void UseDevelopmentSettings()
         {
             Tracing.ConsoleLevel = TraceLevel.Verbose;
             Queues.MaxPollingInterval = TimeSpan.FromSeconds(2);
-            Singleton.ListenerLockPeriod = TimeSpan.FromSeconds(15);
```
brettsam: Note that I removed this restriction now. I think this was there to allow you to kill the process and restart without waiting a minute. Now that's not a concern. This will also keep things more consistent in Functions so debug mode won't shorten your timeout until the host restarts.

brettsam:
@mathewc @fabiocav -- can I get a review on this? I'll need to rebase before getting in, but it shouldn't change much.

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
mathewc: Yes - this lock period issue for local development (frequent restarts) has been very problematic.

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostContextFactory.cs
```diff
@@ -188,8 +188,7 @@ public JobHostContextFactory(IStorageAccountProvider storageAccountProvider, ICo
                         queueConfiguration, exceptionHandler, trace, functions,
                         functionInstanceLogger, functionExecutor);
 
-                    Guid hostInstanceId = Guid.NewGuid();
```
mathewc: Note that technically this will be a breaking change and would affect anyone trying to run multiple instances of a JobHost on a single machine. Not sure how prevalent that is, but in such cases you'd have two separate instances using the same lease ID, and potentially other collisions where we assumed instance IDs were globally unique before.

I do think that spinning up multiple hosts on the same machine is an anti pattern, and the benefits of this change likely outweigh these concerns. I just want to make sure we're not missing anything here.

@MikeStall ?

mathewc: E.g. previously one way I would test singleton support was to spin up multiple instances of the same host on a single machine, so I could watch one get the lease, and the other fail to get it. What would happen now? The second host instance would steal the lease from the first? What happens when the first instance loses its lease?

fabiocav: With this approach, they would both, technically, get a successful lease.

src/Dashboard/Data/FunctionInstanceSnapshot.cs
```diff
@@ -17,7 +17,7 @@ public class FunctionInstanceSnapshot
 
         public Guid Id { get; set; }
 
-        public Guid HostInstanceId { get; set; }
```
fabiocav: Would @MikeStall be able to comment on that?

brettsam: Based on this code, I suspect that it would break something. https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Dashboard/Data/FunctionInstanceSnapshot.cs#L20

@MikeStall -- how would I verify that this stuff works if we change this from Guid to string?

brettsam: Oh, shoot, but the dashboard is deployed at a different cadence. So it's possible that we are writing logs with strings yet the Dashboard reads them as Guids and fails.

mathewc: Yeah, we need to ensure this isn't breaking. Likely that means manual testing of all the various scenarios. I'm not sure my self from looking at the code. We'll also need to get a clean run of the Dashboard tests, which again is manual currently because they're not yet running on CI.

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostContextFactory.cs
mathewc: Yeah, I agree. I'm just pointing out that there is a non-zero chance that someone might get bit by this. But if they do, they're really doing an anti-pattern. Also for core SDK this is a new major release, so we can make a break here.

src/Dashboard/Data/FunctionInstanceSnapshot.cs
mathewc: As far as testing goes, I believe the host instance Id is primarily used for host specific messaging from Dashboard (e.g. host "Abort" for runaway functions, etc.). See https://github.com/Azure/azure-webjobs-sdk/issues/770 for details.

mathewc:
Signing off on the code changes - they look good. Just need to do a lot of testing for Dashboard and Singleton to make sure that everything still works.

src/Dashboard/Data/FunctionInstanceSnapshot.cs
brettsam: I sent a mail to @MikeStall  and @mathewc with the details, but my testing this morning found an issue. The Dashboard uses 'host message queue' name to determine if it has already aborted that host. It expects this name to be different with every invocation of the host so it can mark it as an 'abort receipt' in blobs.

I'm looking now to keep the old behavior for `hostInstanceId` so that everything continues working with the Dashboard and instead rename my property `HostStableId` and use that exclusively for the blob lease ids. That should keep the dashboard behavior the same.

src/Microsoft.Azure.WebJobs.Host/Blobs/Triggers/BlobTriggerAttributeBindingProvider.cs
```diff
@@ -45,7 +46,8 @@ internal class BlobTriggerAttributeBindingProvider : ITriggerBindingProvider
             IContextSetter<IMessageEnqueuedWatcher> messageEnqueuedWatcherSetter,
             ISharedContextProvider sharedContextProvider,
             SingletonManager singletonManager,
-            TraceWriter trace)
+            TraceWriter trace,
+            string hostMachineId)
```
mathewc: Good, I was just about to comment that you update all the param names for clarity :)

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/WebJobs.Host.EndToEndTests.csproj
```diff
@@ -36,6 +36,15 @@
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <LangVersion>default</LangVersion>
   </PropertyGroup>
+  <PropertyGroup>
+    <SignAssembly>true</SignAssembly>
+  </PropertyGroup>
+  <PropertyGroup>
+    <AssemblyOriginatorKeyFile>..\..\src\Common\PublicKey.snk</AssemblyOriginatorKeyFile>
```
mathewc: Curious why this was needed?

brettsam: I access the internal setter for `HostMachineId` for testing, so I had to add `InternalsVisibleTo` to this project. And to do that, I needed it to be signed to match how we're doing it with the other test assemblies. I want to understand why it worked before, though. Suddenly changing the name to `HostMachineId` triggered 'not accessible' error messages. I'll figure it out before I commit.

brettsam: Well, I'm at a loss for why this was able to build before. But we don't have AppVeyor building PRs yet so my guess is it would have failed on my previous PR without it -- I must have set it up that way locally and forgot to commit before I left on vacation a couple weeks ago. 

Long story short -- we need `InternalsVisibleTo` so we can simulate two different machines, so it's required.

mathewc: It was by design that the e2e tests didn't have internals visible, to ensure that the full integration tests are actually using the real product and not doing anything funky.

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
```diff
@@ -123,6 +125,31 @@ public string HostId
...
+                        ?? Environment.MachineName.GetHashCode().ToString("X", CultureInfo.InvariantCulture).PadLeft(32, '0');
+
+                    _hostMachineId = _hostMachineId.Substring(0, 32).ToLowerInvariant();
+                }
+
+                return _hostMachineId;
+            }
+
+            // for testing
+            internal set
```
mathewc: If we make this public, we'll also probably want to put validation in place similar to what we have for host id. Since this value will be used in queue names and storage entities, only valid storage characters are allowed.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/WebJobs.Host.EndToEndTests.csproj
brettsam: I had it public at first (I think you saw the getter/setter comment) but realized that no one should ever really be setting this unless they want to run two hosts on the same machine, which I figured would only be us in the tests. 

But this does open up the possibility of running two hosts locally for testing, which you'd asked about in another comment.

The risk is that someone doesn't understand this and manually sets the HostMachineId and then deploys to two instances -- they'd basically be exchanging the lease back and forth.

But I can add a comment that the setter should only be used for running multiple hosts on the same machine...

mathewc: Well, HostId itself is settable, but users really don't ever need to do that except in advanced scenarios.

See https://github.com/Azure/azure-webjobs-sdk-script/issues/1052. I wonder if for local CLI we won't also need the public ability to configure this as well?

brettsam: I think making it settable is fine (I've already started) -- was just trying to justify why i was hiding it before :-) 

@fabiocav -- it looks like the requirements for a LeaseId are that it must be a GUID. Do you remember why you kept the `HostInstanceId` a string in the Script repo? It seems like the easiest thing to do from a type-checking standpoint is just make it a GUID rather than a string.

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
```diff
@@ -123,6 +125,31 @@ public string HostId
...
+        /// Gets the host machine ID. All host instances on the same machine will have the same HostMachineId.
+        /// </summary>
+        public string HostMachineId
+        {
+            get
+            {
+                if (_hostMachineId == null)
+                {
+                    _hostMachineId = Environment.GetEnvironmentVariable("WEBSITE_INSTANCE_ID")
+                        ?? Environment.MachineName.GetHashCode().ToString("X", CultureInfo.InvariantCulture).PadLeft(32, '0');
```
MikeStall: >GetHashCode [](start = 51, length = 11)

I believe GetHashCode() is not deterministic and could return different results when called in different process invocations or different appdomains.

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
MikeStall: Beware, each storage entity has different naming rules.  A summary here: [https://blogs.msdn.microsoft.com/jmstall/2014/06/12/azure-storage-naming-rules/](https://blogs.msdn.microsoft.com/jmstall/2014/06/12/azure-storage-naming-rules/)

---
In reply to: [92639500](https://github.com/Azure/azure-webjobs-sdk/pull/913#discussion_r92639500) [](ancestors = 92639500)

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
```diff
@@ -123,6 +125,31 @@ public string HostId
             }
         }
 
+        /// <summary>
+        /// Gets the host machine ID. All host instances on the same machine will have the same HostMachineId.
```
MikeStall: >same machine [](start = 64, length = 12)

What is the significant of "same machine"?   (this is potentially dangerous - I'm trying to understand the invariants and what assumptions we're making about this value and what we're using this for)  
If 2 instances are on different machines, can they return the same HostMachineId?

MikeStall:
>                     await _leaseBlob.RenewLeaseAsync(condition, null, null, cancellationToken);

Does Renew() need to be updated with any of this? 

---
Refers to: src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonManager.cs:580 in 17ab79c. [](commit_id = 17ab79c80938f654b75a7e64d90d28ea5d770cc6, deletion_comment = False)

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
brettsam: It's certainly possible to have two machines with the same name (although is that possible in Antares?) -- but highly unlikely. If this happens, the Lease will keep getting exchanged back-and-forth between those machines. We'd have the same issue in the ScriptHost if this happens -- two hosts would think they're the primary host.

@fabiocav -- when doing this from ScriptHost, did you give thought to anything else on the machine that would work better than the machine name?

brettsam:
Renew doesn't need this -- once you get a lease id back from Storage, you use that going forward (even though it's the same). So we just need it for the initial acquisition of the lease.

src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonListener.cs
```diff
@@ -116,7 +118,7 @@ private void OnLockTimer(object sender, ElapsedEventArgs e)
 
         internal async Task TryAcquireLock()
         {
-            _lockHandle = await _singletonManager.TryLockAsync(_lockId, null, _attribute, CancellationToken.None, retry: false);
+            _lockHandle = await _singletonManager.TryLockAsync(_lockId, _hostMachineId, null, _attribute, CancellationToken.None, retry: false);
```
MikeStall: >_hostMachineId [](start = 72, length = 14)

Passing the host machine id straight through as the proposed lease id raises some questions:
- do they have the same naming rules. valid character sets, etc? 
- do they have the same casing rules?

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
mathewc: Why do we even need to use GetHashCode at all? In the end, all we're trying to do is get a stable 32 character string, right?

brettsam: It must be a Guid to be used as a proposed lease id. 

From the docs (https://docs.microsoft.com/en-us/rest/api/storageservices/fileservices/lease-blob) : x-ms-proposed-lease-id: Proposed lease ID, in a GUID string format.

src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonListener.cs
brettsam: This is now a Guid so none of that matters.

From the docs (https://docs.microsoft.com/en-us/rest/api/storageservices/fileservices/lease-blob) : x-ms-proposed-lease-id: Proposed lease ID, in a GUID string format.

MikeStall: If we're relying on it being a guid, can we change _hostMachineId's type from string to a guid? 
For example, even "string that's really a guid"  is dangerous - what are the casing rules, and are dashes allowed?   Some azure things (like blob names)  are case *sensitive*, so the same Guid could be converted to different strings. 

---
In reply to: [92680067](https://github.com/Azure/azure-webjobs-sdk/pull/913#discussion_r92680067) [](ancestors = 92680067)

brettsam:
We decided that this was too disruptive and didn't solve all the problems we wanted to solve. Closing.

Issue #681, opened 2016-03-30
Move SDK and Dashboard CI to App Veyor


mathewc:
We need to migrate our remaining Team City jobs off of Team City. This will also allow us to start using C# 6.

mathewc:
Update: I've done the work to move the tests to the public TestEasy package. That means our tests no longer run on our CI. We should move this project to our new AppVeyor CI setup.

paulbatum:
Here's my thoughts on how to to tackle this:
1. Leave the signed SDK build on TC for now
2. Update the signed SDK job to drop the bits into a location accessible by AppVeyor (we could try mounting azure files as a file share)
3. Move the dashboard job to AppVeyor (it appears they have UI testing support - if it doesnt work out we can fallback to using our internal jenkins server)

mathewc:
This work will also pave the way for us moving the SDK job over. With that job outputting its artifacts to an azure files share, we can move it to App Veyor as well, and set up a TeamCity / Jenkins job to do the signing.

mamaso:
Closing this.

SDK has been moved, signed build has been moved, and dashboard has been deprioritized.

PR #914, opened 2016-11-17
Add appveyor test support

mamaso:
Added 'StartSafe' extension method to avoid JobHost start deadlocks.
Couple test fixes due to UTC time issues.
Couple fixes due to line ending differences

Do we want the same 'UT' test projects for the core sdk?

### DIFF
--- a/WebJobs.proj
+++ b/WebJobs.proj
@@ -141,6 +141,21 @@
     <xunit Assemblies="@(UnitTestAssemblies)"/>
   </Target>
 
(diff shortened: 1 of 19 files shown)


mamaso:
#681

mamaso:
@mathewc think this is a bit cleaner & less error prone, thanks for the suggestions.

Created a SyncContextTestFramework to override XunitTestFramework and marked the test assemblies to use that framework, which allows for global beforeAll and afterAll behavior

mamaso:
temporarily closing this, something's screwy up with the assembly method

PR #915, opened 2016-11-18
Add appveyor test support

mamaso:
@mathewc @fabiocav 

After a bit of thinking, fudging with xunit or creating test only fixes aren't the way to go, we should prevent the deadlock itself, as with a funky setup a dev could run into this (+ it's cleaner).

Read this article, think it gives a good summary of options: https://blogs.msdn.microsoft.com/pfxteam/2012/04/13/should-i-expose-synchronous-wrappers-for-asynchronous-methods/

We could use `configureAwait(false)` all the way down, but that's a huge change and would break anyone calling sync methods from a sync context which requires posting back to the same thread.

`Task.Run(...)` in the publicly exposed Sync() methods executes the async call on a background thread, which keeps it from interacting with the caller's sync context (which is good, as if the caller is using a sync context *and* using a non-async method they aren't following good patterns & could run into this deadlock / post back problem)

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/JobHost.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/JobHost.cs
@@ -105,7 +105,7 @@ namespace Microsoft.Azure.WebJobs
         /// <summary>Starts the host.</summary>
         public void Start()
         {
-            StartAsync().GetAwaiter().GetResult();
+            Task.Run(() => StartAsync()).GetAwaiter().GetResult();
         }
 
         /// <summary>Starts the host.</summary>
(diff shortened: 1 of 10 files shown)


mathewc:
If these changes get the tests passing on AppVeyor and they also run locally, please check get this in.

I think we may want to revisit the Task.Run, but I can do that as a follow up once we have a job that is successfully running.

src/Microsoft.Azure.WebJobs.Host/JobHost.cs
```diff
@@ -105,7 +105,7 @@ internal IListener Listener
         /// <summary>Starts the host.</summary>
         public void Start()
```
fabiocav: @mathewc @mamaso this keeps coming back to mind. As we discussed, this is fine to get us unblocked and things up and running on with CI, but one conversation I'd like to get started is considering removal of **all** of this sync-over-async methods as they're always a source of issues, hide some important details from the caller and really provide very little benefit (a caller could still choose to make the async call if they wished and they'd be in control of how it's being scheduled, how continuations are setup, etc.).
Just planting the seed :)

/cc @MikeStall

PR #925, opened 2016-11-28
Improved stability of using Azure Storage

EranGon:
Improved stability of using Azure Storage:
- Added retry policy to all storage operations
- Fixed exception handling to not use ExtendedErrorInformation


mathewc:
Thanks @EranGon for taking the time to look into this. We'll get the right fix in place.

Issue #916, opened 2016-11-18
Functions show as running forever if end events are missing


fabiocav:
Functions appear as running if end events are missing.

#### Repro steps

Provide the steps required to reproduce the problem

1. Create a C# function that awaits on a long `Task.Delay`
2. In Kudu, kill the host process (ideally, wait for the function to appear in monitoring before doing so)
3. The function will continue to show as running, indefinitely
 
![skypeteams-image](https://cloud.githubusercontent.com/assets/2507935/20449030/d5c6ba7c-ad9b-11e6-9e7f-effc4e16a6fe.png)

#### Expected behavior
The function should show as terminated (failed)

#### Actual behavior

Function shows as running.

fabiocav:
@MikeStall, FYI

MikeStall:
We need a heartbeat so we can tell the difference between Orphaned & Running.  This is how the WebJobs dashboard solves this.

tim-schulze:
What is meant by "end events"?  We discovered a similar issue with some functions we developed that are showing as never ending in the invocation log-- some of these go back for days.  The confusing part is that our functions are simple and don't include loops/wait states so we don't understand why they wouldn't end normally since we aren't manually killing the host process.

MikeStall:
@tim-schulze  - WebJobs (and Functions) record when the function starts and when it ends.   But if the process dies in the middle, before the function finishes, then it only has a "start" recorded and no "end".   Naively, that'd make it look like it was still running. WebJobs SDK handles this by emitting a periodic "heartbeat" that lets us know the function is still alive (this is a similar pattern to how Azure  Queue Messages know the msg is still being processed) .   Thus if we have no End time, but the heartbeat is "recent", we know the function is still running. If the heartbeat is stale, we know the function was abandoned. 

The bug here is that Functions is not doing this heartbeat.

tim-schulze:
@MikeStall Thanks for the explanation, much appreciated.

MikeStall:
Tracked by https://github.com/projectkudu/AzureFunctionsPortal/issues/843

PR #928, opened 2016-11-30
Fix https://github.com/Azure/azure-webjobs-sdk/issues/916

MikeStall:
If a function is terminated, don't leave the function status orphaned as "Running".
Add Heartbeat, and then we can tell the difference between "IsRunning" and "NeverFinished"

Heartbeat is written on the same thread that writes to the existing entities. This avoids a category of evil race conditions where multiple threads update the same entity.

Give FunctionInstanceLogItem a  Refresh() that gets called for heart beats. That hook could also be used to do live parameter updates.

We need to plumb this through both the Function instance and Recent instance index.
Also expose plumb through to expose in existing portal UI.

Switch from DateTimeOffset to UTC DateTime. (These serialize the same way in Azure tables, so not a breaking change)

Don't serialize Status anymore. Instead, compute status based on other properties.
1. This avoids redundancy (Where status and other props were inconsistent),
2. It's more condusive to changing status.

Make some more types non-public. They were only public for unit testing, and the tests are now internalsVisibleTo

Significant improvement in unit test.


src/Dashboard/ApiControllers/FunctionsController.cs
```diff
@@ -341,6 +341,7 @@ private InvocationLogViewModel CreateInvocationEntry(RecentInvocationEntry entry
             metadataSnapshot.EndTime = entry.EndTime;
             metadataSnapshot.Succeeded = entry.Succeeded;
             metadataSnapshot.Heartbeat = entry.Heartbeat;
+            metadataSnapshot.HeartbeatExpiredTime = entry.HeartbeatExpiredTime;
```
mathewc: You can't roll this in to the existing Heartbeat hanging off of the snapshot because that is the legacy blob stuff? It will be somewhat confusing to have both of these properties floating around in the codebase.

src/Dashboard/ApiControllers/FunctionsController.cs
```diff
@@ -626,6 +627,12 @@ public IHttpActionResult Abort(string instanceQueueName)
 
         private bool? HostInstanceHasHeartbeat(FunctionInstanceSnapshot snapshot)
         {
+            if (snapshot.HeartbeatExpiredTime.HasValue)
+            {
+                var now = DateTime.UtcNow;
```
mathewc: nit unecessary local?

src/Dashboard/Data/FunctionInstanceSnapshot.cs
```diff
@@ -45,6 +45,9 @@ public class FunctionInstanceSnapshot
 
         public DateTimeOffset? EndTime { get; set; }
 
+        // For fast-style logging.
```
mathewc: Can you add a better description for this? What exactly does this value represent

src/Dashboard/Data/RecentInvocationEntry.cs
```diff
@@ -64,6 +65,9 @@ public string DisplayTitle
             get { return _succeeded; }
         }
 
+        // Summary heartbeat from fast logger
```
mathewc: nit: same here - better description

src/Dashboard/Data/RecentInvocationEntry.cs
```diff
@@ -16,6 +16,7 @@ public class RecentInvocationEntry
         private const string StartTimeKey = "StartTime";
         private const string EndTimeKey = "EndTime";
         private const string SucceededKey = "Succeeded";
+        private const string HeartbeatExpritedTimeKey = "Heartbeat2Name";
```
mathewc: spelling "Exprited"

src/Dashboard/FastTable/ExtensionHelpers.cs
```diff
@@ -17,6 +17,7 @@ public static FunctionInstanceSnapshot ConvertToSnapshot(this IRecentFunctionEnt
                 FunctionShortName = entity.DisplayName,
                 DisplayTitle = entity.DisplayName, // skips Argument check
                 StartTime = entity.StartTime,
+                HeartbeatExpiredTime = entity.HeartbeatExpireTime,
```
mathewc: nit: naming inconsistency between ExpiredTime and ExpireTime?

src/Dashboard/ViewModels/InvocationLogViewModel.cs
```diff
@@ -28,14 +28,9 @@ internal InvocationLogViewModel(FunctionInstanceSnapshot snapshot, bool? heartbe
...
-            if (heartbeatIsValid.HasValue)
-            {
-                Status = snapshot.GetStatusWithHeartbeat(heartbeatIsValid.Value);
-            }
-            else
-            {
-                Status = snapshot.GetStatusWithoutHeartbeat();
-            }
+
+            Status = snapshot.GetStatusWithHeartbeat(heartbeatIsValid);
```
mathewc: Will we need a new Dashboard release for these changes?

src/Microsoft.Azure.WebJobs.Logging/Entities/InstanceTableEntity.cs
```diff
@@ -90,9 +89,14 @@ internal static string RowKeyFunctionInstanceId(Guid functionInstanceId)
 
         public string TriggerReason { get; set; }
 
-        public DateTimeOffset StartTime { get; set; }
+        public DateTime StartTime { get; set; }
```
mathewc: Are these breaking format changes in the serialization format to table storage? I.e. existing Dashboard can read both old an new?

src/Dashboard/ApiControllers/FunctionsController.cs
```diff
@@ -626,6 +627,12 @@ public IHttpActionResult Abort(string instanceQueueName)
...
+                return snapshot.HeartbeatExpiredTime.Value > now;
```
mathewc: Don't we want the other way around? I.e. on Refresh we add 50 minutes to UtcNow, so we're expired only when now catches up to that value?

src/Microsoft.Azure.WebJobs.Logging/Entities/InstanceTableEntity.cs
MikeStall: I verified this is not a breaking change. I did a test to write in one format, read in the other, and vice versa.  Specifically, Azure Tables only supports EDM.DateTime, and the DateTimeOffset is just client-side sugar. They both write the same underlying value to the table and the server doesn't know the difference. We should be using UTC DateTime everywhere rather than DateTimeOffset.

src/Dashboard/ApiControllers/FunctionsController.cs
MikeStall: I can try - the challenge is: 
a) they're 2 different heartbeat models (this has the heartbeat on the function; the other has the heartbeat on the host, and the function points to the host, so it's an extra IO lookup) 
b) I had already undergone significant pain trying to play well with existing infrastructure and just ran out of steam ;)

src/Dashboard/Data/RecentInvocationEntry.cs
MikeStall: I don't think this is written to storage.  All this metadata stuff is legacy that looks like pure in-memory marshalling - not sure what value it's actually adding. But I tried to cooperate with it to keep this change scoped.

mathewc: Perhaps rename to just FunctionInstanceHeartbeatExpiry (removing the 2 and following the naming convention of the others)

src/Dashboard/Data/RecentInvocationEntry.cs
```diff
@@ -219,6 +236,22 @@ private static string GetMetadataString(IDictionary<string, string> metadata, st
             return metadata[key];
         }
 
+        private static DateTime? GetMetadataNullableDateTime(IDictionary<string, string> metadata, string key)
+        {
+            Debug.Assert(metadata != null);
+
+            if (!metadata.ContainsKey(key))
```
mathewc: nit: metadata.TryGetValue?

src/Dashboard/Data/RecentInvocationEntry.cs
```diff
@@ -219,6 +236,22 @@ private static string GetMetadataString(IDictionary<string, string> metadata, st
...
+            {
+                return null;
+            }
+
+            string unparsed = metadata[key];
+            DateTime parsed = DateTime.Parse(unparsed, CultureInfo.InvariantCulture);
```
mathewc: Should you be defensive and use DateTime.TryParse?

src/Dashboard/FastTable/ExtensionHelpers.cs
mathewc: I'd just rename all these to FunctionInstanceHeartbeatExpiry

src/Microsoft.Azure.WebJobs.Logging/Entities/RecentPerFuncEntity.cs
```diff
@@ -108,20 +114,23 @@ public string FunctionInstanceId
...
+                return _hasError || (this.RawStatus == "CompletedFailed");
+            }
+            set
+            {
+                _hasError = value;
             }
         }
+
+        /// <summary>
+        ///  Legacy for errors. We only read from this to check if it's "CompletedFailed" so we pick up legacy errors.
```
mathewc: This legacy you're referring to is because the enum you're now using has name CompletedFailure instead of CompletedFailed?

src/Microsoft.Azure.WebJobs.Logging/Entities/InstanceTableEntity.cs
```diff
@@ -90,9 +89,14 @@ internal static string RowKeyFunctionInstanceId(Guid functionInstanceId)
...
 
-        public DateTimeOffset? EndTime { get; set; }
+        public DateTime? EndTime { get; set; }
+
+        // Last heart beat. When EndTime is missing, this lets us guess if the function is still alive. 
+        // Whereas Start and End are written very determinsitically in the lifecycle, 
+        // heartbeat is special because it's written at potentially random times by a background poller thread.
```
mathewc: Where does this poller thread actually live? Where are these updates being emitted?

src/Dashboard/ApiControllers/FunctionsController.cs
MikeStall: I have a preference for this style. (splitting out side-effecting or non-idempotent operations into their own locals) 

---
In reply to: [90366513](https://github.com/Azure/azure-webjobs-sdk/pull/928#discussion_r90366513) [](ancestors = 90366513)

src/Dashboard/ApiControllers/FunctionsController.cs
MikeStall: I'll do the rename & comment.  This is not persisted data, so we're not painting ourselves into a corner. 

---
In reply to: [90485175](https://github.com/Azure/azure-webjobs-sdk/pull/928#discussion_r90485175) [](ancestors = 90485175)

src/Microsoft.Azure.WebJobs.Logging/Entities/InstanceTableEntity.cs
MikeStall: This is the existing poller thread used to track some aggregate stats.  It's a task started when we emit our first start event; and it's stopped when we do the final flush. 

---
In reply to: [90490676](https://github.com/Azure/azure-webjobs-sdk/pull/928#discussion_r90490676) [](ancestors = 90490676)

src/Microsoft.Azure.WebJobs.Logging/Entities/RecentPerFuncEntity.cs
MikeStall: I'll clarify the comment. 
        ///  Original ("legacy") log writers would write out the function status. 
        ///  Now, we write separate bits (start time, end time, HasError) and infer the status from that. 
        ///  However, for the legacy case, it didn't have HasError and so we need to detect an error if the status field was set to "CompletedFailed".
        ///  New cases don't even write this status field anymore. 

---
In reply to: [90489837](https://github.com/Azure/azure-webjobs-sdk/pull/928#discussion_r90489837) [](ancestors = 90489837)

src/Microsoft.Azure.WebJobs.Logging/Entities/InstanceTableEntity.cs
MikeStall: Also note this is only changing the type (DateTimeOffset --> DateTime) , which have the same server representation.  It doesn't change the field name. 

---
In reply to: [90368085](https://github.com/Azure/azure-webjobs-sdk/pull/928#discussion_r90368085) [](ancestors = 90368085)

src/Microsoft.Azure.WebJobs.Logging/FunctionInstanceLogItem.cs
```diff
@@ -46,17 +46,38 @@ public class FunctionInstanceLogItem : IFunctionInstanceBaseEntry
...
+        public DateTime? HeartbeatExpireTime { get; set; }
+
         /// <summary>
-        /// Current status of this instance. 
+        /// Given log item a chance to refresh. 
         /// </summary>
-        public FunctionInstanceStatus Status { get; set; }
+        public virtual void Refresh()
+        {
+            this.HeartbeatExpireTime = DateTime.UtcNow.AddMinutes(50);
```
MikeStall: will change to pass this  in. 

---
In reply to: [90368373](https://github.com/Azure/azure-webjobs-sdk/pull/928#discussion_r90368373) [](ancestors = 90368373)

src/Dashboard/Data/RecentInvocationEntry.cs
```diff
@@ -64,11 +65,23 @@ public string DisplayTitle
...
+        // Summary heartbeat from fast logger
+        // This heartbeat is for the function-instance (not for the host).  
+        public DateTime? FunctionInstanceHeartbeatExpiry { get; set; }
+
         public HeartbeatDescriptor Heartbeat
         {
             get { return _heartbeat; }
         }
 
+        public static string FunctionInstanceHeartbeatExpiredTimeKey1
```
mathewc: nit: remove the extraneous "1" from the end of this property name? Actually you could just remove the entire property and make the field internal if all you're trying to do is make the value accessible.

src/Microsoft.Azure.WebJobs.Logging/Internal/LogReader.cs
```diff
@@ -254,6 +254,16 @@ DateTime endTime
                     var segment = await SafeExecuteQuerySegmentedAsync(rangeQuery, startTime, endTime, continuationToken);
 
                     list.AddRange(segment.Results);
+
+                    if (rangeQuery.TakeCount.HasValue)
```
mathewc: I already merged this change in from you before, right?

mathewc:
I left some minor comments, otherwise LGTM. :ship:

src/Microsoft.Azure.WebJobs.Logging/IFunctionInstanceBaseEntry.cs
```diff
@@ -22,8 +23,81 @@ public interface IFunctionInstanceBaseEntry
...
+        /// <summary>
         /// Current status of this instance. 
+        /// We specifically infer the status from the other fields rather than explicitly include status 
+        /// so that we don't duplicate state. 
         /// </summary>
-        FunctionInstanceStatus Status { get; }
-    } 
+        public static FunctionInstanceStatus GetStatus(this IFunctionInstanceBaseEntry item)
+        {
+            if (item.IsCompleted())
```
MikeStall: Yes. See LoggerTests.cs StatusTest. 

---
In reply to: [91567469](https://github.com/Azure/azure-webjobs-sdk/pull/928#discussion_r91567469) [](ancestors = 91567469)

src/Microsoft.Azure.WebJobs.Logging/Entities/InstanceTableEntity.cs
```diff
@@ -112,17 +116,22 @@ internal static string RowKeyFunctionInstanceId(Guid functionInstanceId)
...
 
-        public string RawStatus { get; set; }
         public Guid? ParentId { get; private set; }
 
-        private FunctionInstanceStatus GetStatus()
+        Guid IFunctionInstanceBaseEntry.FunctionInstanceId
+        {
+            get
+            {
+                return Guid.Parse(this.RowKey);
```
MikeStall: Yes. We're doing the same check on line 53. 

---
In reply to: [91566148](https://github.com/Azure/azure-webjobs-sdk/pull/928#discussion_r91566148) [](ancestors = 91566148)

mathewc:
I believe this has been merged? Closing.

MikeStall:
This fixes https://github.com/Azure/azure-webjobs-sdk/issues/916  . It has been merged.

Issue #844, opened 2016-09-23
Allow default Queue visibility timeout parameters to be customized


mathewc:
Currently in our default QueueProcessor we hard code the visibility timeout for failed queue trigger invocations to `TimeSpan.Zero` (code [here](https://github.com/Azure/azure-webjobs-sdk/blob/master/src/Microsoft.Azure.WebJobs.Host/Queues/QueueProcessor.cs#L105)). While you can write and register your own custom queue processor to override this, we should provide a `JobHostQueuesConfiguration` knob at the host level that lets this default be set globally.

A configuration knob is particularly IMPORTANT in scenarios like **Azure Functions** where the possibility for users to write a custom QueueProcessor doesn't exist.

mathewc:
Another question arose [here](https://stackoverflow.com/questions/40006372/azure-function-and-storage-queue-what-to-do-if-function-fails) asking for the ability to customize the invisibility timeout used when messages are dequeued.

lindydonna:
Moving to Next since the request has come up from customers.

kayomarz:
Thanks for adding this... its a nice feature!

I tried using the visibility timeout, setting it to 7 minutes but the queue message seems to become visible after a few seconds.

my host.json is:

    {
      "id": "6c8614e4e7a14c798439271d00bb0d23",
      "queues": {
        "visibilityTimeout": "00:07:00"
      }
    }

Does this seem ok? Does `"visibilityTimeout": "00:07:00"` correspond to 7 minutes?

This is with [azure-webjobs-sdk-script v1.0.10690](https://github.com/Azure/azure-webjobs-sdk-script/releases/tag/v1.0.10690)

mathewc:
Yes, that is the correct timespan format. I just tried this and it works for me as expected. Not that this setting controls the retry visibility for messages that fail processing. I.e. if your function fails, it will retry the message again after 7 minutes. You're not seeing that?

kayomarz:
Thanks for the quick reply.

An App service is scaled out to two instances. Adding 1 single Item in
queue correctly triggers a queueTrigger(ed) job. However, In under a
minute, the second instance of  the queueTrigger also seems to get
triggered with the same queue message as the work item, even though the
first job is still running (the job runs for 5 mins)

Not sure if I'm doing something else wrong.

Now that I know that the visibility timeout is defined correctly, I'll
check if something else might be going wrong.

Thanks!

dnlmcv:
Hi,

[...]

PR #945, opened 2016-12-16
Making VisibilityTimeout configurable (#844)

mathewc:
Addresses https://github.com/Azure/azure-webjobs-sdk/issues/844

I'll be exposing this in Functions as well.

### DIFF
--- a/sample/SampleHost/Program.cs
+++ b/sample/SampleHost/Program.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using System;
 using Microsoft.Azure.WebJobs;
 
 namespace SampleHost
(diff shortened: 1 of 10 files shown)


sample/SampleHost/Program.cs
```diff
@@ -10,6 +11,8 @@ class Program
         static void Main(string[] args)
         {
             var config = new JobHostConfiguration();
+            config.Queues.VisibilityTimeout = TimeSpan.FromSeconds(15);
```
mathewc: For my local testing. With these changes, when the sample function throws an exception, there is a 15 second delay before the next retry now.

src/Microsoft.Azure.WebJobs.Host/Queues/Listeners/QueueListener.cs
```diff
@@ -73,6 +74,10 @@ internal sealed class QueueListener : IListener, ITaskSeriesCommand, INotificati
             _trace = trace;
             _queueConfiguration = queueConfiguration;
 
+            // if the function runs longer than this, the invisibility will be updated
+            // on a timer periodically for the duration of the function execution
+            _visibilityTimeout = TimeSpan.FromMinutes(10);
```
mathewc: Note that this existing internal value is still not configurable, and I don't think it needs to be.

mathewc:
I'll add a test

markanewman:
I see the sample for WebJobs above, but I don't see one for Functions. Can we get a link to that PR here too?

PR #946, opened 2016-12-16
Handling the case when the storage account is not configured

soninaren:


mamaso:
@mathewc I think we had a chat about this, wasn't there a case for HttpTrigger only function accounts which use no storage? Would this break that scenario?

soninaren:
This change wouldn't affect HttpTrigger only functions. It would only affect a storage account function.

src/Microsoft.Azure.WebJobs.Host/Executors/StorageAccountProviderExtensions.cs
```diff
@@ -54,7 +56,25 @@ public static Task<IStorageAccount> GetStorageAccountAsync(this IStorageAccountP
                 }
             }
 
-            return provider.GetAccountAsync(connectionStringName, cancellationToken);
+            IStorageAccount account = await provider.GetAccountAsync(connectionStringName, cancellationToken);
+            ValidateStorageAccount(account, connectionStringName);
+            return account;
+        }
+
+        private static void ValidateStorageAccount(IStorageAccount account, string connectionStringName)
```
mamaso: does it make sense to bump this up into GetAccountAsync, where most storage account validation occurs?

soninaren: GetAccountAsync also handles dashboard storage account. To avoid that will have to introduce additional logic which is already separated by https://github.com/Azure/azure-webjobs-sdk/pull/946/files#diff-4bab545de76b34702affefd1e5180ae9L15

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Queues/QueueTests.cs
```diff
@@ -187,9 +187,14 @@ public void Func([Queue(QueueName)] out string x)
...
             var host = TestHelpers.NewJobHost<ProgramSimple>();  // no storage account!
-
+            string message = "Microsoft Azure WebJobs SDK Storage connection string is missing or empty. " +
+                       "The Microsoft Azure Storage account connection string can be set in the following ways:" + Environment.NewLine +
+                       "1. Set the connection string named 'AzureWebJobsStorage' in the connectionStrings section of the .config file in the following format " +
+                       "<add name=\"AzureWebJobsStorage\" connectionString=\"DefaultEndpointsProtocol=http|https;AccountName=NAME;AccountKey=KEY\" />, or" + Environment.NewLine +
+                       "2. Set the environment variable named 'AzureWebJobsStorage', or" + Environment.NewLine +
+                       "3. Set corresponding property of JobHostConfiguration.";
             TestHelpers.AssertIndexingError(() => host.Call("Func"),
-                "ProgramSimple.Func", "Unable to bind Queue because no storage account has been configured.");
```
mathewc: So there is now some dead code in QueueBindingProvider.ToParameterDescriptorForCollector. Recommend removing that since the account there can't be null anymore.

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Queues/QueueTests.cs
```diff
@@ -187,9 +187,14 @@ public void Func([Queue(QueueName)] out string x)
         public void Fails_When_No_Storage_is_set()
         {
             var host = TestHelpers.NewJobHost<ProgramSimple>();  // no storage account!
-
+            string message = "Microsoft Azure WebJobs SDK Storage connection string is missing or empty. " +
```
mathewc: Consider just calling the FormatParseAccountErrorMessage method to get this string, rather than duplicate it here.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Bindings/Path/BindingParameterResolverTests.cs
```diff
@@ -75,10 +75,10 @@ public void DateTimeResolver_ReturnsExpectedValue()
             Assert.NotNull(DateTime.Parse(resolvedValue));
 
             resolvedValue = resolver.Resolve("datetime:MM/yyyy");
-            Assert.Equal(DateTime.UtcNow.ToString("MM/yyyy"), resolvedValue);
+            Assert.Equal(DateTime.Now.ToString("MM/yyyy"), resolvedValue);
```
mathewc: Please revert these changes?

src/Microsoft.Azure.WebJobs.Host/Executors/StorageAccountProviderExtensions.cs
mathewc: Can we push into GetAccountAsync as @mamaso suggests, adding a new `bool throwIfMissing = true` parameter to DefaultStorageAccountProvider.GetAccountAsync that we can set to `false` when calling for the Dashboard?

mathewc:
Looks good - just one minor comment in the test file, otherwise :ship:

PR #948, opened 2016-12-17
Fixing BindingDataProvider to enable binding to less derived types

fabiocav:
This fix enables scenarios where the value provided is a type derived from the type used for binding. A concrete example is when binding to WebHooks (e.g. Slack) that read the payload as form data, which will result in a value of an internal ASP.NET type (`HttpValueCollection`), derived from NameValueCollection, which is the type intended to be used in user code.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingDataProvider.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingDataProvider.cs
@@ -48,7 +48,7 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
         /// <inheritdoc/>
         public IReadOnlyDictionary<string, object> GetBindingData(object value)
         {
-            if (value != null && value.GetType() != _type)
+            if (value != null && !_type.IsAssignableFrom(value.GetType()))
             {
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.WebJobs.Host/Bindings/BindingDataProvider.cs
```diff
@@ -48,7 +48,7 @@ internal BindingDataProvider(string template, bool ignoreCase = false)
         /// <inheritdoc/>
         public IReadOnlyDictionary<string, object> GetBindingData(object value)
         {
-            if (value != null && value.GetType() != _type)
+            if (value != null && !_type.IsAssignableFrom(value.GetType()))
```
fabiocav: For that specific scenario, yes, this would address the issue. The logic in the runtime should probably be improved so this type is properly handled and the `BindingDataProvider` isn't used, since as you point out, there won't be anything interesting to pull into binding parameters (unless we handle collections like that so that we actually populate the data using its items), but running into this surfaced the issue and I thought making this change would make sense.

Is there a reason why we'd want to enforce exact type matches as opposed tho what is proposed with the change?

Issue #951, opened 2016-12-21
BlobTrigger based webjobs failing due entries in logs


PrzemekGawron:
We experience a serious production issue that makes all webjobs that are using BlobTrigger unusable.
They continuously restarting each time throwing following error:

```
[12/21/2016 10:11:26 > 5053fb: ERR ] Unhandled Exception: System.FormatException: Failed to parse RequestedObjectKey property of the log entry. It should be in one of the supported formats: "https://account.blob.core.windows.net/container/blob", or"/account/container/blob"
[12/21/2016 10:11:26 > 5053fb: ERR ]    at Microsoft.Azure.WebJobs.Host.Blobs.Listeners.StorageAnalyticsLogEntry.ToBlobPath()
[12/21/2016 10:11:26 > 5053fb: ERR ]    at System.Linq.Enumerable.WhereSelectListIterator`2.MoveNext()
[12/21/2016 10:11:26 > 5053fb: ERR ]    at System.Linq.Enumerable.WhereEnumerableIterator`1.MoveNext()
[12/21/2016 10:11:26 > 5053fb: ERR ]    at Microsoft.Azure.WebJobs.Host.Blobs.Listeners.BlobLogListener.<GetRecentBlobWritesAsync>d__9.MoveNext()
[12/21/2016 10:11:26 > 5053fb: ERR ] --- End of stack trace from previous location where exception was thrown ---
[12/21/2016 10:11:27 > 5053fb: ERR ]    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[12/21/2016 10:11:27 > 5053fb: ERR ]    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[12/21/2016 10:11:27 > 5053fb: ERR ]    at Microsoft.Azure.WebJobs.Host.Blobs.Listeners.PollLogsStrategy.<ExecuteAsync>d__6.MoveNext()
[12/21/2016 10:11:28 > 5053fb: ERR ] --- End of stack trace from previous location where exception was thrown ---
[12/21/2016 10:11:28 > 5053fb: ERR ]    at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
[12/21/2016 10:11:28 > 5053fb: ERR ]    at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #952, opened 2016-12-21
Return null (rather than throw) if blob log cannot be parsed

brettsam:
Addresses #951

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/StorageAnalyticsLogEntry.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/StorageAnalyticsLogEntry.cs
@@ -58,7 +58,7 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
                     ServiceType.HasValue &&
                     ServiceType == StorageServiceType.Blob &&
                     OperationType.HasValue &&
-                    ((OperationType == StorageServiceOperationType.ClearPage) || 
+                    ((OperationType == StorageServiceOperationType.ClearPage) ||
                     (OperationType == StorageServiceOperationType.CopyBlob) ||
                     (OperationType == StorageServiceOperationType.CopyBlobDestination) ||
                     (OperationType == StorageServiceOperationType.PutBlob) ||
@@ -87,7 +87,7 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
             var entry = new StorageAnalyticsLogEntry();
 
             DateTime requestStartTime;
-            if (!DateTime.TryParseExact(fields[(int)StorageAnalyticsLogColumnId.RequestStartTime], "o", 
+            if (!DateTime.TryParseExact(fields[(int)StorageAnalyticsLogColumnId.RequestStartTime], "o",
                 CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind, out requestStartTime))
             {
                 return null;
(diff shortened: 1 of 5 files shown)


PrzemekGawron:
@brettsam Can you roughly estimate when this will be available to end-users?

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/StorageAnalyticsLogEntry.cs
```diff
@@ -139,23 +137,19 @@ public BlobPath ToBlobPath()
...
                 path = startPos != -1 ? RequestedObjectKey.Substring(startPos + 1) : String.Empty;
             }
 
             if (String.IsNullOrEmpty(path))
             {
-                throw new FormatException("Failed to parse RequestedObjectKey property of the log entry. " +
-                    "It should be in one of the supported formats: " +
-                    @"""https://account.blob.core.windows.net/container/blob"", or" +
-                    @"""/account/container/blob""");
+                return null;
```
mathewc: I don't think we can simply return null here - there are upstream callers of this method that assume the returned value will not be null (e.g. BlobLogListener).

mathewc: I think we should update the parse code to take into account StorageAnalyticsLogColumnId.RequestStatus. In the repro case, this will be a ThrottlingError. We could then change consumer code so it only processes sucessfull requests. I.e. change BlobLogListener to include entry.IsSuccess in its where clause. Thoughts?

brettsam: It looks like `BlobLogListener` handles null already (I mentioned this in the issue as well) -- https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/BlobLogListener.cs#L67-L75

Although I don't see any tests for that scenario -- I'll look into that now.

I was going to handle ThrottlingError directly, but then saw that there are 20+ other statuses (https://docs.microsoft.com/en-us/rest/api/storageservices/fileservices/storage-analytics-logged-operations-and-status-messages). If another customer finds that one of these also doesn't contain a valid path, we'll be right back in this code patching for that one. It seemed smarter to ignore any log entries that do not contain a valid `BlobPath` as we can't operate on them anyway.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Blobs/Listeners/StorageAnalyticsLogEntryTests.cs
```diff
@@ -111,8 +109,8 @@ public void ToBlobPath_IfMalformedBlobPath_ThrowsFormatException()
             string requestedObjectKey = "/account/malformed-blob-path";
             StorageAnalyticsLogEntry entry = CreateEntry("2014-09-08T18:44:18.9681025Z", StorageServiceOperationType.PutBlob, StorageServiceType.Blob, requestedObjectKey);
 
-            ExceptionAssert.ThrowsFormat(() => entry.ToBlobPath(), "Failed to parse RequestedObjectKey property of the log entry. " +
-                    "Blob identifiers must be in the format container/blob.");
+            BlobPath blob = entry.ToBlobPath();
```
mathewc: nit: update test name since fex is no longer thrown

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/StorageAnalyticsLogEntry.cs
mathewc: Fair enough :)

PR #954, opened 2017-01-03
Change AutoResolve error message

mathewc:
Make it less likely that users will accidentally cause their secret values to be written to error logs.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Globalization;
 using System.Reflection;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host.Bindings.Path;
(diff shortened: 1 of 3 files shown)


src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -134,42 +133,54 @@ private bool TryCreateAutoResolveBindingTemplate(PropertyInfo propInfo, INameRes
         {
             template = null;
 
-            AutoResolveAttribute attr = propInfo.GetCustomAttribute<AutoResolveAttribute>();
+            string resolvedValue = null;
+            if (!TryAutoResolveValue(_source, propInfo, nameResolver, out resolvedValue))
```
mathewc: All the changes here are me making the code more unit testable - pushing logic out into TryAutoResolveValue static

src/src.ruleset
```diff
@@ -8,6 +8,7 @@
     <Rule Id="CA1026" Action="None" />
     <Rule Id="CA1031" Action="None" />
     <Rule Id="CA1033" Action="None" />
+    <Rule Id="CA1305" Action="None" />
```
mathewc: This is a compiler warning for using CultureInfo in string.Format calls. The string interpolation I added above was causing an error. We don't need these errors.

Issue #804, opened 2016-08-30
Poison blob queue message in multi storage account scenario


Fabian-Schmidt:
Hi,
I have setup two storage accounts in the connectionStrings. First one is setup as `AzureWebJobsStorage` and a second as `AzureWebJobsMyToken`.

In the second account a blob trigger is configured via an attribute on the class `[StorageAccount("MyToken")]`.
In case this blob trigger fails multiple times a poison blob message is generated in the queue `webjobs-blobtrigger-poison` of the first storage account.
My expectation was that the same storage account, as used for the trigger, is used for the poison message.

Thanks,
Fabian

mathewc:
Thanks for reporting. What version of the SDK are you using?

Fabian-Schmidt:
Microsoft.Azure.WebJobs Version 1.1.2.0
Microsoft.Azure.WebJobs.Extensions Version 1.0.1.0
Microsoft.Azure.WebJobs.Host Version 1.1.2.0

mamaso:
@christopheranderson @mathewc Unfortunately this scenario and #797 are basically incompatible (if we add the poison queue to the secondary storage account it cannot be a blob-only storage account).

Which issue do you guys want to prioritize?

lindydonna:
@mamaso We were thinking of excluding blob only storage accounts for RC, since it requires perf testing as well. Does that simplify the decision?

mathewc:
@mamaso My thought was that the control queue we use for poison blobs would live in the primary storage account, since it is an internal implementation detail of blob triggers. That's the same place the trigger queue we create lives.

mamaso:
Yep, I think it's fine for the primary storage account to have all the queues.  Perhaps the answer is to document the poison queue behavior better since this didn't match the user's expectation of where it would be created.

mamaso:
Closing this. the current behavior is by design.

mathewc:
What is our behavior for poison queues for QueueTrigger functions in the context of secondary storage accounts? We want to ensure that we have a consistent story.

mamaso:
Just tested, QueueTrigger will create the poison queue in the secondary account, so there's some inconsistency here.

mathewc:
Yeah, that's what I thought. We'll need to drive towards a consistent model for all of these.

mathewc:
[...]

PR #956, opened 2017-01-05
BlobTrigger poison queue fix (#804)

mathewc:
Fix for https://github.com/Azure/azure-webjobs-sdk/issues/804

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/BlobListenerFactory.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/BlobListenerFactory.cs
@@ -136,9 +136,16 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
             IStorageQueueClient queueClient = _hostAccount.CreateQueueClient();
             IStorageBlobClient blobClient = _hostAccount.CreateBlobClient();
 
+            // Important: We're using the "data account" here, which is the account that the
+            // function the listener is for is targeting. This is the account that will be used
+            // to read the trigger blob.
+            IStorageBlobClient userBlobClient = _dataAccount.CreateBlobClient();
+            IStorageQueueClient userQueueClient = _dataAccount.CreateQueueClient();
+
             string hostId = await _hostIdProvider.GetHostIdAsync(cancellationToken);
             string hostBlobTriggerQueueName = HostQueueNames.GetHostBlobTriggerQueueName(hostId);
             IStorageQueue hostBlobTriggerQueue = queueClient.GetQueueReference(hostBlobTriggerQueueName);
+            IStorageQueue poisonQueue = userQueueClient.GetQueueReference(HostQueueNames.BlobTriggerPoisonQueue);
 
             SharedQueueWatcher sharedQueueWatcher = _sharedContextProvider.GetOrCreateInstance<SharedQueueWatcher>(
                 new SharedQueueWatcherFactory(_messageEnqueuedWatcherSetter));
@@ -153,14 +160,9 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
             // Create a "bridge" listener that will monitor the blob
             // notification queue and dispatch to the target job function.
             SharedBlobQueueListener sharedBlobQueueListener = _sharedContextProvider.GetOrCreateInstance<SharedBlobQueueListener>(
-                new SharedBlobQueueListenerFactory(sharedQueueWatcher, queueClient, hostBlobTriggerQueue,
+                new SharedBlobQueueListenerFactory(sharedQueueWatcher, hostBlobTriggerQueue, poisonQueue,
                     _queueConfiguration, _exceptionHandler, _trace, sharedBlobListener.BlobWritterWatcher));
             var queueListener = new BlobListener(sharedBlobQueueListener);
-
-            // Important: We're using the "data account" here, which is the account that the
-            // function the listener is for is targeting. This is the account that will be used
-            // to read the trigger blob.
-            IStorageBlobClient userBlobClient = _dataAccount.CreateBlobClient();
 
             // Register our function with the shared queue listener
             RegisterWithSharedBlobQueueListenerAsync(sharedBlobQueueListener, userBlobClient);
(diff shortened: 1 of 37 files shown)


src/Microsoft.Azure.WebJobs.Host/Queues/Listeners/QueueListener.cs
```diff
@@ -273,9 +274,11 @@ private TaskSeriesCommandResult CreateSucceededResult()
             }
         }
 
-        private void OnMessageAddedToPoisonQueue(object sender, EventArgs e)
+        private void OnMessageAddedToPoisonQueue(object sender, PoisonMessageEventArgs e)
         {
-            _sharedWatcher.Notify(_poisonQueue.Name);
+            // TODO: this is assuming that the poison queue is in the same
+            // storage account
+            _sharedWatcher.Notify(e.PoisonQueue.Name);
```
mathewc: I want to create a bug for this. Not an real issue, but something we should look to improve. This shared watcher notify stuff is just a way to wake up a sleeping queue listener faster when something is written to it on the same instance, so we don't have to wait for the exponential backoff interval.

src/Microsoft.Azure.WebJobs.Host/Queues/PoisonQueueMessageEventArgs.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            Message = message;
+            PoisonQueue = poisonQueue;
+        }
+
+        /// <summary>
+        /// The poison message
+        /// </summary>
+        public CloudQueueMessage Message { get; private set; }
+
```
MikeStall: What does this mean in the Blob case?

src/Microsoft.Azure.WebJobs.Host/Queues/QueueProcessor.cs
```diff
@@ -211,13 +213,9 @@ public QueueProcessor(QueueProcessorFactoryContext context)
         /// Called to raise the MessageAddedToPoisonQueue event
         /// </summary>
         /// <param name="e">The event arguments</param>
-        protected internal virtual void OnMessageAddedToPoisonQueue(EventArgs e)
+        protected internal virtual void OnMessageAddedToPoisonQueue(PoisonMessageEventArgs e)
```
MikeStall: >OnMessageAddedToPoisonQueue [](start = 40, length = 27)

Why is this exposed?  
If somebody wants to react to the poison message, can they just use a [QueueTrigger] on the poison queue?

src/Microsoft.Azure.WebJobs.Host/Queues/PoisonQueueMessageEventArgs.cs
mathewc: Same thing. Remember that blob trigger works by enqueuing messages, which we set up an internal "shared listener" on.

src/Microsoft.Azure.WebJobs.Host/Queues/QueueProcessor.cs
mathewc: Yes, customers set up queue trigger functions to process the poison queue. QueueProcessor is a lower level advanced thing, and if you see how it is used, you'll see that other aspects of our infra need to know when a poison message is added.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/BlobListenerFactory.cs
```diff
@@ -153,17 +159,23 @@ internal class BlobListenerFactory : IListenerFactory
...
 
-            // Important: We're using the "data account" here, which is the account that the
-            // function the listener is for is targeting. This is the account that will be used
-            // to read the trigger blob.
-            IStorageBlobClient userBlobClient = _dataAccount.CreateBlobClient();
+            // determine which client to use for the poison queue
+            // by default this should target the same storage account
+            // as the blob container we're monitoring
+            var poisonQueueClient = userQueueClient;
+            if (_dataAccount.Type == StorageAccountType.BlobOnly)
```
mathewc: Good catch - changed

src/Microsoft.Azure.WebJobs.Host/JobHostBlobsConfiguration.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        /// <summary>
+        /// Gets or sets a value indicating whether a single centralized
+        /// poison queue for poison blobs should be used (in the primary
+        /// storage account) or whether the poison queue for a blob triggered
+        /// function should be co-located with the target blob container.
+        /// This comes into play only when using multiple storage accounts via
+        /// <see cref="StorageAccountAttribute"/>. The default is false.
+        /// </summary>
+        public bool CentralizePoisonQueue { get; private set; }
```
mathewc: Open to naming ideas on this

src/Microsoft.Azure.WebJobs.Host/JobHostBlobsConfiguration.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// Represents configuration for <see cref="BlobTriggerAttribute"/>.
+    /// </summary>
+    public class JobHostBlobsConfiguration
```
mathewc: Likely there will be additional blob scan knobs we want to expose in the future as well. I do remember this coming up before but we didn't have a way previously.

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/BlobTriggerTests.cs
```diff
@@ -51,31 +49,6 @@ public static void Run([BlobTrigger(BlobPath)] ICloudBlob blob)
             }
         }
 
-        [Fact]
-        public void BlobTrigger_IfBindingAlwaysFails_MovesToPoisonQueue()
```
mathewc: Removing this test because we already have test coverage for poison scenarios, and the test no longer runs because it relies on mocking of the internal QueueProcessor which isn't possible anymore because I now have a new SharedBlobQueueProcessor in place. I fiddled too long with getting the test to keep running, and it isn't worth it because we have coverage already.

mathewc:
Merged in https://github.com/Azure/azure-webjobs-sdk/commit/56b5302dda3ff6d671fb7e4c1b9ddcce0594a9a2

PR #961, opened 2017-01-06
updating SingletonManager to fetch storage account via StorageAccountProviderExtensions

soninaren:


src/Microsoft.Azure.WebJobs.Host/Executors/StorageAccountProviderExtensions.cs
```diff
@@ -22,14 +22,19 @@ public static Task<IStorageAccount> GetDashboardAccountAsync(this IStorageAccoun
             return provider.GetAccountAsync(ConnectionStringNames.Dashboard, cancellationToken);
         }
 
-        public static async Task<IStorageAccount> GetStorageAccountAsync(this IStorageAccountProvider provider, CancellationToken cancellationToken)
+        public static async Task<IStorageAccount> GetStorageAccountAsync(this IStorageAccountProvider provider, CancellationToken cancellationToken, string connectionStringName = null)
```
mathewc: How about we rename IStorageAccountProvider.GetAccountAsync to GetAccountOrNullAsync since it can return null. Then we won't have name collisions with our extension methods that validate and throw. You can then leave this method signature as is, and add a new extension method GetAccountAsync that takes the connection name and validates/throws. You'll then have this method and the one below call GetAccountAsync to do the validation and return the account.

The one place where we allow null for Dashboard account, you'll then use the GetAccountOrNull method. All other places are doing validation.l

src/Microsoft.Azure.WebJobs.Host/Executors/IStorageAccountProvider.cs
```diff
@@ -9,6 +9,6 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
 {
     internal interface IStorageAccountProvider
     {
-        Task<IStorageAccount> GetAccountAsync(string connectionStringName, CancellationToken cancellationToken);
+        Task<IStorageAccount> TryGetAccountAsync(string connectionStringName, CancellationToken cancellationToken);
```
mathewc: Please rename this to GetAccountOrNullAsync - TryGet by convention is a method that returns a boolean value and returns its value as an out parameter.

mathewc:
Just one rename, otherwise looks good to me.

src/Microsoft.Azure.WebJobs.Host/Executors/IStorageAccountProvider.cs
soninaren: leave it as is as per our conversation :)

PR #963, opened 2017-01-07
Add Table tests 

MikeStall:
This is filling some test holes in preparation for switching table binding over to use binding rules.
No product change


mathewc:
Is this checked in?

Issue #713, opened 2016-05-06
Make IArgumentBindingProvider<T> internal again?


mathewc:
This interface was opened up to support binding extensibility (the ability to customize an existing binding). See wiki [here](https://github.com/Azure/azure-webjobs-sdk-extensions/wiki/Binding-Customization) for more information.  Specifically, we had users ask for the ability to add an IDictionary binding to our built in TableAttribute binding (see original issue [here](https://github.com/Azure/azure-webjobs-sdk/issues/517)). We used this interface to enable that.

I believe Mikes new binding stuff (ConverterManager) may support this in a more natural way. We should move the TableAttribute extensibility on top of that to verify.

MikeStall:
We can cleanly solve this with the new CoverterManager approach.  Also related to https://github.com/Azure/azure-webjobs-sdk/issues/887

Issue #887, opened 2016-10-26
Switch Table bindings over to use BindingFactory helpers 


MikeStall:
Same thing we did for [Queue]

Issue #790, opened 2016-08-13
Add a singleton output binding for Tables


mathewc:
The Table binding supports a singleton input binding allowing you to bind to a single table record. However, it does **not** support a singleton output binding. I.e. `[Table("test", "PK", "RK")] out MyItem` is not supported. For output, you currently have to use ICollector/IAsyncCollector.

This is inconsistent and confusing.

MikeStall:
This is also relate dto https://github.com/Azure/azure-webjobs-sdk/issues/887  . Making Tables use the binding rules (like Queue) would drive consistency which would naturally fix this.

mathewc:
Moving this to 2.0.0 because Mike is working on a bunch of changes now in Table binding and this will go along with those.

MikeStall:
I think table updates in place. So it's really like an an In/Out binding.

MikeStall:
This was fixed with the "make tables use IAsyncCollector rule" above, and there's a test in TableTests.cs for it.

mathewc:
Awesome, thanks :)

PR #967, opened 2017-01-10
Make Table Bindings use Rule system, delete code. 

MikeStall:
Make Table binders use new rule-based system. 
Expand Converter Manager  to allow specifying rules for "Open Types".  Aggressively use these on Table converters.  TableEntity bindings are still untouched. 

Remove a bunch of now dead table binding code. 
See TableTests.cs for summary. 
Interesting changes are in ConverterManager and TableAttributeBindingProvider

This lets us define rules like:
1.	“CloudTable --> IQueryable<T> where T: ITableEntity”   (this is effectively also solving #713)
2.	T --> ITableEntity where T has a PartitionKey and RowKey property 

Fixes:
https://github.com/Azure/azure-webjobs-sdk/issues/713  - extensibility 
https://github.com/Azure/azure-webjobs-sdk/issues/790  - this happens naturally by making Table's AsyncCollector use the standard rules. 

Mostly solves https://github.com/Azure/azure-webjobs-sdk/issues/887


src/Microsoft.Azure.WebJobs.Host/Tables/TableAttributeBindingProvider.cs
```diff
@@ -68,32 +62,156 @@ private TableAttributeBindingProvider(INameResolver nameResolver, IStorageAccoun
...
+
+            var bindAsyncCollector = bindingFactory.BindToAsyncCollector<TableAttribute, ITableEntity>(
+                original.BuildFromTableAttribute,
+                null,
+                original.CollectAttributeInfo);
+
+            var bindToJobject = bindingFactory.BindToExactAsyncType<TableAttribute, JObject>(
+                original.BuildJObject,
+                null,
+                original.CollectAttributeInfo);
```
brettsam: I just looked up CollectAttributeInfo to see what it does. It looks like ultimately, it is creating the client and storing it in the ResolvedTableAttribute so it can later be extracted. Our other bindings create their clients lazily and store them in ConcurrentDictionaries to be pulled out based on connection strings (so one client per conn str). Couldn't you do the same thing here and remove the concept of the 'postResolveHook' completely?

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
```diff
@@ -176,7 +176,37 @@ public IBindingProvider BindToGenericItem<TAttribute>(Func<TAttribute, Type, Tas
             Func<TAttribute, ParameterInfo, INameResolver, Task<TAttribute>> postResolveHook = null)
             where TAttribute : Attribute
         {
-            var bindingProvider = new ExactTypeBindingProvider<TAttribute, TUserType>(_nameResolver, buildFromAttribute, buildParameterDescriptor, postResolveHook);
+            var bindingProvider = new ExactTypeBindingProvider<TAttribute, TUserType>(
+                _nameResolver, 
+                buildFromAttribute,
+                null,
+                buildParameterDescriptor, 
+                postResolveHook);
```
MikeStall: defaults to null and is ignored. This is an advanced hook for legacy compat for the dashboard. 

---
In reply to: [95845742](https://github.com/Azure/azure-webjobs-sdk/pull/967#discussion_r95845742) [](ancestors = 95845742)

src/Microsoft.Azure.WebJobs.Host/Tables/TableAttributeBindingProvider.cs
MikeStall: The core issue here is that most attributes have all key information within a single attribute plus the configuration.  Storage is a bit different: it's split across multiple attributes (ie, [Table] and [StorageAccount]). This makes storage look like the other bindings by merging everything back into a single attribute. The fact that the newly resolved attribute also has a client in it is secondary. 

---
In reply to: [95850184](https://github.com/Azure/azure-webjobs-sdk/pull/967#discussion_r95850184) [](ancestors = 95850184)

brettsam: Why can't you do that in a binding rule? Is it because you don't have access to the ParameterInfo? Could that get passed into the rule as well? My concern is that we're adding a hook to our 'new' binding rules in order to maintain legacy support. And the more 'hooks' we expose, the more confused I get :-) We really need some kind of pipeline diagram to show people where they can plug in, and how.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
```diff
@@ -176,7 +176,37 @@ public IBindingProvider BindToGenericItem<TAttribute>(Func<TAttribute, Type, Tas
...
+        /// Create a binding provider that binds the parameter to a set of possible types via converter manager. 
+        /// </summary>
+        /// <typeparam name="TAttribute">type of binding attribute on the user's parameter.</typeparam>
+        /// <typeparam name="TMiddleType">An intermediate type to bind to. 
+        /// Converter Manager will then convert this intermediate type to the user's exact type in their signature.</typeparam>
+        /// <param name="buildFromAttribute">builder function to create the object that will get passed to the user function.</param>
+        /// <param name="buildParameterDescriptor">An optional function to create a specific ParameterDescriptor object for the dashboard. If missing, a default ParameterDescriptor is created. </param>
+        /// <param name="postResolveHook">an advanced hook for translating the attribute. </param>
+        /// <returns>A binding provider that applies these semantics.</returns>
+        public IBindingProvider BindToGeneralAsyncType<TAttribute, TMiddleType>(
```
mathewc: For many other methods on this type you use the terminology "BindToGenericxxx" - can you use that here as well? Otherwise the differences between BindToGeneric and BindToGeneral aren't clear

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
```diff
@@ -176,7 +176,37 @@ public IBindingProvider BindToGenericItem<TAttribute>(Func<TAttribute, Type, Tas
...
+                _nameResolver, 
+                buildFromAttribute,
+                null,
+                buildParameterDescriptor, 
+                postResolveHook);
+            return bindingProvider;
+        }
+
+        /// <summary>
+        /// Create a binding provider that binds the parameter to a set of possible types via converter manager.
```
mathewc: nit: can you use consistent capitalization in these blocks? e.g. some sentences start with capital, some don't, etc.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/ExactTypeBindingProvider.cs
```diff
@@ -8,23 +8,28 @@
 
 namespace Microsoft.Azure.WebJobs.Host.Bindings
 {
-    // BindingProvider to bind to an exact type. Useful for binding to a client object. 
-    internal class ExactTypeBindingProvider<TAttribute, TUserType> : IBindingProvider
+    // BindingProvider to bind to type. It can user Converter manager to convert from TMiddleType to the user
```
mathewc: nit: it can use

src/Microsoft.Azure.WebJobs.Host/Bindings/ConverterManager.cs
```diff
@@ -4,23 +4,85 @@
...
+            }
+
+            if (t == typeof(OpenType))
+            {
+                return (type) => true;
+            }
+            var method = t.GetMethod("IsValid", BindingFlags.Static | BindingFlags.Public | BindingFlags.NonPublic);
+            if (method == null)
+            {
+                throw new InvalidOperationException("Type " + t.FullName + "' is derived from " + typeof(OpenType).Name + " but is missing an "
```
mathewc: nit: some $"" string interpolation would make this easier to read

src/Microsoft.Azure.WebJobs.Host/Bindings/ConverterManager.cs
```diff
@@ -32,15 +94,31 @@ private static string DefaultByteArray2String(byte[] bytes)
...
-        {           
+        {
             string key = GetKey<TSrc, TDest, TAttribute>();
-            _funcsWithAttr[key] = converter;            
+            _funcsWithAttr[key] = converter;                        
         }
 
         // Return null if not found. 
-        private FuncConverter<TSrc, TAttribute, TDest> TryGetConverter<TSrc, TAttribute, TDest>()
+        private FuncConverter<TSrc, TAttribute, TDest> TryGetConverter<TSrc, TAttribute, TDest>(
```
mathewc: Generally non async TryGet methods return a bool and return their value as an output parameter.

src/Microsoft.Azure.WebJobs.Host/Bindings/ConverterManager.cs
```diff
@@ -49,13 +127,38 @@ private static string DefaultByteArray2String(byte[] bytes)
...
+                var builder = obj as Func<Type, Type, Func<object, object>>;
+                if (builder != null)
+                {
+                    Func<object, object> converter = builder(typeSrc, typeDest); // converter 
+                    FuncConverter<TSrc, TAttribute, TDest> func3 = (input, attribute, context) =>
+                    {
+                        var output = converter(input);
+                        return (TDest)output;
+                    };
+                    return func3;
```
mathewc: Can you give better names to some of these locals like func3, key1, etc. for better readability?

src/Microsoft.Azure.WebJobs.Host/Bindings/ConverterManagerExtensions.cs
```diff
@@ -0,0 +1,234 @@
...
+    /// <summary>
+    /// Extensions for <see cref="IConverterManager"/> 
+    /// </summary>
+    public static class IConverterManagerExtensions
+    {
+        /// <summary>
+        /// Add a converter for the given Source to Destination conversion.
+        /// The typeConverter type is instantiated with the type arguments and constructorArgs is passed. 
+        /// </summary>
+        /// <typeparam name="TSource"></typeparam>
```
mathewc: Can you fill out all the API help for all the new public surface area?

src/Microsoft.Azure.WebJobs.Host/Tables/TableAttributeBindingProvider.cs
```diff
@@ -68,32 +62,156 @@ private TableAttributeBindingProvider(INameResolver nameResolver, IStorageAccoun
...
+            // IStorageTable --> IQueryable<ITableEntity>
+            converterManager.AddConverterBuilder<IStorageTable, TableQueryableOpenType, TableAttribute>(
+                original.QueryableConverter);
 
             var bindingFactory = new BindingFactory(nameResolver, converterManager);
-            var bindAsyncCollector = bindingFactory.BindToAsyncCollector<TableAttribute, ITableEntity>(original.BuildFromTableAttribute, null, original.CollectAttributeInfo);
 
-            var bindToJobject = bindingFactory.BindToExactAsyncType<TableAttribute, JObject>(original.BuildJObject, null, original.CollectAttributeInfo);
-            var bindToJArray = bindingFactory.BindToExactAsyncType<TableAttribute, JArray>(original.BuildJArray, null, original.CollectAttributeInfo);
+            var bindToExactCloudTable = bindingFactory.BindToGeneralAsyncType<TableAttribute, CloudTable>(
```
MikeStall: The func callback is async. This could benefit from a better name. 

---
In reply to: [95847425](https://github.com/Azure/azure-webjobs-sdk/pull/967#discussion_r95847425) [](ancestors = 95847425)

src/Microsoft.Azure.WebJobs.Host/Tables/TableAttributeBindingProvider.cs
MikeStall: We *don't* want multiple attributes for a single binding- it's inconsistent with Functions.  So only the existing legacy cases would have separate attributes and everybody else would be on the single-attribute plan.   The binding rules push people towards the Functions plan. 

---
In reply to: [95891423](https://github.com/Azure/azure-webjobs-sdk/pull/967#discussion_r95891423) [](ancestors = 95891423)

src/Microsoft.Azure.WebJobs.Host/Bindings/ConverterManager.cs
MikeStall: what's that? 

---
In reply to: [96106526](https://github.com/Azure/azure-webjobs-sdk/pull/967#discussion_r96106526) [](ancestors = 96106526)

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
brettsam: I believe that with the move to support open types in the converters, we may be able to get rid of the 'BindToGeneric' completely -- is that the plan, @MikeStall?

MikeStall: @brettsam  - yes. I'll put an obsolete attribute on it. This should also greatly reduce the need for AddFilter, since the CM now handles filtering logic.

src/Microsoft.Azure.WebJobs.Host/Bindings/ConverterManagerExtensions.cs
```diff
@@ -0,0 +1,234 @@
...
+            {
+                if (method.Name.StartsWith("Convert", StringComparison.OrdinalIgnoreCase))
+                {
+                    if (method.ReturnType == typeDest)
+                    {
+                        var parameters = method.GetParameters();
+                        if (parameters.Length == 1 && parameters[0].ParameterType == typeSource)
+                        {
+                            // Type mismatch 
+                            // var func = (Func<object, object>) method.CreateDelegate(typeof(Func<object, object>));
```
mathewc: remove commented out code?

src/Microsoft.Azure.WebJobs.Host/Bindings/ConverterManagerExtensions.cs
```diff
@@ -0,0 +1,234 @@
...
+            converterManager.AddConverterBuilder<TSource, TDestination, TAttribute>((typeSrc, typeDest) =>
+            {
+                Func<object, object> converter = GetConverter(converterInstance, typeSrc, typeDest);
+                return converter;
+            });
+        }
+
+        // Find a Convert* method that is compatible with the given source and destiation types. 
+        // Signature will be like:
+        //    TIn Convert*(TOut)
```
mathewc: What are the covariance behavior for these converters? E.g. if I have various conversion methods for various points in the hierarchy, will the best match be used?

```csharp
public class Foo { }
public class Bar : Foo { }
public class Baz : Bar { }

public class Converters
{
    public Foo Convert(Bar b) { ... }
    public Foo Convert(Foo f) { ... }

    public Foo Convert(Baz b) { ... }
}
```

src/Microsoft.Azure.WebJobs.Host/Bindings/ConverterManagerExtensions.cs
```diff
@@ -0,0 +1,234 @@
...
+            Type typeExisting;
+            if (genericArgs.TryGetValue(name, out typeExisting))
+            {
+                return typeExisting == type;
+            }
+            genericArgs[name] = type;
+            return true;
+        }
+
+        // Return truen if the types are compatible.
```
mathewc: nit: true

test/Microsoft.Azure.WebJobs.Host.UnitTests/ConverterManagerTests.cs
```diff
@@ -197,6 +199,160 @@ public void AttributeOverloads2()
...
+
+            public TOutput Convert(TInput input)
+            {
+                var str = (string)(object)input;
+                return (TOutput)(object)int.Parse(str);
+            }
+        }
+
+        [Fact]
+        public void OpenType2()
```
mathewc: Can you use test names that make it clear what is being tested?

src/Microsoft.Azure.WebJobs.Host/Bindings/ConverterManager.cs
mathewc: You can write this same throw statement as below using the new C#6 feature:

```csharp
throw new InvalidOperationException($"Type {t.FullName} is derived from {typeof(OpenType).Name} but is missing an IsValid() method"};
```

src/Microsoft.Azure.WebJobs.Host/Bindings/ConverterManagerExtensions.cs
```diff
@@ -0,0 +1,234 @@
...
+            {
+                return typeExisting == type;
+            }
+            genericArgs[name] = type;
+            return true;
+        }
+
+        // Return truen if the types are compatible. 
+        // If openType has generic args, then add a [Name,Type] entry to the genericArgs dictionary. 
+        private static bool CheckArg(Type openType, Type specificType, Dictionary<string, Type> genericArgs)
```
mathewc: There are a bunch of tricky private static helpers in this class that I think would benefit from some good unit tests, since they are so core. I looked quickly at the added tests, and didn't see any that seemed to exercise all the permutations here.

src/Microsoft.Azure.WebJobs.Host/Bindings/ConverterManagerExtensions.cs
```diff
@@ -0,0 +1,234 @@
...
+                            Func<object, object> func = (input) =>
+                            {
+                                return method.Invoke(instance, new object[] { input });
+                            };
+                            return func;
+                        }
+                    }
+                }
+            }
+            throw new InvalidOperationException("No Convert method on type " + t.Name + " to convert from " +
```
mathewc: Another example :)

```csharp
throw new InvalidOperationException($"No Convert method on type {t.Name} to convert from {typeSource.Name} to {typeDest.Name}";
```

src/Microsoft.Azure.WebJobs.Host/Bindings/OpenType.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Host.Bindings
+{
+    /// <summary>
+    /// Placeholder to use with converter manager for describing open types. 
+    /// The type can be filtering via a IsValid() method.
```
mathewc: This summary text isn't really clear to me. I think very clear documentation for this pattern if it is convention based.

src/Microsoft.Azure.WebJobs.Host/Tables/CloudTableArgumentBindingProvider.cs
```diff
@@ -1,48 +0,0 @@
...
-            {
-                get
-                {
-                    return FileAccess.ReadWrite;
-                }
-            }
-
-            public async Task<IValueProvider> BindAsync(IStorageTable value, ValueBindingContext context)
-            {
-                await value.CreateIfNotExistsAsync(context.CancellationToken);
```
mathewc: Where does this CreateIfNotExists logic happen now?

src/Microsoft.Azure.WebJobs.Host/Tables/StringToStorageTableConverter.cs
```diff
@@ -1,38 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the MIT License. See License.txt in the project root for license information.
-
-using System;
-using Microsoft.Azure.WebJobs.Host.Converters;
-using Microsoft.Azure.WebJobs.Host.Storage.Table;
-
-namespace Microsoft.Azure.WebJobs.Host.Tables
-{
-    internal class StringToStorageTableConverter : IConverter<string, IStorageTable>
```
mathewc: I don't have a good sense for the level of tests we had in place for all of these various binding permutations that are being removed. You're sure that we're covered for all of they (they continue to work?)

src/Microsoft.Azure.WebJobs.Host/Tables/TableAttributeBindingProvider.cs
```diff
@@ -68,32 +62,106 @@ private TableAttributeBindingProvider(INameResolver nameResolver, IStorageAccoun
...
         public static IBindingProvider Build(INameResolver nameResolver, IConverterManager converterManager, IStorageAccountProvider accountProvider, IExtensionRegistry extensions)
         {
             var original = new TableAttributeBindingProvider(nameResolver, accountProvider, extensions);
 
-            converterManager.AddConverter<JObject, ITableEntity, TableAttribute>(original.JObjectToTableEntityConverterFunc);
+            converterManager.AddConverter<JObject, ITableEntity, TableAttribute>(original.JObjectToTableEntityConverterFunc);            
+            converterManager.AddConverterBuilder<object, ITableEntity, TableAttribute>(typeof(Object2ITableEntityConverter<>));
 
+            // IStorageTable --> IQueryable<ITableEntity>
+            converterManager.AddConverterBuilder<IStorageTable, TableQueryableOpenType, TableAttribute>(typeof(Table2IQueryableConverter<>));
```
mathewc: nit: can you rename Table2IQueryableConverter to TableToIQueryableConverter?

src/Microsoft.Azure.WebJobs.Host/Tables/TableAttributeBindingProvider.cs
mathewc: Yeah, can you please rename these? I see there are some existing BindToExactAsyncType methods as well. An "async type" doesn't really make sense.

src/Microsoft.Azure.WebJobs.Host/Tables/TableAttributeBindingProvider.cs
```diff
@@ -68,32 +62,106 @@ private TableAttributeBindingProvider(INameResolver nameResolver, IStorageAccoun
...
-                new IBindingProvider[] { bindToJArray, bindToJobject, bindAsyncCollector, original });
+                ValidateAttribute, nameResolver,
+                new IBindingProvider[]
+                {
+                    AllowMultipleRows(bindingFactory, bindAsyncCollector),
+                    AllowMultipleRows(bindingFactory, bindToExactCloudTable),
+                    AllowMultipleRows(bindingFactory, bindToExactTestCloudTable),
+                    bindToJArray,
+                    bindToJobject,
+                    original
```
mathewc: What is left in "original" now that most of the bindings have been converted?

src/Microsoft.Azure.WebJobs.Host/Tables/StringToStorageTableConverter.cs
MikeStall: The majority of the converter logic is unchanged (this brings up another issue: https://github.com/Azure/azure-webjobs-sdk/issues/974) .  These are some top-level converters, also to IStorageTable (which is the internal test interface).  I'll bang at it more before the final commit.

src/Microsoft.Azure.WebJobs.Host/Tables/TableAttributeBindingProvider.cs
MikeStall: Just single-entity converters. Those are especially tricky because Table has dedicated logic for deciding whether to push any changes back after method returns.

src/Microsoft.Azure.WebJobs.Host/Tables/TableAttributeBindingProvider.cs
```diff
@@ -68,32 +62,106 @@ private TableAttributeBindingProvider(INameResolver nameResolver, IStorageAccoun
...
+            IStorageTable table = GetTable(attribute);
+            await table.CreateIfNotExistsAsync(CancellationToken.None);
+
+            var sdkTable = table.SdkObject;
+            return sdkTable;
+        }
+
+        private static void ValidateAttribute(TableAttribute attribute, Type parameterType)
+        {
+            // Queue pre-existing  behavior: if there are { }in the path, then defer validation until runtime.
```
mathewc: Do you mean Table here? What about the presence of %% characters? Can we just remove this validation and rely on the centralized runtime validation?

src/Microsoft.Azure.WebJobs.Host/Tables/CloudTableArgumentBindingProvider.cs
MikeStall: It's moved to the converters.  There's some decent existing test coverage on this too.

src/Microsoft.Azure.WebJobs.Host/Tables/TableAttributeBindingProvider.cs
```diff
@@ -218,18 +286,10 @@ private ITableEntity JObjectToTableEntityConverterFunc(JObject source, TableAttr
...
-
-                if (argumentBinding == null)
-                {
-                    throw new InvalidOperationException("Can't bind Table to type '" + parameter.ParameterType + "'.");
-                }
-
-                binding = new TableBinding(parameter.Name, argumentBinding, client, path);
-            }
+                // This should have been caught by the other rule-based binders. 
+                // We never except this to get thrown.
```
mathewc: expect

src/Microsoft.Azure.WebJobs.Host/Bindings/OpenType.cs
MikeStall: I'll switch it to be a virtual and not convention based.

src/Microsoft.Azure.WebJobs.Host/Tables/TableAttributeBindingProvider.cs
```diff
@@ -365,5 +425,76 @@ public ResolvedTableAttribute(TableAttribute inner, IStorageTableClient client)
 
             internal IStorageTableClient Client { get; private set; }
         }
+
+        // IStorageTable --> IQueryable<T>
+        // ConverterManager's pattern matcher will figure out TElement. 
+        private class Table2IQueryableConverter<TElement> where TElement : ITableEntity, new()
```
mathewc: Why can't these converters be generic types like TableToQueryableConverter<IStorageTable, TElement> that derive from a converter interface that defines the contract? That would seem to make things clearer, and the types would be self describing. Also, that might make the searching for converters easier since you can look at interface types rather than searching by convention.

src/Microsoft.Azure.WebJobs.Host/Tables/TableAttributeBindingProvider.cs
```diff
@@ -365,5 +425,76 @@ public ResolvedTableAttribute(TableAttribute inner, IStorageTableClient client)
...
+
+                TableClient.VerifyDefaultConstructor(entityType);
+            }
+
+            public IQueryable<TElement> Convert(IStorageTable value)
+            {
+                // If Table does not exist, treat it like have zero rows. 
+                // This means return an non-null but empty enumerable.
+                // SDK doesn't do that, so we need to explicitly check. 
+                Task<bool> t = Task.Run(() => value.ExistsAsync(CancellationToken.None));
```
mathewc: Are the majority of converters expected to be sync or async? This just came up recently with @brettsam's changes to IValueBinder (making it async)?

src/Microsoft.Azure.WebJobs.Host/Bindings/ConverterManagerExtensions.cs
MikeStall: We don't have a strong covariance story yet. Although the converters we're adding right now don't really demand that.  Table is aggressive in that we have rules:
```
ITableEntity  --> ITableEntity   : do nothing. Common for DynamicEntity, or C# types that derive from ITE. 
JObject--> ITableEntity : invoke poco serialization logic originally in Script 
T --> ITableEntity:  invoke poco  serialization logic in sdk. 
```

We solve it via ordering. Rules registered first get precedence.

src/Microsoft.Azure.WebJobs.Host/Tables/TableAttributeBindingProvider.cs
MikeStall: We have a hole here - same problem with [Queue] (where this code was copied from). 
The problem is that we have existing (useful) legacy behavior that expects upfront static errors . So the runtime errors are too late. 
This approach here is definitely bad - this seems like something that should be rolled into SDK validation logic; perhaps related to the validation changes you've been looking at.

src/Microsoft.Azure.WebJobs.Host/Tables/TableAttributeBindingProvider.cs
```diff
@@ -365,5 +425,76 @@ public ResolvedTableAttribute(TableAttribute inner, IStorageTableClient client)
...
+            public ITableEntity Convert(TElement item)
+            {
+                return Converter.Convert(item);
+            }
+        }
+
+        // Matches IQueryable<T> where T is derived from ITableEntity
+        private class TableQueryableOpenType : OpenType
+        {
+            public static bool IsValid(Type t)
```
mathewc: The empty OpenType marker type still feels weird to me. If IsValid is required, why isn't is part of the base class?

src/Microsoft.Azure.WebJobs.Host/Tables/TableAttributeBindingProvider.cs
brettsam: I don't think we need 'Async' in the method namel. You are registering an async callback, but I think that 'BindToExactType' works fine -- and I'd vote that we only support async callbacks here so we don't need two versions of every 'BindTo' method.

src/Microsoft.Azure.WebJobs.Host/Tables/TableAttributeBindingProvider.cs
MikeStall: I'll make it a base class virtual.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
mathewc: Rather than Obsolete, can't you just remove? We're still in preview in this release.

MikeStall: I removed them and switched over to a single unified BindToInput

---
In reply to: [96505558](https://github.com/Azure/azure-webjobs-sdk/pull/967#discussion_r96505558) [](ancestors = 96505558)

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/ExactTypeBindingProvider.cs
MikeStall: removed this file. 

---
In reply to: [96106019](https://github.com/Azure/azure-webjobs-sdk/pull/967#discussion_r96106019) [](ancestors = 96106019)

src/Microsoft.Azure.WebJobs.Host/Bindings/ConverterManager.cs
MikeStall: Then I have to worry about (false, !null) possibilities. This is internal. 

---
In reply to: [96106814](https://github.com/Azure/azure-webjobs-sdk/pull/967#discussion_r96106814) [](ancestors = 96106814)

src/Microsoft.Azure.WebJobs.Host/Tables/TableAttributeBindingProvider.cs
MikeStall: These are all replaced with "BindToInput" 

---
In reply to: [96500661](https://github.com/Azure/azure-webjobs-sdk/pull/967#discussion_r96500661) [](ancestors = 96500661)

src/Microsoft.Azure.WebJobs.Host/Tables/TableAttributeBindingProvider.cs
MikeStall: Majority are sync. However, with the pattern matching, we can push this up and allow both sync or async. 

---
In reply to: [96492103](https://github.com/Azure/azure-webjobs-sdk/pull/967#discussion_r96492103) [](ancestors = 96492103)

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/QueueBindingProvider.cs
```diff
@@ -64,9 +64,13 @@ internal class QueueBindingProvider
             var bindingFactory = new BindingFactory(nameResolver, converterManager);
 
             var bindAsyncCollector = bindingFactory.BindToAsyncCollector<QueueAttribute, IStorageQueueMessage>(BuildFromQueueAttribute, ToWriteParameterDescriptorForCollector, CollectAttributeInfo);
```
brettsam: I like how you removed the BuildParameterDescriptor/PostResolveHook from the public API below. Can you do the same for the BindToAsyncCollector?

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
```diff
@@ -136,60 +136,60 @@ public IBindingProvider BindToGenericValueProvider<TAttribute>(Func<TAttribute,
...
+        /// If true, this rule can use the ConverterManager to convert from TType to the final parameter type </param>
+        /// <param name="builderType">A class with builder methods on it. This will get instantiated to perform the builder rule.</param>
+        /// <param name="constructorArgs">constructor arguments to pass to the typeBuilder instantiation. This can be used 
+        /// to flow state (like configuration, secrets, etc) from the configuration to the specific binding</param>
+        /// <returns>A binding rule.</returns>
+        public IBindingProvider BindToInput<TAttribute, TType>(
+            bool allowConversions,
+            Type builderType,
+            params object[] constructorArgs)
+                where TAttribute : Attribute
```
brettsam: This API confuses me now -- at first glance, I thought "what is a class with builder methods on it?" It looks like we require you to have a function with a method that looks like `TType AnyMethodName(TAttribute)`. I think it would be much more self-documenting if we had some interface like `IInputBuilder<TAttribute, TType>` with a single method `Task<TType> BuildInputAsync(TAttribute attribute);` You could use that instead of `object` in the other overload as well.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
```diff
@@ -136,60 +136,60 @@ public IBindingProvider BindToGenericValueProvider<TAttribute>(Func<TAttribute,
...
-        public IBindingProvider BindToExactAsyncType<TAttribute, TUserType>(
-            Func<TAttribute, Task<TUserType>> buildFromAttribute,
-            Func<TAttribute, ParameterInfo, INameResolver, ParameterDescriptor> buildParameterDescriptor = null,
-            Func<TAttribute, ParameterInfo, INameResolver, Task<TAttribute>> postResolveHook = null)
+        /// <typeparam name="TAttribute">Type of binding attribute on the user's parameter.</typeparam>
+        /// <typeparam name="TType">The user type must be compatible with this type for the binding to apply.</typeparam>
+        /// <param name="allowConversions">
+        /// If false, then the user parameter type must exactly match TType. 
+        /// If true, this rule can use the ConverterManager to convert from TType to the final parameter type </param>
+        /// <param name="builderInstance">Instance with converter methods on it.</param>
```
brettsam: It's not an instance with "converter" methods, is it? Shouldn't it be an instance with "builder" methods on it? Also ties in with my comment above.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
```diff
@@ -136,60 +136,60 @@ public IBindingProvider BindToGenericValueProvider<TAttribute>(Func<TAttribute,
...
+        /// <param name="allowConversions">
+        /// If false, then the user parameter type must exactly match TType. 
+        /// If true, this rule can use the ConverterManager to convert from TType to the final parameter type </param>
+        /// <param name="builderType">A class with builder methods on it. This will get instantiated to perform the builder rule.</param>
+        /// <param name="constructorArgs">constructor arguments to pass to the typeBuilder instantiation. This can be used 
+        /// to flow state (like configuration, secrets, etc) from the configuration to the specific binding</param>
+        /// <returns>A binding rule.</returns>
+        public IBindingProvider BindToInput<TAttribute, TType>(
+            bool allowConversions,
+            Type builderType,
```
brettsam: `BindToInput` changed to using a builder Type (which I'm fine with -- it may even make things nicer), but why didn't the other `BindTo` methods change? AsyncCollector and ValueProvider should match this pattern. If I'm writing a new binding, I'd expect the concepts to be similar for each rule I add.

mathewc: I think "builder type" should be an abstract type that defines a set of interface patterns that people can implement. Having it be an open type using method patterns will make things hard to discover and error prone I think.

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/QueueBindingProvider.cs
MikeStall: Yes - but let's do that in a separate change. 

---
In reply to: [97888844](https://github.com/Azure/azure-webjobs-sdk/pull/967#discussion_r97888844) [](ancestors = 97888844)

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/FluidBindingProvider.cs
```diff
@@ -0,0 +1,17 @@
...
+
+using System;
+using System.Reflection;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Protocols;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    // Base class to aide in private backwards compatability hooks for some bindings. 
+    internal class FluidBindingProvider<TAttribute>
```
mathewc: I don't understand what this is

src/Microsoft.Azure.WebJobs.Host/Bindings/PatternMatcher.cs
```diff
@@ -0,0 +1,263 @@
...
+
+                    T result = resultTask.GetAwaiter().GetResult();
+                    return result;
+                };
+                return converter;
+            }
+        }
+
+        // Wrapper for matching against a static type. 
+        private class CreateViaType : PatternMatcher
```
mathewc: I think these types need better names. How about TypeMatcher, InstanceMatcher (below)? Basically nouns instead of verbs.

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
```diff
@@ -197,6 +197,18 @@ public INameResolver NameResolver
         }
 
         /// <summary>
+        /// Get the converter manager, which can be used to register additional conversions for 
+        /// customizing model binding. 
+        /// </summary>
+        public IConverterManager ConverterManager
```
mathewc: Add a test for this to the JobHostConfiguration test suite, similar to the StorageClientFactory_GetterSetter test.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/FluidBindingProviderExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+
+using System;
+using System.Reflection;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Bindings;
+using Microsoft.Azure.WebJobs.Host.Protocols;
+
+namespace Microsoft.Azure.WebJobs.Host
+{    
+    // Internal Extension methods for setting backwards compatibility hooks on certain bindings.
```
mathewc: We don't have any backwards compatability constraints here right? Why don't we just put the model we want in place now?

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/QueueBindingProvider.cs
```diff
@@ -64,9 +64,13 @@ internal class QueueBindingProvider
...
-            var bindClient = bindingFactory.BindToExactAsyncType<QueueAttribute, IStorageQueue>(BuildClientFromQueueAttributeAsync, ToReadWriteParameterDescriptorForCollector, CollectAttributeInfo);
-            var bindSdkClient = bindingFactory.BindToExactAsyncType<QueueAttribute, CloudQueue>(BuildRealClientFromQueueAttributeAsync, ToReadWriteParameterDescriptorForCollector, CollectAttributeInfo);
-                        
+
+            var bindClient = bindingFactory.BindToInput<QueueAttribute, IStorageQueue>(false, typeof(QueueBuilder))
```
mathewc: I think its confusing that we're specifying a QueueBuilder type here containing our own implementation, but we use a different mechanism to specify the "resolve hooks". I think we should have the same simple model for all of these customization points. E.g. I just derive from an abstract case class and override some methods as needed. Required methods are abstract, optional are virtual.

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/QueueBindingProvider.cs
```diff
@@ -210,6 +201,22 @@ private static IStorageQueue GetQueue(QueueAttribute attrResolved)
...
+        private class QueueBuilder
+        {
+            public static async Task<IStorageQueue> ConvertBuildClientFromQueueAttributeAsync(QueueAttribute attrResolved)
+            {
+                IStorageQueue queue = GetQueue(attrResolved);
+                await queue.CreateIfNotExistsAsync(CancellationToken.None);
+                return queue;
+            }
+
+            public static async Task<CloudQueue> ConvertBuildRealClientFromQueueAttributeAsync(QueueAttribute attrResolved)
```
mathewc: This method name (and the one above it) are pretty inscrutable :)

src/Microsoft.Azure.WebJobs.Host/Tables/TableAttributeBindingProvider.cs
```diff
@@ -68,106 +62,81 @@ private TableAttributeBindingProvider(INameResolver nameResolver, IStorageAccoun
...
         }
-        
+
         public static IBindingProvider Build(INameResolver nameResolver, IConverterManager converterManager, IStorageAccountProvider accountProvider, IExtensionRegistry extensions)
         {
             var original = new TableAttributeBindingProvider(nameResolver, accountProvider, extensions);
 
-            converterManager.AddConverter<JObject, ITableEntity, TableAttribute>(original.JObjectToTableEntityConverterFunc);
+            converterManager.AddConverter<JObject, ITableEntity, TableAttribute>(original.JObjectToTableEntityConverterFunc);            
+            converterManager.AddConverterBuilder<object, ITableEntity, TableAttribute>(typeof(ObjectToITableEntityConverter<>));
```
mathewc: Let's rename this to AddConverter - they you simply have various overloads. Each is in the end adding a converter, in some form.

src/Microsoft.Azure.WebJobs.Host/Tables/TableAttributeBindingProvider.cs
```diff
@@ -68,106 +62,81 @@ private TableAttributeBindingProvider(INameResolver nameResolver, IStorageAccoun
...
-            var bindAsyncCollector = bindingFactory.BindToAsyncCollector<TableAttribute, ITableEntity>(original.BuildFromTableAttribute, null, original.CollectAttributeInfo);
 
-            var bindToJobject = bindingFactory.BindToExactAsyncType<TableAttribute, JObject>(original.BuildJObject, null, original.CollectAttributeInfo);
-            var bindToJArray = bindingFactory.BindToExactAsyncType<TableAttribute, JArray>(original.BuildJArray, null, original.CollectAttributeInfo);
+            // Includes converter manager, which provides access to IQueryable<ITableEntity>
+            var bindToExactCloudTable = bindingFactory.BindToInput<TableAttribute, CloudTable>(true, typeof(JObjectBuilder))
+                .SetPostResolveHook<TableAttribute>(original.ToParameterDescriptorForCollector, original.CollectAttributeInfo);
+
+            var bindToExactTestCloudTable = bindingFactory.BindToInput<TableAttribute, IStorageTable>(true, typeof(JObjectBuilder))
+                .SetPostResolveHook<TableAttribute>(original.ToParameterDescriptorForCollector, original.CollectAttributeInfo);
```
mathewc: Can we remove the SetPostResolveHook method and supporting infrastructure, and instead have the builder Type passed in implement an interface providing the hook impls?

src/Microsoft.Azure.WebJobs.Host/IConverterManager.cs
```diff
@@ -72,5 +87,60 @@ public static class IConverterManagerExtensions
...
+        public static void AddConverter<TSource, TDestination, TAttribute>(
+          this IConverterManager converterManager,
+          object converterInstance)
+          where TAttribute : Attribute
+        {
+            var patternMatcher = PatternMatcher.New(converterInstance);
+            converterManager.AddConverterBuilder<TSource, TDestination, TAttribute>(patternMatcher);
+        }
+
+        private static void AddConverterBuilder<TSource, TDestination, TAttribute>(
```
brettsam: Were you going to rename this to AddConverter as well (instead of Builder)?

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
```diff
@@ -136,60 +136,50 @@ public IBindingProvider BindToGenericValueProvider<TAttribute>(Func<TAttribute,
...
-        public IBindingProvider BindToExactAsyncType<TAttribute, TUserType>(
-            Func<TAttribute, Task<TUserType>> buildFromAttribute,
-            Func<TAttribute, ParameterInfo, INameResolver, ParameterDescriptor> buildParameterDescriptor = null,
-            Func<TAttribute, ParameterInfo, INameResolver, Task<TAttribute>> postResolveHook = null)
+        /// <typeparam name="TAttribute">Type of binding attribute on the user's parameter.</typeparam>
+        /// <typeparam name="TType">The user type must be compatible with this type for the binding to apply.</typeparam>
+        /// <param name="builderInstance">Instance with converter methods on it.</param>
+        /// <returns>A binding rule.</returns>
+        public IBindingProvider BindToInput<TAttribute, TType>(
+            object builderInstance)
```
brettsam: Doesn't this have to implment IConverter<TAttribute, TType>?

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
```diff
@@ -136,60 +136,50 @@ public IBindingProvider BindToGenericValueProvider<TAttribute>(Func<TAttribute,
...
         /// </summary>
-        /// <typeparam name="TAttribute">type of binding attribute on the user's parameter.</typeparam>
-        /// <typeparam name="TUserType">The exact type of the user's parameter that this will bind to.</typeparam>
-        /// <param name="buildFromAttribute">builder function to create the object that will get passed to the user function.</param>
-        /// <returns>A binding provider that applies these semantics.</returns>
-        public IBindingProvider BindToExactType<TAttribute, TUserType>(Func<TAttribute, TUserType> buildFromAttribute)
-            where TAttribute : Attribute
+        /// <typeparam name="TAttribute">Type of binding attribute on the user's parameter.</typeparam>
+        /// <typeparam name="TType">The user type must be compatible with this type for the binding to apply.</typeparam>
+        /// <param name="builderType">A class with builder methods on it. This will get instantiated to perform the builder rule.</param>
```
brettsam: This should read "a class that implements IConverter<TAttribute, TType>"

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/FluidBindingProviderExtensions.cs
brettsam: I agree with Mathew's feedback from yesterday. It'd be easier to follow if we had some internal IBuildParameterDescriptor/IPostResolveHook interfaces that you could check for, rather than attaching state like this.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
```diff
@@ -101,11 +101,11 @@ public IBindingProvider BindToGenericValueProvider<TAttribute>(Func<TAttribute,
...
+        /// <typeparam name="TAttribute">Type of binding attribute on the user's parameter.</typeparam>
         /// <typeparam name="TMessage">'core type' for the IAsyncCollector.</typeparam>
-        /// <param name="buildFromAttribute">function to allocate the collector object given a resolved instance of the attribute.</param>
+        /// <param name="buildFromAttribute">Function to allocate the collector object given a resolved instance of the attribute.</param>
         /// <param name="buildParameterDescriptor">An optional function to create a specific ParameterDescriptor object for the dashboard. If missing, a default ParameterDescriptor is created. </param>
-        /// <param name="postResolveHook">an advanced hook for translating the attribute. </param>
+        /// <param name="postResolveHook">An advanced hook for translating the attribute. </param>
         /// <returns>A binding provider that applies these semantics.</returns>
         public IBindingProvider BindToAsyncCollector<TAttribute, TMessage>(
             Func<TAttribute, IAsyncCollector<TMessage>> buildFromAttribute,
```
brettsam: Will you be working on getting AsyncCollector into this pattern next? I'd rather only have to absorb these changes once in Extensions.

PR #971, opened 2017-01-13
Improvements to template parameter resolution

mathewc:
Still want to add some tests, but the new resolve logic for Table.Filter is tested and working.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingContext.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingContext.cs
@@ -14,6 +14,7 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
     {
         private readonly FunctionBindingContext _functionContext;
         private readonly IReadOnlyDictionary<string, object> _bindingData;
+        private readonly IDictionary<string, object> _properties;
         private readonly CancellationToken _cancellationToken;
 
         private AmbientBindingContext _ambientContext;
@@ -35,6 +36,7 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
             _bindingData = bindingData;
             _functionContext = valueContext.FunctionContext;
             _cancellationToken = valueContext.CancellationToken;
+            _properties = new Dictionary<string, object>();
         }
 
         internal BindingContext(AmbientBindingContext ambientContext, CancellationToken cancellationToken)
@@ -48,6 +50,7 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
             _functionContext = ambientContext.FunctionContext;
             _bindingData = ambientContext.BindingData;
             _cancellationToken = cancellationToken;
+            _properties = new Dictionary<string, object>();
         }
 
         /// <summary>
@@ -80,6 +83,18 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
         public IReadOnlyDictionary<string, object> BindingData
         {
             get { return _bindingData; }
+        }
+
+        /// <summary>
+        /// Gets the set of property values for the current bind operation.
+        /// </summary>
+        /// <remarks>
+        /// Bindings can add arbitrary values to this collection as needed to
+        /// flow state around.
+        /// </remarks>
+        public IDictionary<string, object> Properties
+        {
+            get { return _properties; }
         }
 
         /// <summary>
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplate.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplate.cs
@@ -20,8 +20,9 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings.Path
         private readonly string _pattern;
         private readonly IReadOnlyList<BindingTemplateToken> _tokens;
         private readonly bool _ignoreCase;
+        private readonly ParameterResolver _parameterResolver;
 
-        internal BindingTemplate(string pattern, IReadOnlyList<BindingTemplateToken> tokens, bool ignoreCase = false)
+        internal BindingTemplate(string pattern, IReadOnlyList<BindingTemplateToken> tokens, bool ignoreCase = false, ParameterResolver parameterResolver = null)
         {
             if (pattern == null)
             {
@@ -36,6 +37,7 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings.Path
             _pattern = pattern;
             _tokens = tokens;
             _ignoreCase = ignoreCase;
+            _parameterResolver = parameterResolver ?? new DefaultParameterResolver();
         }
 
         /// <summary>
@@ -66,14 +68,14 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings.Path
         /// A factory method to parse input template string and construct a binding template instance using
         /// parsed tokens sequence.
         /// </summary>
-        /// <param name="input">A binding template string in a format supported by <see cref="BindingTemplateParser"/>.
-        /// </param>
+        /// <param name="input">A binding template string in a format supported by <see cref="BindingTemplateParser"/></param>
+        /// <param name="parameterResolver">The optional <see cref="ParameterResolver"/> to use when resolving parameter values.</param>
         /// <param name="ignoreCase">True if matching should be case insensitive.</param>
         /// <returns>Valid ready-to-use instance of <see cref="BindingTemplate"/>.</returns>
-        public static BindingTemplate FromString(string input, bool ignoreCase = false)
+        public static BindingTemplate FromString(string input, bool ignoreCase = false, ParameterResolver parameterResolver = null)
         {
             IReadOnlyList<BindingTemplateToken> tokens = BindingTemplateParser.ParseTemplate(input);
-            return new BindingTemplate(input, tokens, ignoreCase);
+            return new BindingTemplate(input, tokens, ignoreCase: ignoreCase, parameterResolver: parameterResolver);
         }
 
         /// <summary>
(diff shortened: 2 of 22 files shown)


src/Microsoft.Azure.WebJobs.Host/Bindings/BindingContext.cs
```diff
@@ -83,6 +86,18 @@ public TraceWriter Trace
         }
 
         /// <summary>
+        /// Gets the set of property values for the current bind operation.
+        /// </summary>
+        /// <remarks>
+        /// Bindings can add arbitrary values to this collection as needed to
+        /// flow state around.
+        /// </remarks>
+        public IDictionary<string, object> Properties
```
mathewc: This is the planned way for binding specific custom resolvers to communicate state back to the binding pipeline.

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplate.cs
```diff
@@ -99,28 +129,26 @@ public string Bind(IReadOnlyDictionary<string, string> parameters)
...
-                        // parameters (e.g. rand-guid, datetime, etc.)
-                        value = resolver.Resolve(token.Value);
-                        builder.Append(value);
-                    }
-                    else
+                        ParameterIndex = token.Index,
+                        ParameterName = token.Value,
+                        Template = _pattern,
+                        BindingData = parameters,
+                        Properties = bindingContext?.Properties
```
mathewc: Currently this Properties bag is the only reason BindingData needs to flow through. I couldn't add BindingData in its entirety on ParameterBindingContext because the core assembly doesn't have access to the host assembly. I think that is fine though.

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplate.cs
```diff
@@ -99,28 +129,26 @@ public string Bind(IReadOnlyDictionary<string, string> parameters)
                 parameters = caseInsensitive;
             }
 
+            StringBuilder builder = new StringBuilder();
             foreach (BindingTemplateToken token in Tokens)
             {
                 if (token.IsParameter)
                 {
-                    string value;
-                    BindingParameterResolver resolver = null;
```
mathewc: All this internal binding logic has been factored out into a DefaultParameterResolver.

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplate.cs
```diff
@@ -36,6 +37,7 @@ internal BindingTemplate(string pattern, IReadOnlyList<BindingTemplateToken> tok
             _pattern = pattern;
             _tokens = tokens;
             _ignoreCase = ignoreCase;
+            _parameterResolver = parameterResolver ?? new DefaultParameterResolver();
```
mathewc: Eventually we'll want this DefaultParameterResolver to be added to the JobHost service collection so it can be replaced entirely. There are some hurdles to that currently.  First, there isn't any way to get to the services collection from this point in the code, and most existing callers won't be passing a resolver in. Currently its only the extensions build on the new extensibility model. So until we can put a consistent story in place across the board for this, the default resolver remains internal.

src/Microsoft.Azure.WebJobs.Host/DefaultParameterResolver.cs
```diff
@@ -0,0 +1,41 @@
...
+
+using System;
+using Microsoft.Azure.WebJobs.Host.Bindings.Path;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// The default <see cref="ParameterResolver"/> used by the framework.
+    /// </summary>
+    public class DefaultParameterResolver : ParameterResolver
```
mathewc: Until this is truely pluggable by people at the JobHost level, I can just make this internal.

src/Microsoft.Azure.WebJobs/ParameterResolverContext.cs
```diff
@@ -0,0 +1,45 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// Class providing context information for a template parameter resolution.
+    /// </summary>
+    public class ParameterResolverContext
```
MikeStall: >ParameterResolverContext [](start = 17, length = 24)

Does each { } set in a template get its own ParameterResolverContext class?

src/Microsoft.Azure.WebJobs/ParameterResolverContext.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// <summary>
+        /// Gets or sets the name of the parameter being resolved.
+        /// </summary>
+        public string ParameterName { get; set; }
+
+        /// <summary>
+        /// Gets or sets the index into the <see cref="Template"/> where
+        /// the parameter was parsed from.
+        /// </summary>
+        public int ParameterIndex { get; set; }
```
MikeStall: Do we need Template  &ParameterIndex? Can we remove them?

src/Microsoft.Azure.WebJobs.Host/DefaultParameterResolver.cs
```diff
@@ -0,0 +1,41 @@
...
+
+            string value;
+            BindingParameterResolver resolver = null;
+            if (context.BindingData != null && context.BindingData.TryGetValue(context.ParameterName, out value))
+            {
+                // parameter is resolved from binding data
+                context.Value = value;
+                return true;
+            }
+            else if (BindingParameterResolver.TryGetResolver(context.ParameterName, out resolver))
```
MikeStall: Could these builtin resolvers now be moved into DefaultParameterResolver?
IE, if someobody wanted to override the defaults (say remove them; or change how 'datetime' behaved), they could then just change the DefaultParameterResolver?

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplate.cs
MikeStall: I think this is an important point. We need to be able to eventually have this on JobHostConfig. Are we confident there's a clear non-breaking path to do this? 

---
In reply to: [95934502](https://github.com/Azure/azure-webjobs-sdk/pull/971#discussion_r95934502) [](ancestors = 95934502)

test/Microsoft.Azure.WebJobs.Host.TestCommon/TestTraceWriter.cs
```diff
@@ -11,7 +11,7 @@ public class TestTraceWriter : TraceWriter
         public Collection<TraceEvent> Traces = new Collection<TraceEvent>();
         private object _syncLock = new object();
 
-        public TestTraceWriter(TraceLevel level) : base(level)
+        public TestTraceWriter(TraceLevel level = TraceLevel.Verbose) : base(level)
```
MikeStall: >(TraceLevel level = TraceLevel.Verbose [](start = 30, length = 38)

Is this change related to the PR?

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingContext.cs
MikeStall: What's the difference between BindingData and Properties?

---
In reply to: [95933589](https://github.com/Azure/azure-webjobs-sdk/pull/971#discussion_r95933589) [](ancestors = 95933589)

mathewc: BindingData is coming from BindingContext.BindingData and contains the actual invoke values from the request (e.g. what our default resolver uses to resolve parameters from incoming binding data). Open to naming suggestions for "Properties". It's just a bag for arbitrary use.

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplate.cs
mathewc: This logic isn't pluggable today, so I don't think we need to worry about this now. I've taken it 90% of the way. I'm confident we can do this as an add on later.

However, there might still be a way for me to get this in place now, if we made this available statically (e.g. similar to System.Web.Http.GlobalConfiguration). Something like that would allow me to access from BindingTemplate internally and get this working for all bindings.

I'll take this up in a separate PR/work item, since this is an additive feature.

src/Microsoft.Azure.WebJobs.Host/DefaultParameterResolver.cs
mathewc: Yeah, that's a second level of refactoring that can happen. I could do that in a follow up since its all internal. It's interesting that I had already created an internal "parameter resolver" hierarchy for this stuff :) Good that we're finishing this and making it public. I'll unify this stuff.

src/Microsoft.Azure.WebJobs/ParameterResolverContext.cs
mathewc: Yes - see how we new up a ParameterResolverContext on each resolve from the call site in BindingTemplate.Bind

src/Microsoft.Azure.WebJobs/ParameterResolverContext.cs
mathewc: Yes - see how I'm using them in the Table filter resolver

test/Microsoft.Azure.WebJobs.Host.TestCommon/TestTraceWriter.cs
mathewc: In writing my test, I wanted this simplification

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingContext.cs
brettsam: I don't see how I get access to the BindingContext for my binding rules.

PR #977, opened 2017-01-19
Adding a new Bind extension method to Binder

mathewc:
IBinder has a non-async Bind extension method that calls the underlying async impl. Adding a corresponding symmetric one for Binder's BindAsync method that takes a collection of attributes.

This comes up often in Functions and makes the code users have to write harder.

### DIFF
--- a/src/Microsoft.Azure.WebJobs/IBinderExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs/IBinderExtensions.cs
@@ -5,8 +5,10 @@ using System;
 
 namespace Microsoft.Azure.WebJobs
 {
-    /// <summary>Provides extension methods for the <see cref="IBinder"/> interface.</summary>
-    public static class BinderExtensions
+    /// <summary>
+    /// Provides extension methods for the <see cref="IBinder"/> interface.
+    /// </summary>
+    public static class IBinderExtensions
     {
         /// <summary>Binds the specified attribute.</summary>
         /// <typeparam name="T">The type to which to bind.</typeparam>
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/Runtime/BinderExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/Runtime/BinderExtensions.cs
@@ -1,0 +1,31 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// Provides extension methods for <see cref="Binder"/>.
+    /// </summary>
+    public static class BinderExtensions
+    {
+        /// <summary>
+        /// Binds the specified attribute.
+        /// </summary>
+        /// <typeparam name="T">The type to which to bind.</typeparam>
+        /// <param name="binder">The binder to use to bind.</param>
+        /// <param name="attributes">The collection of attributes to bind. The first attribute in the
+        /// collection should be the primary attribute.</param>
+        /// <returns></returns>
+        public static T Bind<T>(this Binder binder, Attribute[] attributes)
+        {
+            if (binder == null)
+            {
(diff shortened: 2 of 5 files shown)


src/Microsoft.Azure.WebJobs.Protocols/BinderParameterDescriptor.cs
```diff
@@ -7,7 +7,9 @@ namespace Microsoft.Azure.WebJobs.Protocols
 namespace Microsoft.Azure.WebJobs.Host.Protocols
 #endif
 {
-    /// <summary>Represents a parameter bound to an IBinder.</summary>
+    /// <summary>
+    /// Represents a parameter bound to a Binder.
+    /// </summary>
     [JsonTypeName("IBinder")]
```
mathewc: Leaving the serialization protocol name as IBinder for back compat (here and below).

src/Microsoft.Azure.WebJobs.Host/Bindings/Runtime/Binder.cs
```diff
@@ -16,7 +16,7 @@ namespace Microsoft.Azure.WebJobs
     /// <summary>
     /// Class providing functionality for dynamically binding to WebJobs SDK attributes at runtime.
     /// </summary>
-    public class Binder : IBinder, IWatchable, IDisposable
+    public class Binder : IWatchable, IDisposable
```
MikeStall: >public [](start = 4, length = 6)

Can we make Binder private and leave IBinder public? 
A public interface is less liability than a public class.

mathewc: The problem is that interfaces are not extensible at all. That's exactly the problem I ran into with IBinder - for Functions I needed to add a new overload as well as access to BindingData but was unable to. Having a concrete class allows us to add new members in the future as needed without breaking changes.

src/Microsoft.Azure.WebJobs/IBinder.cs
```diff
@@ -1,26 +0,0 @@
...
-using System.Threading.Tasks;
-
-namespace Microsoft.Azure.WebJobs
-{
-    /// <summary>
-    /// Interface defining functionality for dynamically binding to WebJobs SDK attributes
-    /// at runtime.
-    /// </summary>
-    /// <remarks>This interface enables imperative binding with attribute information specified at runtime.</remarks>
-    public interface IBinder
```
MikeStall: Removing this interface would be a significant breaking change to all consumers (and not just extension authors). This has been a long-standing interface that we've encouraged consumers to use to do late bindings.

mathewc: Yeah - I also realized that it would be breaking for Functions, though we've encouraged people to use Binder there. I'll rework this PR to just add the new overload I was trying to add in the first place w/o removing this. We might also consider marking it as obsolete so people have time to merge over.

src/Microsoft.Azure.WebJobs/IBinderExtensions.cs
```diff
@@ -5,8 +5,10 @@
 
 namespace Microsoft.Azure.WebJobs
 {
-    /// <summary>Provides extension methods for the <see cref="IBinder"/> interface.</summary>
-    public static class BinderExtensions
+    /// <summary>
+    /// Provides extension methods for the <see cref="IBinder"/> interface.
+    /// </summary>
+    public static class IBinderExtensions
```
mathewc: Renaming this class to avoid type name conflicts with the new extension class I'm adding. Should have been named this to begin with.

Since people don't reference this extension class directly by type name, this technically breaking change should be fine.

mamaso:
I'm not a fan of adding synchronous wrappers to our async methods due to the deadlock problems associated with some of the other synchronous wrappers added to the code.

What does the functions scenario look like? Will they always need to call this function in a synchronous fashion? If this saves a `.GetAwaiter().GetResult()` call in basically every case then its a reasonable change, otherwise I'd prefer to avoid it.

mathewc:
Closing for now due to the unpopularity of the change.

PR #987, opened 2017-01-26
Emit a warning for async void functions

mathewc:
Constantly in the forums and on StackOverflow users report issues with their job functions, and it often turns out to be that they have incorrectly defined an `async void` function. E.g. most recently on SO here: http://stackoverflow.com/questions/41876364/queue-messages-not-being-moved-to-the-poison-queue.

Since there are no good reasons to do this, any instances are most certainly a user error. So I'm throwing an indexing error in these cases.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/TypeUtility.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/TypeUtility.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Globalization;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Azure.WebJobs.Host
@@ -73,6 +74,30 @@ namespace Microsoft.Azure.WebJobs.Host
             }
 
             return null;
+        }
+
+        public static bool IsAsync(MethodInfo methodInfo)
+        {
+            if (methodInfo == null)
+            {
+                throw new ArgumentNullException(nameof(methodInfo));
+            }
+
+            var stateMachineAttribute = methodInfo.GetCustomAttribute<AsyncStateMachineAttribute>();
+            if (stateMachineAttribute != null)
+            {
+                var stateMachineType = stateMachineAttribute.StateMachineType;
+                if (stateMachineType != null)
+                {
+                    return stateMachineType.GetCustomAttribute<CompilerGeneratedAttribute>() != null;
+                }
+            }
+            return false;
+        }
+
+        public static bool IsAsyncVoid(MethodInfo methodInfo)
+        {
+            return IsAsync(methodInfo) && (methodInfo.ReturnType == typeof(void));
         }
     }
 }
--- a/test/Microsoft.Azure.WebJobs.Host.UnitTests/Indexers/FunctionIndexerTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.UnitTests/Indexers/FunctionIndexerTests.cs
@@ -3,17 +3,19 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
+using System.Linq;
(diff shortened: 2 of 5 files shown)


mathewc:
@fabiocav @mamaso You guys did a similar check in Functions. Now that this will be in the core SDK, we need to decide if the higher level check adds any value.

src/Microsoft.Azure.WebJobs.Host/TypeUtility.cs
```diff
@@ -74,5 +75,29 @@ internal static bool IsJObject(Type type)
...
         }
+
+        public static bool IsAsync(MethodInfo methodInfo)
+        {
+            if (methodInfo == null)
+            {
+                throw new ArgumentNullException(nameof(methodInfo));
+            }
+
+            var stateMachineAttribute = methodInfo.GetCustomAttribute<AsyncStateMachineAttribute>();
```
MikeStall: This looks kind of magical. How do we know this?

test/Microsoft.Azure.WebJobs.Host.UnitTests/Indexers/FunctionIndexerTests.cs
```diff
@@ -115,6 +115,22 @@ public void IndexMethod_IfMethodReturnsTask_DoesNotThrow()
         }
 
         [Fact]
+        public void IndexMethod_IfMethodReturnsAsyncVoid_Throws()
+        {
+            // Arrange
+            IFunctionIndexCollector index = CreateDummyFunctionIndex();
```
MikeStall: I added some test helpers that deal with the boiler plate code here and avoids the casts, awaits, mocks, etc. 
Check out Fails_BindingContract_Mismatch() as an example. Can you use that pattern?

src/Microsoft.Azure.WebJobs.Host/TypeUtility.cs
mathewc: Standard attributes applied by the compiler. E.g. http://stackoverflow.com/questions/13183313/how-to-determine-if-method-is-async-at-runtime

MikeStall: Can we include that url in the comments?

mamaso:
Throwing the indexing error is a breaking change, we only log a warning. Otherwise I'd say this check seems to supersede the functions' check.

mathewc:
I'm not concerned with this breaking change - anyone doing this pattern is already broken I believe.

mamaso:
Agreed that most are broken, but not all. An async void function will still execute and bindings such as `CloudBlockBlob` would work without a problem.

fabiocav:
I was going to mention that this was a breaking change, but @mathewc told me I didn't need to comment because he gets it.

mathewc:
Yeah, for Functions we don't want to take a breaking change. I'll look into making this a warning.

src/Microsoft.Azure.WebJobs.Host/TypeUtility.cs
mathewc: I think the documentation on these attributes is clear enough. It says "Indicates whether a method is marked with either the Async (Visual Basic) or async (C# Reference) modifier."

mathewc:
Changed this to a warning. In core SDK usages, this will show up in the console as a yellow warning - enough to get the users attention.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Indexers/FunctionIndexerTests.cs
mathewc: See my expanded test - I need to pass in a mock trace writer. I'm going to just leave this test using the same tools all the other tests in the suite are using for now. If we want to we can convert all of them over later.

Issue #990, opened 2017-01-26
Remove Table.Filter specific logic in AttributeCloner


MikeStall:
This change https://github.com/Azure/azure-webjobs-sdk/commit/53990372268e0fa172a308ac16a99f4a582d4236 let Table.Filter  be an OData-aware resolver string, instead of just a vanilla { } string. 

We need to remove the hardcoded logic and generalize it.  Some sort of "resolver policy" on [AutoResolve] attribute.

PR #1000, opened 2017-02-01
expanding ResolutionPolicy for public use

brettsam:
I partially resurrected Mathew's PR here: https://github.com/Azure/azure-webjobs-sdk/pull/971. I'm going to need this to be able to move forward with DocDB so wanted to get an initial review out before writing lots of tests and plumbing i through everything.

One big change that @MikeStall will be interested in -- now passing the ValueBindingContext along with the resolved attribute to all of the builder methods (only ExactType is doing this for this initial PR).

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
@@ -6,7 +6,6 @@ using System.Collections.Generic;
 using System.Reflection;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host.Bindings.Path;
-using Microsoft.Azure.WebJobs.Host.Tables;
 
 namespace Microsoft.Azure.WebJobs.Host.Bindings
 {
@@ -78,9 +77,9 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
                     BindingTemplate template;
                     if (TryCreateAutoResolveBindingTemplate(propInfo, nameResolver, out template))
                     {
-                        var policy = ResolutionPolicy.GetPolicy(propInfo);
+                        IResolutionPolicy policy = GetPolicy(propInfo);
                         template.ValidateContractCompatibility(bindingDataContract);
-                        getArgFuncs[i] = (bindingData) => policy.TemplateBind(template, bindingData);
+                        getArgFuncs[i] = (bindingData) => TemplateBind(policy, propInfo, source, template, bindingData);
                     }
                     else
                     {
@@ -106,9 +105,9 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
                             BindingTemplate template;
                             if (TryCreateAutoResolveBindingTemplate(prop, nameResolver, out template))
                             {
-                                var policy = ResolutionPolicy.GetPolicy(prop);
+                                IResolutionPolicy policy = GetPolicy(prop);
                                 template.ValidateContractCompatibility(bindingDataContract);
-                                setProperties.Add((newAttr, bindingData) => prop.SetValue(newAttr, policy.TemplateBind(template, bindingData)));
+                                setProperties.Add((newAttr, bindingData) => prop.SetValue(newAttr, TemplateBind(policy, prop, newAttr, template, bindingData)));
                             }
                             else
                             {
@@ -185,7 +184,7 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
             }
 
             return true;
-        }        
+        }
 
         // Get a attribute with %% resolved, but not runtime {} resolved.
         public TAttribute GetNameResolvedAttribute()
@@ -305,47 +304,49 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
             return newAttr;
         }
 
-        // Resolution policy for { } in  binding templates. 
-        // The default policy is just a direct substitution for the binding data. 
-        // Derived policies can enforce formatting / escaping when they do injection. 
-        private class ResolutionPolicy
+        private static string TemplateBind(IResolutionPolicy policy, PropertyInfo prop, Attribute attr, BindingTemplate template, IReadOnlyDictionary<string, object> bindingData)
         {
-            public string TemplateBind(BindingTemplate template, IReadOnlyDictionary<string, object> bindingData)
+            if (bindingData == null)
             {
-                if (bindingData == null)
-                {
-                    return template.Pattern;
-                }
-                return this.TemplateBindWorker(template, bindingData);
+                return template?.Pattern;
             }
 
-            protected virtual string TemplateBindWorker(BindingTemplate template, IReadOnlyDictionary<string, object> bindingData)
-            {                
-                return template.Bind(bindingData);
-            }
-
-            // Expand on [AutoResolve] property to make this pluggable, 
-            public static ResolutionPolicy GetPolicy(PropertyInfo propInfo)
-            {
-                if (propInfo.DeclaringType == typeof(TableAttribute) && propInfo.Name == "Filter")
-                {
-                    return new TableFilterResolutionPolicy();
-                }
-                else
-                {
-                    return new ResolutionPolicy();
-                }
-            }
+            return policy.TemplateBind(prop, attr, template, bindingData);
         }
 
-        // Table.Filter's { } resolution does OData escaping. 
-        private class TableFilterResolutionPolicy : ResolutionPolicy
+        internal static IResolutionPolicy GetPolicy(PropertyInfo propInfo)
         {
-            protected override string TemplateBindWorker(BindingTemplate template, IReadOnlyDictionary<string, object> bindingData)
+            AutoResolveAttribute autoResolveAttribute = propInfo.GetCustomAttribute<AutoResolveAttribute>();
+
+            var formatterType = autoResolveAttribute.ResolutionPolicyType;
+
+            if (formatterType != null)
             {
-                var filter = TableFilterFormatter.Format(template, bindingData);
-                return filter;
+                // Special-case Table as there is no way to declare this ResolutionPolicy
+                // and use BindingTemplate in the Core assembly
+                if (formatterType == typeof(WebJobs.ODataFilterResolutionPolicy))
+                {
+                    return new ODataFilterResolutionPolicy();
+                }
+
+                if (!typeof(IResolutionPolicy).IsAssignableFrom(formatterType))
+                {
+                    throw new InvalidOperationException($"The {nameof(AutoResolveAttribute.ResolutionPolicyType)} on {propInfo.Name} must derive from {typeof(IResolutionPolicy).Name}.");
+                }
+
+                try
+                {
+                    var obj = Activator.CreateInstance(formatterType);
+                    return (IResolutionPolicy)obj;
+                }
+                catch (MissingMethodException)
+                {
+                    throw new InvalidOperationException($"The {nameof(AutoResolveAttribute.ResolutionPolicyType)} on {propInfo.Name} must derive from {typeof(IResolutionPolicy).Name} and have a default constructor.");
+                }
             }
+
+            // return the default policy                        
+            return new DefaultResolutionPolicy();
         }
 
         // If no name resolver is specified, then any %% becomes an error.
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/DefaultResolutionPolicy.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/DefaultResolutionPolicy.cs
@@ -1,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using Microsoft.Azure.WebJobs.Host.Bindings;
+using Microsoft.Azure.WebJobs.Host.Bindings.Path;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// Resolution policy for { } in  binding templates. 
+    /// The default policy is just a direct substitution for the binding data. 
+    /// Derived policies can enforce formatting / escaping when they do injection. 
+    /// </summary>
+    internal class DefaultResolutionPolicy : IResolutionPolicy
+    {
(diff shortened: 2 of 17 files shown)


src/Microsoft.Azure.WebJobs/ResolutionContext.cs
```diff
@@ -0,0 +1,45 @@
...
+            _properties = properties;
+        }
+
+        /// <summary>
+        /// The value of the property.
+        /// </summary>
+        public string PropertyValue { get; set; }
+
+        /// <summary>
+        /// Gets or sets the binding data collection.
```
MikeStall: >or sets [](start = 17, length = 7)

Type is read-only dict. Can data be set?

MikeStall:
>         [AutoResolve]

Did this need a resolution policy marker?  



---
Refers to: src/Microsoft.Azure.WebJobs/TableAttribute.cs:94 in 07f8728. [](commit_id = 07f8728343b137abeb05447e5b4c6b8c170fb2e2, deletion_comment = False)

src/Microsoft.Azure.WebJobs/ResolutionContext.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// </summary>
+        public IReadOnlyDictionary<string, object> BindingData { get; set; }
+
+        /// <summary>
+        /// Gets or sets property values for the current bind operation.
+        /// </summary>
+        public IDictionary<string, object> Properties
```
MikeStall: >Properties [](start = 43, length = 10)

What's the difference between properties and binding data?

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -23,22 +22,26 @@ internal class AttributeCloner<TAttribute>
...
 
+        private IResolutionPolicy _defaultResolutionPolicy;
+
         public AttributeCloner(
             TAttribute source,
             IReadOnlyDictionary<string, Type> bindingDataContract,
             INameResolver nameResolver = null,
-            Func<TAttribute, Task<TAttribute>> hook = null)
+            Func<TAttribute, Task<TAttribute>> hook = null,
+            IResolutionPolicy defaultResolutionPolicy = null)
```
MikeStall: >IResolutionPolicy defaultResolutionPolicy = null [](start = 12, length = 48)

This feels wrong:
1. attribute wide?  What happens when different properties have different resolution policies. 
2. defaulting to null?  Why can't we just grab it off the [AutoResolve] attribute?

brettsam: This is the DefaultResolutionPolicy -- what we fall back to if it isn't specified on the [AutoRenew]. It's not happening yet (I meant to call this out), but this would let someone register their own DefaultResolutionPolicy. Then, they would pull it from the Services and pass it into the BindingFactory like they do with INameResolver today. If we want to allow this (which I believe you called out in the other review), we'd need a way to let someone override the default somewhere.

src/Microsoft.Azure.WebJobs/ODataFilterResolutionPolicy.cs
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// Marker type for Table resolution
+    /// </summary>
+    public sealed class ODataFilterResolutionPolicyMarker<TAttribute> where TAttribute : Attribute
```
MikeStall: >TAttribute [](start = 58, length = 10)

Why is this needed?

src/Microsoft.Azure.WebJobs.Host/Bindings/DefaultResolutionPolicy.cs
```diff
@@ -0,0 +1,29 @@
...
+using Microsoft.Azure.WebJobs.Host.Bindings.Path;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// Resolution policy for { } in  binding templates. 
+    /// The default policy is just a direct substitution for the binding data. 
+    /// Derived policies can enforce formatting / escaping when they do injection. 
+    /// </summary>
+    internal class DefaultResolutionPolicy<TAttribute> : IResolutionPolicy<TAttribute> where TAttribute : Attribute
```
MikeStall: ><TAttribute>  [](start = 42, length = 13)

I'd remove this generic param. 99% of guys don't need it and will ignore it, and the 1% case that needs it can cast.

test/Microsoft.Azure.WebJobs.Host.UnitTests/PublicSurfaceTests.cs
```diff
@@ -126,7 +126,9 @@ public void WebJobsPublicSurface_LimitedToSpecificTypes()
                 "StorageAccountAttribute",
                 "DisableAttribute",
                 "TimeoutAttribute",
-                "TraceLevelAttribute"
+                "TraceLevelAttribute",
+                "IResolutionPolicy",
+                "ResolutionContext"
```
MikeStall: >ResolutionContext [](start = 17, length = 17)

remove? Don't need a new context class anymore

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -31,6 +30,8 @@ internal class AttributeCloner<TAttribute>
         // Optional hook for post-processing the attribute. This is intended for legacy hack rules.
         private readonly Func<TAttribute, Task<TAttribute>> _hook;
 
+        private IResolutionPolicy<TAttribute> _defaultResolutionPolicy;
```
MikeStall: >private IResolutionPolicy<TAttribute> _defaultResolutionPolicy [](start = 8, length = 62)

we can get rid of this field

src/Microsoft.Azure.WebJobs.Host/Bindings/DefaultResolutionPolicy.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        public string TemplateBind(PropertyInfo propInfo, TAttribute resolvedAttribute, BindingTemplate template, IReadOnlyDictionary<string, object> bindingData)
+        {
+            if (bindingData == null)
+            {
+                return template?.Pattern;
+            }
```
MikeStall: I think this null check should be pushed to the caller (rather than to every implementation of TemplateBind).  The current /dev branch has that.

test/Microsoft.Azure.WebJobs.Host.UnitTests/PublicSurfaceTests.cs
brettsam: yeah, i didn't fix those tests -- figured i'd wait till it was all ready (b/c the names kept changing anyway)

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
brettsam: true -- i'll just create it when i need it

src/Microsoft.Azure.WebJobs.Host/Bindings/DefaultResolutionPolicy.cs
brettsam: That means that if bindingData is null, my policy is never called. Maybe someone wants to do something special in that case. I didn't like that it restricted something that I'm trying to handle in a custom way.

src/Microsoft.Azure.WebJobs/ODataFilterResolutionPolicy.cs
brettsam: This is the empty marker class.

src/Microsoft.Azure.WebJobs.Host/Bindings/DefaultResolutionPolicy.cs
brettsam: Seems so weird not to use it when AttributeCloner<TAttribute> already knows the type. But fine.

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -305,47 +307,52 @@ internal TAttribute ResolveFromBindings(IReadOnlyDictionary<string, object> bind
...
+            IResolutionPolicy<TAttribute> policy = null;
+
+            if (autoResolveAttribute != null && autoResolveAttribute.ResolutionPolicyType != null)
             {
-                if (propInfo.DeclaringType == typeof(TableAttribute) && propInfo.Name == "Filter")
+                var ctor = autoResolveAttribute.ResolutionPolicyType.GetConstructor(new Type[] { typeof(IResolutionPolicy<TAttribute>) });
+                if (ctor != null)
                 {
-                    return new TableFilterResolutionPolicy();
+                    policy = (IResolutionPolicy<TAttribute>)Activator.CreateInstance(autoResolveAttribute.ResolutionPolicyType, new object[] { _defaultResolutionPolicy });
```
brettsam: I think I'm going to remove this "nesting" feature to keep the change simpler. None of us need it right now and we can always expand it in the future.

src/Microsoft.Azure.WebJobs.Host/Bindings/DefaultResolutionPolicy.cs
MikeStall: Exactly. Why would somebody do something special in that case? Every chance we give a derived implementation to unnecessarily specialize is just another source of bugs. 

---
In reply to: [99444656](https://github.com/Azure/azure-webjobs-sdk/pull/1000#discussion_r99444656) [](ancestors = 99444656)

src/Microsoft.Azure.WebJobs/TableAttribute.cs
```diff
@@ -91,7 +91,7 @@ public string RowKey
         /// <summary>
         /// Allow arbitrary table filter. RowKey should be null. 
         /// </summary>
-        [AutoResolve]
+        [AutoResolve(ResolutionPolicyType = typeof(ODataFilterResolutionPolicyMarker<TableAttribute>))]
```
MikeStall: ><TableAttribute> [](start = 84, length = 16)

This generic here feels completely out of place. It has no semantic value here; and is only here because it leaked up from the implementation (and the implementation only needs it because of a rare SQL case that no other binding is using)

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -305,47 +304,44 @@ internal TAttribute ResolveFromBindings(IReadOnlyDictionary<string, object> bind
...
+            var formatterType = autoResolveAttribute.ResolutionPolicyType;
 
-            // Expand on [AutoResolve] property to make this pluggable, 
-            public static ResolutionPolicy GetPolicy(PropertyInfo propInfo)
+            if (formatterType != null)
             {
-                if (propInfo.DeclaringType == typeof(TableAttribute) && propInfo.Name == "Filter")
+                // Special-case Table as there is no way to declare this ResolutionPolicy
+                // and use BindingTemplate in the Core assembly
+                if (formatterType.FullName == typeof(ODataFilterResolutionPolicy).FullName)
```
MikeStall: >FullName [](start = 34, length = 8)

do we need the .FullName? 
just do :
  (formatterType == typeof(ODataFilterResolutionPolicy))

src/Microsoft.Azure.WebJobs.Host/Bindings/IResolutionPolicy.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    /// <summary>
+    /// Resolution policy for { } in  binding templates. 
+    /// The default policy is just a direct substitution for the binding data. 
+    /// Derived policies can enforce formatting / escaping when they do injection. 
+    /// </summary>
+    public interface IResolutionPolicy
+    {
+        /// <summary>
+        /// Binding
```
MikeStall: >Binding [](start = 12, length = 7)

nit: fill in comments

src/Microsoft.Azure.WebJobs/ODataFilterResolutionPolicy.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Host.Bindings
+{
+    /// <summary>
+    /// Marker type for Table resolution
+    /// </summary>
+    internal sealed class ODataFilterResolutionPolicy
```
MikeStall: >internal [](start = 4, length = 8)

It'd be fine to make this public so that other attributes could use the same OData resolution logic that Table does .

MikeStall:
:shipit:

src/Microsoft.Azure.WebJobs.Host/Bindings/IResolutionPolicy.cs
brettsam: I'm going to hit tests and comments in the next PR

src/Microsoft.Azure.WebJobs/ODataFilterResolutionPolicy.cs
brettsam: This is the "marker" class. I've actually added "Marker" to the end of it to make that clear. I can make the "real" one public, though.

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
brettsam: I've changed it to `(formatterType == typeof(ODataFilterResolutionPolicyMarker))`. Not sure how I make them use the same type name on both of these without it being confusing.

src/Microsoft.Azure.WebJobs.Host/Bindings/DefaultResolutionPolicy.cs
brettsam: My philosophy is -- they asked to be called during binding resolution so we should call them and let them decide. 

But since this will be so rarely used, I'm not that concerned about it. I'll change it to check null first.

MikeStall:
This would fix  https://github.com/Azure/azure-webjobs-sdk/issues/990

PR #1006, opened 2017-02-02
Add generic IAsyncCollector rule

MikeStall:
This adds a BindToCollector rule that follows a similar pattern to the recent BindToInput rule. 

It leverages the the new OpenType support to collapse the generic and non-generic case into a single rule. 
This  improves a few things:
1. eliminates a bunch of internal reflection code 
2. overall reduction in code. 
3. before, the generic and non-generic rules where separate. This unifies them into a single binding.  

This removes the old BindToGeneric rule so it's a (beta) breaking change similar to BindToInput.


src/Microsoft.Azure.WebJobs.Host/TypeUtility.cs
```diff
@@ -116,5 +118,20 @@ public static bool IsAsyncVoid(MethodInfo methodInfo)
         {
             return IsAsync(methodInfo) && (methodInfo.ReturnType == typeof(void));
         }
+                
+        private static bool HasConverterWorker<TAttribute, TSrc, TDest>(IConverterManager converterManager)
```
brettsam: nit: shouldn't this just be `HasConverter`?

test/Microsoft.Azure.WebJobs.Host.UnitTests/Common/TypedCollectorTests.cs
```diff
@@ -116,7 +116,7 @@ public class FakeQueueTypedClient : IExtensionConfigProvider
         public List<object> _items;
         public string _prefix; // from attribute, to test attribute automatic resolution. 
 
-        public Func<FakeQueueAttribute, Type, bool> Filter;
+        //public Func<FakeQueueAttribute, Type, bool> Filter;
```
brettsam: nit: remove

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
```diff
@@ -104,37 +104,33 @@ public IBindingProvider BindToGenericValueProvider<TAttribute>(Func<TAttribute,
...
         /// <typeparam name="TMessage">'core type' for the IAsyncCollector.</typeparam>
         /// <param name="buildFromAttribute">Function to allocate the collector object given a resolved instance of the attribute.</param>
-        /// <param name="buildParameterDescriptor">An optional function to create a specific ParameterDescriptor object for the dashboard. If missing, a default ParameterDescriptor is created. </param>
-        /// <param name="postResolveHook">An advanced hook for translating the attribute. </param>
         /// <returns>A binding provider that applies these semantics.</returns>
         public IBindingProvider BindToAsyncCollector<TAttribute, TMessage>(
-            Func<TAttribute, IAsyncCollector<TMessage>> buildFromAttribute, 
-            Func<TAttribute, ParameterInfo, INameResolver, ParameterDescriptor> buildParameterDescriptor = null,
-            Func<TAttribute, ParameterInfo, INameResolver, Task<TAttribute>> postResolveHook = null)
+            Func<TAttribute, IAsyncCollector<TMessage>> buildFromAttribute)
```
brettsam: This doesn't match the BindToInput pattern -- I'd expect this to be an `IConverter<Attribute, IAsyncCollector<TMessage>>`.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
```diff
@@ -104,37 +104,33 @@ public IBindingProvider BindToGenericValueProvider<TAttribute>(Func<TAttribute,
...
-        public IBindingProvider BindToGenericAsyncCollector<TAttribute>(
-            Func<TAttribute, Type, object> builder, 
-            Func<TAttribute, Type, bool> filter = null)
+        /// <typeparam name="TType">'core type' for the IAsyncCollector. This can be an OpenType and allow resolving against generics.</typeparam>
+        /// <param name="builderType">
+        /// Type of the builder object. Should expose a conversion from TAttribute to <see cref="IAsyncCollector{TType}"/>
+        /// </param>
+        /// <param name="constructorArgs">Arguments to pass to the constructor for the builderType.</param>
+        /// <returns></returns>
+        public IBindingProvider BindToCollector<TAttribute, TType>(
```
brettsam: I know the underlying type here is an AsyncCollector -- but would calling these methods `BindToOutput` be more general and mirror `BindToInput`?

src/Microsoft.Azure.WebJobs.Host/TypeUtility.cs
MikeStall: it's private. There's a HasConverter below that's the public method.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
MikeStall: This is the existing signature - it's a convenience method for supplying a direct lambda, which is useful for concrete types. The full Converter<> method is below.

brettsam: BindToInput can take either a Type or an IConverter instance.
BindToOutput/Collector can take a Type or a lambda that matches the IConverter signature.

Those APIs should mirror each other. If we're saying that the place to do validation is in your IConverter constructor, we at least need to supply that for BindToOutput/Collector.

brettsam: To add to that -- it's not a part of the review, but can you also change BindToInput take an IConverter<TAttribute, TType> rather than an object? https://github.com/Azure/azure-webjobs-sdk/blob/1345e4e29fd5fc89f72a1afb42a5c02273f079b6/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs#L166

MikeStall: Unfortunately, the problem there is generics. That'd be fine for a concrete type, but not for an OpenType.

brettsam: Can you remind me what the scenario is that breaks that? I thought if you were using generics/OpenType, then you'd use the overload that takes Type.

MikeStall: The problem is that when you add the rule, you can't refer to generic parameters (T), all types must be concrete. So we use a OpenType as a concrete placeholder for a generic parameter T. Ie, you'd add a rule as: 
    IConverter<CloudTable, IQueryable<OpenType>>.  
But your actual implementation uses a generic T instead of the concrete class OpenType.  Nobody uses OpenType in their implementations.

brettsam: Isn't that where you would use the Type overload and pass `typeof(MyConverter<>)`? I'm trying to think through a scenario where you'd want to use OpenType and also had a concrete converter class.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
mathewc: But collectors are inputs the function  that's the sense BindToInput uses the name as I understand it. I think we'll cause confusion if we try to indicate in/out via the naming here. E.g. some bindings may be in/out like Stream.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
```diff
@@ -104,37 +104,33 @@ public IBindingProvider BindToGenericValueProvider<TAttribute>(Func<TAttribute,
...
         /// <typeparam name="TAttribute">Type of binding attribute on the user's parameter.</typeparam>
-        /// <param name="builder">Function that gets passed (Resolved attribute, 'core' collector type) and instantiates an IAsyncCollector</param>
-        /// <param name="filter">Optional. type filter, called once at index time. If missing, assume True.</param>   
-        /// <returns>A binding provider that applies these semantics.</returns>
-        public IBindingProvider BindToGenericAsyncCollector<TAttribute>(
-            Func<TAttribute, Type, object> builder, 
-            Func<TAttribute, Type, bool> filter = null)
+        /// <typeparam name="TType">'core type' for the IAsyncCollector. This can be an OpenType and allow resolving against generics.</typeparam>
+        /// <param name="builderType">
+        /// Type of the builder object. Should expose a conversion from TAttribute to <see cref="IAsyncCollector{TType}"/>
```
mathewc: When you say "should expose a conversion" do we verify this and error? If so then I'd say must.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
```diff
@@ -104,37 +104,33 @@ public IBindingProvider BindToGenericValueProvider<TAttribute>(Func<TAttribute,
...
+        /// Create a binding provider for binding a parameter to an <see cref="IAsyncCollector{TType}"/>. 
         /// </summary>
         /// <typeparam name="TAttribute">Type of binding attribute on the user's parameter.</typeparam>
-        /// <param name="builder">Function that gets passed (Resolved attribute, 'core' collector type) and instantiates an IAsyncCollector</param>
-        /// <param name="filter">Optional. type filter, called once at index time. If missing, assume True.</param>   
-        /// <returns>A binding provider that applies these semantics.</returns>
-        public IBindingProvider BindToGenericAsyncCollector<TAttribute>(
-            Func<TAttribute, Type, object> builder, 
-            Func<TAttribute, Type, bool> filter = null)
+        /// <typeparam name="TType">'core type' for the IAsyncCollector. This can be an OpenType and allow resolving against generics.</typeparam>
```
mathewc: Nit: do you mean "element type", the T in IAsyncCollector<T>? I think "element type" might be easier to understand then

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
```diff
@@ -104,37 +104,33 @@ public IBindingProvider BindToGenericValueProvider<TAttribute>(Func<TAttribute,
...
         public IBindingProvider BindToAsyncCollector<TAttribute, TMessage>(
-            Func<TAttribute, IAsyncCollector<TMessage>> buildFromAttribute, 
-            Func<TAttribute, ParameterInfo, INameResolver, ParameterDescriptor> buildParameterDescriptor = null,
-            Func<TAttribute, ParameterInfo, INameResolver, Task<TAttribute>> postResolveHook = null)
+            Func<TAttribute, IAsyncCollector<TMessage>> buildFromAttribute)
             where TAttribute : Attribute
         {
-            return new AsyncCollectorWithConverterManagerBindingProvider<TAttribute, TMessage>(
-                _nameResolver, _converterManager, buildFromAttribute, buildParameterDescriptor, postResolveHook);
+            var converter = new DelegateAdapterCollectorBuider<TAttribute, TMessage> { BuildFromAttribute = buildFromAttribute };
```
mathewc: Spelling errors in types - "Buider"

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
```diff
@@ -104,37 +104,33 @@ public IBindingProvider BindToGenericValueProvider<TAttribute>(Func<TAttribute,
...
+            params object[] constructorArgs)
             where TAttribute : Attribute
         {
-            if (builder == null)
-            {
-                throw new ArgumentNullException("builder");
-            }
-
-            return new GenericAsyncCollectorBindingProvider<TAttribute>(this._nameResolver, builder, filter);
+            var pm = PatternMatcher.New(builderType, constructorArgs);
```
mathewc: What is a "PatternMatcher" - this type probably needs a better name

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/AsyncCollectorBindingProvider.cs
```diff
@@ -0,0 +1,303 @@
...
+using System.Collections;
+using System.Reflection;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Protocols;
+
+namespace Microsoft.Azure.WebJobs.Host.Bindings
+{
+    // General rule for binding parameters to an AsyncCollector. 
+    // Supports the various flavors like IAsyncCollector, ICollector, out T, out T[]. 
+    internal class AsyncCollectorBindingProvider<TAttribute, TType> : FluidBindingProvider<TAttribute>, IBindingProvider
```
mathewc: Can you rename FluidBindingProvider? What makes it fluid?

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/AsyncCollectorBindingProvider.cs
```diff
@@ -0,0 +1,303 @@
...
+                }
+
+                if (elementType.IsArray)
+                {
+                    // out T[]
+                    var messageType = elementType.GetElementType();
+                    return new BindingMode(Mode.OutArray, messageType);
+                }
+
+                var checker = ConverterManager.GetTypeValidator<TType>();
```
mathewc: nit: var validator?

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/AsyncCollectorBindingProvider.cs
```diff
@@ -0,0 +1,303 @@
...
+                    throw new InvalidOperationException("Object element types are not supported.");
+                }
+            }
+
+            // No match. Let another rule claim it
+            return null;            
+        }
+
+        // Represent the different possible flavors for binding to an async collector
+        private class BindingMode
```
mathewc: You already have a "Mode" enum defined above which is a binding mode. Suggest you rename this class to something else to reduce confusion.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
brettsam: Are you objecting to `BindToInput` as well? I agree that our classifications of bindings makes the naming a little hard to comprehend:

1. To bind to (what we call) "outputs", you give us an IAsyncCollector... which does all the mappings behind-the-scenes.
2. To bind to input-only, you use BindToInput. This is kinda what BindToOutput already does, just specialized.
3. We don't really have an in/out story -- you need to bind to ValueProvider.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/AsyncCollectorBindingProvider.cs
```diff
@@ -0,0 +1,303 @@
...
+                this.Mode = mode;
+                this.ElementType = elementType;
+            }
+            public Mode Mode { get; set; }
+            public Type ElementType { get; set; }
+        }
+
+        // TType - specified in the rule. 
+        // TMessage - extracted from the user's parameter. 
+        private class ExactBinding<TMessage> : BindingBase<TAttribute>
```
mathewc: I don't think TMessage makes sense here? What is "message"?

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/AsyncCollectorBindingProvider.cs
```diff
@@ -0,0 +1,303 @@
...
+        {
+            private readonly Func<object, object> _buildFromAttribute;
+
+            private readonly FuncConverter<TMessage, TAttribute, TType> _converter;
```
mathewc: I think we should rename this FuncConverter type, unless it really is converting a Func. I think it is a converter from one type to another though, so I don't think it should have func in its name?

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/AsyncCollectorBindingProvider.cs
```diff
@@ -0,0 +1,303 @@
...
+                this._mode = mode;
+                this._converter = converter;
+            }
+
+            public static ExactBinding<TMessage> TryBuild(
+                AsyncCollectorBindingProvider<TAttribute, TType> parent,
+                Mode mode,
+                BindingProviderContext context)
+            {
+                var cm = parent._converterManager;
```
mathewc: nit: can you name this better, and also maybe move it down to where it is used?

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/AsyncCollectorBindingProvider.cs
```diff
@@ -0,0 +1,303 @@
...
+                var parameter = context.Parameter;
+                TAttribute attributeSource = parameter.GetCustomAttribute<TAttribute>(inherit: false);
+
+                Func<TAttribute, Task<TAttribute>> hookWrapper = null;
+                if (parent.PostResolveHook != null)
+                {
+                    hookWrapper = (attrResolved) => parent.PostResolveHook(attrResolved, parameter, parent._nameResolver);
+                }
+
+                var cloner = new AttributeCloner<TAttribute>(attributeSource, context.BindingDataContract, parent._nameResolver, hookWrapper);
```
mathewc: can you move this down to where it is needed/used?

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/AsyncCollectorBindingProvider.cs
```diff
@@ -0,0 +1,303 @@
...
+                    param = parent.BuildParameterDescriptor(attributeSource, parameter, parent._nameResolver);
+                }
+                else
+                {
+                    param = new ParameterDescriptor
+                    {
+                        Name = parameter.Name,
+                        DisplayHints = new ParameterDisplayHints
+                        {
+                            Description = "input"
```
mathewc: Where does this "input" show up in the dashboard? Other places (e.g. BindCollector) we're using "output". What should it be?

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/AsyncCollectorBindingProvider.cs
```diff
@@ -0,0 +1,303 @@
...
+            {
+                string invokeString = Cloner.GetInvokeString(attrResolved);
+
+                object obj = _buildFromAttribute(attrResolved);
+                
+                IAsyncCollector<TMessage> collector;
+                if (_converter != null)
+                {
+                    // Apply a converter
+                    var collector2 = (IAsyncCollector<TType>)obj;
```
mathewc: nit: use a better variable name. E.g. originalCollector, baseCollector or something

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/AsyncCollectorBindingProvider.cs
```diff
@@ -0,0 +1,303 @@
...
+                }
+
+                var vp = CoerceValueProvider(_mode, invokeString, collector);
+                return Task.FromResult(vp);
+            }
+
+            // Get a ValueProvider that's in the right mode. 
+            private static IValueProvider CoerceValueProvider(Mode mode, string invokeString, IAsyncCollector<TMessage> collector)
+            {
+                IValueProvider vp;
```
mathewc: nit: remove unecessary local and just return below?

src/Microsoft.Azure.WebJobs.Host/TypeUtility.cs
mathewc: I agree the naming is a little strange. These methods don't seem like they belong on TypeUtility at all. I think they should be non-generic extension methods to IConverterManager.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Common/BindingFactoryHelperTests.cs
```diff
@@ -1,38 +0,0 @@
...
-using Microsoft.Azure.WebJobs.Host.Config;
-using Microsoft.Azure.WebJobs.Host.TestCommon;
-using Xunit;
-using System.Threading.Tasks;
-using System.Reflection;
-using Microsoft.Azure.WebJobs.Host.Bindings;
-
-namespace Microsoft.Azure.WebJobs.Host.UnitTests.Common
-{
-    public class BindingFactoryHelperTests
```
mathewc: I see deleted tests, but what about new test coverage for all the new code you're adding?

src/Microsoft.Azure.WebJobs.Host/TypeUtility.cs
MikeStall: I've moved to extensions methods for IConverterManager. 

---
In reply to: [99681763](https://github.com/Azure/azure-webjobs-sdk/pull/1006#discussion_r99681763) [](ancestors = 99681763)

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
MikeStall: It's an existing internal type so I'll defer this. It finds the IConverter<> overload on the builder.

---
In reply to: [99673891](https://github.com/Azure/azure-webjobs-sdk/pull/1006#discussion_r99673891) [](ancestors = 99673891)

test/Microsoft.Azure.WebJobs.Host.UnitTests/Common/BindingFactoryHelperTests.cs
MikeStall: It's in TypedCollectorTests.cs

---
In reply to: [99682130](https://github.com/Azure/azure-webjobs-sdk/pull/1006#discussion_r99682130) [](ancestors = 99682130)

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/AsyncCollectorBindingProvider.cs
MikeStall: It should be "Fluent", referring to the fluent design pattern for APIs. ([https://en.wikipedia.org/wiki/Fluent_interface](https://en.wikipedia.org/wiki/Fluent_interface))  . 


---
In reply to: [99674533](https://github.com/Azure/azure-webjobs-sdk/pull/1006#discussion_r99674533) [](ancestors = 99674533)

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/AsyncCollectorBindingProvider.cs
MikeStall: I'll go with CollectorBindingPattern

---
In reply to: [99675591](https://github.com/Azure/azure-webjobs-sdk/pull/1006#discussion_r99675591) [](ancestors = 99675591)

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
MikeStall: I'd like to keep with BindToCollector:
1. There can be other output bindings (like writeable streams or ValueProvider)
2. I want people thinking in terms of the collector interface

---
In reply to: [99675747](https://github.com/Azure/azure-webjobs-sdk/pull/1006#discussion_r99675747) [](ancestors = 99675747)

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
MikeStall: I think you're right. 

---
In reply to: [99667749](https://github.com/Azure/azure-webjobs-sdk/pull/1006#discussion_r99667749) [](ancestors = 99667749)

MikeStall: I'll keep the existing sig, but rename to BindToCollector so that it's just a convenience overload. And I'll add the `IConverter<Attribute, IAsyncCollector<TMessage>>`. overload to be consistent. 

---
In reply to: [99659831](https://github.com/Azure/azure-webjobs-sdk/pull/1006#discussion_r99659831) [](ancestors = 99659831)

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
MikeStall: Per Brett's suggestion, I'll add an instance overload so that we get static checking in the common case. 

---
In reply to: [99673251](https://github.com/Azure/azure-webjobs-sdk/pull/1006#discussion_r99673251) [](ancestors = 99673251)

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/AsyncCollectorBindingProvider.cs
mathewc: This is marked as pending, but I see no change?

MikeStall: I think this rename is out of scope. It's a pre-existing type that's orthogonal to this change.

PR #1011, opened 2017-02-08
updating nuget.exe download version

brettsam:


brettsam:
The latest batch of nugets require a higher version, which we need to get from the new(er) official drop location. We already do this in Script.

I'm going to commit this so I can move forward. Feel free to comment.

PR #1012, opened 2017-02-09
Fix bug in resolving generic arrays.

MikeStall:


mathewc:
I assume this needs to go out in the current batch of changes?

src/Microsoft.Azure.WebJobs.Host/Bindings/PatternMatcher.cs
```diff
@@ -217,6 +224,21 @@ private static bool CheckArg(Type openType, Type specificType, Dictionary<string
                 return AddGenericArg(genericArgs, name, specificType);
             }
 
+            // T[]
+            if (openType.IsArray && specificType.IsArray)
+            {
+                var elementType = openType.GetElementType();
+                var specificElementType = specificType.GetElementType();
```
mathewc: nit: move this into if block

MikeStall:
Merged into dev

src/Microsoft.Azure.WebJobs.Host/Bindings/PatternMatcher.cs
MikeStall: done

PR #1020, opened 2017-02-14
Applying Obsolete to unbaked extensibility interfaces

mathewc:
Applying Obsolete attribute to all the new extensibility surface area that is not yet baked. This unblocks us to release the 2.0 version of the SDK.

The full diff between version 1.1.2 (latest stable) and the current codebase is here: https://github.com/Azure/azure-webjobs-sdk/compare/9091e37e00d6e610...f877251dd5a9488d. Particularly interesting in that diff is the diff on PublicSurfaceTests.cs which is our static analysis test for showing all changes to public surface area. I used this to determine the set of APIs to mark Obsolete.


mathewc:
Guys, what do you think? @MikeStall @brettsam

MikeStall:
I think these are fine. I'd also include:
- Binder class should be internal (use IBinder interface instead) 
- IOrderedValueBinder, BindStepOrder (per that other bug, I think this needs to go away completely for extensions) 
- IExtensionRegistry, IExtensionConfigProvider




---
In reply to: [281573779](https://github.com/Azure/azure-webjobs-sdk/pull/1020#issuecomment-281573779) [](ancestors = 281573779)

MikeStall:
ICloudBlobStreamBinder? 

---
In reply to: [281736233](https://github.com/Azure/azure-webjobs-sdk/pull/1020#issuecomment-281736233) [](ancestors = 281736233,281573779)

mathewc:
We can't include any of the additional things you listed. Binder is used by Functions - that would be breaking. All those other interfaces are preexisting interfaces that are used by people. Our goal is not to reevaluate all those now. I'm focused on the NEW surface area we added only.

MikeStall:
These are newly exposed since 1.0? 
Script can still call obsolete stuff, right? 


---
In reply to: [281793314](https://github.com/Azure/azure-webjobs-sdk/pull/1020#issuecomment-281793314) [](ancestors = 281793314)

mathewc:
Let me clarify: The goal of this work is only to mark as obsolete new unbaked work that has been added **since the last public release 1.1.2**. We're not going and reviewing/obsoleting anything further in the past. The bar is that we want to prevent people from taking new dependencies on new unbaked stuff, but we don't want to arbitrarily break people w/o viable replacements. 

Regarding Binder - not only is that used internally in functions, it is also used by user function code. It can't be removed.

MikeStall:
Binder class should not be used by user code - they should just use the interface. That's exactly the sort of public-surface thing we should fix before promoting new surface area to a major release. 

---
In reply to: [281823617](https://github.com/Azure/azure-webjobs-sdk/pull/1020#issuecomment-281823617) [](ancestors = 281823617)

MikeStall:
:shipit:

mathewc:
Merged

PR #1027, opened 2017-02-15
Fix pattern matcher for BindingRules

mamaso:
PatternMatcher would throw (from `specificType.GetGenericTypeDefinition`) if Type -> GenericType IConverter was defined when it should return false and move to evaluate the next IConverter.

Can occur if BindingRule IConverter and general IConverter are on the same class.

### DIFF
--- a/test/Microsoft.Azure.WebJobs.Host.UnitTests/ConverterManagerTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.UnitTests/ConverterManagerTests.cs
@@ -514,6 +514,32 @@ namespace Microsoft.Azure.WebJobs.Host.UnitTests
                 PatternMatcher.ResolveGenerics(int1, genArgs));
         }
 
+        public class TestConverter :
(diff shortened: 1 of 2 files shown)


test/Microsoft.Azure.WebJobs.Host.UnitTests/ConverterManagerTests.cs
```diff
@@ -514,6 +514,28 @@ public void TestPatternMatcher_ResolveGenerics()
...
+                return null;
+            }
+        }
+
+        [Fact]
+        public void PatternMatcher_Succeeds_WhenBindingRuleConverterExists()
+        {
+            var pm = PatternMatcher.New(typeof(TestConverter));
+            var func = pm.TryGetConverterFunc(typeof(string), typeof(byte[]));
+        }
```
MikeStall: can you add a call for the other rule too: 
   pm.TryGetConverterFunc(typeof(Attribute), typeof(IAsyncCollector<string>); 
and assert they're non-null.

Issue #921, opened 2016-11-22
Need to drive consistency for IAsyncCollector<T> batching across bindings


mathewc:
See http://stackoverflow.com/questions/40732715/azure-functions-icollectort-vs-iasynccollectort

We're inconsistent across bindings and extensions in our IAsyncCollector implementations, and whether they batch or add immediately. For example:

- SendGrid : Batches and sends on Flush
- Twilio: Batches
- EventHub: Batches and flushes automatically based on a configurable BatchSize
- ApiHub: Doesn't batch
- DocumentDB: Doesn't batch
- NotificationHub: Doesn't batch
- MobileTable: Doesn't batch
- Queue: Doesn't batch
- ServiceBus: Doesn't batch
-

MikeStall:
We identified 2 canonical scenarios for why this is important:

1. Error handling - Are events sent if there's an error in the function ?
```
void Error([Queue] IAsyncCollector<Payload> q) 
{
                q.AddAsync(…);
                throw Exception(); // Is message queued?
}
```

2. Ordering - important when an event refers to another output. IE, we write a blob, and then enqueue a message that refers to that blob. 

```
Work([Queue] IAsyncCollector<Payload> q, [Blob] TextWriter output) 
{
                Output.WriteLine(…);
                var msg = …  // refers to output blob!
                q.AddAsync(msg);
}
```

What’s even more sinister is that this can lead to non-determinisim. (I believe solving these non-determinism issues is also a huge-value add for WebJobs/Functions). There’s a race, maybe the blob is written first during testing; and the order is different in production.  

This is easy when if we only send a small number of rows via IAsyncCollector because we could just buffer in -memory and force an ordering. But it’s prohibitive if we send huge payloads.

lindydonna:
Putting in 2.0.0-release since it's a breaking change

waldekmastykarz:
FWIW, I noticed the same issue with queues and tables, where messages would be added to the queue  directly but rows in the table only after the function completed. As a workaround I switched to adding rows to tables manually rather than using outputs but it would definitely be better if we could use outputs for this kind of scenarios.

MikeStall:
A proposed solution is:
1. the first N (1000?) events get buffered and flushed at the end.  A function could emit an unbounded number of events (ie, run for 6 hours, emit 5 million events), so we have to have a limit N somewhere. But at that scale, the author knows what they're dealing with. 
[...]

PR #1032, opened 2017-02-17
Provide consistent buffering semantics for IAsyncCollector across all…

MikeStall:
… extension types.

Fix https://github.com/Azure/azure-webjobs-sdk/issues/921

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/AsyncCollector/BufferedIAsyncCollector.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/AsyncCollector/BufferedIAsyncCollector.cs
@@ -1,0 +1,67 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Host.Bindings
+{
+    /// <summary>
+    /// Buffer up to the first N items. 
+    /// If Flush() is called, stop buffering. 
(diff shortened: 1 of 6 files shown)


mamaso:
Tricky problem, I don't like the arbitrary in-memory buffering. However, if the user is adding massive objects I suppose they can flush in order to turn off buffering.

Why do we revert to the extension collector buffering after the first flush is called? I'd prefer to maintain the same buffering behavior to make things easier to reason about.

MikeStall:
@mamaso  - the idea here is to provide consistent default behavior, which I expect will cover the majority of cases. Once they call Flush, the user is explicitly assuming control of the flush semantics.   Put another way, this gives them a clean way to "turn off" our automatic buffering.

mathewc:
What should we do with this PR - it's half a year old? I don't think we have design consensus on this yet. Should we close this for now?

mathewc:
Didn't mean to close this - meant to ask Mike if he thought we should close this.

MikeStall:
I think we should leave this open, discuss the design, and make decision.

fabiocav:
Bringing this up again. Should we just open a design issue for discussion and close this?

MikeStall:
We already have an issue opened for this (https://github.com/Azure/azure-webjobs-sdk/issues/921), but any solution is technically breaking so we held off in v1. This PR is my proposal for solving it. Basically, we buffer the items.

mathewc:
This PR is a year old. Can we close it?

MikeStall:
@mathewc  - we should discuss.  The issue is:
1. given a specific binding type, want a consistent output flushing semantics between C#, CSX, and Node
2. want consistent output flushing semantics across different binding types.

fabiocav:
Closing this for now. If we intend to continue working on this, let's reopen.

PR #1033, opened 2017-02-18
Make MaxPollingInterval configurable for each queue processor

### LINKED ISSUE #1030: Make MaxPollingInterval configurable for each queue processor
In some cases there is a need to poll queues with different latencies (e.g. chat queue with low latency or image processing queue with moderate latency). In such cases one global MaxPollingInterval makes it impossible to maintain only the needed amount of polling, i.e. it leads to increased storage costs.

Dia6lo:
Fixes #1030

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Queues/Listeners/HostMessageListenerFactory.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Queues/Listeners/HostMessageListenerFactory.cs
@@ -15,9 +15,6 @@ namespace Microsoft.Azure.WebJobs.Host.Queues.Listeners
 {
     internal class HostMessageListenerFactory : IListenerFactory
     {
-        private static readonly TimeSpan Minimum = QueuePollingIntervals.Minimum;
-        private static readonly TimeSpan DefaultMaximum = QueuePollingIntervals.DefaultMaximum;
-
         private readonly IStorageQueue _queue;
         private readonly IQueueConfiguration _queueConfiguration;
         private readonly IWebJobsExceptionHandler _exceptionHandler;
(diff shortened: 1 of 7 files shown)


src/Microsoft.Azure.WebJobs.Host/Queues/Listeners/HostMessageListenerFactory.cs
```diff
@@ -83,16 +80,9 @@ public Task<IListener> CreateAsync(CancellationToken cancellationToken)
         {
             ITriggerExecutor<IStorageQueueMessage> triggerExecutor = new HostMessageExecutor(_executor, _functionLookup, _functionInstanceLogger);
 
-            TimeSpan configuredMaximum = _queueConfiguration.MaxPollingInterval;
-            // Provide an upper bound on the maximum polling interval for run/abort from dashboard.
```
mathewc: So you're removing this upper bound here. Previously we were bounding this. Might not be an issue - what was your thought here?

mathewc:
Merged, thanks :)

src/Microsoft.Azure.WebJobs.Host/Queues/Listeners/HostMessageListenerFactory.cs
Dia6lo: Yeah, I see now the point of this maximum - without it when we set high MaxPollingInterval, Dashboard become less responsive. I think that complete removal of IDelayStrategy injection is a mistake. We can bring it back and create new strategy only if none was passed. Second option (and I think the best option) is to pass custom QueueProcessor in this case. Another option, that would be great too (for extensibility), is adding custom IDelayStrategy to QueueProcessor itself and fetch it in QueueListener constructor (but it's sort of completely separate feature).

Dia6lo: Anyway, one of these options should be applied, or there will be troubles in the future.

mathewc: Fixed

PR #1035, opened 2017-02-22
Propagate trigger listener exceptions

### LINKED ISSUE #996: Allow listener startup failures to be non fatal
Currently when the JobHost starts up, all listeners are also started, and if any listeners fail (e.g. due to an invalid ServiceBus connection string, etc.) the exception will bubble up and bring down the host.

We should consider allowing the host to have a mode where this is not the case. I mention a mode, because likely we want the default to remain the way it is. However, for advanced scenarios (e.g. Azure Functions) it is useful to be able to treat individual functions as more isolated units and allow the host to start even if one or more function listeners fail to start. We'd log the errors well, but allow functions non in error to start.

mamaso:
resolves #996 

Wrap function trigger listeners to catch errors and trace FunctionIndexingExceptions (which will be picked up by script). If handled, swallow, otherwise throw.

Also changed the `CreateTriggeredFunctionDefinition` method to be a bit less unwieldy (static -> instance and removed intermediate method to lessen parameter passing)

**REQUIRES CHANGES IN SCRIPT TO HANDLE NEW EXCEPTIONS**

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Exceptions/RecoverableException.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Exceptions/RecoverableException.cs
@@ -1,0 +1,88 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Runtime.Serialization;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// A recoverable exception, i.e. can be handled.
+    /// </summary>
+    [Serializable]
+    public class RecoverableException : Exception
+    {
+        /// <inheritdoc/>
+        public RecoverableException() : base()
+        {
+        }
+
+        /// <inheritdoc/>
+        public RecoverableException(string message) : base(message)
+        {
+        }
+
+        /// <summary>
+        /// Constructs a new instance.
+        /// </summary>
+        /// <param name="message">The exception message.</param>
+        /// <param name="innerException">The inner exception.</param>
+        public RecoverableException(string message, Exception innerException)
+            : base(message, innerException)
+        {
+        }
+
+        /// <summary>
+        /// Constructs a new instance.
+        /// </summary>
+        /// <param name="info">The <see cref="SerializationInfo"/>.</param>
+        /// <param name="context">The <see cref="StreamingContext"/>.</param>
+        protected RecoverableException(SerializationInfo info, StreamingContext context) : base(info, context)
+        {
+            if (info == null)
+            {
+                throw new ArgumentNullException("info");
+            }
+
+            Handled = bool.Parse(info.GetString("Handled"));
+        }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether the exception should be treated
+        /// as handled.
+        /// </summary>
+        public bool Handled { get; set; }
+
+        /// <inheritdoc/>
+        public override void GetObjectData(SerializationInfo info, StreamingContext context)
+        {
+            if (info == null)
+            {
+                throw new ArgumentNullException("info");
+            }
+
+            info.AddValue("Handled", this.Handled);
+
+            base.GetObjectData(info, context);
+        }
+
+        /// <summary>
+        /// Tries to recover by propagating exception through trace pipeline.
+        /// Recovers if exception is handled by trace pipeline, else throws.
+        /// </summary>
+        /// <param name="trace"></param>
+        public void TryRecover(TraceWriter trace)
+        {
+            if (trace == null)
+            {
+                throw this;
+            }
+
+            trace.Error(Message, this);
+            if (!Handled)
+            {
+                throw this;
+            }
+        }
+    }
+}
--- a/src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
@@ -85,20 +85,10 @@ namespace Microsoft.Azure.WebJobs.Host.Indexers
                 }
                 catch (FunctionIndexingException fex)
                 {
-                    // Route the indexing exception through the TraceWriter pipeline
-                    _trace.Error(fex.Message, fex);
-
-                    // If the error has been marked as handled, ignore the indexing exception
-                    // and continue on to the rest of the methods. In this case the method in
-                    // error simply won't be running in the JobHost.
-                    if (fex.Handled)
-                    {
-                        continue;
-                    }
-                    else
-                    {
-                        throw;
-                    }
+                    fex.TryRecover(_trace);
+                    // If recoverable, continue to the rest of the methods.
+                    // The method in error simply won't be running in the JobHost.
+                    continue;
                 }
             }
         }
@@ -284,7 +274,9 @@ namespace Microsoft.Azure.WebJobs.Host.Indexers
 
             if (triggerBinding != null)
             {
-                functionDefinition = CreateTriggeredFunctionDefinition(triggerBinding, triggerParameterName, _executor, functionDescriptor, nonTriggerBindings, invoker, _singletonManager);
+                Type triggerValueType = triggerBinding.TriggerValueType;
+                var methodInfo = typeof(FunctionIndexer).GetMethod("CreateTriggeredFunctionDefinition", BindingFlags.Instance | BindingFlags.NonPublic).MakeGenericMethod(triggerValueType);
+                functionDefinition = (FunctionDefinition)methodInfo.Invoke(null, new object[] { triggerBinding, triggerParameterName, functionDescriptor, nonTriggerBindings, invoker });
 
                 if (hasNoAutomaticTriggerAttribute && functionDefinition != null)
                 {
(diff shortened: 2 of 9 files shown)


src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
```diff
@@ -351,23 +343,72 @@ private bool HasJobAttribute(CustomAttributeData attributeData)
...
+
+                public async Task StartAsync(CancellationToken cancellationToken)
+                {
+                    try
+                    {
+                        await _listener.StartAsync(cancellationToken);
+                    }
+                    catch (Exception e)
+                    {
+                        var fex = new FunctionIndexingException(_descriptor.ShortName, e);
```
MikeStall: >FunctionIndexingException [](start = 38, length = 25)

Double checking if this is the right exception type. "Indexing" seems like a static operation (ie, something that could result in squiggles in the VS IDE), whereas the listeners startup could be any runtime operation (bad connection string, network down, etc);

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
```diff
@@ -351,23 +343,72 @@ private bool HasJobAttribute(CustomAttributeData attributeData)
...
+                        _trace.Error(e.Message, fex);
+                        if (!fex.Handled)
+                        {
+                            throw;
+                        }
+                    }
+                }
+
+                public Task StopAsync(CancellationToken cancellationToken)
+                {
```
MikeStall: If Start() threw an exception, is it safe to call Stop() ?

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
```diff
@@ -351,23 +343,72 @@ private bool HasJobAttribute(CustomAttributeData attributeData)
...
-            public Task<IListener> CreateAsync(CancellationToken cancellationToken)
+            public async Task<IListener> CreateAsync(CancellationToken cancellationToken)
             {
                 ListenerFactoryContext context = new ListenerFactoryContext(_descriptor, _executor, cancellationToken);
-                return _binding.CreateListenerAsync(context);
+                IListener listener = await _binding.CreateListenerAsync(context);
+                return new TriggerListener(listener, _descriptor, _trace);
+            }
+
+            private class TriggerListener : IListener
```
MikeStall: >TriggerListener [](start = 26, length = 15)

1. There's a lot of listeners scattered throughout the code.  Is this the only one that needs to be wrapped?
2. Would it be simpler to try/catch the place that calls listener.start() instead of creating a wrapper?

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
```diff
@@ -284,7 +284,9 @@ private bool HasJobAttribute(CustomAttributeData attributeData)
 
             if (triggerBinding != null)
             {
-                functionDefinition = CreateTriggeredFunctionDefinition(triggerBinding, triggerParameterName, _executor, functionDescriptor, nonTriggerBindings, invoker, _singletonManager);
+                Type triggerValueType = triggerBinding.TriggerValueType;
+                MethodInfo createTriggeredFunctionDefinitionMethodInfo = typeof(FunctionIndexer).GetMethod("CreateTriggeredFunctionDefinitionImpl", BindingFlags.Instance | BindingFlags.NonPublic).MakeGenericMethod(triggerValueType);
```
mathewc: nit: var methodInfo =

mathewc: Might as well rename CreateTriggeredFunctionDefinitionImpl by removing the "Impl" suffix?

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
mathewc: I had asked the same question before, but now I can't remember the justification Matt gave on why this was better.

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
mathewc: Yeah, I initially suggested that he piggyback on FEX since we already use that for similar purposes. I did also suggest the possibility of creating a base class that has the "handle" capability, e.g. HandleableException, with our existing FunctionIndexingException and a new FunctionListenerException deriving from it. Not sure the reuse of FEX is that big of an issue really - I can see thinking of listener startup as an index time thing in a loose way.

However, since I had these thoughts before, and Mike did also, lets just put the base class in place.

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
mathewc: If you look at all our other listeners, I don't think we're very good about this in general. If we think we have improvements in this area to make to listeners, I'd say its a new bug. Traditionally from the core SDK point of view, listener startup exceptions are intended to bring down the entire host and process. In that case, stop would never be called.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Listeners/ListenerFactoryTests.cs
```diff
@@ -0,0 +1,79 @@
...
+using Xunit;
+using System.Collections.ObjectModel;
+
+namespace Microsoft.Azure.WebJobs.Host.UnitTests.Listeners
+{
+    public class ListenerFactoryTests
+    {
+        CancellationToken ct = default(CancellationToken);
+
+        async Task<IListener> GetListener(TraceWriter trace)
```
mathewc: nit: here and below, can you put explicit access modifiers on your functions? "private "

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
mamaso: Sounds good, it makes sense.  I piggybacked on FEX in order to avoid any required changes in script, but this will make things more flexible so it's probably worthwhile.

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
mamaso: I'm ensuring that only listeners associated with function triggers could be handled, as other listeners might have unintended side effects. Also, we only have information for the function metadata (name, etc) for these listeners, so we can throw errors that will be routed to the errant function.

Also, it ensures that I don't need to do any hacking with CompositeListener to pass aggregate exceptions.

I'll do a pass over the other listeners to determine if they need to be added.

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
mamaso: Yep. 

Question @mathewc @MikeStall : do we have host startup perf tests in place?

 Changing this method to instance instead of static reduces a lot of confusing clutter, but it may have been there to avoid some perf hit related to invoking a generic method (only reason I can imagine it was coded as it was)

test/Microsoft.Azure.WebJobs.Host.UnitTests/Listeners/ListenerFactoryTests.cs
mamaso: Out of curiosity, any particular reasoning? Agree with prod code, for tests I don't use as much care as I maybe should. Speed up test indexing perhaps?

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
mathewc: No we don't have perf tests. I don't see a perf issue here.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Listeners/ListenerFactoryTests.cs
mathewc: No, it's just style. Were you getting style cop warnings for this? I thought there were checks for this.

mamaso: Perhaps they don't run for the unit test project?

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
mamaso: @mathewc @MikeStall Did some investigation, there are no trigger-related listeners that exist outside this context. I want to limit this in scope to only function triggers, as they are the errors that we can recover from & that are explicitly tied to a method

src/Microsoft.Azure.WebJobs.Host/Exceptions/RecoverableException.cs
```diff
@@ -0,0 +1,61 @@
...
+using System;
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.Serialization;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// A recoverable exception, i.e. can be handled.
+    /// </summary>
+    [SuppressMessage("Microsoft.Design", "CA1032:ImplementStandardExceptionConstructors")]
```
mathewc: why this exclusion? it seems like you've implemented everything you need. Note that our other exceptions are following the design guidelines

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
mathewc: Sounds ok to me

src/Microsoft.Azure.WebJobs.Host/Exceptions/JobMethodException.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Runtime.Serialization;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// Exception that is tied to a specific job method
+    /// </summary>
+    [SuppressMessage("Microsoft.Design", "CA1032:ImplementStandardExceptionConstructors")]
+    [Serializable]
+    public class JobMethodException : RecoverableException
```
mathewc: Can you rename this FunctionListenerException?

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
```diff
@@ -351,23 +343,74 @@ private bool HasJobAttribute(CustomAttributeData attributeData)
...
+                public async Task StartAsync(CancellationToken cancellationToken)
+                {
+                    try
+                    {
+                        await _listener.StartAsync(cancellationToken);
+                    }
+                    catch (Exception e)
+                    {
+                        var name = _descriptor.ShortName;
+                        var message = $"The trigger for function '{name}' was unable to start.";
```
mathewc: "The listener for function ..."

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
```diff
@@ -351,23 +343,74 @@ private bool HasJobAttribute(CustomAttributeData attributeData)
...
+                        var exc = new JobMethodException(name, message, e);
+                        _trace.Error(e.Message, exc);
+                        if (!exc.Handled)
+                        {
+                            throw;
```
mathewc: Maybe you should throw your new exception here?

src/Microsoft.Azure.WebJobs.Host/Exceptions/JobMethodException.cs
mathewc: Can you rename this FunctionException? That naming makes sense with the derived FunctionIndexingException and FunctionInvocationException. The latter should also probably derive from your new base class?

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
mamaso: It's only used for trigger listeners, and I don't think that users of the webjobs sdk have a great concept of listeners. I'd rather be more informative and point to 'oh I have a problem with this method trigger'

src/Microsoft.Azure.WebJobs.Host/Exceptions/RecoverableException.cs
mamaso: Ah copypasta, might not be necessary

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
mamaso: agreed

src/Microsoft.Azure.WebJobs.Host/Exceptions/JobMethodException.cs
mamaso: Sure, considered `FunctionException` but it seemed too generic

src/Microsoft.Azure.WebJobs.Host/Exceptions/RecoverableException.cs
mamaso: Other exceptions are not following the guidlines (in particular `FunctionIndexingException`)

mamaso: Also, I don't think that following the guidelines is good in this case: for a function indexing exception, you'll always have access to and want a method name.

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
mathewc: Yeah they do know about listeners - I think using listener would be clearer. Listeners start - when you talk about a trigger starting it's not really clear - sounds like you're talking about a particular function invocation

src/Microsoft.Azure.WebJobs.Host/Exceptions/RecoverableException.cs
mathewc: All right - I wasn't aware. For the exceptions I added, I followed all the rules :)

mathewc:
Shipit!

PR #1037, opened 2017-02-23
Add conversion from IEnumerable types to JArray

mamaso:
@MikeStall @brettsam 

Initially I was going to add an `IEnumerable<JObject> -> JArray` converter [here](https://github.com/Azure/azure-webjobs-sdk/blob/78608688a0a1a0b9b270dd5b16becd60f1721cf2/src/Microsoft.Azure.WebJobs.Host/Bindings/ConverterManager.cs#L27) in the `ConverterManager` constructor.

However, this approach will take any IEnumerable implementor and convert it to JObject, so it's more flexible. Feels like more of a hack because it's not conforming to the `IConverter` interface, but it'll support script better.

I still think there's some value to explore generic matching:
i.e. `ConcreteType<OpenType>` should match `ConcreteType<SpecificType>`


src/Microsoft.Azure.WebJobs.Host/Bindings/ConverterManager.cs
```diff
@@ -226,6 +226,11 @@ private static OpenType GetOpenType(Type t)
                 }
             }
 
+            if (typeDest == typeof(JArray) && typeof(IEnumerable).IsAssignableFrom(typeSource))
```
MikeStall: >typeof(IEnumerable) [](start = 46, length = 19)

This may be too general. See checks at line 290 below.  For example, this may pull in System.String. 
For now, we should restrict this to IEnumebale<T> where T = JObject or JToken.  We can always make more general later.

mamaso: Ok, good point on too general (we don't want string).

However, I think T = JObject or JToken is too restrictive. If a user defines a builder to IEnumerable<CustomType> I think we should convert that without the extra JObject/JToken constraint.

MikeStall: I think in that case, it involves chaining & composing converters. If the user defines a T-->S converter, and we can bind to IEnumerable<T>, then we could also bind to IEnumerable<S>. , but that may be invoking the converter. 
But this can get tricky quick - so I'm a little shy to get into that unless we have a clear case to guide us. 




---
In reply to: [102849948](https://github.com/Azure/azure-webjobs-sdk/pull/1037#discussion_r102849948) [](ancestors = 102849948)

mamaso: T->S and its usage to convert enumerable types makes sense, but I agree that there's a lot of complexity there.

I'm proposing that we support Type -> JObject and IEnumerable<Type> -> JArray out of box with the FromObject methods. It will light up the script functionality out of the box without the extension author needing to worry about it, and if it doesn't work for their particular scenario they can define converters to override the behavior. What drawbacks do you see to this?



---
In reply to: [102851157](https://github.com/Azure/azure-webjobs-sdk/pull/1037#discussion_r102851157) [](ancestors = 102851157,102849948)

MikeStall: That's way too broad. For example, CloudTable should not bind to JObject.

mamaso: Ah, you're right.  Way too locked into the POCO mindset.

---
In reply to: [103005608](https://github.com/Azure/azure-webjobs-sdk/pull/1037#discussion_r103005608) [](ancestors = 103005608)

src/Microsoft.Azure.WebJobs.Host/Bindings/ConverterManager.cs
```diff
@@ -25,6 +25,7 @@ internal class ConverterManager : IConverterManager
         public ConverterManager()
         {
             this.AddConverter<byte[], string>(DefaultByteArrayToString);
+            this.AddConverter<IEnumerable<JObject>, JArray>((enumerable) => JArray.FromObject(enumerable));
```
MikeStall: >AddConverter [](start = 17, length = 12)

This is good - one catch is that converters don't automatically chain. So the TryGet logic below may need to do a special check to chain this case.

test/Microsoft.Azure.WebJobs.Host.UnitTests/ConverterManagerTests.cs
```diff
@@ -40,11 +40,37 @@ public void ExactMatchOverride()
...
             var x1 = func("x", null, Context);
             Assert.Equal("*x*", x1);
         }
 
+        public T TestDefaultConverter<F, T>(F from, T to)
+        {
+            var cm = new ConverterManager();
+            var converter = cm.GetConverter<F, T, Attribute>();
+            Assert.NotNull(converter);
+            var result = converter.Invoke(from, null, null);
```
MikeStall: >Invoke [](start = 35, length = 6)

nit: why .Invoke?  Can you just use () to call directly?

test/Microsoft.Azure.WebJobs.Host.UnitTests/ConverterManagerTests.cs
```diff
@@ -40,11 +40,37 @@ public void ExactMatchOverride()
...
+
+        [Fact]
+        public void ByteToString_DefaultConverter()
+        {
+            var s = "ab";
+            TestDefaultConverter(Encoding.UTF8.GetBytes(s), s);
+        }
+
+        [Fact]
+        public void IEnumerableToJArray_DefaultConverter()
```
MikeStall: >IEnumerableToJArray_DefaultConverter [](start = 20, length = 36)

Can you include another test to demonstrate chaining? 

Ie, add a converter from Widget --> IEnumerable<JObject> 
and then verify we can do a get (Widget, JArray)?

mamaso:
@MikeStall can you take a look at this when you get a chance? Made the updates you requested

PR #1039, opened 2017-02-25
Config

MikeStall:
For script extensibility, we need to be able to get the static information (specifically Binding providers and converters) before we spin up the listeners. 

JobHostContextFactory breaks all layering and makes this impossible. 

There are currently 4 classes:

- Public JobHost – 
- Public JobHostConfiguration – all the config  (binders, name resolvers, etc). This is immutable once we startup the JobHost. 
- Private JobHostContext – a few key types used by JobHost. This is the result of indexing.  
- Private JobHostContextFactory – take in the Configuration and produce the Context. 

The first two make perfect sense, the last two are confusing.  JobHostContext is really just the fields that should be on JobHost, so that’s not so bad; but JobHostContextFactory is super confusing and there’s a lot of hackery around it.   

I think the extra abstractions where introduced to support a goal that JobHostConfiguration could be viewed as a general purpose property bag and swapped out with an IServicePvovider  (a BCL type). JobHostConfig even impls that interface:
```
   public sealed class JobHostConfiguration : IServiceProvider
```
However, that never **goal could have worked** because: JobHost asks JobHostConfig for a JobHostContextFactory.  Since Factory is private, there’s no way a public implementation of IServicePvovider could ever provide it. 

So bottom line is that _it looks like we have a bunch of extra abstractions that complicate the system, block our script scenarios, and don’t actually provide any tangible value_. 

This removes JobHostContextFactory, and fold the logic into JobHostConfiguration; and then a bunch of indirections just disappear.  It also significantly reduces some duplication in the tests.


src/Microsoft.Azure.WebJobs.Host/Executors/IJobHostContextFactory.cs
```diff
@@ -1,13 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the MIT License. See License.txt in the project root for license information.
-
-using System.Threading;
-using System.Threading.Tasks;
-
-namespace Microsoft.Azure.WebJobs.Host.Executors
-{
-    internal interface IJobHostContextFactory
```
MikeStall: This goes away - use JobHostConfiguration directly instead.

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
```diff
@@ -323,13 +311,6 @@ public object GetService(Type serviceType)
             object service = null;
             _services.TryGetValue(serviceType, out service);
 
-            if (service == null && serviceType == typeof(IJobHostContextFactory))
```
MikeStall: :)

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/FunctionalTest.cs
```diff
@@ -213,73 +213,28 @@ internal static class FunctionalTest
                 StorageAccount = storageAccount
             };
             IHostIdProvider hostIdProvider = new FakeHostIdProvider();
-            INameResolver nameResolver = null;
```
MikeStall: also very happy to remove this.  Large section of very tricky test code that duplicated product code with subtle changes ... good to remove it and just use the product code.

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/JobHostTests.cs
```diff
@@ -0,0 +1,59 @@
...
+using Microsoft.Azure.WebJobs.Host.TestCommon;
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Host.FunctionalTests
+{
+    public class JobHostTests
+    {
+        // Test that we can do static initialization before runtime initialization. 
+        [Fact]
+        public async Task StaticInit()
```
MikeStall: This new SDK test demonstrates how I expect Script to leverage this work. It's verifying all this churn actually accomplishes something practical :)

test/Microsoft.Azure.WebJobs.Host.TestCommon/JobHostFactory.cs
```diff
@@ -33,57 +33,16 @@ public static TestJobHost<TProgram> Create<TProgram>(CloudStorageAccount storage
...
+            JobHostConfiguration config = TestHelpers.NewConfig<TProgram>(hostIdProvider);
 
             IStorageAccountProvider storageAccountProvider = new SimpleStorageAccountProvider(config)
             {
                 StorageAccount = storageAccount,
                 // use null logging string since unit tests don't need logs.
                 DashboardAccount = null
             };
-
-            IExtensionTypeLocator extensionTypeLocator = new NullExtensionTypeLocator();
```
MikeStall: Here's another scary section of test code that duplicated the product.

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostContextFactory.cs
```diff
@@ -27,121 +27,162 @@
...
+
+            return context;
         }
 
-        public async Task<JobHostContext> CreateAndLogHostStartedAsync(JobHost host, CancellationToken shutdownToken, CancellationToken cancellationToken)
+        // Static initialization. Returns a service provider with some new services initialized. 
+        // The new services:
+        // - can retrieve static config like binders and converters; but the listeners haven't yet started.
+        // - can be flowed into the runtime initialization to get a JobHost spun up and running.
+        // This shouldn't be async beccause we shouldn't need to do network calls for static init.
```
mathewc: Async doesn't imply network calls - I don't understand this comment

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostContextFactory.cs
```diff
@@ -27,121 +27,162 @@
...
+            return context;
         }
 
-        public async Task<JobHostContext> CreateAndLogHostStartedAsync(JobHost host, CancellationToken shutdownToken, CancellationToken cancellationToken)
+        // Static initialization. Returns a service provider with some new services initialized. 
+        // The new services:
+        // - can retrieve static config like binders and converters; but the listeners haven't yet started.
+        // - can be flowed into the runtime initialization to get a JobHost spun up and running.
+        // This shouldn't be async beccause we shouldn't need to do network calls for static init.        
+        // This can be called multiple times on a config, which is why it returns a new ServiceProviderWrapper
```
mathewc: How is this called multiple times on a config? It is called only once for the lifetime of a JobHost, right? Are you talking about scenarios like Functions, where we start/stop hosts using the same config?

MikeStall: We have existing SDK unit tests that use multiple JobHosts on a single Config object (see BlobGetsProcessedOnlyOnce_SingleHost ). Caught me be surprise too.

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostContextFactory.cs
MikeStall: It's a conscious choice that this method is not Async.  This is just static initialization (like adding converters, binders, etc)  so it should not need to make network calls.   IE, making this method async would be an indication that something is wrong.

src/Microsoft.Azure.WebJobs.Host/Config/ServiceProviderWrapper.cs
```diff
@@ -0,0 +1,37 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Azure.WebJobs.Host.Config
+{    
+    // Used to wrap a JobHostConfiguration. Let's JobHost startup modify services  
+    // without modifying the underlying configuration. 
+    internal class ServiceProviderWrapper : IServiceProvider
```
mathewc: Can we rename this to JobHostServiceProvider?

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostContextFactory.cs
```diff
@@ -27,121 +27,162 @@
...
+        // instead of modifying the config.
+        public static ServiceProviderWrapper DoStaticInitialization(this JobHostConfiguration config)
```
mathewc: With the suggested rename of ServiceProviderWrapper => JobHostServiceProvider, I'd just rename this method to CreateServiceProvider, and inline all the stuff currently in StaticInitWorker (that method isn't needed).

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostContextFactory.cs
```diff
@@ -27,121 +27,162 @@
...
+        public static async Task<JobHostContext> CreateAndLogHostStartedAsync(
+            this JobHostConfiguration config, 
+            JobHost host, 
+            CancellationToken shutdownToken, 
+            CancellationToken cancellationToken)
         {
-            _storageAccountProvider = storageAccountProvider;
-            _consoleProvider = consoleProvider;
-            _config = config;
+            var newServices = config.DoStaticInitialization();
```
mathewc: var jobHostServices = CreateServiceProvider(config);

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostContextFactory.cs
```diff
@@ -27,121 +27,162 @@
...
+            JobHostContext context = await newServices.DoRuntimeInitialization(config, host, shutdownToken, cancellationToken);
```
mathewc: In line with the other renames I suggested for clarity, I'd rename stuff so you can do:

var context = await CreateJobHostContextAsync(jobHostServices, host, ...);

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostContextFactory.cs
```diff
@@ -27,121 +27,162 @@
...
+        }
 
-            bool hasFastTableHook = config.GetService<IAsyncCollector<FunctionInstanceLogEntry>>() != null;
-            bool noDashboardStorage = config.DashboardConnectionString == null;
-            if (hasFastTableHook && noDashboardStorage)
+        // Do the runtime intitialization. This happens after the static initialization. 
+        // This mainly means:
+        // - indexing the functions 
+        // - spinning up the listeners (so connecting to the services)
+        private static async Task<JobHostContext> DoRuntimeInitialization(this ServiceProviderWrapper config, JobHostConfiguration rawConfig, JobHost host, CancellationToken shutdownToken, CancellationToken cancellationToken)
```
mathewc: Rename to CreateJobHostContextAsync, which better describes what the method actually does

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostContextFactory.cs
```diff
@@ -27,121 +27,162 @@
...
+        // The new services:
+        // - can retrieve static config like binders and converters; but the listeners haven't yet started.
+        // - can be flowed into the runtime initialization to get a JobHost spun up and running.
+        // This shouldn't be async beccause we shouldn't need to do network calls for static init.        
+        // This can be called multiple times on a config, which is why it returns a new ServiceProviderWrapper
+        // instead of modifying the config.
+        public static ServiceProviderWrapper DoStaticInitialization(this JobHostConfiguration config)
+        {
+            var ctx = new ServiceProviderWrapper(config);
+            StaticInitWorker(ctx, config);
```
mathewc: It's confusing that you are creating a wrapper from a config, then in the next line passing both that wrapper as well as the raw config into this function. Why?

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostContextFactory.cs
```diff
@@ -27,121 +27,162 @@
...
+        // This can be called multiple times on a config, which is why it returns a new ServiceProviderWrapper
+        // instead of modifying the config.
+        public static ServiceProviderWrapper DoStaticInitialization(this JobHostConfiguration config)
+        {
+            var ctx = new ServiceProviderWrapper(config);
+            StaticInitWorker(ctx, config);        
+            return ctx;
+        }
+                
+        public static void StaticInitWorker(this ServiceProviderWrapper config, JobHostConfiguration rawConfig)
```
mathewc: nit: Once ServiceProviderWrapper is renamed to JobHostServiceProvider, these params can be named more naturally serviceProvider, config.

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
```diff
@@ -20,12 +21,14 @@ public sealed class JobHostConfiguration : IServiceProvider
     {
         private static readonly IConsoleProvider ConsoleProvider = new DefaultConsoleProvider();
 
+        // In some legacy cases, JobHostConfiguration specifically exposes properties on the derived object 
+        // (and not on the core interface in the _services collection).
```
mathewc: What legacy situations are you referring to?

mathewc: I don't understand this comment at all - I wouldn't classify the below config types as legacy if that is your meaning. Can you just remove this comment? It may cause more confusion than help.

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/FunctionalTest.cs
```diff
@@ -196,15 +196,15 @@ internal static class FunctionalTest
...
+        private static JobHostConfiguration CreateServiceProvider<TResult>(IStorageAccount storageAccount, Type programType,
             IExtensionTypeLocator extensionTypeLocator, TaskCompletionSource<TResult> taskSource,
             IFunctionInstanceLogger functionInstanceLogger, IExtensionRegistry extensions = null)
         {
             return CreateServiceProvider<TResult>(storageAccount, programType, extensionTypeLocator,
                 DefaultJobActivator.Instance, taskSource, functionInstanceLogger, extensions);
         }
 
-        private static IServiceProvider CreateServiceProvider<TResult>(IStorageAccount storageAccount, Type programType,
+        private static JobHostConfiguration CreateServiceProvider<TResult>(IStorageAccount storageAccount, Type programType,
```
mathewc: Can we rename these methods to CreateConfiguration or CreateTestConfiguration since that is what they now do?

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/FunctionalTest.cs
```diff
@@ -327,7 +282,7 @@ public static void RunTrigger(IStorageAccount account, Type programType, IExtens
             }
         }
 
-        public static TResult RunTrigger<TResult>(IServiceProvider serviceProvider, Task<TResult> task)
+        public static TResult RunTrigger<TResult>(JobHostConfiguration serviceProvider, Task<TResult> task)
```
mathewc: Can you fix up the param names so they match with the refactors you've done?

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/JobHostTests.cs
mathewc: I don't see why Script will need to execute any funky init calls on the config - that seems problematic to me. I think we can just rely on the fact that that init has been done internally at the right time by the underlying ScriptHost, and we flow the config around and use it naturally. I.e. I don't think we should have any new public extension methods on config.

src/Microsoft.Azure.WebJobs.Host/Config/ServiceProviderWrapper.cs
MikeStall: my one reservation was that this class is totally generic and has no connection to JobHost.

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostContextFactory.cs
MikeStall: The current initialization logic is pretty tangled - I couldn't fully untangle it. What's happening here is:
1. initialization creates some new state (tracked via the wrapper). Specifically, static init must carry over some objects to runtime init. 
2. but both inits also depend directly on properties in JobHostConfig (that aren't surfaced as services).

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
MikeStall: Sure. I'll remove it.  But the point I'm making is that there's a gross inconsistency in here that JobHsotConfig is an IServiceProvider, but has several objects that are outside the service provider for now good reason. For example, `DefaultStorageAccountProvider _storageAccountProvider`.  DefaultStorageAccountProvider implements IStorageAccountProvider, but JobHostConfig exposes the derived class and accesses properties on DefaultStorageAccountProvider that aren't on IStorageAccountProvider.

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
```diff
@@ -27,67 +27,69 @@
...
+        public static async Task<JobHostContext> CreateAndLogHostStartedAsync(
+            this JobHostConfiguration config, 
+            JobHost host, 
+            CancellationToken shutdownToken, 
+            CancellationToken cancellationToken)
         {
-            _storageAccountProvider = storageAccountProvider;
-            _consoleProvider = consoleProvider;
-            _config = config;
+            var newServices = config.CreateStaticServices();
```
mathewc: I think you should move this call into CreateJobHostContextAsync, since it always needs to be done to create a JobHostContext.

mathewc:
Closing since it looks like this is merged.

Issue #1042, opened 2017-03-06
Use provided TextWriter or TraceWriter make webjob crash


EmmanuelEngel:
Simple webjob that use provided logger crash. Exception is not catchable.

#### Repro steps

Provide the steps required to reproduce the problem

1. Set up project with this simple code into Program.cs 

using System.Collections.Generic;
using System.IO;
using System.Threading.Tasks;
using System.Threading;
using Microsoft.Azure.WebJobs;
using System;
using Microsoft.Azure.WebJobs.Host;

namespace WebJob
{
    public class Functions
    {
        // This function will get triggered/executed when a new message is written 
        // on an Azure Queue called queue.

        public static object locker = new object();

        [NoAutomaticTrigger]
        //public static async Task ProcessQueueMessage(TextWriter log)
        public static async Task ProcessQueueMessage(TraceWriter log)
        {
            try
            {
                var w = log;
                w.Level = System.Diagnostics.TraceLevel.Verbose;
                //using (TextWriter w = TextWriter.Synchronized(log))
                {
                    List<Task> tasks = new List<Task>();
                    int count = 0;
                    for (int i = 0; i < 500; ++i)
                    {
                        int k = i;
                        Task pending = Task.Run(() =>
                        {
                            for (int j = 0; j < 500; ++j)
                            {
                                try
                                {
                                    //w.WriteLine($"[{Interlocked.Increment(ref count)}][{k}] : {j}");
                                    w.Info($"[{Interlocked.Increment(ref count)}][{k}] : {j}");
                                }
                                catch (Exception)
                                {                                 
                                }
                            }
                        });
                        tasks.Add(pending);
                    }
                    await Task.WhenAll(tasks);
                }
            } catch (Exception)
            {
                // nop
            }
        }
    }
}

2. Step B

Deploy it on a web job and start it

#### Expected behavior

No crash

#### Actual behavior

Crash: 

[...]

PR #1043, opened 2017-03-06
making TextWriterTraceAdapter synchronized

brettsam:
Addresses #1042


src/Microsoft.Azure.WebJobs.Host/Bindings/TraceWriter/TextWriterTraceAdapter.cs
```diff
@@ -27,14 +27,18 @@ public override Encoding Encoding
...
+        }
 
         public override void Write(char value)
         {
             // buffer all output
             _text.Append(value);
 
             int len = _text.Length;
-            if (len > 2 && _text[len - 2] == '\r' && _text[len - 1] == '\n')
+            if (len >= 2 && _text[len - 2] == '\r' && _text[len - 1] == '\n')
```
mathewc: Can you explain this change?

test/Microsoft.Azure.WebJobs.Host.UnitTests/Bindings/TraceWriter/TextWriterTraceAdapterTests.cs
```diff
@@ -70,5 +73,24 @@ public void Flush_FlushesRemainingBuffer()
...
+
+        [Fact]
+        public async Task TestMultipleThreads()
+        {
+            // This validates a bug where writing from multiple threads throws an exception.             
+            TestTraceWriter trace = new TestTraceWriter(TraceLevel.Verbose);
+            TextWriter adapter = TextWriterTraceAdapter.Synchronized(trace);
+
+            // Start Tasks to write
+            List<Task> tasks = new List<Task>();
```
mathewc: Did this test repro the problem reliably?

src/Microsoft.Azure.WebJobs.Host/Bindings/TraceWriter/TextWriterTraceAdapter.cs
brettsam: In my test, I just log WriteLine(string.Empty) -- which looks like it just sends '\r\n'. In that case, the string length was only 2, so this logic didn't kick in like I would have expected. Instead, it waited until it had '\r\n\r\n' before flushing. It's a minor change -- but I thought it made sense.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Bindings/TraceWriter/TextWriterTraceAdapterTests.cs
brettsam: Yeah, every time.

PR #1044, opened 2017-03-07
fixing broken cast after previous commit

brettsam:


brettsam:
I'm going to commit this to get CI passing. I'll still address comments.

PR #1048, opened 2017-03-08
Added AppSettingAttribute

### LINKED ISSUE #1031: Replace AutoResolve AllowTokens false with AppSetting Attribute
`[AppSetting]` would be clearer than `[AutoResolve(AllowTokens = false)]`

`[AppSetting(Default=”AzureWebJobsIOTXXX”)]` to support default resolution

Useful on extension attributes to resolve connection string information or other settings.

mamaso:
resolves #1031

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
@@ -3,12 +3,19 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Reflection;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host.Bindings.Path;
 
 namespace Microsoft.Azure.WebJobs.Host.Bindings
 {
+    using BindingData = IReadOnlyDictionary<string, object>;
+    using BindingDataContract = IReadOnlyDictionary<string, System.Type>;
+    // Func to transform Attribute,BindingData into value for cloned attribute property/constructor arg
+    // Attribute is the new cloned attribute - null if constructor arg (new cloned attr not created yet)
+    using BindingDataResolver = Func<Attribute, IReadOnlyDictionary<string, object>, object>;
+
     // Clone an attribute and resolve it.
     // This can be tricky since some read-only properties are set via the constructor.
     // This assumes that the property name matches the constructor argument name.
@@ -19,20 +26,22 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
 
         // Which constructor do we invoke to instantiate the new attribute?
         // The attribute is configured through a) constructor arguments, b) settable properties.
-        private readonly ConstructorInfo _bestCtor;
+        private readonly ConstructorInfo _matchedCtor;
 
         // Compute the arguments to pass to the chosen constructor. Arguments are based on binding data.
-        private readonly Func<IReadOnlyDictionary<string, object>, object>[] _bestCtorArgBuilder;
+        private readonly BindingDataResolver[] _ctorParamResolvers;
 
         // Compute the values to apply to Settable properties on newly created attribute.
-        private readonly Action<TAttribute, IReadOnlyDictionary<string, object>>[] _setProperties;
+        private readonly Action<TAttribute, BindingData>[] _propertySetters;
 
         // Optional hook for post-processing the attribute. This is intended for legacy hack rules.
         private readonly Func<TAttribute, Task<TAttribute>> _hook;
 
+        private static readonly BindingFlags Flags = BindingFlags.Instance | BindingFlags.Public;
+
         public AttributeCloner(
             TAttribute source,
-            IReadOnlyDictionary<string, Type> bindingDataContract,
+            BindingDataContract bindingDataContract,
             INameResolver nameResolver = null,
             Func<TAttribute, Task<TAttribute>> hook = null)
         {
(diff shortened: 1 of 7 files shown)


src/Microsoft.Azure.WebJobs.Host/Extensions/AppSettingAttributeExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace Microsoft.Azure.WebJobs.Host
+{
+    internal static class AppSettingAttributeExtensions
+    {
+        public static AutoResolveAttribute ToAutoResolve(this AppSettingAttribute attribute)
+        {
+            return new AutoResolveAttribute
+            {
+                Default = attribute.Default,
+                AllowTokens = false
```
mathewc: Is the crux of this whole proposal a way to make the terminology of AllowTokens more understandable? If so, rather than introduce a new attribute and have 2 mechanisms around (further complicating things I think), what if we instead renamed AllowTokens to IsAppSetting? @brettsam I remember when we initially named this we had a lot of back and forth on the right name - we might have even had a name like this proposed at one time.

What I don't think we want to do is keep both ways - either we remove AllowTokens completely, or we rename this. `[AutoResolve(IsAppSetting = true)]` seems clear doesn't it?

mamaso: That and default values. The rename to `IsAppSetting` with the addition of a `Default` property seems like it would keep things simpler.

@MikeStall any thoughts?

mamaso:
@mathewc @MikeStall new commit does a full split into `AppSettingAttribute` and `AutoResolveAttribute`, removing `AllowTokens` entirely.  Did some refactoring of the `AttributeCloner` constructor to make the logic more straightforward

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -3,12 +3,18 @@
...
 using System.Collections.Generic;
+using System.Linq;
 using System.Reflection;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host.Bindings.Path;
 
 namespace Microsoft.Azure.WebJobs.Host.Bindings
 {
+    using BindingData = IReadOnlyDictionary<string, object>;
+    using BindingDataContract = IReadOnlyDictionary<string, System.Type>;
```
MikeStall: that's a nice maintainability touch.

MikeStall:
>     public class AttributeClonerTests

I think you're fixing another case here too. 
[AppSetting] shouldn't allow { } in them.   
Before, I think these got resolved automatically by merit of it being an AutoResolve.  

Can you add a test case to verify the { } in [AppSetting] are NOT resolved?


---
Refers to: test/Microsoft.Azure.WebJobs.Host.UnitTests/AttributeClonerTests.cs:17 in 4cbda0a. [](commit_id = 4cbda0acde05d1dbd56e8a030d578e77cd0abb81, deletion_comment = False)

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -19,171 +25,113 @@ internal class AttributeCloner<TAttribute>
...
+        private BindingDataResolver GetResolver(PropertyInfo propInfo, INameResolver nameResolver, BindingDataContract contract)
         {
-            resolvedValue = null;
-
-            AutoResolveAttribute attr = propInfo.GetCustomAttribute<AutoResolveAttribute>();
-            if (attr == null)
+            object originalValue = propInfo.GetValue(_source);
+            // first try to resolve with app setting
+            AppSettingAttribute appSettingAttr = propInfo.GetCustomAttribute<AppSettingAttribute>();
+            if (appSettingAttr != null)
```
MikeStall: >appSettingAttr [](start = 16, length = 14)

nit: it should be illegal to have both attributes on a single property. Here, it looks like [Appsetting] takes precedence over [AutoResolve]. 
Can you make it an error to have both and add a test case?

mamaso: good idea, I'll add.

mamaso:
@MikeStall a bit confused in regards to:

>[AppSetting] shouldn't allow { } in them. 
>Can you add a test case to verify the { } in [AppSetting] are NOT resolved?

Should we error in the case that AppSetting has `{}`? Or should we ignore them and not resolve?

MikeStall:
Ignore and not resolve { } .   For example, imagine if the appsetting contents were a C# format string .

test/Microsoft.Azure.WebJobs.Host.UnitTests/AttributeClonerTests.cs
```diff
@@ -40,8 +40,27 @@ public Attr2(string resolvedProp2, string constantProp)
...
+        public class Attr3: Attribute
+        {
+            [AppSetting]
+            public string Required { get; set; }
+
+            [AppSetting(Default = "default")]
+            public string Default { get; set; }
+        }
+
+        public class InvalidAnnotation: Attribute
```
MikeStall: minor nit: can you comment why it's invalid.

test/Microsoft.Azure.WebJobs.Host.UnitTests/AttributeClonerTests.cs
```diff
@@ -251,6 +312,27 @@ public void Setting_WithNoValueInResolver_Throws()
         }
 
         [Fact]
+        public void AppSettingAttribute_DoesNotResolveBrackets()
```
MikeStall: >AppSettingAttribute_DoesNotResolveBrackets [](start = 20, length = 42)

For the test here, I'd suggest:

class TestAttr { 
   [AppSetting] Prop1 ;
   [AutoResolve] Prop2;
}

Prop1 = prop2 = "test{x}and%y%"
NameResolver has x --> "somevalue", y-->"othervalue"

And then verify that Prop1 is still unchanged; and Prop2 gets the resolution. 


For the current test'
- "req{}" is not valid since the {} is empty.  We don't want somebody to think we skipped resolution because it was an error
- include the %% since Appsetting shouldn't touch those either.  Essentially, [appsetting] is a literal.

mamaso: Prop1 won't be unchanged. It will resolve to EnvironmentVariable["test{x}and%y%"], if such a variable exists, otherwise it will throw.

PR #1054, opened 2017-03-10
Disable listener stopasync if exception

mamaso:
So @MikeStall was right, we should have better handling for the stop case.

Propose that if we get a handled listener exception we attempt to stop the listener immediately.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
@@ -13,6 +13,7 @@ namespace Microsoft.Azure.WebJobs.Host.Listeners
         private readonly IListener _listener;
         private readonly FunctionDescriptor _descriptor;
         private readonly TraceWriter _trace;
+        private bool _started = false;
 
         /// <summary>
         /// Wraps a listener. If the listener throws an exception OnStart,
@@ -43,16 +44,21 @@ namespace Microsoft.Azure.WebJobs.Host.Listeners
             try
             {
                 await _listener.StartAsync(cancellationToken);
+                _started = true;
             }
             catch (Exception e)
             {
                 new FunctionListenerException(_descriptor.ShortName, e).TryRecover(_trace);
             }
         }
-
-        public Task StopAsync(CancellationToken cancellationToken)
+        
+        public async Task StopAsync(CancellationToken cancellationToken)
         {
-            return _listener.StopAsync(cancellationToken);
+            if (_started)
+            {
+                await _listener.StopAsync(cancellationToken);
+                _started = false;
+            }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
```diff
@@ -47,12 +48,23 @@ public void Dispose()
             catch (Exception e)
             {
                 new FunctionListenerException(_descriptor.ShortName, e).TryRecover(_trace);
+
+                // if recovered, disable and attempt a best effort stop of the listener
+                _disabled = true;
```
mathewc: nit: _started?

src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
```diff
@@ -47,12 +48,23 @@ public void Dispose()
...
+                {
+                }
             }
         }
-
+        
         public Task StopAsync(CancellationToken cancellationToken)
         {
-            return _listener.StopAsync(cancellationToken);
+            return _disabled ? Task.FromResult(false) :
```
mathewc: nit: return `_started ? _listener.StopAsync(cancellationToken) : Task.Completed;`

test/Microsoft.Azure.WebJobs.Host.UnitTests/Listeners/FunctionListenerTests.cs
```diff
@@ -20,41 +20,67 @@ public class FunctionListenerTests
...
+
             Assert.Equal("The listener for function 'testfunc' was unable to start.", trace.Traces[0].Message);
             var exc = trace.Traces[0].Exception as FunctionException;
             Assert.Equal("testfunc", exc.MethodName);
             Assert.True(exc.Handled);
+            badListener.VerifyAll();
+        }
+
+        [Fact]
+        public async Task FunctionListener_DoesNotStop_IfDisabled()
```
mathewc: IfNotStarted

src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
mathewc: nit: return `_started ? _listener.StopAsync(cancellationToken) : Task.Completed;`

src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
```diff
@@ -47,12 +48,23 @@ public void Dispose()
             catch (Exception e)
             {
                 new FunctionListenerException(_descriptor.ShortName, e).TryRecover(_trace);
+
+                // if recovered, disable and attempt a best effort stop of the listener
+                _disabled = true;
+                try
+                {
+                    await _listener.StopAsync(cancellationToken);
```
mathewc: Not sure we should call Stop here in Start. Either we just don't call Stop at all if we never Started completely (which seems correct to me), or you move this Stop handling into the Stop method below.

PR #1055, opened 2017-03-10
ILogger and App Insights; move to .NET 4.6

brettsam:
What this enables: 
- Nothing by default -- there is no registered `ILoggerFactory`, so nothing new happens.
- The ability to register an `ILoggerFactory` that you construct yourself (for now). Eventually we can build it for you. See the [sample here](https://github.com/Azure/azure-webjobs-sdk/pull/1055/files#diff-da61be77068f6c3dd7503be7a1094456) for how you would build your LoggerFactory and pass it to us.
- Once a logger factory is registered, we'll log all function results to it. All traces will also go through it.
- By adding our AppInsightsLogger, that means all traces and function results will go to App Insights.
- A function result aggregator that batches function results and sends them periodically rather than sending every individual result. There is an `Aggregator` configuration object on `JobHostConfiguration` to control this.
- By applying filters with the `FilterBuilder`, you can selectively turn on and off the different logging pieces.


abatishchev:
I would upgrade to 4.6 in a separate pr, to make the diff cleaner

sample/SampleHost/Properties/PublishProfiles/brettsam-webjob - Web Deploy.pubxml
```diff
@@ -0,0 +1,22 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
abatishchev: (If you don't mind comments on your pr) likely you don't want this file to go in

brettsam: Thanks :-)

brettsam:
I just had another comment to do that as well :-) I'll separate them out -- likely won't happen until this weekend, though.

christopheranderson:
Related: https://github.com/Azure/azure-webjobs-sdk-script/issues/727

brettsam:
Closing this in favor of #1058 and #1059.

PR #1056, opened 2017-03-11
Stop special-casing StorageAccountAttribute.

MikeStall:
Make the storage attributes (Blob,Table,Queue) derive from StorageAccountAttribute.  This gives Blob* attribute a Connection property, just like all the other attributes.
This is not a breaking change - and no change to public tests.

The key is in TypeUtility's new GetAttr function. It still walks the hierarchy and collapses attributes into a single attr.

The motivation here is:
1. Stop special-casing blob storage. All other resources need something similar here.
2. Provides a 1:1 between Attributes and Function.json that lets us removes corner cases in Script and tooling

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/TypeUtility.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/TypeUtility.cs
@@ -98,6 +98,68 @@ namespace Microsoft.Azure.WebJobs.Host
             return null;
         }
 
+        // $$$
+        internal static TAttribute GetAttr<TAttribute>(ParameterInfo parameter) where TAttribute : Attribute
+        {
+            TAttribute attributeSource = parameter.GetCustomAttribute<TAttribute>(inherit: false);
+
+            // Walk hierachy.
+            // It's possible attribute is not on the parameter but is still on the something 
+            // higher up in the hiearchy
+            ICustomAttributeProvider provider = parameter;
+
+            while (provider != null)
+            {
+                var attrs = provider.GetCustomAttributes(typeof(Attribute), false);
+                foreach (var attr in attrs)
+                {
+                    if (attributeSource == null)
+                    {
+                        attributeSource = attr as TAttribute;
+                        continue;
+                    }
+
+                    if (attr.GetType().IsAssignableFrom(typeof(TAttribute)))
+                    {
+                        // Merge properties. Give precedence to "closest" one. 
+                        foreach (var prop in attr.GetType().GetProperties(BindingFlags.Instance | BindingFlags.Public))
+                        {
+                            if (prop.CanRead && prop.CanWrite)
+                            {
+                                var currentValue = prop.GetValue(attributeSource);
+                                if (currentValue == null)
+                                {
+                                    var value = prop.GetValue(attr);
+                                    prop.SetValue(attributeSource, value);
+                                }
+                            }
+                        }
+                    }
+                }
+
+                provider = GetParent(provider);
+            }
+
+            return attributeSource;
+        }
+
+        private static ICustomAttributeProvider GetParent(ICustomAttributeProvider x)
+        {
+            ParameterInfo p = x as ParameterInfo;
+            if (p != null)
+            {
+                return p.Member as MethodInfo;
+            }
+
+            MethodInfo m = x as MethodInfo;
+            if (m != null)
+            {
+                return m.DeclaringType;
+            }
+
+            return null;
+        }
+
         public static bool IsAsync(MethodInfo methodInfo)
         {
             if (methodInfo == null)
--- a/src/Microsoft.Azure.WebJobs/QueueAttribute.cs
+++ b/src/Microsoft.Azure.WebJobs/QueueAttribute.cs
@@ -26,7 +26,7 @@ namespace Microsoft.Azure.WebJobs
     [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Performance", "CA1813:AvoidUnsealedAttributes")]
     [AttributeUsage(AttributeTargets.Parameter)]
     [DebuggerDisplay("{QueueName,nq}")]
-    public class QueueAttribute : Attribute
+    public class QueueAttribute : StorageAccountAttribute
     {
         private readonly string _queueName;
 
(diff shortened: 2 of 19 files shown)


src/Microsoft.Azure.WebJobs.Host/TypeUtility.cs
```diff
@@ -98,6 +98,68 @@ internal static Type UnwrapTaskType(Type type)
...
+        internal static TAttribute GetAttr<TAttribute>(ParameterInfo parameter) where TAttribute : Attribute
+        {
+            TAttribute attributeSource = parameter.GetCustomAttribute<TAttribute>(inherit: false);
+
+            // Walk hierachy.
+            // It's possible attribute is not on the parameter but is still on the something 
+            // higher up in the hiearchy
+            ICustomAttributeProvider provider = parameter;
+
+            while (provider != null)
```
mathewc: There is already an existing GetHierarchicalAttributeOrNull that was doing this walk up from parameter/method/class. Why the new code for this?

src/Microsoft.Azure.WebJobs.Host/TypeUtility.cs
```diff
@@ -98,6 +98,68 @@ internal static Type UnwrapTaskType(Type type)
...
+                        // Merge properties. Give precedence to "closest" one. 
+                        foreach (var prop in attr.GetType().GetProperties(BindingFlags.Instance | BindingFlags.Public))
+                        {
+                            if (prop.CanRead && prop.CanWrite)
+                            {
+                                var currentValue = prop.GetValue(attributeSource);
+                                if (currentValue == null)
+                                {
+                                    var value = prop.GetValue(attr);
+                                    prop.SetValue(attributeSource, value);
```
mathewc: Why is this getter function setting things?

src/Microsoft.Azure.WebJobs.Host/TypeUtility.cs
mathewc: Also the naming makes it clear from the caller that a hierarchical walk for the attribute will be performed. This isn't clear from "GetAttr"

MikeStall: GetHierarchicalAttributeOrNull  doesn't handle base classes. We do leverage existing tests on the new code.  I can probably remove GetHierarchicalAttributeOrNull  so that there's only one copy of this.

src/Microsoft.Azure.WebJobs.Host/TypeUtility.cs
MikeStall: We have to merge the properties together into a single attr.   Say we have:
   [Blob] [Storage(account="X"] 
We need to set Blob.account = x

src/Microsoft.Azure.WebJobs/QueueAttribute.cs
```diff
@@ -26,7 +26,7 @@ namespace Microsoft.Azure.WebJobs
     [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Performance", "CA1813:AvoidUnsealedAttributes")]
     [AttributeUsage(AttributeTargets.Parameter)]
     [DebuggerDisplay("{QueueName,nq}")]
-    public class QueueAttribute : Attribute
+    public class QueueAttribute : StorageAccountAttribute
```
mathewc: The "is a" relationship here kind of doesn't make sense. If we had a StorageAttribute (rather than StorageAccountAttribute) it does make more sense. QueueAttribute "is a" StorageAttribute. What if we just introduced this new StorageAccount base type and used that?

MikeStall: Ideally, we want (and I think these are in priority order):
1. No breaking changes
2. no redundant attributes (having StorageAccount and Storage would be confusing and is unnecessary from a binding perspective)
3. good names 

I don't think we can get all 3.  The current PR does 1&2.

mathewc: Adding the new StorageAttribute base class would be non breaking and would give us all 3. All the types make sense together, and in the next major version, we can simply remove the Parameter level usage from StorageAccountAttribute - it then stays around for class/method level overrides.

MikeStall: I think it violates #2 ... a separate Storage and StorageAccount attribute seems confusing. What's the difference between them? What's the customer value in having 2 separate attributes?

mathewc: The value is to us as framework authors. We have a base attribute class that is extensible and makes more sense from a type perspective. StorageAccountAttribute is for specifying a storage account name value - that's why I named it the way I did. It was not designed as a base class - only to specify a single account name. StorageAttribute (abstract base class) is the base class for all storage binding attributes.

This won't be confusing for our users - the base class is abstract and they won't even think about it.

MikeStall: Does StorageAccount : Storage; Blob : Storage? What's a specific difference between Storage and StorageAccount?

src/Microsoft.Azure.WebJobs.Host/TypeUtility.cs
```diff
@@ -98,6 +98,68 @@ internal static Type UnwrapTaskType(Type type)
...
+            // It's possible attribute is not on the parameter but is still on the something 
+            // higher up in the hiearchy
+            ICustomAttributeProvider provider = parameter;
+
+            while (provider != null)
+            {
+                var attrs = provider.GetCustomAttributes(typeof(Attribute), false);
+                foreach (var attr in attrs)
+                {
+                    if (attributeSource == null)
```
mathewc: It's not clear to me at all what this code is doing. Some good code comments in this method would go a long way

src/Microsoft.Azure.WebJobs.Host/TypeUtility.cs
mathewc: I think we need to come up with a better way to do this. This attribute merging is hacky, and isn't a pattern we want to set in stone like this.

src/Microsoft.Azure.WebJobs.Host/TypeUtility.cs
MikeStall: Correct

src/Microsoft.Azure.WebJobs.Host/TypeUtility.cs
MikeStall: That's the heart of the proposal...    We started down this path as soon as we split out a single attribute into multiple attributes. We need some means to associated the different attributes together.

ahmelsayed:
doesn't [`ServiceBusAccountAttribute.cs`](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.ServiceBus/ServiceBusAccountAttribute.cs) have similar special handling?

PR #1058, opened 2017-03-11
upgrade all projects to .NET 4.6

brettsam:


src/Dashboard/Web.config
```diff
@@ -1,100 +1,107 @@
...
+      </system.Web>
+  -->
   <system.web>
-    <customErrors mode="RemoteOnly" />
-    <compilation debug="true" targetFramework="4.5" />
-    <httpRuntime targetFramework="4.5" />
-    <authentication mode="None" />
+    <customErrors mode="RemoteOnly"/>
+    <compilation debug="true" targetFramework="4.6"/>
+    <httpRuntime targetFramework="4.5"/>
```
mathewc: This stays 4.5?

mathewc:
Are you going to do the same for the Extensions repo?

src/Dashboard/Web.config
brettsam: It auto-added the runtime change above in a comment. I didn't want to do any manual adjusting here without a great way to test it out. My plan was just to leave it this way, figuring we could update it later.

brettsam:
@mathewc -- yeah, i'm doing that now. Pulling this in and moving everything to 4.6

PR #1059, opened 2017-03-11
adding initial ILogger/AppInsights support

brettsam:
_Note: This builds on PR #1058. I will check that in first and make sure all CI works before moving forward with this. To see the true diff, just view the second commit in this PR._ 

Related: Azure/azure-webjobs-sdk-script#727

What this enables:

- Nothing by default -- there is no registered `ILoggerFactory`, so nothing new happens.
- The ability to register an `ILoggerFactory` that you construct yourself (for now). Eventually we can build it for you. See [the sample here](https://github.com/Azure/azure-webjobs-sdk/pull/1059/files#diff-da61be77068f6c3dd7503be7a1094456) for how you would build your `LoggerFactory` and pass it to us.
- Once a logger factory is registered, we'll log all function results to it. All traces will also go through it.
- By adding our `AppInsightsLogger`, that means all traces and function results will go to App Insights.
- A function result aggregator that batches function results and sends them periodically rather than sending every individual result. There is an `Aggregator` configuration object on `JobHostConfiguration` to control this.
- By applying filters with the `FilterBuilder`, you can selectively turn on and off the different logging pieces.


src/Microsoft.Azure.WebJobs.Host/Extensions/DictionaryExtensions.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace System.Collections.Generic
+{
+    internal static class DictionaryExtensions
+    {
+        public static T GetValueOrNull<TKey, TValue, T>(this IDictionary<TKey, TValue> dictionary, TKey key) where T : class
```
mamaso: may want to special case `string` keys and add ordinal ignore case comparison

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/LoggerConfiguration.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Linq;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Host.Loggers
+{
+    /// <summary>
+    /// Provides a filter for use with an <see cref="ILogger"/>.
+    /// </summary>
+    [System.CLSCompliant(false)]
+    public class FilterBuilder
```
mamaso: Not a fan of this name, makes me think of builder pattern. Maybe categorical filter, idk

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Aggregator/FunctionResultAggregator.cs
```diff
@@ -0,0 +1,125 @@
...
+                _batcher.LinkTo(aggregator),
+                aggregator.LinkTo(publisher)
+            };
+
+            _windowTimer = new Timer(async (o) => await FlushAsync(), null, maxFlushInterval, maxFlushInterval);
+        }
+
+        public Task FlushAsync(CancellationToken cancellationToken = default(CancellationToken))
+        {
+            _batcher?.TriggerBatch();
```
mamaso: A bit confused on this - does the batch trigger a publish? It seems that FlushAsync will aggregate the current buffer, but I'm not sure when things will publish

brettsam: Eventually, yes, it triggers a publish. More accurately, a `BatchBlock` collects items until it hits the max, at which point it passes a collection of those items onto the next block in the chain. In our case, that's the `TransformBlock`, which does the aggregation. When that's done, it passes its output to the next block, which for us is `ActionBlock` that does the publishing. Calling [`TriggerBatch()`](https://msdn.microsoft.com/en-us/library/hh160310(v=vs.110).aspx) short circuits the "wait for max items" behavior and auto-flushes the items in the `BatchBlock` to the next block, which then eventually get published.

Another pre-emptive question I'll try to answer -- is the `BufferBlock` at the front of the pipeline necessary? It's there in the chance that the Batch block is processing items and not accepting more -- it gives them a place to be stored until the Batch block is ready for them. I wasn't able to get it to break with this configuration so I didn't want to break it down any further.

Take a read through this document -- it gives a pretty good overview: https://msdn.microsoft.com/en-us/library/hh228603(v=vs.110).aspx

@tohling and I were actually discussing some of this the other day -- the Dataflow library is really powerful and provides great patterns like this. But it's really hard to validate the logic just by looking at the code. Even writing tests is unit tests is tough -- I focused on the ends of the pipeline and assumed the other blocks were behaving as they were configured.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/LoggerConfiguration.cs
brettsam: Yeah, good call. I pulled this out into a class pretty late and didn't give the name much thought. How about `CategoryFilter`? And I can change the `CategoryFilters` property to `Levels` (or `CategoryLevels`?)

src/Microsoft.Azure.WebJobs.Host/Extensions/DictionaryExtensions.cs
brettsam: That's still possible -- the Dictionary constructor is where you define `StringComparer`. So you can create a dictionary that way and use it with this method.

src/Microsoft.Azure.WebJobs.Host/Bindings/ILogger/ILoggerBinding.cs
```diff
@@ -0,0 +1,93 @@
...
+using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Loggers;
+using Microsoft.Azure.WebJobs.Host.Protocols;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Host.Bindings
+{
+    internal class ILoggerBinding : IBinding
```
brettsam: I didn't even notice that until you pointed it out -- it does make it look like an interface. I think I'll switch it to `LoggerBinding` -- that still conveys the right information, but doesn't look like an interface.

brettsam:
@mamaso -- I added EBMs to the classes that I owned. I didn't want to go back and change a bunch of existing stuff. New commit has those changes as well as some of the naming suggestions.

brettsam:
@mathewc -- Can you take a look at this as well?

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Aggregator/FunctionResultAggregator.cs
mamaso: Yeah, I was impressed with the power that these patterns seem to expose, bummer if testing is rough though :/

Taking into account your explanation, I think that we may need to do more than just a `FlushAsync` in `JobHost.StopAsync` if we want to ensure that a publish happens.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/LoggerConfiguration.cs
mamaso: CategoryFilter and CategoryLevels sounds good to me

brettsam:
@SergeyKanzhelev -- here's the PR for our Core SDK App Insights integration. Most of the AI-specific stuff is in AppInsightsLogger.cs.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/LogValueCollection.cs
```diff
@@ -0,0 +1,80 @@
...
+namespace Microsoft.Azure.WebJobs.Host.Loggers
+{
+    /// <summary>
+    /// The <see cref="FormattedLogValues"/> object created by the framework <see cref="ILogger"/> extensions
+    /// (LogInformation(), for example) require that all key-value pairs be included in the string message
+    /// passed to the method. We'd like to have short strings with a subset of the data, so this class wraps
+    /// <see cref="FormattedLogValuesCollection"/> and allows us to use the same behavior, but with an additional
+    /// payload not included in the message.
+    /// </summary>
+    internal class FormattedLogValuesCollection : IReadOnlyList<KeyValuePair<string, object>>
```
brettsam: Note: I created this class to work around this limitation: https://github.com/aspnet/Logging/issues/533

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/LogValueCollection.cs
```diff
@@ -5,19 +5,32 @@
...
         private IReadOnlyList<KeyValuePair<string, object>> _additionalValues;
 
-        public LogValueCollection(string format, object[] formatValues, IReadOnlyDictionary<string, object> additionalValues)
+        public FormattedLogValuesCollection(string format, object[] formatValues, IReadOnlyDictionary<string, object> additionalValues)
         {
             _formatter = new FormattedLogValues(format, formatValues);
-            _additionalValues = additionalValues.ToList();
+            _additionalValues = additionalValues?.ToList();
+
+            if (_additionalValues == null)
```
mamaso: nit: could use null-coalescing above, `??`

sample/SampleHost/Program.cs
```diff
@@ -19,6 +21,24 @@ static void Main(string[] args)
                 config.UseDevelopmentSettings();
             }
 
+            // Build up a LoggerFactory to log to App Insights, but only if this key exists.
+            string instrumentationKey = Environment.GetEnvironmentVariable("ApplicationInsightsInstrumentationKey");
```
SergeyKanzhelev: Application Insights will use this env var: [APPINSIGHTS_INSTRUMENTATIONKEY](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/42d178c83d0cc72fa5270448e14e2044f1a790bf/src/Core/Managed/Net40/Extensibility/Implementation/TelemetryConfigurationFactory.cs#L22) Also the same is used by Web Apps (actually they came up with the name

brettsam: Yes, this is just a sample to show people they can build up the logger however they want. In our Functions implementation (in another repo -- coming once this PR is done), we do the config automatically and I use APPINSIGHTS_INSTRUMENTATIONKEY.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Aggregator/AggregatorConfiguration.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// <summary>
+        /// Gets or sets a value indicating the the maximum batch size for aggregations. When this number is hit, the results are
+        /// aggregated and sent to every registered <see cref="ILogger"/>.
+        /// </summary>
+        public int BatchSize { get; set; } = 1000;
+
+        /// <summary>
+        /// Gets or sets a value indicating when the aggregator will send results to every registered <see cref="ILogger"/>.
+        /// </summary>
+        public TimeSpan FlushTimeout { get; set; } = TimeSpan.FromSeconds(30);
```
SergeyKanzhelev: In Application Insights we typically send aggregated metrics once a minute, not 30 seconds. I think some UI features may have dependency on this... Worth checking with Kevin Blasko

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/AppInsights/AppInsightsLogger.cs
```diff
@@ -0,0 +1,312 @@
...
+using System.Net.Http;
+using System.Web;
+using Microsoft.ApplicationInsights;
+using Microsoft.ApplicationInsights.Channel;
+using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Host.Loggers
+{
+    internal class AppInsightsLogger : ILogger
```
SergeyKanzhelev: We try to keep the full name for the application insights - related classes (mostly because of marketing and there are other things called AppInsights)

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/AppInsights/AppInsightsLogger.cs
```diff
@@ -0,0 +1,312 @@
...
+
+            foreach (var property in values)
+            {
+                string stringValue = null;
+
+                // Format dates
+                Type propertyType = property.Value?.GetType();
+                if (propertyType == typeof(DateTime))
+                {
+                    stringValue = ((DateTime)property.Value).ToString(DateTimeFormatString);
```
SergeyKanzhelev: It is a best practice to switch to `DateTimeOffset` everywhere

SergeyKanzhelev: Why not UTC?

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/AppInsights/AppInsightsLogger.cs
```diff
@@ -0,0 +1,312 @@
...
+            }
+
+            requestTelemetry.Url = request.RequestUri;
+            requestTelemetry.Properties[LoggingKeys.HttpMethod] = request.Method.ToString();
+
+            requestTelemetry.Context.Location.Ip = GetIpAddress(request);
+            requestTelemetry.Context.User.UserAgent = request.Headers.UserAgent?.ToString();
+
+            HttpResponseMessage response = GetResponse(request);
+            requestTelemetry.ResponseCode = ((int)response?.StatusCode).ToString();
```
SergeyKanzhelev: You may want to override the Success of the telemetry as `201` or `204` with exception is still a success

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/AppInsights/AppInsightsLogger.cs
```diff
@@ -0,0 +1,312 @@
...
+
+        private static void ApplyHttpRequestProperties(RequestTelemetry requestTelemetry, HttpRequestMessage request)
+        {
+            if (request == null)
+            {
+                return;
+            }
+
+            requestTelemetry.Url = request.RequestUri;
+            requestTelemetry.Properties[LoggingKeys.HttpMethod] = request.Method.ToString();
```
SergeyKanzhelev: Not necessary. We typically put it into the `Name` like `GET /Home/Index`. Does it make sense for functions?

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/AppInsights/AppInsightsLogger.cs
```diff
@@ -0,0 +1,312 @@
...
+                    stringValue = ((DateTimeOffset)property.Value).UtcDateTime.ToString(DateTimeFormatString);
+                }
+                else
+                {
+                    stringValue = property.Value?.ToString();
+                }
+
+                // Since there is no nesting of properties, apply a prefix before the property name to lessen
+                // the chance of collisions.
+                telemetry.Properties.Add(LoggingKeys.CustomPropertyPrefix + property.Key, stringValue);
```
SergeyKanzhelev: not sure it will like to have `null` as a value. Worth checking

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/AppInsights/AppInsightsLogger.cs
```diff
@@ -0,0 +1,312 @@
...
+            telemetry.Timestamp = DateTimeOffset.UtcNow;
+            ApplyScopeProperties(telemetry);
+            ApplyCustomProperties(telemetry, logLevel, values);
+            _telemetryClient.TrackTrace(telemetry);
+        }
+
+        private void ApplyCustomProperties(ISupportProperties telemetry, LogLevel logLevel, IEnumerable<KeyValuePair<string, object>> values)
+        {
+            telemetry.Properties.Add(LoggingKeys.CategoryName, _categoryName);
+            telemetry.Properties.Add(LoggingKeys.Level, logLevel.ToString());
```
SergeyKanzhelev: I think Application Insights traces and exceptions already has severity level as a separate field

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/AppInsights/AppInsightsLogger.cs
```diff
@@ -0,0 +1,312 @@
...
+                throw new ArgumentNullException(nameof(state));
+            }
+
+            return AppInsightsScope.Push(state);
+        }
+
+        internal static string GetIpAddress(HttpRequestMessage httpRequest)
+        {
+            HttpContextBase context = httpRequest.Properties.GetValueOrNull<string, object, HttpContextBase>(ScopeKeys.HttpContext);
+            return context?.Request?.UserHostAddress ?? "0.0.0.0";
```
SergeyKanzhelev: Use `X-Forwarded-For` as azure web apps mostly uses load balancers https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/75373e57dcf8d646c54ee188461c373f2cc98939/Src/Web/Web.Shared.Net/ClientIpHeaderTelemetryInitializer.cs#L18

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/AppInsights/AppInsightsLoggerProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+        private readonly Func<string, LogLevel, bool> _filter;
+
+        public AppInsightsLoggerProvider(string instrumentationKey, Func<string, LogLevel, bool> filter)
+        {
+            if (instrumentationKey == null)
+            {
+                throw new ArgumentNullException(nameof(instrumentationKey));
+            }
+
+            TelemetryConfiguration config = TelemetryConfiguration.Active;
```
SergeyKanzhelev: You may want to construct your own instead of using the singleton

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/AppInsights/AppInsightsLoggerProvider.cs
```diff
@@ -0,0 +1,60 @@
...
+            QuickPulseTelemetryModule quickPulse = new QuickPulseTelemetryModule();
+            quickPulse.Initialize(config);
+            quickPulse.RegisterTelemetryProcessor(processor);
+
+            // Plug in perf counters
+            PerformanceCollectorModule perfCounterCollectorModule = new PerformanceCollectorModule();
+            perfCounterCollectorModule.Initialize(config);
+
+            // Instantiate client, which picks up the active config
+            _client = new TelemetryClient();
```
SergeyKanzhelev: pass the confir esxplictitly once you'll stop using singleton

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/AppInsights/AppInsightsScope.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        public static IDisposable Push(object state)
+        {
+            Current = new AppInsightsScope(state, Current);
+            return new DisposableScope();
+        }
+
+        // Builds a state dictionary of all scopes. If an inner scope
+        // contains the same key as an outer scope, it overwrites the value.
+        public static IDictionary<string, object> GetMergedStateDictionary()
```
SergeyKanzhelev: We would typically do it using ITelemetryInitializer. You can have one attached to your TelemetryConfiguration that will initialize ALL telemetry items without an explicit call to this method

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/AppInsights/WebJobsTelemetryInitializer.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        public void Initialize(ITelemetry telemetry)
+        {
+            if (telemetry == null)
+            {
+                return;
+            }
+
+            // RoleName is the app name
+            telemetry.Context.Cloud.RoleName = Environment.GetEnvironmentVariable(AzureWebsiteName) ?? "[Unknown]";
```
SergeyKanzhelev: I always confusing those. Is this one that has slot name suffix or not? Ideally - include slot name suffix

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Constants/LoggingKeys.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Host.Loggers
+{
+    internal static class LoggingKeys
+    {
+        // These are publicly visible as property names or prefixes
```
SergeyKanzhelev: You have some extra alongside the code. Like `[Unknwon]` for the function name. Consider including here

src/Microsoft.Azure.WebJobs.Host/WebJobs.Host.csproj
```diff
@@ -54,6 +54,14 @@
     <AssemblyOriginatorKeyFile>..\Common\PublicKey.snk</AssemblyOriginatorKeyFile>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Microsoft.AI.PerfCounterCollector, Version=2.2.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\Microsoft.ApplicationInsights.PerfCounterCollector.2.2.0\lib\net45\Microsoft.AI.PerfCounterCollector.dll</HintPath>
```
SergeyKanzhelev: Consider including [Microsoft.ApplicationInsights.WindowsApps](https://www.nuget.org/packages/Microsoft.ApplicationInsights.WindowsApps/) and using `Microsoft.ApplicationInsights.WindowsServer.UnhandledExceptionTelemetryModule` and `Microsoft.ApplicationInsights.WindowsServer.UnobservedExceptionTelemetryModule`. 

https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/8c51d66c08c1baf0844613637fa5d3ccfc7590f9/Test/Web/FunctionalTests/TestApps/ConsoleAppFW40/ApplicationInsights.config#L9-L10

Those will handle everything that crashes the process. In future I'd recommend also to subscribe to `FirstChanceExceptionStatisticsTelemetryModule`, but this is only available in beta right now

src/Microsoft.Azure.WebJobs.Host/WebJobs.Host.csproj
```diff
@@ -952,6 +1060,7 @@
       <Link>Properties\PublicKey.snk</Link>
     </None>
     <None Include="app.config" />
+    <None Include="ApplicationInsights.config" />
```
SergeyKanzhelev: You do not need this one. Please remove to avoid possible strange bugs of unexpected telemetry configuration.

src/Microsoft.Azure.WebJobs.Host/app.config
```diff
@@ -1,23 +1,23 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
SergeyKanzhelev: Just in case - redirect Application Insights of earlier versions to the latest to avoid conflicts with the customer code

src/Microsoft.Azure.WebJobs.Host/packages.config
```diff
@@ -1,17 +1,67 @@
...
   <package id="System.Spatial" version="5.8.2" targetFramework="net45" />
+  <package id="System.Text.Encoding" version="4.3.0" targetFramework="net46" />
+  <package id="System.Text.Encoding.Extensions" version="4.3.0" targetFramework="net46" />
+  <package id="System.Text.RegularExpressions" version="4.3.0" targetFramework="net46" />
+  <package id="System.Threading" version="4.3.0" targetFramework="net46" />
+  <package id="System.Threading.Tasks" version="4.3.0" targetFramework="net46" />
+  <package id="System.Threading.Timer" version="4.3.0" targetFramework="net46" />
+  <package id="System.Xml.ReaderWriter" version="4.3.0" targetFramework="net46" />
+  <package id="System.Xml.XDocument" version="4.3.0" targetFramework="net46" />
   <package id="WindowsAzure.Storage" version="7.2.1" targetFramework="net45" />
```
SergeyKanzhelev: You may need to update targetFramework for those as well... not sure why you left them out

SergeyKanzhelev:
I finished with review. Let me know if you'll need me to review this again after you'll do changes.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/AppInsights/AppInsightsLogger.cs
brettsam: This is this odd world we're in where you can have an `HttpTrigger` function. That function has a name -- like `MyHttpTrigger` -- but it is also registered as an HTTP endpoint with one or more verbs allowed. I'm currently using the name (`MyHttpTrigger`) as the Request name. This was my way to capture the verb somewhere as well. The user could change the route mappings and have the URI/verb change, but the `Name` would remain the same.

Do you think that's an appropriate approach?

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/AppInsights/AppInsightsLogger.cs
brettsam: This is another part where some of our "Functions as HttpTriggers" scenarios have interesting properties. Functions may succeed -- as in they did exactly what they were supposed to -- yet return something like a 404. Right now I call that a "Success". They may even return a 500 yet I log it as a "Success" because technically the Function did everything right and didn't throw unexpectedly. Let me send a mail with all of these scenarios to the team and see what they think.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/AppInsights/AppInsightsLogger.cs
SergeyKanzhelev: We try to set a name to something that distinct the "group of executions". Like if you have a controller and action and they'll return a list (fast query) by `GET` and insert into the table by `PUT` (which is slower) we'd like to split those two by name.

If this is the case with functions you may consider doing the same.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/AppInsights/AppInsightsLogger.cs
brettsam: Great point. It looks like you guys first check for `X-Forwarded-For`, and then fall back to looking in the `HttpContext`. Is that right?

SergeyKanzhelev: Yes. For Azure Web Apps IP is always in that header. Make sure to cut out the port - if I'm not mistaken it may be an issue with web apps's load balancer

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/AppInsights/WebJobsTelemetryInitializer.cs
brettsam: It looks like it's not. I created a staging slot and my WEBSITE_SITE_NAME matched on both slots. I'll follow along with what you guys do here: https://github.com/Microsoft/ApplicationInsights-dotnet-server/blob/91016d62f3181e10d4cf589ef8fd64dadb6b54a2/Src/WindowsServer/WindowsServer.Shared/AzureWebAppRoleEnvironmentTelemetryInitializer.cs#L17

SergeyKanzhelev: Yeah, I complained about it already: https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/248 Not sure it was properly addressed.

If you just follow the suite - you can just enable `AzureWebAppRoleEnvironmentTelemetryInitializer` on your telemetry configuration by doing 
`config.TelemetryInitializers.Add(new AzureWebAppRoleEnvironmentTelemetryInitializer())` so you'd avoid code duplication and the implementation will be aligned with what we are doing for web apps.

brettsam: That's what I was just about to do -- it looks like we want exactly that behavior. I'll just add the existing initializer.

src/Microsoft.Azure.WebJobs.Host/WebJobs.Host.csproj
brettsam: This won't work for Web Apps though, will it? If someone is using WebJobs, which runs as a console app, it'll work, but it wouldn't get hit when we run this in the context of a Web app like we do in Functions. I'm going to punt this -- we can probably come up with a solution to log all critical crashing failures in a follow-up.

SergeyKanzhelev: UnobservedTaskException can happen when you forgot to add `.ContinueWith`. So it may be useful. You can postpone it to the next version

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -205,7 +223,7 @@ private static void ApplyHttpRequestProperties(RequestTelemetry requestTelemetry
                 return;
             }
 
-            requestTelemetry.Url = request.RequestUri;
+            requestTelemetry.Url = new Uri(request.RequestUri.GetLeftPart(UriPartial.Path));
```
SergeyKanzhelev: I commented on `.Name` to be path. `.Url` is supposed to be a full url or any other data arguments for the request call (if not http).

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ApplicationInsights/TelemetryClientFactory.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            QuickPulseTelemetryModule quickPulse = new QuickPulseTelemetryModule();
+            quickPulse.Initialize(config);
+            quickPulse.RegisterTelemetryProcessor(processor);
+
+            // Plug in perf counters
+            PerformanceCollectorModule perfCounterCollectorModule = new PerformanceCollectorModule();
+            perfCounterCollectorModule.Initialize(config);
+
+            return config;
```
SergeyKanzhelev: since you included ServerTelemetryChannel package - consider overriding `config.TelemetryChannel` with the instance of it for retries and local storage on failure... Do not forget to call `.Initialize` on that channel. Sorry didn't comment on it earlier

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ApplicationInsights/WebJobsTelemetryInitializer.cs
```diff
@@ -0,0 +1,55 @@
...
+        private const string ComputerNameKey = "COMPUTERNAME";
+
+        public void Initialize(ITelemetry telemetry)
+        {
+            if (telemetry == null)
+            {
+                return;
+            }
+
+            telemetry.Context.Cloud.RoleInstance = GetRoleInstanceName();
```
SergeyKanzhelev: please cache this value in class's private field. If you need to re-read it periodically - cache it only for certain time

sample/SampleHost/SampleHost.csproj
```diff
@@ -34,18 +34,116 @@
...
+    </Reference>
+    <Reference Include="System.IO.FileSystem, Version=4.0.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\System.IO.FileSystem.4.3.0\lib\net46\System.IO.FileSystem.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="System.IO.FileSystem.Primitives, Version=4.0.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\System.IO.FileSystem.Primitives.4.3.0\lib\net46\System.IO.FileSystem.Primitives.dll</HintPath>
+      <Private>True</Private>
+    </Reference>
+    <Reference Include="System.Net.Http, Version=4.1.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
```
mathewc: This is one of the suspect packages I mentioned in email

src/Microsoft.Azure.WebJobs.Host/Bindings/Logger/LoggerBinding.cs
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        public Task<IValueProvider> BindAsync(BindingContext context)
+        {
+            if (context == null)
+            {
+                throw new ArgumentNullException("context");
+            }
+
+            ILogger logger = _loggerFactory.CreateLogger(LoggingCategories.Function);
```
mathewc: This has to be created anew on each bind, because the logger instance needs to be lifetime scoped to the function invocation?

src/Microsoft.Azure.WebJobs.Host/Bindings/TraceWriter/TraceWriterBinding.cs
```diff
@@ -45,11 +49,21 @@ public Task<IValueProvider> BindAsync(BindingContext context)
...
+                    ILogger logger = _loggerFactory.CreateLogger(LoggingCategories.Function);
+                    tracer = new CompositeTraceWriter(new[] { trace, new LoggerTraceWriter(context.Trace.Level, logger) }, null, context.Trace.Level);
+                }
+                else
+                {
+                    // bind directly to the context TraceWriter
+                    tracer = trace;
+                }
             }
             else
```
mathewc: The else case handles bindings to TextWriter which most of our customers still use. Don't we want those logs to also go to AppInsights?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -84,9 +93,13 @@ public HostOutputMessage HostOutputMessage
                 StartTime = functionStartedMessage.StartTime.DateTime
             };
 
+            Stopwatch sw = Stopwatch.StartNew();
             try
             {
-                functionStartedMessageId = await ExecuteWithLoggingAsync(functionInstance, functionStartedMessage, fastItem, parameterLogCollector, functionTraceLevel, cancellationToken);
+                using (_logger?.BeginScope(CreateLogScope(functionInstance.Id, functionInstance.FunctionDescriptor.Method.Name)))
```
mathewc: I wonder if you shouldn't add your own WebJobs specific extension methods, so you can simply do _logger?.BeginFunctionScope(...), since I see this pattern below as well.

src/Microsoft.Azure.WebJobs.Host/app.config
brettsam: The real change there will come in the Web.config of our Script repo. I'll make sure to do this.

src/Microsoft.Azure.WebJobs.Host/Bindings/Logger/LoggerBinding.cs
brettsam: The `CreateLogger` method name is a little misleading as a new logger is not actually created each time -- LoggerFactory maintains a cache of loggers using the category as the key: https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging/LoggerFactory.cs#L17

src/Microsoft.Azure.WebJobs.Host/Bindings/TraceWriter/TraceWriterBinding.cs
brettsam: Fixed in latest.

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostContext.cs
```diff
@@ -17,20 +18,26 @@ internal sealed class JobHostContext : IDisposable
         private readonly IListener _listener;
         private readonly TraceWriter _trace;
         private readonly IAsyncCollector<FunctionInstanceLogEntry> _fastLogger; // optional
+        private readonly IAsyncCollector<FunctionResultLog> _aggregator;
```
MikeStall: Do we need both of these? can we get rid of _fastLogger and just use _aggregator?

src/Microsoft.Azure.WebJobs.Host/Loggers/UpdateParameterLogCommand.cs
```diff
@@ -18,11 +19,12 @@ internal sealed class UpdateParameterLogCommand : IRecurrentCommand
         private readonly IReadOnlyDictionary<string, IWatcher> _watches;
         private readonly IStorageBlockBlob _parameterLogBlob;
         private readonly TraceWriter _trace;
+        private readonly ILogger _logger;
 
         private string _lastContent;
 
         public UpdateParameterLogCommand(IReadOnlyDictionary<string, IWatcher> watches,
-            IStorageBlockBlob parameterLogBlob, TraceWriter trace)
+            IStorageBlockBlob parameterLogBlob, TraceWriter trace, ILogger logger)
```
MikeStall: >TraceWriter trace, ILogger logger [](start = 48, length = 33)

I see this pair getting passed around a lot. Can we get rid of TraceWriter?

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Aggregator/FunctionResultLog.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs.Host.Loggers
+{
+    internal class FunctionResultLog
```
MikeStall: >FunctionResultLog [](start = 19, length = 17)

Can this be merged with the existing FunctionInstanceLogEntry?

src/Microsoft.Azure.WebJobs.Host/Loggers/UpdateParameterLogCommand.cs
brettsam: That's the plan eventually. I did it this way so it would be fairly easy to rip out TraceWriter and have ILogger live in its place. That change will have to happen in a major release since it's a big break. I believe we wanted to do it when we move everything to .NET Core.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
brettsam: Good idea. Done.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ApplicationInsights/TelemetryClientFactory.cs
brettsam: Thanks. Just read this: http://apmtips.com/blog/2015/09/03/more-telemetry-channels/ for some background. That definitely looks like a better option for us.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ApplicationInsights/ITelemetryClientFactory.cs
```diff
@@ -0,0 +1,24 @@
...
+using Microsoft.ApplicationInsights;
+using Microsoft.ApplicationInsights.WindowsServer.Channel.Implementation;
+
+namespace Microsoft.Azure.WebJobs.Host.Loggers
+{
+    /// <summary>
+    /// Interface for creating <see cref="TelemetryClient"/> instances.
+    /// </summary>
+    [CLSCompliant(false)]
+    public interface ITelemetryClientFactory
```
MikeStall: What's the difference between this and ILoggerProvider?

src/Microsoft.Azure.WebJobs.Host/packages.config
brettsam: When you upgrade, Nuget checks to see if they need to be updated (which is only required if the package has a net46 version). Otherwise, you can leave them alone -- changing to net46 would still bring in the same (highest available) package.

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostContext.cs
brettsam: I wanted them to live SxS for the time being. I didn't want to require it to be a "choose one logging strategy" just yet.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Aggregator/FunctionResultLog.cs
brettsam: I remember looking at it and if I recall, it looked kinda ugly, but let me take another look since it's been a while.

brettsam: I'm not sure why I didn't do that. I can add `Duration` and `Success` (or even figure out `Success` by looking at `ErrorDetails`) and just get rid of `FunctionResultLog`. I'll do that.

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostContext.cs
MikeStall: But both of these are just publish endpoints from the runtime (and then Script or somebody else consumes it); and they're publishing semantically same information.  
So we're not forcing a choice. 



---
In reply to: [112053494](https://github.com/Azure/azure-webjobs-sdk/pull/1059#discussion_r112053494) [](ancestors = 112053494)

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ApplicationInsights/ITelemetryClientFactory.cs
brettsam: The 'ApplicationInsightsLoggerProvider` is part of the `Microsoft.Extensions.Logging` infrastructure and I kept it internal just to limit the exposed surface area.

`ITelemetryClientFactory` is public and allows you to provide your own `TelemetryClient` for use with the provider. Script uses this to override a couple of things (notably the `SdkVersion` so the AppInsights team can track usage by WebJobs and Functions differently).

I first made the Provider public and gave it a virtual `CreateTelemetryClient()`, but you're not allowed to call virtual methods from within a constructor. I could have lazily-called the method when it was first used, but thought this was simpler and allowed everything to be initialized earlier.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ApplicationInsights/TelemetryClientFactory.cs
```diff
@@ -0,0 +1,92 @@
...
+using Microsoft.ApplicationInsights.WindowsServer.Channel.Implementation;
+using Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel;
+
+namespace Microsoft.Azure.WebJobs.Host.Loggers
+{
+    /// <summary>
+    /// Creates a <see cref="TelemetryClient"/> for use by the <see cref="ApplicationInsightsLogger"/>. 
+    /// </summary>
+    [CLSCompliant(false)]
+    public class TelemetryClientFactory : ITelemetryClientFactory
```
mathewc: nit: for these default factory implementations, perhaps we should prefix them with "Default"? e.g. DefaultTelemetryClientFactory? That's a pattern we follow elsewhere. Also, does this need to be public?

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ApplicationInsights/WebJobsTelemetryInitializer.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.Collections.Generic;
+using Microsoft.ApplicationInsights.Channel;
+using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.ApplicationInsights.Extensibility;
+
+namespace Microsoft.Azure.WebJobs.Host.Loggers
+{
+    internal class WebJobsTelemetryInitializer : ITelemetryInitializer
+    {
+        private const string WebsiteInstanceIdKey = "WEBSITE_INSTANCE_ID";
```
mathewc: Maybe add this to WebSitesKnownKeyNames?

src/Microsoft.Azure.WebJobs.Host/Queues/QueueProcessorFactoryContext.cs
```diff
@@ -32,17 +35,19 @@ public QueueProcessorFactoryContext(CloudQueue queue, TraceWriter trace, CloudQu
             Queue = queue;
             PoisonQueue = poisonQueue;
             Trace = trace;
+            Logger = loggerFactory?.CreateLogger(LoggingCategories.CreateTriggerCategory("Queue"));
```
mathewc: What is the benefit of scoping these logs to "Queue"? Also, I don't see you doing this for other triggers like blob/servicebus?

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Constants/ScopeKeys.cs
```diff
@@ -0,0 +1,21 @@
...
+        // These are used internally for passing values via scopes
+        public const string FunctionInvocationId = "MS_FunctionInvocationId";
+        public const string FunctionName = "MS_FunctionName";
+        public const string HttpRequest = "MS_HttpRequest";
+
+        // HTTP context is set automatically by ASP.NET, this isn't ours.
+        internal const string HttpContext = "MS_HttpContext";
+
+        // This is set by Functions
+        internal const string FunctionsHttpResponse = "MS_AzureFunctionsHttpResponse";
```
mathewc: Thus far, we've had the flexibility of renaming this key, since it is currently only used internally in Functions. If we do rename it in the future, how can we have a test that will catch that?

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/LoggerConfiguration.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Linq;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Host.Loggers
+{
+    /// <summary>
+    /// Provides a filter for use with an <see cref="ILogger"/>.
+    /// </summary>
+    [System.CLSCompliant(false)]
+    public class LogCategoryFilter
```
mathewc: Looks like the file name is incorrect for this file?

sample/SampleHost/Program.cs
```diff
@@ -20,6 +23,23 @@ static void Main(string[] args)
...
 
             config.CreateMetadataProvider().DebugDumpGraph(Console.Out);
+            config.Tracing.ConsoleLevel = TraceLevel.Off;
+
+            // Build up a LoggerFactory to log to App Insights, but only if this key exists.
+            string instrumentationKey = Environment.GetEnvironmentVariable("APPINSIGHTS_INSTRUMENTATIONKEY");
+            if (!string.IsNullOrEmpty(instrumentationKey))
+            {
+                // build up log levels for any category
+                LogCategoryFilter filterBuilder = new LogCategoryFilter();
```
mathewc: Since the type name is LogCategoryFilter, calling the local filterBuilder seems like a misnomer. If the type is a filter and not a builder/factory, perhaps you should pass it directly into .AddApplicationInsights below, rather than passing in filterBuilder.Filter?

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/LoggerConfiguration.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// <see cref="CategoryLevels"/> and return true if the level is equal to or greater than the filter. If
+        /// there is no match, the value of <see cref="DefaultLevel"/> is used.
+        /// </summary>
+        /// <param name="category">The category of the current log message.</param>
+        /// <param name="level">The <see cref="LogLevel"/> of the current log message.</param>
+        /// <returns>True if the level is equal to or greater than the matched filter. Otherwise, false.</returns>
+        public bool Filter(string category, LogLevel level)
+        {
+            // find the longest loglevel that matches the category
+            IEnumerable<string> matches = CategoryLevels.Keys.Where(k => category.StartsWith(k, System.StringComparison.CurrentCulture));
```
mathewc: Why are we doing this prefix search rather than a direct match? Is this to allow for example a catetory level like "A.B" to apply to "A.B.X" and "A.B.Y"?

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/LoggerConfiguration.cs
mathewc: Rather than a filter interface that is specific to category, is there a way to make this a general filter that has access to category and other properties of the log event "object"? I'm anticipating that in the future people may want to filter on something other than just category. Does the ILogger fx have a built in notion of filters?

E.g. if this filter implemented ILogger, it would receive all event properties and could filter as needed, skipping any logs as needed. I believe we've used a similar pattern with TraceWriter and binding providers.

sample/SampleHost/Program.cs
```diff
@@ -20,6 +23,23 @@ static void Main(string[] args)
             }
 
             config.CreateMetadataProvider().DebugDumpGraph(Console.Out);
+            config.Tracing.ConsoleLevel = TraceLevel.Off;
+
+            // Build up a LoggerFactory to log to App Insights, but only if this key exists.
+            string instrumentationKey = Environment.GetEnvironmentVariable("APPINSIGHTS_INSTRUMENTATIONKEY");
+            if (!string.IsNullOrEmpty(instrumentationKey))
+            {
+                // build up log levels for any category
```
mathewc: "build up log levels for various categories"?

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Aggregator/FunctionResultLog.cs
brettsam: Upon further inspection, doing this will require me to add a new interface like `IAggregator : IAsyncCollector<FunctionInstanceLogEntry>`, with no new members. That would let me continue to differentiate between the `FastLogger` and the `Aggregator`. Otherwise, you can only have one that matches that interface registered with the `ServiceProvider`. It's less code, so maybe it's better that way -- and it does make it easier to find all your aggregators since there's a specific interface.

src/Microsoft.Azure.WebJobs.Host/Loggers/UpdateParameterLogCommand.cs
mathewc: I think it will be a long time before we end up removing TraceWriter, so we'll be living with this for a while. It would be good if we can find a way to remove all this duplication until then,

src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonListener.cs
```diff
@@ -33,6 +36,7 @@ public SingletonListener(MethodInfo method, SingletonAttribute attribute, Single
             _lockId += ".Listener";
 
             _trace = trace;
+            _logger = loggerFactory?.CreateLogger(LoggingCategories.Singleton);
```
mathewc: Why do we need a special logger for singleton log entries?

sample/SampleHost/SampleHost.csproj
mathewc: So what is the resolution here - ngen is no longer an issue for these assemblies?

brettsam: Right. We redirect back to 4.0.0. See mails I've sent -- we're already doing it for the Roslyn update in production (as of last week).

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostContext.cs
brettsam: It's temporary. We can't just delete support for loggers -- next major version update we'll pull what we can.

src/Microsoft.Azure.WebJobs.Host/Loggers/UpdateParameterLogCommand.cs
brettsam: I thought we'd decided on next major version.

sample/SampleHost/Program.cs
brettsam: should be "build up log filters"

sample/SampleHost/Program.cs
brettsam: It used to be called FilterBuilder in a previous iteration. Someone recommended I rename it. I can change that local.

I want to keep it generic enough so that everyone doesn't have to build their own Func from scratch. If you look at other Loggers, they all take `Func<string, LogLevel, bool>`. You can see that here, where I pass the Filter in. It makes it usable in other standard loggers.

Some examples: 
- https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs#L33
- https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Debug/DebugLogger.cs#L31

sample/SampleHost/Program.cs
```diff
@@ -20,6 +23,23 @@ static void Main(string[] args)
...
+                LogCategoryFilter filterBuilder = new LogCategoryFilter();
+                filterBuilder.DefaultLevel = LogLevel.Debug;
+                filterBuilder.CategoryLevels[LoggingCategories.Function] = LogLevel.Debug;
```
brettsam: Sure.

sample/SampleHost/packages.config
```diff
@@ -1,4 +1,58 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
+  <package id="Microsoft.Extensions.Configuration.Abstractions" version="1.1.1" targetFramework="net46" />
```
brettsam: Yup.

src/Microsoft.Azure.WebJobs.Host/Exceptions/RecoverableException.cs
```diff
@@ -71,14 +72,16 @@ public override void GetObjectData(SerializationInfo info, StreamingContext cont
         /// Recovers if exception is handled by trace pipeline, else throws.
         /// </summary>
         /// <param name="trace"></param>
-        internal void TryRecover(TraceWriter trace)
+        /// <param name="logger"></param>
+        internal void TryRecover(TraceWriter trace, ILogger logger)
```
brettsam: See email I sent with a bigger explanation.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -27,11 +28,16 @@ internal class FunctionExecutor : IFunctionExecutor
         private readonly IWebJobsExceptionHandler _exceptionHandler;
         private readonly TraceWriter _trace;
         private readonly IAsyncCollector<FunctionInstanceLogEntry> _fastLogger;
+        private readonly ILogger _logger;
```
brettsam: ILogger has a different design than TraceWriter so you have to think about it a little differently. Every unique category gets its own logger. It's different than TraceWriter where each Trace can get its own category and you only use one TraceWriter.

So one of these logs Result details -- which are unique and handled special and require their own category -- and one is for the Executor itself.

brettsam: And yes, this is one of the most confusing classes wrt logging. There's so much going on here that we'll need to pull some of it in the next major release.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ApplicationInsights/ApplicationInsightsLoggerExtensions.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// <summary>
+        /// Registers an <see cref="ApplicationInsightsLoggerProvider"/> with an <see cref="ILoggerFactory"/>.
+        /// </summary>
+        /// <param name="loggerFactory">The factory.</param>
+        /// <param name="instrumentationKey">The Application Insights instrumentation key.</param>
+        /// <param name="filter">A filter that returns true if a message with the specified <see cref="LogLevel"/>
+        /// and category should be logged. You can use <see cref="LogCategoryFilter.Filter(string, LogLevel)"/>
+        /// or write a custom filter.</param>
+        /// <returns>A <see cref="ILoggerFactory"/> for chaining additional operations.</returns>
+        public static ILoggerFactory AddApplicationInsights(
```
brettsam: Yes, LoggerFactory extensions all use `Add...`

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ApplicationInsights/ApplicationInsightsLoggerProvider.cs
```diff
@@ -0,0 +1,34 @@
...
+            }
+
+            _client = clientFactory.Create(instrumentationKey, samplingSettings);
+            _filter = filter;
+        }
+
+        public ILogger CreateLogger(string categoryName) => new ApplicationInsightsLogger(_client, categoryName, _filter);
+
+        public void Dispose()
+        {
```
brettsam: Nope.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ApplicationInsights/ApplicationInsightsScope.cs
```diff
@@ -0,0 +1,75 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+
+namespace Microsoft.Azure.WebJobs.Host.Loggers
+{
+    internal class ApplicationInsightsScope
```
brettsam: I'm actually going to make it more generic to be like `PropertyBagScope` or something. It has nothing to do with App Insights when you look at what it does.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ApplicationInsights/TelemetryClientFactory.cs
brettsam: Yes, it needs to be public b/c I override it in Script. Good point about `Default`. I do that elsewhere and just missed it here.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ApplicationInsights/WebJobsTelemetryInitializer.cs
brettsam: Yeah -- I didn't know that existed.

src/Microsoft.Azure.WebJobs.Host/Queues/QueueProcessorFactoryContext.cs
brettsam: I don't think there's any logging from other triggers. I want someone to be able to filter to *only* see Queue triggers if they're trying to debug something. So you'd turn on Trace-level (very verbose) logging for your Queues and then you could look through those logs without getting swamped.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Constants/ScopeKeys.cs
brettsam: We'll have a test in Script. There's really no way to do it here.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/LoggerConfiguration.cs
brettsam: Yeah, not sure what happened there.

brettsam: Nothing in the .NET Core library does that -- I didn't want to get too off in the weeds with a design that only applies to us. I want to be able to pass this Filter to the `ConsoleLogger` and `DebugLogger` and the other standard loggers, like I do in the sample. There's nothing inherently built-in, but the pattern is to takes a `Func<string, LogLevel, bool>` as a filter, which is what we produce here.

Nothing is stopping us from adding new ways to filter in the future -- this is just matching with the current patterns already in place.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/LoggerConfiguration.cs
brettsam: Yeah, it lets you control your filtering without having to know every category. You can turn on/off all Host logs or mute certain broad categories.

src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonListener.cs
brettsam: It didn't really fit in the other buckets. We may need a "Host.General" or something for messages that the host writes during normal execution like this. But since this was the only thing that fit that category, I decided to give it its own category.

These categories are one thing that we need to iterate on -- they don't have to be perfect right now.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
mamaso: Yep, I was thinking we could have the executor emit events (started, completed, completedWithError) and each logger could subscribe to them in order to move all this out of the executor. Killing off tracewriter would be a big help.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -0,0 +1,345 @@
...
+                        }
+
+                        // do nothing otherwise
+                        break;
+                }
+            }
+
+            foreach (KeyValuePair<string, double> metric in metrics)
+            {
+                _telemetryClient.TrackMetric($"{functionName} {metric.Key}", metric.Value);
```
mamaso: Have you thought any more about function name / metric.Key as individual columns, using MetricsManager or applying min, max, avg properties manually to the MetricEvent?

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ApplicationInsights/TelemetryClientFactory.cs
```diff
@@ -0,0 +1,92 @@
...
+            return client;
+        }
+
+        internal static string GetAssemblyFileVersion(Assembly assembly)
+        {
+            AssemblyFileVersionAttribute fileVersionAttr = assembly.GetCustomAttribute<AssemblyFileVersionAttribute>();
+            return fileVersionAttr?.Version ?? LoggingConstants.Unknown;
+        }
+
+        internal static TelemetryConfiguration InitializeConfiguration(string instrumentationKey,
```
mamaso: How do we plan to expose configuration options for http logging, perf counters, adaptive sampling in functions?

Also, I think we need to be careful if we enable adaptive sampling transparently by default (and perhaps other options) - it might surprise customers.

brettsam: I've added config settings to let you turn *some* stuff on/off (like sampling), with the idea that we'll open it up with more knobs in the future. We run the risk of having a very complicated config section, though.

As for adaptive sampling, we've been going back-and-forth. With it off, there's a pretty legitimate risk of someone turning on App Insights and hitting their limit within an hour. With it on, they'll lose some specific records and see partial data (Chris hit this in a demo, actually). So far we keep coming down on the side of "enabled by default" to prevent the throttling. But let me grab a few people and see if we can all agree.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ApplicationInsights/ApplicationInsightsLogger.cs
brettsam: I was about to incorporate it last week and then realized there may be drawbacks with building charts in the Portal (I've never seen it display these details before). I just reached out to Sergey about this, so let's see what he says.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ApplicationInsights/ApplicationInsightsScope.cs
brettsam: Made it `DictionaryLoggerScope`

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Aggregator/AggregatorConfiguration.cs
```diff
@@ -0,0 +1,30 @@
...
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// Configuration options for function result aggregation.
+    /// </summary>
+    public class AggregatorConfiguration
```
brettsam: Done. Named it `FunctionResultAggregatorConfiguration` to match with other classes.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Aggregator/AggregatorConfiguration.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// <summary>
+        /// Gets or sets a value indicating whether the aggregator is enabled.
+        /// </summary>
+        public bool IsEnabled { get; set; } = true;
+
+        /// <summary>
+        /// Gets or sets a value indicating the the maximum batch size for aggregations. When this number is hit, the results are
+        /// aggregated and sent to every registered <see cref="ILogger"/>.
+        /// </summary>
+        public int BatchSize { get; set; } = 1000;
```
brettsam: Done.

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
```diff
@@ -161,12 +162,34 @@ public static ServiceProviderWrapper CreateStaticServices(this JobHostConfigurat
...
             var blobsConfiguration = config.Blobs;
 
             TraceWriter trace = services.GetService<TraceWriter>();
-            IAsyncCollector<FunctionInstanceLogEntry> fastLogger = services.GetService<IAsyncCollector<FunctionInstanceLogEntry>>();                               
+            IAsyncCollector<FunctionInstanceLogEntry> registeredFunctionEventCollector = services.GetService<IAsyncCollector<FunctionInstanceLogEntry>>();
+
+            if (registeredFunctionEventCollector != null && aggregator != null)
+            {
+                // If there are both an aggregator and a registered FunctionEventCollector, wrap them in a composite
+                functionEventCollector = new CompositeFunctionEventCollector(new[] { registeredFunctionEventCollector, aggregator });
```
MikeStall: >CompositeFunctionEventCollector [](start = 45, length = 31)

Can this be in Script (ie, up the stack)? IE, the SDK is a produce that just emits events to the registered IAsyncCollector<FunctionInstanceLogEntry>. 
Script is a consumer that pulls these events and does whatever it needs (pushes to fast logger; pushes to AppInsight; etc)

MikeStall:
:shipit:

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Aggregator/FunctionResultAggregate.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            return new ReadOnlyDictionary<string, object>(new Dictionary<string, object>
+            {
+                [LoggingKeys.Name] = Name,
+                [LoggingKeys.Count] = Count,
+                [LoggingKeys.Timestamp] = Timestamp,
+                [LoggingKeys.AvgDuration] = AverageDuration,
+                [LoggingKeys.MaxDuration] = MaxDuration,
+                [LoggingKeys.MinDuration] = MinDuration,
+                [LoggingKeys.Successes] = Successes,
```
brettsam: I'll come back to this. There will be some naming cleanup as we get feedback. That'll be part of our 'Preview' period.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ApplicationInsights/ApplicationInsightsLogger.cs
brettsam: Update -- it's recommended that we move towards this. I'll do this during the Preview period while we're adjusting our data.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -0,0 +1,345 @@
...
+
+            // Log a function result
+            if (_categoryName == LoggingCategories.Results)
+            {
+                LogFunctionResult(stateValues, exception);
+                return;
+            }
+
+            // Log an aggregate record
+            if (_categoryName == LoggingCategories.Aggregator)
```
brettsam: Yeah, could be. I'll come back and clean that up in a subsequent checkin.

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
```diff
@@ -62,6 +63,7 @@ public JobHostConfiguration(string dashboardAndStorageConnectionString)
             }
 
             Singleton = new SingletonConfiguration();
+            Aggregator = new AggregatorConfiguration();
```
brettsam: I had that originally -- but then realized there's really no default logging configuration at the moment, so I was left with just the Aggregation config. I didn't think about putting the LoggingFactory on there, though (it's currently on the HostConfig itself). This is a good idea -- I can come back and clean that up in the next PR if you're okay with it.

xt0rted:
Any idea when this will make it onto NuGet?

Issue #893, opened 2016-11-02
fwd link for missing connection string is wrong in Dashboard


christopheranderson:
Currently we point to a bad link for a bad connection string error in the Dashboard.

Example:
> Make sure that you are setting a connection string named AzureWebJobsDashboard in your Microsoft Azure Website configuration by using the following format DefaultEndpointsProtocol=https;AccountName=NAME;AccountKey=KEY pointing to the Microsoft Azure Storage account where the Microsoft Azure WebJobs Runtime logs are stored.
Please visit the article [about configuring connection strings](https://go.microsoft.com/fwlink/?linkid=320957) for more information on how you can configure connection strings in your Microsoft Azure Website.

We should point to this one: https://go.microsoft.com/fwlink/?linkid=834346

FinVamp1:
Issued https://github.com/Azure/azure-webjobs-sdk/pull/1068 for this issue.

PR #1068, opened 2017-03-16
Update for Issue #893

FinVamp1:
As discussed with @christopheranderson I've fixed this issue as it's been an outlier for a while.


mamaso:
Looks good!

Issue #1004, opened 2017-02-02
Consider expanding EventHub/ServiceBus binding data to include common properties


mathewc:
This comes up often, most recently here on SO: https://stackoverflow.com/questions/42003600/azure-iot-hub-eventhub-and-functions. People need to access the event DeviceId, SequenceNumber, etc. and end up having to change their method signature to drop down to a lower EventData binding to access it.

If we include this (and other interesting properties) in the binding data, it would facilitate these scenarios. Similar to the way our Azure Queue trigger exposes binding data for dequeue count, enqueue time, and other message properties.

For languages other than C# this would be very useful, because in those languages there isn't even a workaround currently to bind to EventData. Adding these to BindingData would also allow the properties to be used in input/output binding expressions.

mathewc:
Also came up recently for BrokeredMessage.ReplyTo http://stackoverflow.com/questions/42143133/how-to-dynamically-bind-service-bus-brokeredmessage-in-an-azure-function/42143407#42143407. We should also add some of these useful BM properties to the static binding data collection as well to facilitate these scenarios.

brettsam:
This also just came up again for access to SB BrokeredMessage properties: http://stackoverflow.com/questions/42385995/azure-functions-get-service-bus-brokeredmessage-with-nodejs

prashanthmadi:
Thanks for starting this thread. I have a similar requirement to receive message properties of Service Bus queue along with message in functions app

####Sample message with its properties :

![servicebusqueue](https://cloud.githubusercontent.com/assets/1486808/24055723/3116a15a-0b0f-11e7-937f-019bcf766e0f.png)

Current implementation only provides message

Samples available at below link :
https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-service-bus#trigger-usage

mathewc:
Added first class binding data support for many of the important ServiceBus and EventHub message/event properties.

For EventHub:

- PartitionContext
- PartitionKey
- Offset
- SequenceNumber
- EnqueuedTimeUtc
- Properties
- SystemProperties

For ServiceBus:

- DeliveryCount
- DeadLetterSource
- ExpiresAtUtc
- EnqueuedTimeUtc
- MessageId
- ContentType
- ReplyTo
- SequenceNumber
- To
- Label
- CorrelationId
- Properties

michaeljqzq:
[...]

PR #1069, opened 2017-03-17
Adding ServiceBus default binding data values (#1004)

mathewc:
Following the exact same pattern that we do for QueueTrigger

### DIFF
--- a/src/Microsoft.Azure.WebJobs.ServiceBus/Triggers/ServiceBusTriggerBinding.cs
+++ b/src/Microsoft.Azure.WebJobs.ServiceBus/Triggers/ServiceBusTriggerBinding.cs
@@ -21,6 +21,7 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Triggers
         private readonly string _parameterName;
         private readonly IObjectToTypeConverter<BrokeredMessage> _converter;
         private readonly ITriggerDataArgumentBinding<BrokeredMessage> _argumentBinding;
+        private readonly IReadOnlyDictionary<string, Type> _bindingDataContract;
         private readonly ServiceBusAccount _account;
         private readonly string _namespaceName;
         private readonly string _queueName;
@@ -53,6 +54,7 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Triggers
             _parameterName = parameterName;
             _converter = CreateConverter(parameterType);
             _argumentBinding = argumentBinding;
+            _bindingDataContract = CreateBindingDataContract(argumentBinding.BindingDataContract);
             _account = account;
             _namespaceName = ServiceBusClient.GetNamespaceName(account);
             _accessRights = accessRights;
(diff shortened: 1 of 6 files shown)


src/Microsoft.Azure.WebJobs.ServiceBus/Triggers/ServiceBusTriggerBinding.cs
```diff
@@ -122,6 +127,62 @@ public Task<IListener> CreateListenerAsync(ListenerFactoryContext context)
...
+            bindingData.Add("DeadLetterSource", value.DeadLetterSource);
+            bindingData.Add("ExpiresAtUtc", value.ExpiresAtUtc);
+            bindingData.Add("EnqueuedTimeUtc", value.EnqueuedTimeUtc);
+            bindingData.Add("MessageId", value.MessageId);
+            bindingData.Add("ContentType", value.ContentType);
+            bindingData.Add("ReplyTo", value.ReplyTo);
+            bindingData.Add("SequenceNumber", value.SequenceNumber);
+            bindingData.Add("To", value.To);
+            bindingData.Add("Label", value.Label);
+            bindingData.Add("CorrelationId", value.CorrelationId);
```
brettsam: nit: could use nameof() here to make sure they all match perfectly.

mathewc:
Added tests

src/Microsoft.Azure.WebJobs.ServiceBus/Triggers/ServiceBusTriggerBinding.cs
```diff
@@ -122,6 +127,75 @@ public Task<IListener> CreateListenerAsync(ListenerFactoryContext context)
...
+
+            return contract;
+        }
+
+        internal static IReadOnlyDictionary<string, object> CreateBindingData(BrokeredMessage value,
+            IReadOnlyDictionary<string, object> bindingDataFromValueType)
+        {
+            var bindingData = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
+
+            SafeAddValue(() => bindingData.Add(nameof(value.DeliveryCount), value.DeliveryCount));
```
mathewc: BrokeredMessage is not very unit testable. Getters on local message properties will throw in many cases, because the message hasn't actually been sent/receive. E.g properties like ExpiresAt, EnqueuedTime, etc. In a full e2e test however, none of these throw.

src/Microsoft.Azure.WebJobs.ServiceBus/Triggers/ServiceBusTriggerBinding.cs
```diff
@@ -122,6 +127,75 @@ public Task<IListener> CreateListenerAsync(ListenerFactoryContext context)
             return factory.CreateAsync(context.CancellationToken);
         }
 
+        internal static IReadOnlyDictionary<string, Type> CreateBindingDataContract(IReadOnlyDictionary<string, Type> argumentBindingContract)
+        {
+            var contract = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);
+            contract.Add("DeliveryCount", typeof(int));
```
mathewc: I picked a subset of all the properties to add, based on what I think might be useful. I didn't add everything. I did add the ones customers have explicitly asked for though. Let me know if you think any others should be added.

mathewc:
I'll be doing the same for EventHub trigger in a separate PR.

mathewc:
Added a quick check to the e2e tests.

PR #1070, opened 2017-03-17
Add ServiceBusEntityType to specify whether Queue/Topic is created if not exists

### LINKED ISSUE #1052: Option to create service bus topic or queue when using attributes
When the output of a webjob function is a service bus, and a queue/topic does not already exists, there appears to be no way to specify whether a Queue or a Topic should be created.

In a environment where the sender and consumer are running in different webjobs, the startup sequence determines whether a queue of a topic is created.

#### Repro steps
Given the following sender and receiver :

**Sender**

`public async Task ProcessQueueMessage([ServiceBusTrigger("Queue1")] string message,
             [ServiceBus("Topic1")] IAsyncCollector<SomethingMessage> outputMessages)`

**Receiver**

`public async Task ProcessTopicMessage([ServiceBusTrigger("Topic1", "Subscription1")] SomethingMessage message)`

If the sender starts up first, "Topic1" is created as a queue.
but if the Receiver starts up first "Topic1" is created as a topic.

It is also worth noting that once the sender creates the queue, no topic subscribers are able to function.

#### Expected behavior

Having the ability to specify whether the output is a queue or a topic via an attribute property
```
 public async Task ProcessQueueMessage([ServiceBusTrigger("Queue1")] string message,
             [ServiceBus("Topic1", IsTopic : true)] IAsyncCollector<SomethingMessage> outputMessages)
```

or perhaps via a different attribute

```
 public async Task ProcessQueueMessage([ServiceBusTrigger("Queue1")] string message,
             [ServiceBusTopic("Topic1")] IAsyncCollector<SomethingMessage> outputMessages)
```

#### Known workarounds

On startup of the sending webjob worker, manually create topics if they dont exist

mamaso:
resolves #1052


src/Microsoft.Azure.WebJobs.ServiceBus/Bindings/ServiceBusBinding.cs
```diff
@@ -48,7 +51,8 @@ public bool FromAttribute
             {
                 Account = _account,
                 MessageSender = messageSender,
-                AccessRights = _accessRights
+                AccessRights = _accessRights,
```
mamaso: very tempted to just throw the actual attribute on ServiceBusEntity and pull AccessRights & isQueue from that.  Also fine with changing isQueue to an enum of EntityTypes or something

src/Microsoft.Azure.WebJobs.ServiceBus/Bindings/ServiceBusBinding.cs
```diff
@@ -20,16 +20,19 @@ internal class ServiceBusBinding : IBinding
...
         {
             _parameterName = parameterName;
             _argumentBinding = argumentBinding;
             _account = account;
             _namespaceName = ServiceBusClient.GetNamespaceName(account);
             _path = path;
-            _accessRights = accessRights;
-            _converter = CreateConverter(account, path, accessRights);
+            _accessRights = attr.Access;
+            _isQueue = !(attr is ServiceBusTopicAttribute);
```
MikeStall: >!(attr is ServiceBusTopicAttribute); [](start = 23, length = 36)

nit: type checks like this can get brittle - somebody else adds another derived class; or the type hierarchy gets juggled - it can easily break. 
Consider an internal virtual IsQueue property on ServiceBusAttribute.

src/Microsoft.Azure.WebJobs.ServiceBus/ServiceBusTopicAttribute.cs
```diff
@@ -0,0 +1,30 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.ServiceBus.Messaging;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <inherit/>
+    [AttributeUsage(AttributeTargets.Parameter)]
+    public sealed class ServiceBusTopicAttribute : ServiceBusAttribute
```
MikeStall: >ServiceBusTopicAttribute [](start = 24, length = 24)

What's the guidance here. Is the base class now for queues and the new derived class for topics?  (but we still need to allow the existing SBA to work on Topics for backwards compat?)

mamaso: That's what I was envisioning, I prefer having a separate attribute `ServiceBusTopic` to having an `IsTopic` flag on the main attribute constructor.

I think I should also add a `ServiceBusQueue` attribute to make things more symmetrical.

Does that sound reasonable?

src/Microsoft.Azure.WebJobs.ServiceBus/Bindings/ServiceBusBinding.cs
mamaso: great point, I'll add that

mamaso:
(close & open to get the new ci running)

mamaso:
@MikeStall added IsQueue virtual and a mirroring ServiceBusQueueAttribute

src/Microsoft.Azure.WebJobs.ServiceBus/ServiceBusQueueAttribute.cs
```diff
@@ -0,0 +1,32 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.ServiceBus.Messaging;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <inherit/>
+    [AttributeUsage(AttributeTargets.Parameter)]
+    public sealed class ServiceBusQueueAttribute : ServiceBusAttribute
```
mathewc: I don't understand why we're adding a separate attribute if you've already added a boolean hint. Isn't that sufficient for us to distinguish what we need to do internally? I really think we should avoid adding a new attribute for this.

src/Microsoft.Azure.WebJobs.ServiceBus/ServiceBusAttribute.cs
```diff
@@ -58,5 +58,8 @@ public ServiceBusAttribute(string queueOrTopicName, AccessRights access)
         /// The default is "Manage".
         /// </summary>
         public AccessRights Access { get; private set; }
+
+        // Used to choose the correct entity type to create, if necessary
+        internal virtual bool IsQueue => true;
```
mathewc: I wonder if there isn't a SB entity type enum we can use here. Or perhaps we define our own.

mamaso: *shrug* https://docs.microsoft.com/en-us/dotnet/api/microsoft.servicebus.management.portalresourcetype

src/Microsoft.Azure.WebJobs.ServiceBus/ServiceBusQueueAttribute.cs
mamaso: It's not necessary, but I think it reads better than boolean hints.

mamaso:
@mathewc Removed the new attributes (couple good reasons, doesn't match the trigger attribute semantics & @MikeStall has concerns about the binding pipeline & inheritance), went with an enum hint.

New hint doesn't match the trigger attribute either, but it's closer.

src/Microsoft.Azure.WebJobs.ServiceBus/ServiceBusEntityType.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.ServiceBus
+{
+    /// <summary>
+    /// Service Bus entity type.
+    /// </summary>
+    public enum ServiceBusEntityType
```
mathewc: Let's just name this EntityType - it's in the ServiceBus namespace so it is scoped.

src/Microsoft.Azure.WebJobs.ServiceBus/ServiceBusAttribute.cs
```diff
@@ -54,6 +55,11 @@ public ServiceBusAttribute(string queueOrTopicName, AccessRights access)
         public string QueueOrTopicName { get; private set; }
 
         /// <summary>
+        /// The name of the queue to bind to.
```
mathewc: "Value indicating the type of the entity to bind to"

src/Microsoft.Azure.WebJobs.ServiceBus/Bindings/MessageSenderExtensions.cs
```diff
@@ -52,7 +52,17 @@ internal static class MessageSenderExtensions
 
             try
             {
-                await namespaceManager.CreateQueueAsync(sender.Path);
+                switch (entityType)
```
mathewc: nit: fix up the comment above "required rights to create the queue"

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ServiceBusEndToEndTests.cs
```diff
@@ -445,6 +489,30 @@ public class ServiceBusTestJobs_RestrictedAccess : ServiceBusTestJobsBase
...
+            }
+
+            public static void SBQueueDefault(
+                [ServiceBusTrigger(StartQueueName + "1")] string message,
+                [ServiceBus(QueueNamePrefix + "1")] out string output)
+            {
+                output = "should create queue";
+            }
+
+            public static void SBQueueQueue(
```
mathewc: ? what is SBQueueQueue?

mamaso: Queue trigger -> explicit Queue output, but condensed into gibberish ;)

PR #1071, opened 2017-03-18
Adding binding data contract for EventHub trigger (#1004)

mathewc:
For issue https://github.com/Azure/azure-webjobs-sdk/issues/1004. See related work I did for ServiceBus here: https://github.com/Azure/azure-webjobs-sdk/pull/1069. I'm also adding the `IDictionary<string, object>` property bags (`BrokeredMessage.Properties`, `EventData.Properties`, `EventData.SystemProperties`) in this PR, which has been a customer ask, particularly in the context of Azure Functions. This particularly helps unblock non C# scenarios for Functions - Node.js functions will now be able to access all these properties, whereas they couldn't before.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubTriggerInputBindingStrategy.cs
+++ b/src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubTriggerInputBindingStrategy.cs
@@ -13,25 +13,14 @@ namespace Microsoft.Azure.WebJobs.ServiceBus
     // Binding strategy for an event hub triggers. 
     internal class EventHubTriggerBindingStrategy : ITriggerBindingStrategy<EventData, EventHubTriggerInput>
     {
-        private const string DataContractPartitionContext = "partitionContext";
-
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Reliability", "CA2000:Dispose objects before losing scope")]
-        public EventHubTriggerInput ConvertFromString(string x)
+        public EventHubTriggerInput ConvertFromString(string input)
         {
-            byte[] bytes = Encoding.UTF8.GetBytes(x);
+            byte[] bytes = Encoding.UTF8.GetBytes(input);
             EventData eventData = new EventData(bytes);
 
             // Return a single event. Doesn't support multiple dispatch 
             return EventHubTriggerInput.New(eventData);            
-        }
-
-        // Get the static binding contract
-        //  - gets augmented 
-        public Dictionary<string, Type> GetStaticBindingContract()
-        {
-            Dictionary<string, Type> contract = new Dictionary<string, Type>();
-            contract[DataContractPartitionContext] = typeof(PartitionContext);
-            return contract;
         }
 
         // Single instance: Core --> EventData
@@ -41,8 +30,7 @@ namespace Microsoft.Azure.WebJobs.ServiceBus
             {
                 throw new ArgumentNullException("value");
             }
-            EventData eventData = value.GetSingleEventData();
-            return eventData;
+            return value.GetSingleEventData();
         }
 
         public EventData[] BindMultiple(EventHubTriggerInput value, ValueBindingContext context)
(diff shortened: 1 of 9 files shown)


src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubTriggerInputBindingStrategy.cs
```diff
@@ -54,7 +42,26 @@ public EventData[] BindMultiple(EventHubTriggerInput value, ValueBindingContext
...
 
-        // GEt runtime instance of binding contract 
+        // Get the static binding contract
+        //  - gets augmented 
+        public Dictionary<string, Type> GetStaticBindingContract()
+        {
+            Dictionary<string, Type> contract = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);
+            contract.Add("PartitionContext", typeof(PartitionContext));
+
+            // TODO: can we determine here if the binding is single dispatch?
```
mathewc: I'd really like to be able to determine single dispatch here - not sure how. If we can't, it won't be a huge issue - even if a customer tried to use these properties for an EventData[], it would fail at runtime.

src/Microsoft.Azure.WebJobs.ServiceBus/Triggers/ServiceBusTriggerBinding.cs
```diff
@@ -141,6 +141,7 @@ public Task<IListener> CreateListenerAsync(ListenerFactoryContext context)
             contract.Add("To", typeof(string));
             contract.Add("Label", typeof(string));
             contract.Add("CorrelationId", typeof(string));
+            contract.Add("Properties", typeof(IDictionary<string, object>));
```
mathewc: Doing the same thing for SB - this will allow people to access custom message properties.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/EventHubEndToEndTests.cs
```diff
@@ -0,0 +1,103 @@
...
+        private const string TestHubName = "testhub";
+
+        public EventHubEndToEndTests()
+        {
+            var config = new JobHostConfiguration()
+            {
+                TypeLocator = new FakeTypeLocator(typeof(EventHubTestJobs))
+            };
+            var eventHubConfig = new EventHubConfiguration();
+            eventHubConfig.AddSender(TestHubName, Environment.GetEnvironmentVariable("AzureWebJobsEventHubSender"));
```
mathewc: TODO: I'll need to get the build machines set up with these connections

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubTriggerInputBindingStrategy.cs
MikeStall: that's a good point - I'm not sure if this is possible today. The single strategy object handles both, and the caller is determining whether to call BindSingle or BindMultiple. Can you file a bug to me to followup on that?

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubTriggerInputBindingStrategy.cs
```diff
@@ -63,9 +70,34 @@ public EventData[] BindMultiple(EventHubTriggerInput value, ValueBindingContext
...
+            if (value.IsSingleDispatch)
+            {
+                EventData eventData = value.GetSingleEventData();
+                SafeAddValue(() => bindingData.Add(nameof(eventData.PartitionKey), eventData.PartitionKey));
+                SafeAddValue(() => bindingData.Add(nameof(eventData.Offset), eventData.Offset));
+                SafeAddValue(() => bindingData.Add(nameof(eventData.SequenceNumber), eventData.SequenceNumber));
+                SafeAddValue(() => bindingData.Add(nameof(eventData.EnqueuedTimeUtc), eventData.EnqueuedTimeUtc));
+                SafeAddValue(() => bindingData.Add(nameof(eventData.Properties), eventData.Properties));
+                SafeAddValue(() => bindingData.Add(nameof(eventData.SystemProperties), eventData.SystemProperties));
+            }
```
MikeStall: Is it ok to not have this binding data in the multiple case? The alternative would be include it as some array. 
Our guidance is:
- Use multiple dispatch because it's faster
- use single dispatch to get the binding data.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubTriggerInputBindingStrategy.cs
mathewc: Created issue https://github.com/Azure/azure-webjobs-sdk/issues/1072

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubTriggerInputBindingStrategy.cs
mathewc: I've updated the code such that it will use arrays in the multiple dispatch case

PR #1075, opened 2017-03-20
Add FunctionNameAttribute to WebJobs SDK

ahmelsayed:
This is for the new programming model for .NET where users would need to specify a function name as an attribute. 
The original ask was just for [`FunctionNameAttribute.cs`](https://github.com/Azure/azure-webjobs-sdk/compare/ahmels-function-name?expand=1#diff-9f2e8d02d4f5a4798dae27001ecc212b) though it seemed straight forward to get that consumed by the host itself, so I thought I'd include it here. If it's not as simple as I thought, then we can remove that. if this is indeed all that's needed I can update the comment that says this is not consumed by the SDK.

### DIFF
--- a/src/Microsoft.Azure.WebJobs/FunctionNameAttribute.cs
+++ b/src/Microsoft.Azure.WebJobs/FunctionNameAttribute.cs
@@ -1,0 +1,29 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false, Inherited = false)]
+    public sealed class FunctionNameAttribute : Attribute
+    {
+        private string _name;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="FunctionNameAttribute"/> class with a given name.
+        /// </summary>
+        /// <param name="name">Name of the function.</param>
+        public FunctionNameAttribute(string name)
+        {
+            this._name = name;
+        }
+
+        /// <summary>
+        /// Gets the function name.
+        /// </summary>
+        public string Name => _name;
+    }
+}
--- a/src/Microsoft.Azure.WebJobs/WebJobs.csproj
+++ b/src/Microsoft.Azure.WebJobs/WebJobs.csproj
@@ -63,6 +63,7 @@
     <Compile Include="BlobTriggerAttribute.cs" />
     <Compile Include="BinderExtensions.cs" />
     <Compile Include="DisableAttribute.cs" />
+    <Compile Include="FunctionNameAttribute.cs" />
     <Compile Include="GlobalSuppressions.cs" />
     <Compile Include="IAsyncCollector.cs" />
     <Compile Include="IAttributeInvokeDescriptor.cs" />
--- a/test/Microsoft.Azure.WebJobs.Host.UnitTests/PublicSurfaceTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.UnitTests/PublicSurfaceTests.cs
@@ -128,7 +128,8 @@ namespace Microsoft.Azure.WebJobs.Host.UnitTests
                 "DisableAttribute",
                 "TimeoutAttribute",
                 "TraceLevelAttribute",
-                "ODataFilterResolutionPolicy"
+                "ODataFilterResolutionPolicy",
+                "FunctionNameAttribute"
             };
 
             AssertPublicTypes(expected, assembly);


src/Microsoft.Azure.WebJobs/FunctionNameAttribute.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <param name="name">Name of the function.</param>
+        public FunctionNameAttribute(string name)
+        {
+            this._name = name;
+        }
+
+        /// <summary>
+        /// Gets the function name.
+        /// </summary>
+        public string Name
```
mamaso: nit: `public string Name => _name`;

src/Microsoft.Azure.WebJobs.Host/Indexers/MethodInfoExtensions.cs
```diff
@@ -16,7 +16,9 @@ public static string GetFullName(this MethodInfo methodInfo)
                 throw new ArgumentNullException("methodInfo");
             }
 
-            return String.Format(CultureInfo.InvariantCulture, "{0}.{1}", methodInfo.DeclaringType.FullName, methodInfo.Name);
+            return TryGetCustomName(methodInfo, out string customName)
```
mamaso: isn't it necessary to declare `customName` before you pass it as an out parameter? I think the appveyor errors are because of this, though I'd love this shorthand

mamaso (approved):

src/Microsoft.Azure.WebJobs.Host/Indexers/MethodInfoExtensions.cs
```diff
@@ -26,7 +28,15 @@ public static string GetShortName(this MethodInfo methodInfo)
                 throw new ArgumentNullException("methodInfo");
             }
 
-            return String.Format(CultureInfo.InvariantCulture, "{0}.{1}", methodInfo.DeclaringType.Name, methodInfo.Name);
+            return TryGetCustomName(methodInfo, out string customName)
+                ? customName
+                : String.Format(CultureInfo.InvariantCulture, "{0}.{1}", methodInfo.DeclaringType.Name, methodInfo.Name);
+        }
+
+        private static bool TryGetCustomName(MethodInfo methodInfo, out string customName)
```
mathewc: What guarantees are we making for this attribute as far as where the custom name takes effect? I see other places in the codebase that use function name w/o going through this helper, meaning they'd use the actual name, not attribute provided name.

src/Microsoft.Azure.WebJobs/FunctionNameAttribute.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// This is for providing a custom name to for an SDK function.
```
mathewc: Rather than introduce our own new attribute for this, I wonder if we shouldn't instead just use DisplayNameAttribute https://msdn.microsoft.com/en-us/library/system.componentmodel.displaynameattribute(v=vs.110).aspx

mathewc: Also, we'll want to verify that things work end to end when this is applied in the traditional webjobs Dashboard. E.g. if you override the display name, that name is shown, and everything still continues to work - Dashboard lookups etc. still function.

src/Microsoft.Azure.WebJobs.Host/Indexers/MethodInfoExtensions.cs
ahmelsayed: I only saw FunctionDescriptor being created in one place with a functionName, and that calls GetFullName and GetShortName. I might be wrong though.

src/Microsoft.Azure.WebJobs/FunctionNameAttribute.cs
ahmelsayed: That's a good question. I think the limitations we need for that name is that it has to be a valid fileName since, at least short term, that'll get translated into a folder with a function.json inside of it. I don't know how much sense DisplayName would make for that since it'll be in the folderName, http route, etc. also it's going to be a required attribute for an SDK function to be transformed into a function.json.

src/Microsoft.Azure.WebJobs.Host/Indexers/MethodInfoExtensions.cs
ahmelsayed: I thought I overheard Fabio saying appVeyor supported C#7. I probably miss understood, I'll fix it.

src/Microsoft.Azure.WebJobs/FunctionNameAttribute.cs
ahmelsayed: This was meant to completely replace the name the host is using when referring to this function. As I said I only added that because it looked simple enough, but if you think it's more work to get it working end to end, then I will remove it and leave just the Attribute for the tooling to consume.

src/Microsoft.Azure.WebJobs.Host/Indexers/MethodInfoExtensions.cs
mamaso: build must not be set up for C#7, I'll take a look at that

src/Microsoft.Azure.WebJobs/FunctionNameAttribute.cs
mathewc: All features we add to the core SDK are usable in the SDK outside of functions. So based on some of the changes above, it looks like you're overriding function indexing behavior, etc. We use function name in lots of other places too, including in the Dashboard, etc. So I'm just saying that if we're adding a new core SDK feature we need to verify it end to end in all the various usages.

src/Microsoft.Azure.WebJobs/FunctionNameAttribute.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// This is not actually consumed by the SDK runtime currently, but rather by function tooling
```
mathewc: Can we remove this description block? I'd rather not have any Azure Functions specific notes in here.

mathewc (approved):
One small comment

Issue #1072, opened 2017-03-20
EventHub Trigger Single/Multiple info


mathewc:
The EH trigger binding is unique from all other current triggers in that it supports multiple dispatch - i.e. binding to an array of events. However, the information on whether the binding is single or multiple is not made available yet in the pipeline in all places where it is needed (e.g. binding contract creation / resolution).

I've added a TODO in the code linking to this issue number. For now the code will always add contract members to the static contract, though we want to avoid that.

PR #1079, opened 2017-03-21
EventHub multi-dispatch binding data support

mathewc:
Completes the work started in https://github.com/Azure/azure-webjobs-sdk/pull/1071. Also addresses https://github.com/Azure/azure-webjobs-sdk/issues/1072.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Triggers/ITriggerBindingStrategy.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Triggers/ITriggerBindingStrategy.cs
@@ -36,8 +36,9 @@ namespace Microsoft.Azure.WebJobs.Host.Triggers
         /// then the properties on that type are route parameters that can feed into other bindings. 
         /// Intentionally make this mutable so that callers can add more items to it and override defaults. 
         /// </summary>
+        /// <param name="isSingleDispatch">true if this is the BindSingle contract; false if it's the BindMultiple contract</param>
         /// <returns>A map of the names of items in the binding contract and their types.</returns>
-        Dictionary<string, Type> GetStaticBindingContract();
+        Dictionary<string, Type> GetBindingContract(bool isSingleDispatch);
 
         /// <summary>
         /// Get the values of the route-parameters given an instance of the trigger value. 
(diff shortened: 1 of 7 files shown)


src/Microsoft.Azure.WebJobs.Host/Triggers/ITriggerBindingStrategy.cs
```diff
@@ -36,8 +36,9 @@ public interface ITriggerBindingStrategy<TMessage, TTriggerValue>
         /// then the properties on that type are route parameters that can feed into other bindings. 
         /// Intentionally make this mutable so that callers can add more items to it and override defaults. 
         /// </summary>
+        /// <param name="isSingleDispatch">true if this is the BindSingle contract; false if it's the BindMultiple contract</param>
         /// <returns>A map of the names of items in the binding contract and their types.</returns>
-        Dictionary<string, Type> GetStaticBindingContract();
+        Dictionary<string, Type> GetBindingContract(bool isSingleDispatch);
```
mathewc: More closely matches naming scheme used in underlying infrastructure elsewhere

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubTriggerInputBindingStrategy.cs
```diff
@@ -42,72 +42,80 @@ public EventData[] BindMultiple(EventHubTriggerInput value, ValueBindingContext
...
+            AddBindingContractMember(contract, "SystemProperties", typeof(IDictionary<string, object>), isSingleDispatch);
 
             return contract;
         }
 
         private static void AddBindingContractMember(Dictionary<string, Type> contract, string name, Type type, bool isSingleDispatch)
         {
+            if (!isSingleDispatch)
+            {
+                name += "Array";
```
mathewc: Chose a naming pattern for array based binding data. If other bindings do this in the future, we'd enforce this pattern, so we need to decide on a good name. Rather than try to apply some pluralization heuristic (always error prone), I just append "Array".

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/EventHubEndToEndTests.cs
```diff
@@ -148,10 +151,22 @@ public static void SendEvents_TestHub2(int numEvents, string input, [EventHub(Te
                 }
             }
 
-            public static void ProcessMultipleEvents([EventHubTrigger(TestHub2Name)] string[] events)
+            public static void ProcessMultipleEvents([EventHubTrigger(TestHub2Name)] string[] events,
+                string[] partitionKeyArray, DateTime[] enqueuedTimeUtcArray, IDictionary<string, object>[] propertiesArray,
```
mathewc: Here is the new binding support in action. In Azure Functions for a Node.js function, this binding data will be accessible via:

```javascript
module.exports = function (context, events) {
    for (int i = 0; i < events.length; i++
    {
        var properties = context.bindingData.propertiesArray[i];
        . . .
    }

    . . .
}
```

PR #1081, opened 2017-03-22
Fix null setting name bug

mamaso:
Fixed nameResolver.resolve(null) in `GetAppSettingResolver`

Also added null handling to ConfigurationUtility


mathewc:
Shipit

PR #1084, opened 2017-03-23
Adding Account properties onto base Azure Storage binding attributes

mathewc:
Alternate proposal for https://github.com/Azure/azure-webjobs-sdk/pull/1056 that achieves the same user facing goals, while:

- avoiding commiting to any new funky generalized attribute inheritance/merging logic. This logic remains targeted to our own StorageAccountAttribute handling. I.e. we keep this pattern internal.
- avoiding overloading the existing StorageAccountAttribute as a base class for storage binding attributes


src/Microsoft.Azure.WebJobs/StorageAttribute.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    /// <summary>
+    /// Base class for all Azure Storage binding attributes.
+    /// </summary>
+    public abstract class StorageAttribute : Attribute
+    {
+        /// <summary>
+        /// Gets the name of the Azure Storage connection string to use.
+        /// </summary>
+        public string Account { get; set; }
```
mamaso: it'd be nice if you can use `[AppSetting]` here, maybe could cut out some of the name resolver code you have.  Not sure if the storage providers are hooked up to the extensibility pipeline

src/Microsoft.Azure.WebJobs.Host/Blobs/Bindings/BlobAttributeBindingProvider.cs
```diff
@@ -60,7 +60,7 @@ internal class BlobAttributeBindingProvider : IBindingProvider
 
             string resolvedPath = Resolve(blobAttribute.BlobPath);
             IBindableBlobPath path = null;
-            IStorageAccount account = await _accountProvider.GetStorageAccountAsync(context.Parameter, context.CancellationToken, _nameResolver);
```
mamaso: are there still valid uses of the `GetStorageAccountAsync(parameterInfo ... )` call? might be able to remove it

src/Microsoft.Azure.WebJobs/StorageAttribute.cs
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// Base class for all Azure Storage binding attributes.
+    /// </summary>
+    public abstract class StorageAttribute : Attribute
```
MikeStall: >StorageAttribute [](start = 26, length = 16)

I don't understand the difference between StorageAccountAttribute and StorageAttribute. Similar names and structurally equivalent.  Why can't we use the existing attribute?

src/Microsoft.Azure.WebJobs.Host/TypeUtility.cs
```diff
@@ -98,6 +98,25 @@ internal static Type UnwrapTaskType(Type type)
             return null;
         }
 
+        internal static TAttribute GetResolvedAttribute<TAttribute>(ParameterInfo parameter) where TAttribute : Attribute
+        {
+            var attribute = parameter.GetCustomAttribute<TAttribute>();
+
+            var storageAttribute = attribute as StorageAttribute;
```
MikeStall: >as StorageAttribute [](start = 45, length = 19)

This sort of special casing was what I wanted to avoid. (I think this is the crux of the difference between the 2 PRs). 
This is hard-coding the checks that I had previously generalized. In this case, it degenerates to the same runtime calls as the generalized algorithm. 
We'd need to duplicate these checks for ServiceBus.

src/Microsoft.Azure.WebJobs.Host/Executors/StorageAccountProviderExtensions.cs
```diff
@@ -82,16 +86,27 @@ private static void ValidateStorageAccount(IStorageAccount account, string conne
         }
 
         /// <summary>
-        /// Walk from the parameter up to the containing type, looking for a
-        /// <see cref="StorageAccountAttribute"/>. If found, return the account.
+        /// Determine whether the specified parameter declares an explicit storage
+        /// account to use, either by specifying a value for <see cref="StorageAttribute.Account"/> or
+        /// if <see cref="StorageAccountAttribute"/> has been applied.
         /// </summary>
         internal static string GetAccountOverrideOrNull(ParameterInfo parameter)
```
MikeStall: >GetAccountOverrideOrNull(ParameterInfo parameter) [](start = 31, length = 49)

Can we get rid of the overload that takes a ParameterInfo? (the other PR removed this function completely). Part of the goal here is to move away form ParameterInfo and rely exclusively on the Attribute, since the attribute is 1:1 with Function.json.

src/Microsoft.Azure.WebJobs.Host/TypeUtility.cs
mathewc: Yes, this is the main difference. I strongly prefer this to the other because it keeps this issue internal - doesn't introduce a new "hierarchical attribute merging" feature that customers can start depending on. This proposal solves all our feature needs w/o introducing any new feature surface area.

src/Microsoft.Azure.WebJobs.Host/Executors/StorageAccountProviderExtensions.cs
mathewc: Yes, this is still needed for CloudStorageAccount/CloudQueueClient type bindings, e.g. https://github.com/Azure/azure-webjobs-sdk/blob/dev/test/Microsoft.Azure.WebJobs.Host.EndToEndTests/MultipleStorageAccountsEndToEndTests.cs#L230. In such cases, there is no Queue attribute.

src/Microsoft.Azure.WebJobs.Host/TypeUtility.cs
MikeStall: 1. You could still accomplish that goals using the existing StorageAccountAttribute  (and not introducing a new StorageAttribute) 
2. What about ServiceBus? Do we copy this pattern for SB too? 
3. What about tooling? Does it special case? Actually, it'd be great if tooling can just call this method, and thne it matters much less. 

---
In reply to: [107949364](https://github.com/Azure/azure-webjobs-sdk/pull/1084#discussion_r107949364) [](ancestors = 107949364)

src/Microsoft.Azure.WebJobs/StorageAttribute.cs
mathewc: StorageAccountAttribute is a class/method/parameter level attribute that as its name indicates is solely for specifying an account name. Rather than overload that, I think its cleaner to introduce an abstract base that is not specific to storage account and could have other shared properties in the future. It's just cleaner. Note that customers can't use this base attribute so it won't cause any issues/confusion - it is just for us to share common properties.

As an alternative, we could also introduce a "ConnectionSetting" attribute similar or deriving from AppSettingAttribute that can be applied to such properties. E.g.:

```csharp
public class QueueAttribute : Attribute
{
    [ConnectionSetting]
}
```

That attribute would apply to both Storage and ServiceBus. The drawback there is that it would require reflection to access/set the value.

src/Microsoft.Azure.WebJobs.Host/Executors/StorageAccountProviderExtensions.cs
mathewc: Yes, I think I can remove these two methods - verifying that now.

mathewc: No, it is still used for CloudStorageAccount binding: https://github.com/Azure/azure-webjobs-sdk/blob/dev/test/Microsoft.Azure.WebJobs.Host.EndToEndTests/MultipleStorageAccountsEndToEndTests.cs#L230

src/Microsoft.Azure.WebJobs.Host/TypeUtility.cs
mamaso: Agree that limiting scope of attribute merging is in our best interest, so I think a concrete implementation is probably better in this case than a general one.

MikeStall: I'm fine with limiting it, but we still need to address the 3 points above. 

---
In reply to: [107999544](https://github.com/Azure/azure-webjobs-sdk/pull/1084#discussion_r107999544) [](ancestors = 107999544)

src/Microsoft.Azure.WebJobs.Host/Bindings/StorageAccount/CloudStorageAccountBindingProvider.cs
```diff
@@ -28,14 +28,15 @@ public CloudStorageAccountBindingProvider(IStorageAccountProvider accountProvide
         {
             ParameterInfo parameter = context.Parameter;
 
-            if (context.Parameter.ParameterType != typeof(CloudStorageAccount))
+            if (parameter.ParameterType != typeof(CloudStorageAccount))
             {
                 return null;
             }
 
-            IStorageAccount account = await _accountProvider.GetStorageAccountAsync(context.Parameter, context.CancellationToken);
```
mathewc: Reworked this so I could remove the ParameterInfo based overload - this was the last usage.

src/Microsoft.Azure.WebJobs.Host/Blobs/Bindings/BlobAttributeBindingProvider.cs
mathewc: Removed

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/QueueBindingProvider.cs
```diff
@@ -195,22 +181,33 @@ private IAsyncCollector<IStorageQueueMessage> BuildFromQueueAttribute(QueueAttri
             return new QueueAsyncCollector(queue, _messageEnqueuedWatcherGetter.Value);
         }
 
-        private static IStorageQueue GetQueue(QueueAttribute attrResolved)
+        private IStorageQueue GetQueue(QueueAttribute attrResolved)
```
mathewc: That would then require all the callers upstream to be async - I think the intent was to keep this localized.

src/Microsoft.Azure.WebJobs.Host/TypeUtility.cs
```diff
@@ -98,6 +121,29 @@ internal static Type UnwrapTaskType(Type type)
             return null;
         }
 
+        internal static TAttribute GetResolvedAttribute<TAttribute>(ParameterInfo parameter) where TAttribute : Attribute
+        {
+            var attribute = parameter.GetCustomAttribute<TAttribute>();
+
+            var attributeConnectionProvider = attribute as IConnectionProvider;
```
mathewc: IConnectionProvider allows us to generically interface with all the attributes to get/set their connection properties.

src/Microsoft.Azure.WebJobs.Host/TypeUtility.cs
```diff
@@ -98,6 +121,29 @@ internal static Type UnwrapTaskType(Type type)
...
+            if (attributeConnectionProvider != null && string.IsNullOrEmpty(attributeConnectionProvider.Connection))
+            {
+                // if the attribute doesn't specify an explicit connnection, walk up
+                // the hierarchy looking for an override specified via attribute
+                var connectionProviderAttribute = attribute.GetType().GetCustomAttribute<ConnectionProviderAttribute>();
```
mathewc: ConnectionProviderAttribute is the way the various attributes indicate which override attribute they are bound to. E.g. Queue/Blob/Table attributes are marked with [ConnectionProviderAttribute(typeof(StorageAccountAttribute))]. This is the key to enabling this code to generalize and handle both SAA/SBAA.

src/Microsoft.Azure.WebJobs.ServiceBus/ServiceBusAccountAttribute.cs
```diff
@@ -31,5 +31,18 @@ public ServiceBusAccountAttribute(string account)
         /// Gets the name of the ServiceBus connection string to use.
         /// </summary>
         public string Account { get; private set; }
+
+        /// <inheritdoc />
+        string IConnectionProvider.Connection
```
mathewc: I wanted the property name on the various trigger/binding attributes to be named "Connection" which I think is better than "Account". So I implement this explicitly here to avoid confusion on this attribute.

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/QueueBindingProvider.cs
mamaso: I believe you can use `IAsyncConverters` for all of the usages

src/Microsoft.Azure.WebJobs.ServiceBus/ServiceBusAccount.cs
```diff
@@ -12,28 +11,5 @@ internal class ServiceBusAccount
         public MessagingFactory MessagingFactory { get; set; }
 
         public NamespaceManager NamespaceManager { get; set; }
-
-        internal static string GetAccountOverrideOrNull(ParameterInfo parameter)
```
mathewc: Cross compiling TypeUtility allows me to reuse the single generalized method for this, removing duplication/special casing

src/Microsoft.Azure.WebJobs.Host/Tables/TableAttributeBindingProvider.cs
```diff
@@ -146,11 +125,11 @@ private IAsyncCollector<ITableEntity> BuildFromTableAttribute(TableAttribute att
         }
 
         // Get the storage table from the attribute.
-        private static IStorageTable GetTable(TableAttribute attribute)
+        private IStorageTable GetTable(TableAttribute attribute)
```
mamaso: same comment here, all the usages are within IAsyncCollectors

src/Microsoft.Azure.WebJobs/ConnectionProviderAttribute.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// For Attributes that implement <see cref="IConnectionProvider"/> this attribute can be
+    /// applied to the Attribute to specify the attribute Type used for hierarchical
+    /// declarative overrides.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class)]
+    public sealed class ConnectionProviderAttribute : Attribute
```
mamaso: Assuming we want to mark the ConnectionProvider classes as obsolete. Is this intended as a mechanism for an internal fix that's only public so that we don't duplicate across assemblies? Or do we want to expose this to extension authors as a mechanism for hierarchical connection overrides (I'd rather avoid this)?

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/QueueBindingProvider.cs
mathewc: Won't fix. Refactoring all this to be async can be done in a separate change. I'm retaining existing behavior here.

src/Microsoft.Azure.WebJobs/ConnectionProviderAttribute.cs
mathewc: My plan is to leave this public. I don't have any reservations doing that - it's already a public pattern in that ServiceBus is doing it. Rather than hack some internal thing, I scoped it to something that I think makes general sense and could be cleanly used by other extensions if they wanted to. I.e. we have (and could surface) attribute resolution methods that know how to "fold" these properly.

src/Microsoft.Azure.WebJobs.ServiceBus/ServiceBusAccountAttribute.cs
MikeStall: yeah, that will give us better impedence with the conversion back and forth from Functions.json. 

---
In reply to: [108546447](https://github.com/Azure/azure-webjobs-sdk/pull/1084#discussion_r108546447) [](ancestors = 108546447)

src/Microsoft.Azure.WebJobs/ConnectionProviderAttribute.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs
```
MikeStall: >WebJobs [](start = 26, length = 7)

Nit: I think we should move these (ConnectionProviderAttr, IConnectionProvider) out of Microsoft.Azure.WebJobs namespace and one level deep, probably Microsoft.Azure.WebJobs.Config. 
User code won't refer to these; so it pollutes their namespace. 
We could apply that rule to any "meta" attribute used to define the binding attributes (AutoResolve, AppSetting, etc). - I should do something similar in my byob PR.

MikeStall:
:shipit:

src/Microsoft.Azure.WebJobs.ServiceBus/WebJobs.ServiceBus.csproj
```diff
@@ -131,6 +131,9 @@
     <Compile Include="..\Microsoft.Azure.WebJobs.Host\Triggers\ITriggerDataArgumentBinding.cs">
       <Link>Triggers\ITriggerDataArgumentBinding.cs</Link>
     </Compile>
+    <Compile Include="..\Microsoft.Azure.WebJobs.Host\TypeUtility.cs">
+      <Link>TypeUtility.cs</Link>
+    </Compile>
```
MikeStall: My hope is that when we switch ServiceBus over to the binding rules model, this would go away since the SDK would fold the attribute before handing it to SB.

src/Microsoft.Azure.WebJobs/ConnectionProviderAttribute.cs
mathewc: Hmmm, there are lots of other types that are currently in the root namespace but fall into this same category.  If we want to introduce a new namespace strategy, we should do that in a separate PR and examine all the new recently added types and move as necessary (though many would likely be breaking).

E.g. IAttributeInvokeDescriptor, ODataFilterResolutionPolicy, etc.

MikeStall: Yes, I think those should be included too.

Issue #1083, opened 2017-03-23
Expand BlobTrigger/QueueTrigger binding metadata


mathewc:
Recently I expanded the set of built in binding metadata for ServiceBusTrigger and EventHubTrigger (PRs https://github.com/Azure/azure-webjobs-sdk/pull/1071 and https://github.com/Azure/azure-webjobs-sdk/pull/1069). It would be good to do the same for BlobTrigger - e.g. add Uri, Metadata (custom properties), etc.

Our binding data for queues is already pretty good, I don't think there are any more properties that we need.

mathewc:
@davidebbo FYI - this would address the blob URI scenario you had recently

davidebbo:
Relates to http://stackoverflow.com/questions/42981770/azure-functions-c-sharp-get-blob-uri/42982455#42982455

PR #1087, opened 2017-03-24
Expanding BlobTrigger binding metadata (#1083)

mathewc:


src/Microsoft.Azure.WebJobs.Host/Blobs/Triggers/BlobTriggerBinding.cs
```diff
@@ -190,8 +190,11 @@ private FileAccess Access
 
         private static IReadOnlyDictionary<string, Type> CreateBindingDataContract(IBlobPathSource path)
         {
-            Dictionary<string, Type> contract = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);
+            var contract = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);
             contract.Add("BlobTrigger", typeof(string));
+            contract.Add("Uri", typeof(Uri));
+            contract.Add("Properties", typeof(BlobProperties));
```
mathewc: This BlobProperties object has a bunch of useful info on it, Content-Type, etc. This object is useful in core SDK scenarios where the function might be binding to a Stream for example.

The challenge will be making this available for example to Node.js functions in Azure Functions. I didn't want to cherry pick properties and hoist them as top level. I think what I'll do is add the capability for simple property bag objects like this to be translated into a dictionary, so for a Node function, it would be accessible as a regular object. Those changes will be in Functions.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
```diff
@@ -550,9 +550,19 @@ public static void DisabledJob([QueueTrigger(Queue1Name)] string message)
...
             [BlobTrigger(ContainerName + "/" + Blob1Name)] Stream inputStream,
+            string blobTrigger,
+            Uri uri,
+            IDictionary<string, string> metadata,
+            BlobProperties properties,
             [Blob(ContainerName + "/" + Blob2Name, FileAccess.Write)] Stream outputStream,
             CancellationToken token)
         {
+            Assert.True(uri.ToString().EndsWith(blobTrigger));
+            string parentId = metadata["AzureWebJobsParentId"];
```
mathewc: Interesting that we use this custom metadata collection for SDK properties like the parent function ID, which is how we do the Dashboard correlation capabilities.

src/Microsoft.Azure.WebJobs.Host/Blobs/Triggers/BlobTriggerBinding.cs
mamaso: Out of scope, but it's also worth thinking how to expose binding data for input/output bindings in script languages, i.e. set content type on an output blob. This plan seems like it takes us a step closer.

The dictionary translation seems reasonable.  Can probably use JObject, it implements IDictionary<string, JToken>

src/Microsoft.Azure.WebJobs.Host/Blobs/Triggers/BlobTriggerBinding.cs
```diff
@@ -190,8 +190,11 @@ private FileAccess Access
 
         private static IReadOnlyDictionary<string, Type> CreateBindingDataContract(IBlobPathSource path)
         {
-            Dictionary<string, Type> contract = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);
+            var contract = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);
```
mamaso: nit: could use C#6 initializer if you like
``` 
new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase) {
  ["blobtrigger"] = typeof(string)
}

File: src/Microsoft.Azure.WebJobs.Storage/Blob/IStorageBlobProperties.cs [hunk 4]
```diff
@@ -37,5 +37,10 @@ internal interface IStorageBlobProperties
         /// Gets the LeaseStatus of the blob.
         /// </summary>
         LeaseStatus LeaseStatus { get; }
+
+        /// <summary>
+        /// Gets the underlying SDK object.
+        /// </summary>
+        BlobProperties SdkObject { get; }
```
mamaso: Not a fan of this naming, since it's basically a poco object I'd rather see it named as BlobProperties.  Basically passing 'sdkobject' to a script language reads weirdly to me.

src/Microsoft.Azure.WebJobs.Host/Blobs/Triggers/BlobTriggerBinding.cs
mathewc: Yeah, that is on my list to tackle next :)

src/Microsoft.Azure.WebJobs.Storage/Blob/IStorageBlobProperties.cs
mathewc: Yeah, I'm following the existing naming pattern for these mock types. SdkObject is used all over.

src/Microsoft.Azure.WebJobs.Host/Blobs/Triggers/BlobTriggerBinding.cs
mathewc: I'm an old fashioned kind of guy :)

PR #1093, opened 2017-03-28
SDK support for Script and Tooling

MikeStall:
Preview of public surface area changes. 

1. Add a new Binding marker attribute for marking SDK attributes. Tooling can use this. 
2. bring-your-own-bindings: Add an ITooling interface that Script can use so that script can directly consume an SDK binding without any additional glue code. 
This works by leveraging the binding and converter graph to answer questions Script runtime needs. 
3. Remove some obsolete attributes. Add a fluent interface on ExtensionConfig.


src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeConverter.cs
```diff
@@ -0,0 +1,114 @@
...
+using System.Reflection;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Bindings.Path;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Host.Bindings
+{
+    internal class AttributeCloner
+    {
+        // $$$ Merge this with other methods on this class.
```
MikeStall: todo: look at better factoring / sharing with AttributeCloner here

src/Microsoft.Azure.WebJobs.Host/Config/ExtensionConfigContext.cs
```diff
@@ -9,6 +12,9 @@ namespace Microsoft.Azure.WebJobs.Host.Config
     /// </summary>
     public class ExtensionConfigContext
```
MikeStall: Additions  here are to expose a fluent binding interface that lets us also remove BindingFactory completely.  If we go ahead with this, we just need to add the rest of the Bind* methods here. 
See usage example in test.

src/Microsoft.Azure.WebJobs.Host/Extensions/ITooling.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.IO;
+using System.Reflection;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// Tooling interface for script 
+    /// </summary>
+    public interface ITooling
```
MikeStall: This is the critical new interface that we expose to Script.

src/Microsoft.Azure.WebJobs/BindingAttribute.cs
```diff
@@ -0,0 +1,16 @@
...
+
+using System;
+
+namespace Microsoft.Azure.WebJobs.Host.Config
+{
+    /// <summary>
+    /// Place this on an attribute to note that it's a binding attribute. 
+    /// An extension should then claim this and bind it. 
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class)]
```
MikeStall: This is the new VS tooling attribute.

test/Microsoft.Azure.WebJobs.Host.UnitTests/ToolingTests.cs
```diff
@@ -0,0 +1,179 @@
...
+        }
+
+        public class Widget
+        {
+            public string Value;
+        }
+
+        public class TestExtension : IExtensionConfigProvider            
+        {  
+            public void Initialize(ExtensionConfigContext context)
```
MikeStall: This is an example of the fluent interface.

src/Microsoft.Azure.WebJobs.Host/Extensions/Tooling.cs
```diff
@@ -0,0 +1,397 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Bindings;
+using Microsoft.Azure.WebJobs.Host.Config;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    // Provides additional bookkeeping on extensions 
+    internal class Tooling : ITooling
```
mamaso: prefer a more descriptive name for this, maybe `ScriptMetadataTooling`

src/Microsoft.Azure.WebJobs.Host/Extensions/Tooling.cs
```diff
@@ -0,0 +1,397 @@
...
+
+        internal void Init(IBindingProvider root)
+        {
+            this._root = root;
+
+            // Populate assembly resolution from converters.
+            var converter = this._config.GetService<IConverterManager>() as ConverterManager;
+            if (converter != null)
+            {
+                converter.AddAssemblies(this.AddAssembly);
```
mamaso: nit: `(type) => AddAssembly(type.Assembly)` and remove the instance method one liner

src/Microsoft.Azure.WebJobs.Host/Extensions/Tooling.cs
```diff
@@ -0,0 +1,397 @@
...
+
+        // By convention, typeof(EventHubAttribute) --> "EventHub"
+        private static string GetNameFromAttribute(Type attributeType)
+        {
+            string fullname = attributeType.Name; // no namespace
+            const string Suffix = "Attribute";
+
+            if (!fullname.EndsWith(Suffix, StringComparison.OrdinalIgnoreCase))
+            {
+                throw new InvalidOperationException("Attribute type '" + fullname + "' must end in 'Attribute'");
```
mamaso: nit: interpolated strings

src/Microsoft.Azure.WebJobs.Host/Extensions/Tooling.cs
```diff
@@ -0,0 +1,397 @@
...
+            }
+            string name = fullname.Substring(0, fullname.Length - Suffix.Length);
+            return name;
+        }
+
+        public Type GetAttributeTypeFromName(string name)
+        {
+            Type attrType;
```
mamaso: nit: `attrType = null; TryGetValue(...); return attrType;`

also, remove commented code if unused

src/Microsoft.Azure.WebJobs.Host/Extensions/Tooling.cs
```diff
@@ -0,0 +1,397 @@
...
+        {
+            IBindingProvider root = this._root;
+
+            bool result = Task.Run(() => ScriptHelpers.CanBindAsync(root, attribute, parameterType)).GetAwaiter().GetResult();
+
+            return result;
+        }
+               
+        // Get a better implementation 
+        public Type GetDefaultType(
```
mamaso: Are there any blockers to making this async? Prefer not to have the blocking CanBind call if possible.

src/Microsoft.Azure.WebJobs.Host/Extensions/Tooling.cs
```diff
@@ -0,0 +1,397 @@
...
+                throw new InvalidOperationException($"Can't bind {attribute.GetType().Name} to a script-compatible  output.");
+            }
+            else if (access == FileAccess.Read)
+            {
+                if (IsBinary(requestedType))
+                {
+                    return typeof(byte[]);
+                }
+
+                var array = IsCardinalityMany(requestedType);
```
mamaso: confusing naming / usage, thought array might be a bool

src/Microsoft.Azure.WebJobs.Host/Extensions/Tooling.cs
```diff
@@ -0,0 +1,397 @@
...
+            return null;
+        }
+
+        private static bool IsBinary(Type requestedType)
+        {
+            return requestedType == typeof(byte[]);
+        }
+
+        // Helpers for testing some Script Scenarios 
+        public static class ScriptHelpers
```
mamaso: Move these classes into their own files as appropriate

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
```diff
@@ -228,12 +228,12 @@ public IBindingProvider BindToGenericValueProvider<TAttribute>(Func<TAttribute,
 
         // Adapter to expose a delegate veneer over IAsyncCollector builders.
```
MikeStall: Yes. It's a general pattern for making an IConverter around a lambda. 

---
In reply to: [108521945](https://github.com/Azure/azure-webjobs-sdk/pull/1093#discussion_r108521945) [](ancestors = 108521945)

src/Microsoft.Azure.WebJobs.Host/Extensions/Tooling.cs
MikeStall: can you elaborate on the motivation? 

---
In reply to: [108524510](https://github.com/Azure/azure-webjobs-sdk/pull/1093#discussion_r108524510) [](ancestors = 108524510)

src/Microsoft.Azure.WebJobs.Host/Extensions/Tooling.cs
MikeStall: this is all static analysis, so it shouldn't be async. Ie, we explicitly shouldn't be making any network calls. 
I actually want to push the other direction and replace the CanBindAsync with a non-async version.  


---
In reply to: [108527441](https://github.com/Azure/azure-webjobs-sdk/pull/1093#discussion_r108527441) [](ancestors = 108527441)

src/Microsoft.Azure.WebJobs.Host/Extensions/Tooling.cs
mamaso: There are two overloaded `AddAssembly` functions, so it's hard to tell which is being passed here. `AddAssembly(Type)` did not immediately make sense to me, unlike `AddAssembly(Assembly)`. I'd prefer to see `AddAssemblyFromType` or the lambda route as opposed to the overload, but it is a nit :)

src/Microsoft.Azure.WebJobs.Host/Extensions/Tooling.cs
mamaso: If there's a non async version then that solves this as well, just looking for consistency. If we have to go async I prefer async all the way up so that it's the consumer's choice

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeConverter.cs
MikeStall: It turns out this path it doesn't need to do any resolution.  Next iteration will be even simpler. 

---
In reply to: [108521724](https://github.com/Azure/azure-webjobs-sdk/pull/1093#discussion_r108521724) [](ancestors = 108521724)

src/Microsoft.Azure.WebJobs.Host/Extensions/Tooling.cs
MikeStall: The problem is that async is viral. If we make this async, then non-asnc functions can't call it. So there's a real pressure to make it sync. 

---
In reply to: [108730382](https://github.com/Azure/azure-webjobs-sdk/pull/1093#discussion_r108730382) [](ancestors = 108730382)

src/Microsoft.Azure.WebJobs.Host/Extensions/Tooling.cs
MikeStall: renaming to: GetElementTypeIfCardinalityMany


---
In reply to: [108527761](https://github.com/Azure/azure-webjobs-sdk/pull/1093#discussion_r108527761) [](ancestors = 108527761)

src/Microsoft.Azure.WebJobs.Host/Extensions/Tooling.cs
MikeStall: I'll change to private. This is only used as a helper in this class. I prefer a private nested class since that limits the bread of access.

---
In reply to: [108534601](https://github.com/Azure/azure-webjobs-sdk/pull/1093#discussion_r108534601) [](ancestors = 108534601)

src/Microsoft.Azure.WebJobs.Host/Extensions/Tooling.cs
MikeStall: ok. It's private, so we can always change later. 

---
In reply to: [108524012](https://github.com/Azure/azure-webjobs-sdk/pull/1093#discussion_r108524012) [](ancestors = 108524012)

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeConverter.cs
```diff
@@ -0,0 +1,96 @@
...
+                if (prop.CanWrite)
+                {
+                    JToken token;
+                    if (properties.TryGetValue(prop.Name, StringComparison.OrdinalIgnoreCase, out token))
+                    {
+                        var obj = ApplyNameResolver(token, prop.PropertyType);
+                        prop.SetValue(newAttr, obj);
+                    }
+                }
+            }
```
mamaso: I'd love to see this using the current AttributeCloner logic. I think all you need to do is replace GetResolver with ApplyNameResolver

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/BindToInputBindingProvider.cs
```diff
@@ -50,6 +53,63 @@ public Task<IBinding> TryCreateAsync(BindingProviderContext context)
...
+                    SourceAttribute = typeof(TAttribute),
+                    Converters = converters,
+                    UserType = type
+                };
+            }
+        }
+
+        // DefaultTypes (in precedence order) that we check for input bindings. 
+        private static readonly Type[] DefaultTypes = new Type[] 
+        {
```
brettsam: It looked like the ordering is different in AsyncCollector (it goes byte[], string, JObject). Should they be consistent?

src/Microsoft.Azure.WebJobs.Host/Config/FluentBindingRule.cs
```diff
@@ -0,0 +1,311 @@
...
+                throw new InvalidOperationException("Validator already set");
+            }
+            _validator = validator;
+            return this;
+        }
+
+        // Called by infrastructure after the extension is invoked.
+        // This applies all changes accumulated on the fluent object. 
+        internal void Flush()
+        {
```
brettsam: Flush seems like the wrong name -- could this be something like Build or Initialize?

src/Microsoft.Azure.WebJobs.Host/Extensions/ITooling.cs
brettsam: Maybe I'm the only one, but when I see Tooling, I think of Visual Studio. Is there a different name we could give this to better reflect what it's for?

src/Microsoft.Azure.WebJobs.Host/Tables/TableExtension.cs
```diff
@@ -0,0 +1,371 @@
...
+                    .SetPostResolveHook(ToParameterDescriptorForCollector)
+                    .BindToInput<CloudTable>(builder)                    
+                .WhenIsNull(RowKeyProperty)
+                    .SetPostResolveHook(ToParameterDescriptorForCollector)
+                    .BindToInput<IStorageTable>(builder)                    
+                .BindToCollector<ITableEntity>(BuildFromTableAttribute)
+                .WhenIsNotNull(PartitionKeyProperty).WhenIsNotNull(RowKeyProperty)
+                    .BindToInput<JObject>(builder)
+                .BindToInput<JArray>(builder)
+                .Bind(original);
```
brettsam: How strict is the ordering here? Can I get myself into trouble if I call BindToInput().WhenIsNull()? What kinds of errors do I get?

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
```diff
@@ -261,6 +267,12 @@ public JobHostTraceConfiguration Tracing
         }
 
         /// <summary>
+        /// get host-level metadata which the extension may Read to do configuration. 
+        /// </summary>
+        [Obsolete("preview")]
+        public JObject HostConfigMetadata { get; set; }
```
MikeStall: >public [](start = 8, length = 6)

This is important.  It will get passed to the extension's Initialize function to let it do what ScriptBinding.Initialize did.

src/Microsoft.Azure.WebJobs.Host/Config/ExtensionConfigContextExtensions.cs
```diff
@@ -69,5 +71,77 @@ public static class ExtensionConfigContextExtensions
             IExtensionRegistry extensions = context.Config.GetService<IExtensionRegistry>();
             extensions.RegisterBindingRules<TAttribute>(bindingProviders);
         }
+
+        /// <summary>
```
MikeStall: These are helpers for plumbing host.json through to the extension. 
Some extensions need just a specific section, some need the entire host.json (since they handle multiple sections). 
If the host.json is consistent with the config (same property names, etc), then the extension can just call the Apply() helper.

src/Microsoft.Azure.WebJobs.Host/Tables/TableExtension.cs
```diff
@@ -0,0 +1,371 @@
...
+using Microsoft.Azure.WebJobs.Host.Protocols;
+using Microsoft.Azure.WebJobs.Host.Storage;
+using Microsoft.Azure.WebJobs.Host.Storage.Table;
+using Microsoft.WindowsAzure.Storage.Table;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Host.Tables
+{
+    // Extension for adding Azure Table  bindings. 
+    internal class TableExtension : IExtensionConfigProvider
```
MikeStall: >TableExtension [](start = 19, length = 14)

Most of this code is moved from TableAttributeBindingProvider.

test/Microsoft.Azure.WebJobs.Host.UnitTests/FunctionIndexerFactory.cs
```diff
@@ -24,44 +24,38 @@ internal static class FunctionIndexerFactory
...
-            IExtensionTypeLocator extensionTypeLocator = new NullExtensionTypeLocator();
-            ContextAccessor<IMessageEnqueuedWatcher> messageEnqueuedWatcherAccessor =
-                new ContextAccessor<IMessageEnqueuedWatcher>();
-            ContextAccessor<IBlobWrittenWatcher> blobWrittenWatcherAccessor =
-                new ContextAccessor<IBlobWrittenWatcher>();
-            ISharedContextProvider sharedContextProvider = new SharedContextProvider();
-            TraceWriter logger = traceWriter ?? new TestTraceWriter(TraceLevel.Verbose);
+            IStorageAccountProvider storageAccountProvider = GetStorageAccountProvider(account);
+
+            var config = TestHelpers.NewConfig(storageAccountProvider, nameResolver, extensionRegistry);
```
MikeStall: >TestHelpers.NewConfig [](start = 25, length = 21)

This function is another example of code that's redundant with JobHostConfiguration setup.  The key thing is that it now calls the TestHelpers.NewConfig() rather than duplicate all the logic and glue on its own.

src/Microsoft.Azure.WebJobs/BindingAttribute.cs
MikeStall: To lightup with BYOB and VS tooling, yes. So existing scenarios aren't broken.  


---
In reply to: [110183092](https://github.com/Azure/azure-webjobs-sdk/pull/1093#discussion_r110183092) [](ancestors = 110183092)

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeConverter.cs
```diff
@@ -0,0 +1,96 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Host.Bindings
+{
+    internal class AttributeCloner
```
MikeStall: I blame StyleCop. 
We already have a AttributeClone<T> and AttributeClone.
StyleCop won't allow both in the same file AttributeCloner.cs.

---
In reply to: [110198953](https://github.com/Azure/azure-webjobs-sdk/pull/1093#discussion_r110198953) [](ancestors = 110198953)

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
MikeStall: fixed comment.

src/Microsoft.Azure.WebJobs.Host/Tables/TableExtension.cs
MikeStall: we'll give a config-time error.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/IRuleProvider.cs
```diff
@@ -0,0 +1,34 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+
+namespace Microsoft.Azure.WebJobs.Host.Bindings
+{
+    /// <summary>
+    /// Enable binding rules (like <see cref="IBindingProvider"/> to be self-describing for tooling. 
+    /// </summary>
+    internal interface IRuleProvider
```
mamaso: To clarify: an IBindingProvider must implement IRuleProvider in order to be a script supported binding?

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/FilteringBindingProvider.cs
```diff
@@ -2,28 +2,33 @@
...
     // Useful for extension  that need some special case type restriction. 
     // If the user parameter type passes the predicate, then chain to an inner an provider. 
-    internal class FilteringBindingProvider<TAttribute> : IBindingProvider
+    internal class FilteringBindingProvider<TAttribute> : IBindingProvider, IRuleProvider
         where TAttribute : Attribute
     {
         private readonly Func<TAttribute, Type, bool> _predicate;
         private readonly INameResolver _nameResolver;
         private readonly IBindingProvider _inner;
+        private readonly string _description;
```
mamaso: is description used only for display purposes?

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/GenericCompositeBindingProvider.cs
```diff
@@ -67,5 +69,34 @@ public GenericCompositeBindingProvider(IEnumerable<IBindingProvider> providers)
             }
             throw new InvalidOperationException("Can't bind " + resourceName + " to type '" + context.Parameter.ParameterType + "'.");
         }
+
+        public IEnumerable<Rule> GetRules()
+        {
+            foreach (var provider in _providers.OfType<IRuleProvider>())
```
mamaso: nit: could use the following, but it probably doesn't improve readability (I like linq too much)

```
_providers
  .OfType<IRuleProvider>()
  .SelectMany(rule => rule.GetRules())
```

src/Microsoft.Azure.WebJobs.Host/Bindings/ConverterManager.cs
```diff
@@ -416,6 +534,40 @@ public override bool IsMatch(Type type)
             {
                 return type == _type;
             }
+
+            internal override string GetDisplayName()
+            {
+                return TypeToString(_type);
+            }
+            public static string TypeToString(Type t)
```
mamaso: nit: maybe move this to `TypeUtility`

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
```diff
@@ -261,6 +267,12 @@ public JobHostTraceConfiguration Tracing
         }
 
         /// <summary>
+        /// get host-level metadata which the extension may read to do configuration.
```
mamaso: Currently get/set - also JObject is mutable. Do we want to lock the configuration down so that an extension can read entire config but only modify its own section?

src/Microsoft.Azure.WebJobs.Host/Tables/TableExtension.cs
```diff
@@ -0,0 +1,371 @@
...
+    {
+        private IStorageAccountProvider _accountProvider;
+
+        // Property names on TableAttribute
+        private const string RowKeyProperty = "RowKey";
```
mamaso: nit: `nameof(TableAttribute.RowKey/PartitionKey)`

src/Microsoft.Azure.WebJobs.Host/Tables/TableExtension.cs
mamaso: To add on to this, how do you know when a filter (WhenIsNull, etc) ends? Is it the next bind call?  What happens if you add a converter after a filter?

src/Microsoft.Azure.WebJobs.Host/Tables/TableExtension.cs
```diff
@@ -0,0 +1,371 @@
...
+                        break;
+                    case EdmType.Binary:
+                        value = new JValue(entityProperty.Value.BinaryValue);
+                        break;
+                }
+
+                jsonObject.Add(entityProperty.Key, value);
+            }
+
+            jsonObject.Add("PartitionKey", tableEntity.PartitionKey);
```
mamaso: same `nameof` suggestion here

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
```diff
@@ -369,25 +369,22 @@ void IExtensionConfigProvider.Initialize(ExtensionConfigContext context)
...
+            IConverterManager cm = context.Config.GetService<IConverterManager>();
             var triggerBindingProvider = new EventHubTriggerAttributeBindingProvider(nameResolver, cm, this);
-            extensions.RegisterExtension<ITriggerBindingProvider>(triggerBindingProvider);
+            context.AddBindingRule<EventHubTriggerAttribute>()
+                .BindToTrigger(triggerBindingProvider);
 
             // register our binding provider
-            var ruleOutput = bf.BindToCollector<EventHubAttribute, EventData>(BuildFromAttribute);
-            extensions.RegisterBindingRules<EventHubAttribute>(ruleOutput);
+            context.AddBindingRule<EventHubAttribute>()
```
mamaso: I like how the fluent interface helps do away with a bunch of the type parameter specification :)

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
```diff
@@ -59,6 +59,12 @@ public IBindingProvider AddFilter<TAttribute>(Func<TAttribute, Type, bool> predi
             return new FilteringBindingProvider<TAttribute>(predicate, this._nameResolver, innerProvider);
         }
 
+        internal IBindingProvider AddFilter<TAttribute>(Func<TAttribute, Type, bool> predicate, IBindingProvider innerProvider, string description)
```
MikeStall: good catch. victim of code churn.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/FilteringBindingProvider.cs
MikeStall: Yes.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/IRuleProvider.cs
MikeStall: Basically yes. But using the v2 binding interfaces (BindToInput, BindToCollecotr, BindToStream), this happens automatically because the extension author no longer implements IBindingProvider directly anymore.

src/Microsoft.Azure.WebJobs.Host/Tables/TableExtension.cs
MikeStall: I'm make Bind*() return void to guarantee we avoid the confusion.

test/Microsoft.Azure.WebJobs.Host.UnitTests/ExtensionConfigContextTests.cs
```diff
@@ -0,0 +1,116 @@
...
+
+            // Simulates extension initialization scope.
+            {
+                var binding = ctx.AddBindingRule<TestAttribute>();
+
+                binding.WhenIsNull("Mode").
+                        BindToInput<int>(attr => attr.Value);
+
+                binding.WhenIsNotNull("Mode").
+                        BindToInput<string>(attr => attr.ToString());
```
brettsam: Nice. This does seem easier to parse.

PR #1095, opened 2017-03-29
Logging: add Exception callback for LogWriter

brettsam:
This will enable better logging for https://github.com/Azure/azure-webjobs-sdk-script/issues/1245

In Functions, under load, you can get to a place where the logger is unable to write to Table storage because it is being throttled. Exceptions are thrown, but because they're coming from an unmonitored background Task, we never log them anywhere. This provides a callback so Exceptions can be logged.

There will be a corresponding change in Script that will log these to the TraceWriter. This will allow us to see how often this occurs. I'm not sure that there's a great solution to remove the issue completely (it would involve splitting the logs across more partitions) -- but the move to logging via Application Insights should make this less of an issue.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging/Internal/LogWriter.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging/Internal/LogWriter.cs
@@ -49,7 +49,9 @@ namespace Microsoft.Azure.WebJobs.Logging
         static ContainerActiveLogger _container;
         private CloudTableInstanceCountLogger _instanceLogger;
 
-        public LogWriter(string hostName, string machineName, ILogTableProvider logTableProvider)
+        private Action<Exception> _onException;
+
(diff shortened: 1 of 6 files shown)


src/Microsoft.Azure.WebJobs.Logging/Internal/LogWriter.cs
```diff
@@ -293,8 +296,19 @@ private FunctionInstanceLogItem[] Update()
 
         private async Task FlushCoreAsync()
         {
-            await FlushTimelineAggregateAsync(true);
-            await FlushIntancesAsync();
+            try
+            {
+                await FlushTimelineAggregateAsync(true);
+                await FlushIntancesAsync();
```
mamaso: nit: rename this if it's a quick fix

brettsam:
re: ignoring the logs -- true. But we should always get them in our backend logs, which I'm more interested in at the moment.

PR #1098, opened 2017-03-30
Clear the internal TextWriterTraceAdapter when calling Flush

flagbug:
I hope my assumption that calling `Flush` twice shouldn't result in the buffer being written twice is correct


mathewc:
Thanks for fixing this. How did you discover the issue out of curiosity?

mathewc:
Merged, thanks for the contribution!

flagbug:
@mathewc It was completely by accident, I was investigating some performance issues around logging in my Webjobs and while digging into the Webjobs SDK codebase, I found this

PR #1100, opened 2017-04-01
Add exception handler to listener factory context

### LINKED ISSUE #1318: Office integrations

mamaso:
resolves https://github.com/Azure/azure-webjobs-sdk-script/issues/1318

Essentially, listeners have no way to propagate exceptions during the listening process. Queue listener was previously hooked up to IWebJobsExceptionHandler via constructor param passing, but service bus listener & others were not.

- Added IWebJobsExceptionHandler to ListenerFactoryContext
- Modified QueueListener to use listener factory context handler instead of previous routing through constructor params
- Modified ServiceBusListener to use the handler
  - required some gross code cloning OnMessageOptions

Exceptions should now flow here: https://github.com/Azure/azure-webjobs-sdk-script/blob/dev/src/WebJobs.Script.WebHost/WebScriptHostExceptionHandler.cs#L70, which gets them into the trace pipeline which will log them per function


src/Microsoft.Azure.WebJobs.Host/Extensions/IWebjobsExceptionHandlerExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// Helper methods for IWebJobsExceptionHandler
+    /// </summary>
+    public static class IWebJobsExceptionHandlerExtensions
+    {
+        /// <summary>
+        /// Captures the exception and synchronously runs OnUnhandledExceptionAsync
+        /// </summary>
+        /// <param name="handler"></param>
+        /// <param name="e"></param>
+        public static void Capture(this IWebJobsExceptionHandler handler, Exception e)
```
mamaso: removes some clutter

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexProvider.cs
```diff
@@ -2,27 +2,18 @@
...
-        private readonly ITypeLocator _typeLocator;
-        private readonly ITriggerBindingProvider _triggerBindingProvider;
-        private readonly IBindingProvider _bindingProvider;
-        private readonly IJobActivator _activator;
-        private readonly IFunctionExecutor _executor;
-        private readonly IExtensionRegistry _extensions;
-        private readonly SingletonManager _singletonManager;
-        private readonly TraceWriter _trace;
-
+        private Func<CancellationToken, Task<IFunctionIndex>> _initializeIndex;
```
mamaso: was routing parameters through this class, saw that it was only serving as an intialization wrapper, encapsulated in a lambda

src/Microsoft.Azure.WebJobs.Host/Listeners/ListenerFactoryContext.cs
```diff
@@ -39,6 +43,11 @@ public ListenerFactoryContext(FunctionDescriptor descriptor, ITriggeredFunctionE
         public ITriggeredFunctionExecutor Executor { get; private set; }
 
         /// <summary>
+        /// Gets the web jobs exception handler
+        /// </summary>
+        public IWebJobsExceptionHandler ExceptionHandler { get; private set; }
```
mamaso: meat of the change is adding the IWebJobsExceptionHandler to ListenerFactoryContext so that listeners can hook into the webjobs exception pipeline during normal execution (not just during `StartAsync`).

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
```diff
@@ -338,17 +347,19 @@ private class ListenerFactory : IListenerFactory
...
             private readonly ITriggerBinding _binding;
+            private readonly IWebJobsExceptionHandler _handler;
 
-            public ListenerFactory(FunctionDescriptor descriptor, ITriggeredFunctionExecutor executor, ITriggerBinding binding)
+            public ListenerFactory(FunctionDescriptor descriptor, ITriggeredFunctionExecutor executor, ITriggerBinding binding, IWebJobsExceptionHandler handler)
             {
                 _descriptor = descriptor;
                 _executor = executor;
                 _binding = binding;
+                _handler = new FunctionExceptionHandler(descriptor, handler);
```
mamaso: wrap the existing handler in a function exception handler, which wraps exceptions in `FunctionInvocationExceptions` so that they are tied to a method and can be picked up by script

src/Microsoft.Azure.WebJobs.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -51,11 +54,34 @@ private Task StartAsyncCore(CancellationToken cancellationToken)
             cancellationToken.ThrowIfCancellationRequested();
 
             _receiver = _messagingProvider.CreateMessageReceiver(_messagingFactory, _entityPath);
-            _receiver.OnMessageAsync(ProcessMessageAsync, _messageProcessor.MessageOptions);
+            var cloned = CloneMessageOptions(_messageProcessor.MessageOptions);
+            cloned.ExceptionReceived += (sender, args) => _handler.Capture(args.Exception);
```
mamaso: add our function-specific event handler

src/Microsoft.Azure.WebJobs.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -51,11 +54,34 @@ private Task StartAsyncCore(CancellationToken cancellationToken)
...
 
+        internal static OnMessageOptions CloneMessageOptions(OnMessageOptions original)
+        {
+            var clone = new OnMessageOptions()
+            {
+                AutoComplete = original.AutoComplete,
+                AutoRenewTimeout = original.AutoRenewTimeout,
+                MaxConcurrentCalls = original.MaxConcurrentCalls
+            };
+            var field = typeof(OnMessageOptions).GetField("ExceptionReceived", BindingFlags.Instance | BindingFlags.NonPublic);
```
mamaso: deep clone of event handler. gross

src/Microsoft.Azure.WebJobs.ServiceBus/Listeners/ServiceBusQueueListenerFactory.cs
```diff
@@ -1,46 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
```
mamaso: didn't find that these listener factories provided much value besides testability, so canned them.

src/Microsoft.Azure.WebJobs.ServiceBus/Listeners/ServiceBusListener.cs
brettsam: Oy. It doesn't feel right doing reflection to clone this. I'm not sure if there's a better alternative, though. @MikeStall -- any recommendations on this?

Just to reiterate -- we're doing this because without it, every ServiceBusTrigger will wire up a new ExceptionReceived handler -- so any exception will get broadcast to every trigger. Is that right?

mamaso: @brettsam exactly, this is to get a handler per function & get around the global handler issue.

src/Microsoft.Azure.WebJobs.Host/Extensions/IWebjobsExceptionHandlerExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+
+using System;
+using System.Runtime.ExceptionServices;
+
+namespace Microsoft.Azure.WebJobs.Host.Timers
+{
+    /// <summary>
+    /// Helper methods for IWebJobsExceptionHandler
+    /// </summary>
+    public static class IWebJobsExceptionHandlerExtensions
```
mamaso: Good call

src/Microsoft.Azure.WebJobs.Host/Extensions/IWebjobsExceptionHandlerExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        /// <summary>
+        /// Captures the exception and synchronously runs OnUnhandledExceptionAsync
+        /// </summary>
+        /// <param name="handler"></param>
+        /// <param name="e"></param>
+        public static void Capture(this IWebJobsExceptionHandler handler, Exception e)
+        {
+            var info = ExceptionDispatchInfo.Capture(e);
+            handler?.OnUnhandledExceptionAsync(info).GetAwaiter().GetResult();
```
MikeStall: +1.  Either OnUnhandledExceptionAsync shouldn't be async; or this method should be async (this is the reverse viral nature of async) 

---
In reply to: [111592392](https://github.com/Azure/azure-webjobs-sdk/pull/1100#discussion_r111592392) [](ancestors = 111592392)

src/Microsoft.Azure.WebJobs.Host/Extensions/IWebjobsExceptionHandlerExtensions.cs
MikeStall: I think the convention is to actually *keep* the I since this is an extension class. 

Consider if you had class Foo, and interface IFoo, and then extensions classes on each, you'd need "FooExtension" and "IFooExtension" to avoid conflict.

src/Microsoft.Azure.WebJobs.Host/Extensions/IWebjobsExceptionHandlerExtensions.cs
mamaso: I made this synchronous because all callers previously used `OnUnhandledExceptionAsync(info).GetAwaiter().GetResult()`. I've switched over to async and changed the callers to use the asynchronous method without GetAwaiter... have a look and let me know if you see any issues with removing the sync call.

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexProvider.cs
```diff
@@ -32,80 +23,37 @@ internal class FunctionIndexProvider : IFunctionIndexProvider
...
-            {
-                throw new ArgumentNullException("triggerBindingProvider");
-            }
-
-            if (bindingProvider == null)
-            {
-                throw new ArgumentNullException("bindingProvider");
-            }
-
-            if (activator == null)
```
MikeStall: > if (activator == null) [](start = 11, length = 23)

are we losing the null checks?

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexProvider.cs
```diff
@@ -32,80 +23,37 @@ internal class FunctionIndexProvider : IFunctionIndexProvider
...
-                throw new ArgumentNullException("extensions");
-            }
+                FunctionIndex index = new FunctionIndex();
+                FunctionIndexer indexer = new FunctionIndexer(triggerBindingProvider, bindingProvider, activator, executor, extensions, singletonManager, trace, handler);
 
-            if (singletonManager == null)
-            {
-                throw new ArgumentNullException("singletonManager");
-            }
+                // should we parallelize this?
```
MikeStall: >// should we parallelize this? [](start = 16, length = 30)

remove?

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexProvider.cs
MikeStall: fwiw, I think the generated code here is will be unequivocally worse than before. The compiler will generate an extra class with a closure that does all the same thing that we're removing here.  
I kind of wonder why we're passing all of these types in there in the first place.  


---
In reply to: [109277385](https://github.com/Azure/azure-webjobs-sdk/pull/1100#discussion_r109277385) [](ancestors = 109277385)

src/Microsoft.Azure.WebJobs.Host/Listeners/ListenerFactoryContext.cs
MikeStall: The configs create the listeners. So could they just cache this and pass to the listeners?  Put another way, is adding this here enabling something new, or just encouraging an existing pattern?

---
In reply to: [109277405](https://github.com/Azure/azure-webjobs-sdk/pull/1100#discussion_r109277405) [](ancestors = 109277405)

MikeStall:
>         void Initialize(JobHost host);

We don't really want all the extensions getting access to this method.  I wonder if we can either remove it? OR split it onto a separate interface?


---
Refers to: src/Microsoft.Azure.WebJobs.Host/Timers/IWebJobsExceptionHandler.cs:19 in be1bea3. [](commit_id = be1bea3df5bc532978a7158faa05ea22a3ec1d27, deletion_comment = False)

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexProvider.cs
mamaso: FunctionIndexer had all null checks for its parameters so I decided that it's fine to omit them here

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexProvider.cs
mamaso: existing comment, good eye :)

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexProvider.cs
mamaso: Agree that the generated code is 'worse', but it's better in the sense that we don't need to write it, don't need to add boilerplate to pass through another parameter, etc. 

The root of the problem/my frustration is the large number of parameters that are passed throughout the sdk - perhaps moving to C# 7 we could group some of these in tuples and use destructuring in parameters.

src/Microsoft.Azure.WebJobs.Host/Listeners/ListenerFactoryContext.cs
mamaso: Good question - this is encouraging an existing pattern which we see in blob listener, queue listener, and (now) service bus listener.  However, it all could be done via caching from the job host config - I'll do an iteration to see how complicated that path is.

Didn't realize that the core listeners / extensions used the config.Initialize call!

mamaso:
@MikeStall I agree that the IWebJobsExceptionHandler class could use some iteration, it didn't feel like a general exception handler, was in the timers namespace, etc - but it was the best existing method for exception handling. Agree on the `initialize` method, at present the extensions already have access to this through the jobhostconfig.

mamaso:
holding off on this until @brettsam completes the 'LogContext' changes

Issue #1067, opened 2017-03-16
AzureWebJobsDashboard not working with ZRS storage


baywet:
#### Repro steps

1. Create a ZRS, Resource Manager, General Purpose storage

1. Use the connection string as input parameter of the JobHostConfiguration

2. Receive a message (Service Bus in my case but I don't think it matters)

#### Expected behavior

Webjob logging entries to the dashboard

#### Actual behavior

Getting an exception: `Unhandled Exception: System.InvalidOperationException: Storage account 'accountname' is of unsupported type 'BlobOnly'. Supported types are 'GeneralPurpose'`
Web job when published to an Azure app service, keeps restarting.

#### Known workarounds

Creating and using a GRS instead.

#### Related information 
WebJob SDK 2.0.0.0, .NET "classic" 4.5.2

christopheranderson:
@mamaso - can you confirm this repros for all ZRS accounts?

mamaso:
Looks like this is all ZRS, see [here](https://blogs.msdn.microsoft.com/windowsazurestorage/2014/08/01/introducing-zone-redundant-storage/)

> Since ZRS accounts do not support page blob, file, table or queue, any attempt to create or manipulate those objects on a ZRS storage account will fail.

mamaso:
Closing this as by design.

Related to https://github.com/Azure/azure-webjobs-sdk/issues/810

Just one more piece of info, these ZRS accounts are usable, but not as primary storage accounts - i.e. `AzureWebJobsDashboard` or `AzureWebJobsStorage`.  You can create another app setting `MyZRSStorageAccount` and create a function something like this:

```
public static void BlobToBlob_DifferentAccounts_SecondaryToPrimary(
            [BlobTrigger(Input + "/{name}"), StorageAccount("MyZRSStorageAccount")] string input,
            [Blob(Output + "/{name}")] out string output)
        {
            output = input;
        }
```

baywet:
Thanks for investigating. Maybe the error should be more explicit though and documentation should include the fact that ZRS is not supported? It took quite some time to figure out what was the issue on a production workload, and it used to be working fine before we updated the SDK. Having proper error and doc would have helped a lot.

mamaso:
Yep, the error could be improved to indicate that it could be a blob-only or ZRS account.  We unfortunately have no way to discover exactly what type of account you have from the connection string.

I can update the docs to indicate that ZRS accounts are not supported as well

[...]

PR #1101, opened 2017-04-02
Add ZRS warning to blob-only accounts

mamaso:
improved warning for #1067


test/Microsoft.Azure.WebJobs.Host.UnitTests/Executors/DefaultStorageAccountProviderTests.cs
```diff
@@ -265,7 +265,7 @@ public void StorageAccountOverrides_MultipleLevels()
             DefaultStorageAccountProvider provider = CreateProductUnderTest(services, connectionStringProvider, parser, validator);
             var exception = await Assert.ThrowsAsync<InvalidOperationException>(() => provider.GetDashboardAccountAsync(CancellationToken.None));
 
-            Assert.Equal("Storage account 'name' is of unsupported type 'Premium'. Supported types are 'GeneralPurpose'", exception.Message);
+            Assert.Equal("Storage account 'name' is of unsupported type 'Premium'. Supported types are 'General Purpose'", exception.Message);
```
brettsam: nit: Can you have one test that uses multiple supported types? The message looks good, it'd just be nice to have a validation in there.

Issue #1060, opened 2017-03-13
Event Hubs / Executor Performance Investigation


mamaso:
Currently, the function executor creates a large number of timers with relatively long durations.

This isn't a problem for long running functions, however in the case of many short-lived functions (IoT, event hub, service bus) the timers are not doing meaningful work and are causing lock contention (in the .net TimerQueue implementation), as well as possibly when callbacks fire and post back to the threadpool (less convinced that this is an issue).

A couple improvements that could be made to the executor:

- maintain a measure of function execution time (perhaps moving average). If the execution time is low enough, remove certain timers (update log, update parameter log, etc)
- add timer jitter to make these timer callbacks fall in different batches

There could also be:
- improve clarity, it's currently very hard to follow the logic
- move any initialization logic out of executor into indexer

Implementation of .net timers can be found here: https://referencesource.microsoft.com/#mscorlib/system/threading/timer.cs,b15414c13eba8b8f

mamaso:
Updates:

I haven't found the expected contention in the timer pool - that doesn't appear to be an issue.

After doing some performance profiling, I found that we spend ~25% of execution in reflection calls. I moved those to the indexing path in #1102 so that they only run at initialization.

I found a few improvements related to contention after profiling as well:
- Storage ops are a significant source of contention. They create a number of worker threads which contend in a [custom finalizer](https://github.com/Azure/azure-storage-net/blob/afc5c6d99a805fb4165631ab5cfd139238b00e71/Lib/ClassLibraryCommon/Core/Util/AsyncStreamCopier.cs#L165). These operations primarily come from:
  - FastTable logging
    - When removing the dashboard connection string to disable logging we can see throughput improvements of ~25%. For high throughput scenarios like event hubs table logging has a high operation latency (~500ms) and high volume (GBs per hour).
  - EventProcessorHost
    - Event processor host does a lot of storage operations under the hood for partition lease managment and checkpointing. We have control over checkpointing - decreasing the checkpoint frequency increased throughput 10-15%.
- Storage ops in dashboard logging path should be async (see [here](https://github.com/Azure/azure-webjobs-sdk/blob/663a508e8a851629c26a51e7de3af36629dfd120/src/Microsoft.Azure.WebJobs.Logging/Entities/ContainerActiveEntity.cs#L78))

mamaso:
Closing this, will reopen if I start more investigation.

PR #1106, opened 2017-04-12
Configurable event hub checkpointing

mamaso:
related to #1060

### DIFF
--- a/src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
+++ b/src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
@@ -33,6 +33,7 @@ namespace Microsoft.Azure.WebJobs.ServiceBus
         private readonly PartitionManagerOptions _partitionOptions; // optional, used to create EventProcessorHost
 
         private string _defaultStorageString; // set to JobHostConfig.StorageConnectionString
+        private int _batchCheckpointFrequency = 1;
 
         /// <summary>
         /// Name of the blob container that the EventHostProcessor instances uses to coordinate load balancing listening on an event hub. 
@@ -66,6 +67,26 @@ namespace Microsoft.Azure.WebJobs.ServiceBus
             _partitionOptions = partitionOptions;
 
             _options = options;
+        }
+
+        /// <summary>
+        /// Gets or sets the number of batches to process before creating an EventHub cursor checkpoint. Default 1.
+        /// </summary>
+        public int BatchCheckpointFrequency
+        {
+            get
+            {
+                return _batchCheckpointFrequency;
+            }
+
+            set
+            {
+                if (value <= 0)
+                {
+                    throw new InvalidOperationException("Batch checkpoint frequency must be larger than 0.");
+                }
+                BatchCheckpointFrequency = value;
+            }
         }
 
         /// <summary>
--- a/src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubListener.cs
+++ b/src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubListener.cs
@@ -19,13 +19,15 @@ namespace Microsoft.Azure.WebJobs.ServiceBus
         private readonly EventProcessorHost _eventListener;
         private readonly bool _singleDispatch;
         private readonly EventProcessorOptions _options;
+        private readonly EventHubConfiguration _config;
 
-        public EventHubListener(ITriggeredFunctionExecutor executor, EventProcessorHost eventListener, EventProcessorOptions options, bool single)
+        public EventHubListener(ITriggeredFunctionExecutor executor, EventProcessorHost eventListener, bool single, EventHubConfiguration config)
         {
             this._executor = executor;
             this._eventListener = eventListener;
             this._singleDispatch = single;
-            this._options = options;
+            this._options = config.GetOptions();
+            this._config = config;
         }
 
         void IListener.Cancel()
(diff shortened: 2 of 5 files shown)


paulbatum:
Do you have a script PR that would let me set the batches per checkpoint via host.json?

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
```diff
@@ -69,6 +69,11 @@ public EventHubConfiguration()
         }
 
         /// <summary>
+        /// Sets the frequency at which to checkpoint event batches.
+        /// </summary>
+        public int BatchesPerCheckpoint { get; set; } = 1;
```
tohling: Would it be more self-explanatory to name is BatchCountPerCheckpoint?

mamaso: Naming is totally up in the air, I struggled to come up with something I liked :)

mamaso:
@paulbatum at present no. It also might make sense to expose this at the trigger attribute level (aka function level) instead of host level config @MikeStall probably has some input on that front.

I'd probably prefer the function granularity, but current event hub listener settings are all host level.

paulbatum:
@mamaso Yeah I can see the argument for being this configurable per function. It feels a bit weird though, because all of our performance knobs are currently host level.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubListener.cs
```diff
@@ -55,16 +57,46 @@ IEventProcessor IEventProcessorFactory.CreateEventProcessor(PartitionContext con
...
+            }
+        }
+
         // We get a new instance each time Start() is called. 
         // We'll get a listener per partition - so they can potentialy run in parallel even on a single machine.
         private class Listener : IEventProcessor
         {
             private readonly EventHubListener _parent;
             private readonly CancellationTokenSource _cts = new CancellationTokenSource();
+            private Func<PartitionContext, Task> _checkpoint;
```
fabiocav: Make this `readonly`

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
fabiocav: I also like `CheckpointFrequency`.... `BatchesCheckpointFrequency` perhaps?

mamaso:
@mathewc or @MikeStall can you take a look at this?

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
```diff
@@ -69,6 +69,11 @@ public EventHubConfiguration()
         }
 
         /// <summary>
+        /// Sets the frequency at which to checkpoint event batches.
```
MikeStall: >/// Sets the frequency at which to checkpoint event batches. [](start = 8, length = 60)

Can you expand on the comment to say what exactly this is counting?

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubListener.cs
```diff
@@ -55,16 +57,46 @@ IEventProcessor IEventProcessorFactory.CreateEventProcessor(PartitionContext con
             return new Listener(this);
         }
 
+        internal static Func<Func<Task>, Task> CreateCheckpointStrategy(int batchesPerCheckpoint)
+        {
+            switch (batchesPerCheckpoint)
+            {
+                // no checkpointing
```
MikeStall: > // no checkpointing [](start = 15, length = 20)

Why would we allow this?

MikeStall:
With the recent BYOB extensibility work, the extension's initialize method gets the host.json and it should be able to handle this with no Script PR changes. 
[https://github.com/Azure/azure-webjobs-sdk-script/issues/1409](https://github.com/Azure/azure-webjobs-sdk-script/issues/1409)  is tracking this.l

---
In reply to: [293703567](https://github.com/Azure/azure-webjobs-sdk/pull/1106#issuecomment-293703567) [](ancestors = 293703567)

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubListener.cs
mamaso: You're right, it shouldn't be allowed - need to move the cursor somehow.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/IEventHubProvider.cs
```diff
@@ -8,6 +8,9 @@ namespace Microsoft.Azure.WebJobs.ServiceBus
     // Expose to binders / attributes so they can get the EventHub connections. 
     internal interface IEventHubProvider
     {
+        // Gets the checkpointing frequency
```
mathewc: Are you rebased on dev? I don't even see this IEventHubProvider type existing anymore

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubTriggerAttributeBindingProvider.cs
```diff
@@ -62,7 +62,7 @@ public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext contex
             Func<ListenerFactoryContext, bool, Task<IListener>> createListener =
              (factoryContext, singleDispatch) =>
              {
-                 IListener listener = new EventHubListener(factoryContext.Executor, eventHostListener, options, singleDispatch);
+                 IListener listener = new EventHubListener(factoryContext.Executor, eventHostListener, options, singleDispatch, _eventHubConfig.BatchCheckpointFrequency);
```
mathewc: Why not just pass in the entire config object to the listener, so we don't have to modify this param list each time we add a new knob?

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
mathewc: nit: "Gets or sets..."

mathewc: perhaps: "Gets or sets the number of batches to process before checkpointing"

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
```diff
@@ -69,6 +69,11 @@ public EventHubConfiguration()
...
+        /// </summary>
+        public int BatchCheckpointFrequency { get; set; } = 1;
```
mathewc: This setting applies to single dispatch mode as well?

mamaso: Yep - single dispatch will pull a batch -> spawn batchSize # of tasks -> checkpoint after all batches complete

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubTriggerAttributeBindingProvider.cs
mamaso: Reasonable - we source eventHostListener, options, and the checkpoint frequency from the config. A lot of this is testability, but as long as we're only pulling poco properties off the config it should be no big deal.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/IEventHubProvider.cs
mamaso: haven't rebased, must have changed since then.

PR #1113, opened 2017-04-17
First step towards enabling on prem specific triggers

shrishrirang:
**What change is this?**
- This commit is the first set of changes for running Azure Functions without Azure Storage
- With this change and similar changes in the azure-webjobs-script repo (details below), triggers and bindings that aren't storage specific would be able to work without a storage account (would need SQL instead for the lease management).

**Details**
- BlobLeaseProxy and SqlLeaseProxy provide blob and sql based implementations respectively.
- A complete SQL based implementation will be added in a subsequent commit
- With just AzureWebJobsStorage defined in the environment variables, there won't be any change in behavior
- Optionally, AzureWebJobsLease can be set to either a SQL connection string or an Azure Storage connection string. LeaseFactory will initialize the appropriate lease implementation accordingly.
- ILeaseProxy is made public interface so that BlobLeaseManager in the azure-webjobs-script repo can start using the same implementation. Ideally, the lease implementation can be in a repo of its own but I am just keeping everything here to keep things simple.

**Testing**
- BlobLeaseProxyTests.cs re-works a few existing tests. Full fledged functional testing of the ILeaseProxy implementations will be added in a follow-up commit. (PR to be merged only after that)
- Some of the tests rely on AzureWebJobsLease to be same as AzureWebJobsStorage (if AzureWebJobsLease is defined). They will be updated separately.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Lease/LeaseFactory.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Lease/LeaseFactory.cs
@@ -1,0 +1,34 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using Microsoft.Azure.WebJobs.Host.Executors;
+using Microsoft.Azure.WebJobs.Host.Lease;
+using Microsoft.Azure.WebJobs.Host.Storage;
+
+namespace Microsoft.Azure.WebJobs.Host.Lease
+{
+    /// <summary>
+    /// ILeaseProxy factory
+    /// </summary>
+    internal class LeaseFactory
+    {
+        public static ILeaseProxy CreateLeaseProxy(IStorageAccountProvider storageAccountProvider)
+        {
+            ILeaseProxy leaseProxy = null;
+
+            // Choose between Sql and Blob based lease implementations
+            if (SqlLeaseProxy.IsSqlLeaseType())
+            {
+                leaseProxy = new SqlLeaseProxy();
+            }
+            else
+            {
+                leaseProxy = new BlobLeaseProxy(storageAccountProvider);
+            }
+
+            return leaseProxy;
+        }
+    }
+}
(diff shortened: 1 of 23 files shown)


shrishrirang:
@fabiocav, @mathewc

shrishrirang:
@jvano

src/Microsoft.Azure.WebJobs.Host/Lease/LeaseFactory.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// ILeaseProxy factory
+    /// </summary>
+    internal class LeaseFactory
+    {
+        public static ILeaseProxy CreateLeaseProxy(IStorageAccountProvider storageAccountProvider)
+        {
+            ILeaseProxy leaseProxy = null;
+
+            // Choose between Sql and Blob based lease implementations
+            if (SqlLeaseProxy.IsSqlLeaseType())
```
jvano: >if (SqlLeaseProxy.IsSqlLeaseType()) [](start = 12, length = 35)

this will try to connect using SQL which is an expensive call just to decide the implementation type. The default connection timeout is 15 seconds. I think we should just parse the connection string to detect the implementation type

mamaso: Agree with @jvano, this would hurt functions startup time. We may still want to test the sql connection after we've established that it looks like a sql connection string, however, in order to flow exceptions to the runtime / fall back to storage if necessary.

shrishrirang: `IsSqlLeaseType()` will not attempt to make any connection. It just parses the connection string.

shrishrirang: In case of an invalid connection string, attempting to perform any operation on the `ILeaseProxy` interface would fail at a later point. I don't think we should be attempting to fallback to storage at this point.

shrishrirang:
Hey @mathewc, were you able to take a look at this? Let me know if you have any questions.

fabiocav:
@shrishrirang sorry about the delay. Will be looking at this as well ASAP.

mathewc:
Closing this, since it is being subsumed by https://github.com/Azure/azure-webjobs-sdk/pull/1202

PR #1116, opened 2017-04-19
Update documentation for StorageAccountAttribute for Functions

lindydonna:


mathewc:
+1 on what Matt said - I don't think we need to make any change here. The help text is valid as is.

lindydonna:
@mamaso @mathewc We're adding support for StorageAccountAttribute to MakeFunctionJson.

The help text isn't valid; we're not adding the "AzureWebJobs" prefix if you're using Functions. If you use WebJobs, an attribute "MyStorage" will look for an app setting "AzureWebJobsMyStorage". For functions, we're not doing this.

mathewc:
Yes, even in functions we apply those same rules. That's why the default storage connection strings which we prefix with "AzureWebJobs" work. The same resolution logic runs for both Functions and WebJobs. E.g. see https://github.com/Azure/azure-webjobs-sdk/blob/663a508e8a851629c26a51e7de3af36629dfd120/src/Microsoft.Azure.WebJobs.Host/AmbientConnectionStringProvider.cs#L37.

I don't think we have any comment updates to make here.

lindydonna:
@mathewc Agreed. I'll instead do another PR to clarify how the "AzureWebJobs" prefix is used. As Matt explained it to me, if you put in a string "Foo", it will search for an appsetting named "Foo" and then one named "AzureWebJobsFoo". That wasn't clear from the current doccomment.

lindydonna:
@mathewc Ah, I see it's the other way around, it looks for AzureWebJobsFoo, then Foo

Issue #798, opened 2016-08-23
Support Table Bindings (input) from ServiceBus Custom Properties


JoeBrockhaus:
I have [a S/O question](http://stackoverflow.com/questions/39110874/) which highlights my current issue. 

If I json-serialize the model representing the message body, and construct the BrokeredMessage from that string, subsequent Table bindings that rely on properties in that message body fail. 

Being able to bind to those BrokeredMessage Custom Properties would solve this problem.

MikeStall:
Is this saying that the binding data for a SB trigger should include the "custom properties"? 
Normally binding data needs to be static. Maybe related to https://github.com/Azure/azure-webjobs-sdk/issues/968

mathewc:
Yeah, making arbitrary/variable custom properties violates a bunch of rules we have now in the system that binding data is static. However, if there are some non-variable properties in the property bag, those could be promoted to binding data (similar to what we do with Queue dequeue count, etc.). Related issue: https://github.com/Azure/azure-webjobs-sdk/issues/1004

mathewc:
We recently added support for binding to BrokeredMessage.Properties as documented here: https://github.com/Azure/azure-webjobs-sdk/wiki/Trigger-Binding-Data. I believe that gets us half way where you want to go, however there is still no way for you to "dot into" the Properties data in your table binding expression (e.g. can't do `{Properties.PK}`).

PR #1117, opened 2017-04-19
Allow dot expressions  in binding TemplateBind

MikeStall:
Supports:
https://github.com/Azure/azure-webjobs-sdk-script/issues/1418
https://github.com/Azure/azure-webjobs-sdk/issues/798

this lets you say:
"{a.b.c}" 
where it now does property dereferences.


src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplate.cs
```diff
@@ -84,54 +87,54 @@ public static BindingTemplate FromString(string input, bool ignoreCase = false)
...
-        public string Bind(IReadOnlyDictionary<string, string> parameters)
+        public string Bind(IReadOnlyDictionary<string, object> parameters)
         {
             StringBuilder builder = new StringBuilder();
 
             if (_ignoreCase && parameters != null)
             {
                 // convert to a case insensitive dictionary
-                var caseInsensitive = new Dictionary<string, string>(parameters.Count, StringComparer.OrdinalIgnoreCase);
+                var caseInsensitive = new Dictionary<string, object>(parameters.Count, StringComparer.OrdinalIgnoreCase);
```
mamaso: nit: Linq ToDictionary reads a bit cleaner to me, same below. 

Also, I don't like that we have at least 3 separate dictionaries created if <string, string> is used, and at least 2 for <string, object> - if we provided a different api to the extension which added individual binding data values instead of one dictionary we could remove all these conversions

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
```diff
@@ -1,32 +1,273 @@
...
+                Type scope = o.GetType();
+                JObject jobj = o as JObject;
+                if (jobj != null)
+                {
+                    JToken propValue;
+                    // JObject's normal dot operator returns null for missing properties; 
+                    // we need to explicitly call TryGetValue.
+                    if (!jobj.TryGetValue(member, StringComparison.OrdinalIgnoreCase, out propValue))
+                    {
+                        throw new InvalidOperationException($"property doesn't exist.");
```
mamaso: in these error cases it might be useful to display the member name where processing failed

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
```diff
@@ -1,32 +1,273 @@
...
+                }
+
+                var strValue = BindingDataPathHelper.ConvertParameterValueToString(current);
+                return strValue;
+            }
+
+            // Case insensitive lookup.
+            // Helper to get a property from an object.             
+            // This supports both static types and JObject (for dynamic) 
+            public static object GetProperty(object o, string member)
```
mamaso: This seems like a lot of reflection to run for each execution. Look at the `PropertyHelper` class - this would likely be more performant if we cached an expression tree of fast property getters to do the dot expression evaluation

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
```diff
@@ -1,32 +1,273 @@
...
+
+            public override string ParameterName
+            {
+                get
+                {
+                    return this._parts[0];
+                }
+            }
+
+            protected override void BuildCapturePattern(StringBuilder builder)
```
MikeStall: No. The caller up the stack has a bigger comment. This is used on triggers. IE, 
    [BlobTrigger("container/{name}") 
Will match against he incoming blob path and assign a value to 'name'


---
In reply to: [112484166](https://github.com/Azure/azure-webjobs-sdk/pull/1117#discussion_r112484166) [](ancestors = 112484166)

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
```diff
@@ -1,32 +1,273 @@
...
         }
 
-        public string Value 
-        { 
-            get { return _value; } 
+        // Represent an expression. 
+        //    "literal{expr}literal" 
+        private class ExpressionToken : BindingTemplateToken
+        {
+            private readonly string[] _parts;
```
MikeStall: I'll rename. It's all internal, so I'll stick with the flat [] since that's easiest. If we ever expand in the future (ie, add array indexing, etc) this may become more complex. 

---
In reply to: [112484373](https://github.com/Azure/azure-webjobs-sdk/pull/1117#discussion_r112484373) [](ancestors = 112484373)

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplate.cs
```diff
@@ -84,54 +87,54 @@ public static BindingTemplate FromString(string input, bool ignoreCase = false)
         /// <returns>Resolved string if succeeded.</returns>
         /// <exception cref="InvalidOperationException">Thrown when required parameter value is not available.
         /// </exception>
-        public string Bind(IReadOnlyDictionary<string, string> parameters)
+        public string Bind(IReadOnlyDictionary<string, object> parameters)
```
MikeStall: good catch. I didn't see the extension method. C# will bind to the instance method over the extension method.  However, we'll still need the extension method for backwards compat of already compiled code. I'll add an obsolete on the extension method.  

---
In reply to: [112491637](https://github.com/Azure/azure-webjobs-sdk/pull/1117#discussion_r112491637) [](ancestors = 112491637)

MikeStall: Of course, it doesn't help that the extension methods are in a different namespace :|


---
In reply to: [112509482](https://github.com/Azure/azure-webjobs-sdk/pull/1117#discussion_r112509482) [](ancestors = 112509482,112491637)

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
MikeStall: caller catches this exception and wraps it with that information. That ensures that all failure points report in a uniform way. 

---
In reply to: [112495171](https://github.com/Azure/azure-webjobs-sdk/pull/1117#discussion_r112495171) [](ancestors = 112495171)

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
MikeStall: I didn't know we had that class. 
One challenge here is that the property access needs to be dynamic. We may have a jobject who's schema changes each time (ie, "Request.Query.Name") 

---
In reply to: [112497496](https://github.com/Azure/azure-webjobs-sdk/pull/1117#discussion_r112497496) [](ancestors = 112497496)

mamaso: The schema may change each time, but the access expression will not (it would always require request.query.name, or it would fail) so I think we can cache the relevant expressions.  You're right that this may work differently for JObject, but it should be consistent for C#. What scenarios would use JObject?

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
```diff
@@ -1,32 +1,273 @@
...
+            // Helper to get a property from an object.             
+            // This supports both static types and JObject (for dynamic) 
+            public static object GetProperty(object o, string member)
+            {
+                Type scope = o.GetType();
+                JObject jobj = o as JObject;
+                if (jobj != null)
+                {
+                    JToken propValue;
+                    // JObject's normal dot operator returns null for missing properties;
```
mamaso: I also think we might want to match JObjects sematics here of returning null/undefined for missing properties. It's not how a C# object would work, but it's what you'd expect from json

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
MikeStall: Anything dynamic (like request) would be jobject.   IE, request.query.name.

---
In reply to: [112513122](https://github.com/Azure/azure-webjobs-sdk/pull/1117#discussion_r112513122) [](ancestors = 112513122)

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingDataPathHelper.cs
```diff
@@ -80,6 +50,10 @@ public static string ConvertParameterValueToString(object parameterValue)
                         {
                             return parameterValue.ToString();
                         }
+                        if (parameterValue is Newtonsoft.Json.Linq.JToken)
```
mathewc: Can you update the method comment above to include JObject?

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplate.cs
```diff
@@ -84,54 +87,54 @@ public static BindingTemplate FromString(string input, bool ignoreCase = false)
...
 
         /// <summary>
+        /// Resolves original parameterized template into a string by replacing parameters with values provided as
+        /// a dictionary.
+        /// </summary>
+        /// <param name="parameters">Dictionary providing parameter values.</param>
+        /// <returns>Resolved string if succeeded.</returns>
+        /// <exception cref="InvalidOperationException">Thrown when required parameter value is not available.
+        /// </exception>
+        [Obsolete("Switch to the overload accepting a IReadOnlyDictionary<string, object>.")]
```
mathewc: Why are you putting Obsolete on a new method?

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingTemplateExtensions.cs
```diff
@@ -46,23 +46,15 @@ public static void ValidateContractCompatibility(this BindingTemplate bindingTem
         /// <param name="bindingTemplate">The binding template to validate.</param>
         /// <param name="bindingData">The binding data to apply to the template.</param>
         /// <returns>The bound template string.</returns>
+        [Obsolete("Call instance method directly")]
         public static string Bind(this BindingTemplate bindingTemplate, IReadOnlyDictionary<string, object> bindingData)
```
mathewc: All the extensions in the extensions repo call this currently. It looks like you've made the method you're directing to internal so extensions won't be able to call it.

MikeStall: Good catch. The extensions compiled because they just bound to this obsolete version (no warnings). I'll make the new instance method public. 

---
In reply to: [114632812](https://github.com/Azure/azure-webjobs-sdk/pull/1117#discussion_r114632812) [](ancestors = 114632812)

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplate.cs
MikeStall: Diff is playing some tricks here. The <string,string> is an actually an existing method. <string,obj> is the new one. 

---
In reply to: [114588153](https://github.com/Azure/azure-webjobs-sdk/pull/1117#discussion_r114588153) [](ancestors = 114588153)

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplate.cs
MikeStall: given the extension method is in a different namespace; I think an adding an instance method is actually a good way to go moving forward. 

---
In reply to: [112510756](https://github.com/Azure/azure-webjobs-sdk/pull/1117#discussion_r112510756) [](ancestors = 112510756,112509482,112491637)

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
MikeStall: We should have consistency between JObject, C# object, and potentially other dynamic sources (expando, IMetataObjectProvider, etc - if we add that in the future) 

---
In reply to: [112513522](https://github.com/Azure/azure-webjobs-sdk/pull/1117#discussion_r112513522) [](ancestors = 112513522)

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplate.cs
MikeStall: Good idea about improving the copying, but changing the trigger contracts is a larger change out of scope for this. Happy to tackle that separately. 


---
In reply to: [112492894](https://github.com/Azure/azure-webjobs-sdk/pull/1117#discussion_r112492894) [](ancestors = 112492894)

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
```diff
@@ -1,32 +1,271 @@
...
     /// binding template string. Then generated token may be consumed to create instance of 
     /// <see cref="BindingTemplate"/> or <see cref="BindingTemplateSource"/>.
+    /// The token may be a literal value or an expression. 
     /// </summary>
-    internal struct BindingTemplateToken
+    internal abstract class BindingTemplateToken
     {
-        private readonly string _value;
-        private readonly bool _isParameter;
+        // If this is expression, get the prameter name
```
mathewc: nit: spelling. For a dotted expression x.y.z, this value will be the terminal member "z"? Perhaps you should just generalize this property to "Name"

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
```diff
@@ -1,32 +1,271 @@
...
+        public abstract string AsLiteral { get; }
+
+        public static BindingTemplateToken NewLiteral(string literalValue)
+        {
+            return new LiteralToken(literalValue);
+        }
+
+        public static BindingTemplateToken NewExpression(string expression)
+        {
+            if (!BindingTemplateParser.IsValidIdentifier(expression))
```
mathewc: I think we should rename IsValidIdentifier to ValidateTemplate, since we've been expanding the syntax over time

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
```diff
@@ -1,32 +1,271 @@
...
+            {
+                throw new FormatException($"The parameter name '{expression}' is invalid.");
```
mathewc: Should this be reworded to "Invalid template expression '{expression}'"?

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
MikeStall: The semantics here are closer to IsValidateIdentifier than ValidateTEmplate. It doesn't quite validate the template (that's higher in the stack that does things like match the { } ).  this is just a first lexical pass on the validation on the stuff between {  }. 


---
In reply to: [114848982](https://github.com/Azure/azure-webjobs-sdk/pull/1117#discussion_r114848982) [](ancestors = 114848982)

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
MikeStall: for 'x.y,z', it returns 'x'.  It's providing the top-level key that we use for static validation.   Since the expression can be dynamic (request.query.name1), we can only validate the first term. 

---
In reply to: [114848605](https://github.com/Azure/azure-webjobs-sdk/pull/1117#discussion_r114848605) [](ancestors = 114848605)

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
MikeStall: sure. 

---
In reply to: [114849420](https://github.com/Azure/azure-webjobs-sdk/pull/1117#discussion_r114849420) [](ancestors = 114849420)

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
```diff
@@ -1,32 +1,271 @@
...
+            }
+
+            // BindingData takes precedence over builtins.  
+            BindingParameterResolver builtin;
+            BindingParameterResolver.TryGetResolver(expression, out builtin);
```
mathewc: Can you make this built-in v.s. expression distinction stricter. They can't be mixed, so we should have:

```csharp
if (BindingParameterResolver.TryGetResolver(expression, out builtIn))
{
}
else
{
}
```

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
```diff
@@ -1,32 +1,271 @@
...
+                }
+            }
+
+            if (builtin == null)
+            {
+                return new ExpressionToken(parts);
+            }
+            else
+            {
+                return new ExpressionTokenOrBuiltin(parts, builtin);
```
mathewc: Can you do away with this ExpressionTokenOrBuiltin type? It doesn't seem right. In this code you should simply be able to create the right concrete type and let polymorphism do its thing

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
```diff
@@ -1,32 +1,271 @@
...
+                if (this._expressionParts.Length != 1)
+                {
+                    throw new InvalidOperationException($"Capture expressions can't include dot operators");
+                }
+
+                var value = this._expressionParts[0];
+                builder.Append(String.Format(CultureInfo.InvariantCulture, "(?<{0}>.*)", value));
+            }
+
+            protected virtual string Fallback()
```
mathewc: This Fallback pattern seems like a hack

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
MikeStall: I wish I could, but the problemw as our smeantics are too dynamic.  binding  info takes precedence over builtins, so we don't know which until after we get runtime binding info. 

---
In reply to: [114852808](https://github.com/Azure/azure-webjobs-sdk/pull/1117#discussion_r114852808) [](ancestors = 114852808)

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
```diff
@@ -1,32 +1,271 @@
...
+            public static object GetProperty(object o, string member)
+            {
+                Type scope = o.GetType();
+                JObject jobj = o as JObject;
+                if (jobj != null)
+                {
+                    JToken propValue;
+                    // JObject's normal dot operator returns null for missing properties; 
+                    // we need to explicitly call TryGetValue.
+                    if (!jobj.TryGetValue(member, StringComparison.OrdinalIgnoreCase, out propValue))
```
mathewc: Need to add support for dictionaries. E.g. the many of the existing bindings include IDictionary<string, object> as part of their contract, e.g. Blobs,EventHub,ServiceBus. E.g. this code should work (I tried it and it doesn't):

```csharp
public static void BlobTriggerToBlob(
    [BlobTrigger("test")] string blob,
    [Blob("output/{metadata.foo}")] string outBlob)
{
    Console.WriteLine("Processed blob: " + blob);
}
```

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
```diff
@@ -1,32 +1,271 @@
...
+                        throw new InvalidOperationException($"property doesn't exist.");
+                    }
+                    return propValue;
+                }
+                else
+                {
+                    var prop = scope.GetProperty(member, BindingFlags.Public | BindingFlags.Instance | BindingFlags.IgnoreCase);
+                    if (prop == null)
+                    {
+                        throw new InvalidOperationException($"property doesn't exist.");
```
mathewc: nit: need better exception messages here and below

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Queues/QueueTests.cs
```diff
@@ -175,6 +175,60 @@ public void InvokeWithBindingData()
...
+            public class SubOject
+            {
+                public string xyz { get; set; }
+            }
+
+
+            // BindingData is case insensitive. 
+            // And queue name is normalized to lowercase. 
+            public const string QueueOutName = "qName-{prop1.xyz}";
+            public void Func([QueueTrigger(QueueName)] Poco triggers, [Queue(QueueOutName)] ICollector<string> q)
```
mathewc: Can you also add a `string prop1` parameter to this method? I.e., hopefully these binding expressions can be bound to as normal parameters? In this case, I think we might already have poco handling which will take care of this - my question is more for properties that are resolved from metadata other than from the poco - e.g. CloudBlockBlob.Metadata which the blob trigger adds. Can we dot into the metadata as a parameter?

MikeStall: prop1 is of type 'SubObject', not string.   So I'd expect 'string prop1' would fail with some sort of type-mismatch. 
But 'SubObject prop1' makes sense.  Is that what you'd expect?


---
In reply to: [114873333](https://github.com/Azure/azure-webjobs-sdk/pull/1117#discussion_r114873333) [](ancestors = 114873333)

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
MikeStall: I'll look at IDict support. It would require some reflection ninja skills to handle an arbitrary IDict<string,T>  where T is obj, string, JToken, etc. 


---
In reply to: [114872172](https://github.com/Azure/azure-webjobs-sdk/pull/1117#discussion_r114872172) [](ancestors = 114872172)

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
MikeStall: these will get wrapper by the caller and it will append additional information like the method name. Is that what you had in mind?

---
In reply to: [114872462](https://github.com/Azure/azure-webjobs-sdk/pull/1117#discussion_r114872462) [](ancestors = 114872462)

mathewc: I was thinking like "Property '{name}' doesn't exist.". However I might not be aware of your stack based concat strategy. However this message in isolation is not very informative :)

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
MikeStall: added in next iteration, complete with Blob unit test for metadata.

---
In reply to: [114885358](https://github.com/Azure/azure-webjobs-sdk/pull/1117#discussion_r114885358) [](ancestors = 114885358,114872172)

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
MikeStall: the caller will add {name}. I have some unit tests for this too. 

---
In reply to: [114904170](https://github.com/Azure/azure-webjobs-sdk/pull/1117#discussion_r114904170) [](ancestors = 114904170)

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Queues/QueueTests.cs
MikeStall: discussed in person. Going the 'subobject prop1' approach. Added to tests. 

---
In reply to: [114880093](https://github.com/Azure/azure-webjobs-sdk/pull/1117#discussion_r114880093) [](ancestors = 114880093,114873333)

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
MikeStall: The problem is they *can* be mixed  ("datetime" is both a builtin and possible binding data) 

---
In reply to: [114852342](https://github.com/Azure/azure-webjobs-sdk/pull/1117#discussion_r114852342) [](ancestors = 114852342)

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
MikeStall: see latest iteration.

---
In reply to: [114856116](https://github.com/Azure/azure-webjobs-sdk/pull/1117#discussion_r114856116) [](ancestors = 114856116,114852808)

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplateToken.cs
```diff
@@ -1,32 +1,293 @@
...
+            // Support:
+            //   IDictionary<string, T>;
+            //   IReadOnlyDictionary<string, T>
+            // There's no common base interface here, and T can be { string, object, poco }, 
+            // so we just directly reflect over 'bool TryGetValue(string, out T)' and invoke that. 
+            private static bool TryGetValue(object o, string member, out object result)
+            {
+                Type scope = o.GetType();
+
+                var method = scope.GetMethod("TryGetValue", BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic);
```
mathewc: Matt pointed it out before, that this is a lot of reflection to do for every single property access. Since the 95% case will be one of IDictionary<string, string>, IDictionary<string, object>, or a read only variety, I wonder if we shouldn't special case those?

Another idea might be to use the indexer:
```csharp
private static bool TryGetIndexedValueOrNull(object target, string name, out object value)
{
    value = null;

    var indexer = target.GetType().GetProperties()
        .Where(p => p.Name == "Item" && 
                p.GetIndexParameters().Count() == 1 && 
                p.GetIndexParameters()[0].ParameterType == typeof(string))
        .FirstOrDefault();
            
    if (indexer != null)
    {
        try
        {
            value = indexer.GetValue(target, new object[] { name });
            return true;
        }
        catch
        {
        }
    }

    return false;
}
```

PR #1118, opened 2017-04-21
Adding Return Value attribute for bindings

soninaren:


fabiocav:
My concern with this is that it won't work for WebJobs SDK users, so the attribute usage will be misleading in those scenarios.
@mathewc , thoughts?

fabiocav:
My recommendation would be to hold on this changes until the proper support is in place so that it works as expected.

PR #1119, opened 2017-04-21
Implement event hub connection

### LINKED ISSUE #1114: Add a "connection" property to EventHubTriggerAttribute
For the Visual Studio tooling to produce function.json, we need a "connection" property in [EventHubTriggerAttribute](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubTriggerAttribute.cs). For SDK customers, we need to wire up the connection string.

It should be called `Connection` to make serialization easy.

mamaso:
resolves #1114, implementation of #1115 

Also added Connection to EventHubAttribute


src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubAttribute.cs
```diff
@@ -27,5 +27,10 @@ public EventHubAttribute(string eventHubName)
         /// </summary>
         [AutoResolve]
         public string EventHubName { get; private set; }
+
+        /// <summary>
+        /// Optional connection name. If missing, tries to use a registered event hub sender.
+        /// </summary>
+        public string Connection { get; set; }
```
mathewc: Why aren't you using the IConnectionProvider approach all the other attributes use? We now handle these in a consistent way across all attributes.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
```diff
@@ -201,17 +216,24 @@ public void AddReceiver(string eventHubName, string receiverConnectionString, st
...
+        internal EventHubClient GetEventHubClient(string eventHubName, string connection)
         {
             EventHubClient client;
             if (_senders.TryGetValue(eventHubName, out client))             
             {
                 return client;
             }
+            else if (!string.IsNullOrWhiteSpace(connection) && NameResolver != null)
+            {
+                var resolvedConnection = NameResolver.Resolve(connection);
```
mathewc: In the IConnectionProvider implementation, the Connection property is marked as [AppSetting] so you shouldn't need to resolve this on your own?

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubAttribute.cs
mamaso: Yeah, I considered it but had some misconceptions about how AppSettingAttribute handled null values. Investigated further, it should work, I'll make the change.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
mamaso: Misconception on my part - thought that AppSetting attribute would throw if there was no value specified. It does, assuming that 1. there is a connection set, explicitly or via default and 2. there is no app setting/env variable with that connection name.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubAttribute.cs
MikeStall: It's good that this is [AppSetting]. IConnectionProvider just gives it a way to pull the connection from the existing ServiceBusAccountAttribute.  But they seem like they could be used together. 

---
In reply to: [113313470](https://github.com/Azure/azure-webjobs-sdk/pull/1119#discussion_r113313470) [](ancestors = 113313470)

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
```diff
@@ -201,17 +201,23 @@ public void AddReceiver(string eventHubName, string receiverConnectionString, st
...
             EventHubClient client;
             if (_senders.TryGetValue(eventHubName, out client))             
             {
                 return client;
             }
+            else if (!string.IsNullOrWhiteSpace(connection))
+            {
+                AddSender(eventHubName, connection);
+                return _senders[eventHubName];
+            }
```
MikeStall: This is the key part of the change, right?

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubTriggerAttribute.cs
```diff
@@ -35,6 +36,7 @@ public EventHubTriggerAttribute(string eventHubName)
         /// <summary>
         /// Optional connection name. If missing, tries to use a registered event hub receiver.
         /// </summary>
+        [AppSetting]
         public string Connection { get; set; }
```
MikeStall: >Connection [](start = 22, length = 10)

Where is this used?

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
mamaso: Yep

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubAttribute.cs
```diff
@@ -10,8 +10,9 @@ namespace Microsoft.Azure.WebJobs.ServiceBus
     /// Setup an 'output' binding to an EventHub. This can be any output type compatible with an IAsyncCollector.
     /// </summary>
     [AttributeUsage(AttributeTargets.Parameter)]
+    [ConnectionProvider(typeof(ServiceBusAccountAttribute))]
```
mathewc: Remove this - ServiceBus doesn't really apply here, and I don't think we need to support method/class level overrides. The SBAA is only for service bus.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubTriggerAttribute.cs
mamaso: It's automatically resolved when the attribute is cloned and passed to BindToInput rule

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubAttribute.cs
```diff
@@ -27,5 +28,11 @@ public EventHubAttribute(string eventHubName)
         /// </summary>
         [AutoResolve]
         public string EventHubName { get; private set; }
+
+        /// <summary>
+        /// Optional connection name. If missing, tries to use a registered event hub sender.
+        /// </summary>
+        [AppSetting]
```
mathewc: None of the other attributes are applying [AppSetting] at this level. We should make these all consistent after testing/verifying what the correct behavior should be.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubAttribute.cs
mamaso: Fine with me to disallow method/class lvl - though EventHubs are service bus entities, so it's an option.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubAttribute.cs
mamaso: 'Correct' behavior would be [AppSetting] to use build in resolution - all the other built ins use custom nameResolution code, which could likely be converted to AppSetting but is out of scope here

mamaso: #1125

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubAttribute.cs
mathewc: I was just thinking that applying SBAA and having it affect EventHub bindings might be confusing for people.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubAttribute.cs
mathewc: Ok. But you're saying that the [AppSetting] that I have applied on the base interface actually doesn't do anything and it can be removed with no effect?

mamaso: Correct

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubAttribute.cs
mamaso: Yep, good point - I'm happy with having no hierarchical pieces to this

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubTriggerAttributeBindingProvider.cs
```diff
@@ -44,20 +44,19 @@ public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext contex
...
 
-            string consumerGroup = attribute.ConsumerGroup;
-            if (consumerGroup == null)
-            {
-                consumerGroup = Microsoft.ServiceBus.Messaging.EventHubConsumerGroup.DefaultGroupName;
-            }
+            string consumerGroup = attribute.ConsumerGroup ?? EventHubConsumerGroup.DefaultGroupName;
             string resolvedConsumerGroup = _nameResolver.ResolveWholeString(consumerGroup);
 
+            if (!string.IsNullOrWhiteSpace(attribute.Connection))
```
MikeStall: >if (!string.IsNullOrWhiteSpace(attribute.Connection)) [](start = 12, length = 53)

This is what I was looking for! So this is the key change for triggering, right?

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubTriggerAttribute.cs
MikeStall: I was looking for what used it. I found it in EventHubTriggerAttributeBinding.

---
In reply to: [113331328](https://github.com/Azure/azure-webjobs-sdk/pull/1119#discussion_r113331328) [](ancestors = 113331328)

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubAttribute.cs
```diff
@@ -27,5 +27,11 @@ public EventHubAttribute(string eventHubName)
         /// </summary>
         [AutoResolve]
         public string EventHubName { get; private set; }
+
+        /// <summary>
```
mathewc: nit: <inheritdoc/>

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubTriggerAttribute.cs
```diff
@@ -35,6 +35,7 @@ public EventHubTriggerAttribute(string eventHubName)
         /// <summary>
```
mathewc: nit: <inheritdoc/>

Issue #1120, opened 2017-04-22
Exceptions in QueueProcessor.DeleteMessageAsync() are not logged & there is no retry logic


mahedley:
QueueProcessor.DeleteMessageAsync() catches exceptions, and does not log that they occurred or retry the deletion of a message. This leads to unexpected duplicates that are difficult to diagnose.

In some cases a storage account queue may be temporarily unavailable due to noisy neighbors or transient issues with the queue. Currently WebJobs will catch an exception due to the queue being unavailable, and then drop the exception without logging it. This can lead to messages that have been processed returning to the queue unexpectedly causing duplicates. See QueueProcessor line 215. 

Since this is outside the user's logic they have no way to catch these exceptions. Since the behavior isn't logged it makes it difficult to troubleshoot and pinpoint these duplicate messages

I propose updating DeleteMessageAsync to log  when they are catching an exception, and also expanding JobHostQueuesConfiguration to allow for a retry count for message deletion.

christopheranderson:
Thanks for the PR. @brettsam  and @MikeStall will look into this.

brettsam:
I'm modifying this -- it's still a good idea to log the errors even if we don't add new retry logic (see #1121 for details there).

PR #1121, opened 2017-04-23
Add Logging, and Retry Logic Delete in Queue Processor. 

mahedley:
This is added in response to the issue I logged #1120 .

I've added retry logic for message, or queue not found, tied to a new parameter in IQueueConfiguration. This is defaulted to 0 so basic behavior is unchanged, however retries will be attempted if the number increases.

I've also added log entries for exceptions so that if a logger is provided it is easier to track an issue caused by the queue being temporarily available. 

The goal is to make it easier to troubleshoot and resolve issues related to transient 503 errors being raised by the storage layer. In some cases with heavily used queues I've seen 503's being raised for messages that still exist. This can cause the existing logic to leave the message on the queue, because those exceptions are being dropped.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/JobHostQueuesConfiguration.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/JobHostQueuesConfiguration.cs
@@ -21,11 +21,14 @@ namespace Microsoft.Azure.WebJobs.Host
         // the runtime error message the user would receive from the SDK otherwise is not as helpful.
         private const int MaxBatchSize = 32;
 
+        private const int DefaultRetryCount = 0;
+
         private int _batchSize = DefaultBatchSize;
         private int _newBatchThreshold;
         private TimeSpan _maxPollingInterval = QueuePollingIntervals.DefaultMaximum;
         private TimeSpan _visibilityTimeout = TimeSpan.Zero;
         private int _maxDequeueCount = DefaultMaxDequeueCount;
+        private int _deleteRetryCount = DefaultRetryCount;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="JobHostQueuesConfiguration"/> class.
@@ -64,14 +67,14 @@ namespace Microsoft.Azure.WebJobs.Host
         /// </summary>
         public int NewBatchThreshold
         {
-            get 
-            { 
+            get
+            {
                 if (_newBatchThreshold == -1)
                 {
                     // if this hasn't been set explicitly, default it
                     return _batchSize / 2;
                 }
-                return _newBatchThreshold; 
+                return _newBatchThreshold;
             }
 
             set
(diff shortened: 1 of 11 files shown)


mahedley:
Reproducing today and the exception may actually being handled in the final piece of exception logic, where the exception is re-thrown. I'll validate and submit a new commit if appropriate.

src/Microsoft.Azure.WebJobs.Host/JobHostQueuesConfiguration.cs
```diff
@@ -172,10 +175,28 @@ public TimeSpan VisibilityTimeout
...
+            set;
+        }
+
+        /// <summary>
+        /// Get's or sets the delete retry count
+        /// In some cases a queue may be unavailable temporarily. Retrying in some cases can
+        /// reduce the chance that a successfully processed message can return to the queue.
+        /// The default is 0
+        /// </summary>
+        public int DeleteRetryCount
```
mahedley: A default makes sense. I was thinking 5. Does it make sense to just put this in a private variable inside of QueueProcessor, or is there a better place to keep something like this?

brettsam:
Thanks for putting this together @mahedley! 

I have one overall comment first -- I remembered somewhere that the Storage clients allowed you to plug in your own retry options that they could handle internally for all operations. I looked around and found this: https://docs.microsoft.com/en-us/azure/architecture/best-practices/retry-service-specific#azure-storage-retry-guidelines. When I look at a freshly-crated CloudQueueClient, I see that it already defaults to using `ExponentialRetry` (with a default backoff of 4 seconds and max retries of 3), which means that this is a retry wrapped in a retry, which generally isn't a good idea.

I'd suggest removing the retries completely from here but leaving the logging for errors (the silent handling of errors has been problematic). If you want to increase your own retries, you can do that using your own `StorageClientFactory` and plugging in your own `RetryPolicy` to the client.

As a secondary question -- were you getting your 503s due to being throttled by the Storage service? You mentioned it being a heavily-used queue. If you're being throttled, I believe it's expected that you'll have a period of downtime until the throttling period is over, which explains why you see 503s here (and why even the built-in exponential retries don't help).

mahedley:
@brettsam Thanks Brett, this actually makes a lot of sense. Once we implemented a custom `QueueProcessor `that logged things out we started to see a slightly different picture. The system logs do occasionally show a 503, but these are not bubbling up to the retry logic, which makes sense based on what you've found in `CloudQueueClient`. 

What we are seeing within the logs are `TaskCancelledException`s, and when we review the system log we can see that the` JobHost` was shutdown. Having a look at the app service I can see all of instances recycle around the time that the customer is seeing the duplicates being generated. 

I'll take my existing changes and back off to just logging, and make an additional catch around the base `Exception` so that it can be logged as well. 

It would seem to also make sense to wrap `CompleteProcessingMessageAsync` in a try catch to attempt to complete the message deletion in the event of the JobHost going down. I can take a run at this and add a note in Issue #1120 if you agree.

brettsam:
Hey @mahedley -- this has been sitting for a while. Did you get a chance to walk this back to just the logging bits for now?

brettsam:
I'm going to close this PR out -- feel free to re-open a new PR with logging pieces. Thanks!

PR #1123, opened 2017-04-25
Fix handly empty connection string, 

MikeStall:
Fixes https://github.com/Azure/azure-webjobs-sdk-script/issues/1456
Add unit tests

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/StorageAccountProviderExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/StorageAccountProviderExtensions.cs
@@ -41,7 +41,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                 throw new ArgumentNullException("provider");
             }
 
-            connectionStringName = connectionStringName ?? ConnectionStringNames.Storage;
+            connectionStringName = string.IsNullOrWhiteSpace(connectionStringName) ? ConnectionStringNames.Storage : connectionStringName;
 
             if (nameResolver != null)
             {
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/StorageAccountProviderExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/StorageAccountProviderExtensions.cs
@@ -41,7 +41,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                 throw new ArgumentNullException("provider");
             }
 
-            connectionStringName = string.IsNullOrWhiteSpace(connectionStringName) ? ConnectionStringNames.Storage : connectionStringName;
+            connectionStringName = string.IsNullOrEmpty(connectionStringName) ? ConnectionStringNames.Storage : connectionStringName;
 
             if (nameResolver != null)
             {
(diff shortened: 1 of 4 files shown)


src/Microsoft.Azure.WebJobs.Host/Executors/StorageAccountProviderExtensions.cs
```diff
@@ -41,7 +41,7 @@ public static Task<IStorageAccount> GetDashboardAccountAsync(this IStorageAccoun
                 throw new ArgumentNullException("provider");
             }
 
-            connectionStringName = connectionStringName ?? ConnectionStringNames.Storage;
+            connectionStringName = string.IsNullOrWhiteSpace(connectionStringName) ? ConnectionStringNames.Storage : connectionStringName;
```
mathewc: nit: I'd change this to IsNullOrEmpty

MikeStall: ok

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Queues/QueueTests.cs
```diff
@@ -143,8 +143,9 @@ public class Poco
 
             // BindingData is case insensitive. 
             // And queue name is normalized to lowercase. 
+            // Connection="" is same as Connection=null
             public const string QueueOutName = "qName-{XYZ}";
-            public void Func([QueueTrigger(QueueName)] Poco triggers,  [Queue(QueueOutName)] ICollector<string> q)
+            public void Func([QueueTrigger(QueueName, Connection ="")] Poco triggers,  [Queue(QueueOutName)] ICollector<string> q)
```
davidebbo: Do we still have coverage elsewhere for the case with no connection?

mathewc: Rather than have the test hidden here, it would be better to have a simple targeted unit test on the GetStorageAccountAsync method itself. Having it hear doesn't make it clear that this is something we're trying to regression test.

MikeStall: Yes, the vast majority of are missing the Connection property altogether. So I just commandeered a case and tweaked it.  A little pair-wise testing.

MikeStall: Let's do both.   The dedicated unit test is goodness. But it's also good to have the end-2-end usage.

davidebbo: Might make sense, but we can do that later. Right now, we need to move forward asap to get this through the stack.

mathewc: Then I'd just remove this from here and rely on the tooling test below which tests this, until an actual unit test for this method is added.

MikeStall:
Both unit tests are now added. :)

MikeStall:
squashed and merged

Issue #1013, opened 2017-02-09
Need proper solution for SetPostResolveHook


mathewc:
We have several bindings that currently rely on this behavior, including 2 of our own as well as an external 3rd party binding (Bot Framework).

We recently made this functionality internal, but we ourselves continue to use it because it is useful. However that has broken external customers. We shouldn't have our own bindings using any substantial internal infrastructure like this - if we need it, it is likely useful to eternal binding authors as well.

To unblock the Bot guys (and avoid breaking changes for them) I put in a temporary hack using reflection to invoke this internal method. We need to remove that.

MikeStall:
For background, this hook does 2 very tactical things, mainly around supporting a fixed set of backwards-compat scenarios that shouldn't apply to extensions: 
1. Preserve existing rich Dashboard support in the core bindings. Our new bindings don’t customize this. We can transfer this responsibility to the binding rules and find good ways to customizations for other bindings – but that was a lower priority and so not done yet.  
2. Support cases where an attribute is split across multiple attributes, ie BlobStorage and StorageAccount.   From the Function's perspective, this is an anti-pattern since it breaks the 1:1 mapping between attributes and Function.json. New extensions shouldn't do this.  Instead, new extensions should have all relevant properties within a single attribute. 

Both of these motivations are limited in scope and don’t apply to new extensions, so new extensions shouldn’t need the hook.

mathewc:
If we can do away with the hooks completely then great. This issue is mainly about making any changes we need to remove the private reflection hack in the Bot extension. If that means getting them to take the breaking change, then we might not have to give them any new API. However it's tricky because we made the API public, they built on it, then we did a take back.

mathewc:
I think this can be closed now, right @MikeStall ?

Issue #1110, opened 2017-04-13
Need better error on misusing [AutoResolve]


MikeStall:
[AutoResolve] should only be used on string properties. IF you use it on a non-string , you get a cryptic crash in the SDK runtime.  It should give you a user error.

MikeStall:
Open question:
- is autoresolve forbidden on- non-string? Or do we coerce to a the target value. 

We fail here where we directly cast the value to a string. 
https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs#L110

PR #1124, opened 2017-04-25
Default to member name

MikeStall:
1. This lets us remove that  hook.
2. Removes evil hack: 
https://github.com/Azure/azure-webjobs-sdk/issues/1013 
3. Remove a bunch of frivolous Async paths that should be sync


src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/FluidBindingProviderExtensions.cs
```diff
@@ -15,12 +15,10 @@ internal static class FluidBindingProviderExtensions
     {
         public static IBindingProvider SetPostResolveHook<TAttribute>(
             this IBindingProvider binder,
-            Func<TAttribute, ParameterInfo, INameResolver, ParameterDescriptor> buildParameterDescriptor = null,
-            Func<TAttribute, ParameterInfo, INameResolver, Task<TAttribute>> postResolveHook = null)
+            Func<TAttribute, ParameterInfo, INameResolver, ParameterDescriptor> buildParameterDescriptor = null)
```
mathewc: nit: can you search for "fluid" and fix with "fluent"? :)

MikeStall: good catch.

test/Microsoft.Azure.WebJobs.Host.UnitTests/AttributeClonerTests.cs
```diff
@@ -64,6 +65,13 @@ public class Attr4: Attribute
             [AutoResolve]
             public string AutoResolve { get; set; }
         }
+        
+        // Test with DefaultValue.MemberName
+        public class Attr5 : Attribute
+        {
+            [AutoResolve(Default = DefaultResolveValue.MemberName)]
```
mathewc: For the purposes of bindings, the member will ALWAYS be a method name, right? Your code even hard codes to method internally. E.g. we won't every use this for properties, events, etc. So I think we should rename to MethodName?

mathewc: Also, perhaps rename the enum to AutoResolveValue?

src/Microsoft.Azure.WebJobs/DefaultResolveValue.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Description
+{
+    /// <summary>
+    /// If set,
```
mathewc: Please fill out xml text fully

src/Microsoft.Azure.WebJobs/DefaultResolveValue.cs
```diff
@@ -0,0 +1,22 @@
...
+    public enum DefaultResolveValue
+    {
+        /// <summary>
+        /// Don't default 
+        /// </summary>
+        None,
+
+        /// <summary>
+        /// Default to this function name. 
+        /// This may use the <see cref="FunctionNameAttribute"/>
```
mathewc: This FunctionNameAttribute comment is not clear. I'd just remove it.

src/Microsoft.Azure.WebJobs.Host/Bindings/ClonerContext.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Description;
+using Microsoft.Azure.WebJobs.Host.Bindings.Path;
+
+namespace Microsoft.Azure.WebJobs.Host.Bindings
+{
+    /// <summary>
+    ///  Context passed to <see cref="AutoResolveAttribute"/> to aide in resolution. 
+    /// </summary>
+    internal class ClonerContext
```
mathewc: Rename to AttributeClonerContext

src/Microsoft.Azure.WebJobs.Host/Bindings/ClonerContext.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        public static ClonerContext New(INameResolver resolver)
+        {
+            return new ClonerContext
+            {
+                NameResolver = resolver,
+            };
+        }
+
+        public static ClonerContext New(INameResolver resolver, ParameterInfo parameter)
```
mathewc: What happens in the case of Binder.BindAsync calls where the ParameterInfo will be fake - it still has a valid MemberInfo, right? Probably be good to do a Functions Node test to make sure

mathewc: Also, for the return value bindings we're talking about, there won't be a ParameterInfo. What are we going to do in those cases?

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -34,21 +35,31 @@ internal class AttributeCloner<TAttribute>
...
-            Func<TAttribute, Task<TAttribute>> hook = null)
+            ClonerContext context = null)
         {
-            _hook = hook;
-            nameResolver = nameResolver ?? new EmptyNameResolver();
+            if (context == null)
+            {
+                context = new ClonerContext();
+            }
+            _methodName = context.MethodName;
```
mathewc: recommend a field for the context so as the context grows over time we don't have to copy individual values off?

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -107,10 +118,21 @@ private BindingDataResolver GetResolver(PropertyInfo propInfo, INameResolver nam
...
             }
             // try to resolve with auto resolve ({...}, %...%)
-            if (autoResolveAttr != null && originalValue != null)
+            if (autoResolveAttr != null)
             {
-                _autoResolves[propInfo] = autoResolveAttr;
-                return GetTemplateResolver((string)originalValue, autoResolveAttr, nameResolver, propInfo, contract);
+                if (autoResolveAttr.Default == DefaultResolveValue.MemberName)
+                {
+                    if (originalValue == null || string.IsNullOrWhiteSpace(originalValue.ToString()))
```
mathewc: Seems like these should be part of the if predicate - we only want to do this default resolution if the target doesn't already have a value

src/Microsoft.Azure.WebJobs.Host/Bindings/ClonerContext.cs
MikeStall: This is all internal. When we implement return values, we can adjust this as needed. (perhaps we pass in the member.NAme directly instead of ParameterInfo)

---
In reply to: [113583638](https://github.com/Azure/azure-webjobs-sdk/pull/1124#discussion_r113583638) [](ancestors = 113583638)

src/Microsoft.Azure.WebJobs/AutoResolveAttribute.cs
```diff
@@ -25,5 +26,10 @@ public AutoResolveAttribute()
         /// in the Microsoft.Azure.WebJobs.Host assembly.
         /// </summary>
         public Type ResolutionPolicyType { get; set; }
+
+        /// <summary>
+        /// If this is empty, then specify a default. 
+        /// </summary>
+        public DefaultResolveValue Default { get; set; } = DefaultResolveValue.None;
```
mathewc: Initially I was thinking that we wanted this to behave like AppSetting.DefaultValue which allows a string literal to be specified, which will be used if the target doesn't already have a value.

By using this enum, we can't express the same thing here. E.g. what if I want [AutoResolve(DefaultValue = "Mathew")], where "Mathew" is only used if no value comes from binding data, app settings, etc.?

src/Microsoft.Azure.WebJobs.Host/Bindings/ClonerContext.cs
MikeStall: This problem exists today (Bot already hits this).  The fake parameterInfo would get a fake MemberInfo with the name.  I can add a test in Script that illustrates this. 

---
In reply to: [113583425](https://github.com/Azure/azure-webjobs-sdk/pull/1124#discussion_r113583425) [](ancestors = 113583425)

src/Microsoft.Azure.WebJobs/AutoResolveAttribute.cs
MikeStall: I'm not seeing any extensions that have meaningful literal default values, so I'm hesitant to add that. 
I'd recommend that extension authors just check for null and handle appropriately. 


---
In reply to: [113586086](https://github.com/Azure/azure-webjobs-sdk/pull/1124#discussion_r113586086) [](ancestors = 113586086)

test/Microsoft.Azure.WebJobs.Host.UnitTests/AttributeClonerTests.cs
MikeStall: Yes, it's method. I'll rename.  I'm just taking what the exsiting usage is. 

---
In reply to: [113582170](https://github.com/Azure/azure-webjobs-sdk/pull/1124#discussion_r113582170) [](ancestors = 113582170)

src/Microsoft.Azure.WebJobs/DefaultResolveValue.cs
MikeStall: let me try to clarify it

---
In reply to: [113582917](https://github.com/Azure/azure-webjobs-sdk/pull/1124#discussion_r113582917) [](ancestors = 113582917)

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
MikeStall: This was tricky since originalValue isn't a string. 
That actually is directly related to:  
[https://github.com/Azure/azure-webjobs-sdk/issues/1110](https://github.com/Azure/azure-webjobs-sdk/issues/1110) 

I'll just fix them both. 

---
In reply to: [113585460](https://github.com/Azure/azure-webjobs-sdk/pull/1124#discussion_r113585460) [](ancestors = 113585460)

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -107,10 +118,27 @@ private BindingDataResolver GetResolver(PropertyInfo propInfo, INameResolver nam
                 return GetAppSettingResolver((string)originalValue, appSettingAttr, nameResolver, propInfo);
             }
             // try to resolve with auto resolve ({...}, %...%)
-            if (autoResolveAttr != null && originalValue != null)
+            if (autoResolveAttr != null)
             {
-                _autoResolves[propInfo] = autoResolveAttr;
-                return GetTemplateResolver((string)originalValue, autoResolveAttr, nameResolver, propInfo, contract);
+                if (propInfo.PropertyType != typeof(string))
```
mamaso: can you move this logic into its own GetAutoResolveResolver function? I think it improves readability & is better with single responsibility

src/Microsoft.Azure.WebJobs/AutoResolveValue.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Description
+{
+    /// <summary>
+    /// Specify the default behavior for an <see cref="AutoResolveAttribute"/> property when it has an empty value. 
+    /// These only apply if the property is empty. They don't apply if the property is an expression that evaluates to empty. 
+    /// </summary>
+    public enum AutoResolveValue
```
mamaso: it seems strange to me that AppSetting.Default is of string type and AutoResolve.Default is an enum

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeClonerContext.cs
```diff
@@ -0,0 +1,50 @@
...
+                 NameResolver = resolver,
+                 MethodName = parameter.Member.Name
+            };
+        }
+        public INameResolver GetResolverOrDefault()
+        {
+            return NameResolver ?? new EmptyNameResolver();
+        }
+
+        // If no name resolver is specified, then any %% becomes an error.
```
mamaso: should we throw an error here? is the null handled?

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -34,21 +35,31 @@ internal class AttributeCloner<TAttribute>
...
         private readonly Action<TAttribute, BindingData>[] _propertySetters;
 
-        // Optional hook for post-processing the attribute. This is intended for legacy hack rules.
-        private readonly Func<TAttribute, Task<TAttribute>> _hook;
-
         private readonly Dictionary<PropertyInfo, AutoResolveAttribute> _autoResolves = new Dictionary<PropertyInfo, AutoResolveAttribute>();
 
         private static readonly BindingFlags Flags = BindingFlags.Instance | BindingFlags.Public;
 
+        private readonly AttributeClonerContext _context;
```
mamaso: This private variable is not required

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -34,21 +35,31 @@ internal class AttributeCloner<TAttribute>
...
+        }
+
+        public AttributeCloner(
+            TAttribute source,
+            BindingDataContract bindingDataContract,
+            AttributeClonerContext context = null)
+        {
+            if (context == null)
+            {
+                context = new AttributeClonerContext();
```
mamaso: can you use a constructed object as a default parameter?

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeClonerContext.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        public static AttributeClonerContext New(INameResolver resolver, ParameterInfo parameter)
+        {
+            return new AttributeClonerContext
+            {
+                 NameResolver = resolver,
+                 MethodName = parameter.Member.Name
+            };
+        }
+        public INameResolver GetResolverOrDefault()
```
mamaso: nit: perhaps make this the getter for NameResolver instead of a separate method

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/AsyncCollectorBindingProvider.cs
```diff
@@ -316,7 +310,8 @@ private class ExactBinding<TMessage> : BindingBase<TAttribute>
                     };
                 }
 
-                var cloner = new AttributeCloner<TAttribute>(attributeSource, context.BindingDataContract, parent._nameResolver, hookWrapper);
+                var clonerCtx = AttributeClonerContext.New(parent._nameResolver, context.Parameter);
```
mamaso: since this occurs in so many places, I think I'd prefer an overload of AttributeCloner which accepts the name resolver and parameter

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -34,21 +35,31 @@ internal class AttributeCloner<TAttribute>
...
-        // Optional hook for post-processing the attribute. This is intended for legacy hack rules.
-        private readonly Func<TAttribute, Task<TAttribute>> _hook;
-
         private readonly Dictionary<PropertyInfo, AutoResolveAttribute> _autoResolves = new Dictionary<PropertyInfo, AutoResolveAttribute>();
 
         private static readonly BindingFlags Flags = BindingFlags.Instance | BindingFlags.Public;
 
+        private readonly AttributeClonerContext _context;
+
         public AttributeCloner(
```
mamaso: Do we want to maintain this constructor, or can we entirely move to the new overload?

src/Microsoft.Azure.WebJobs/AutoResolveValue.cs
mathewc: Yeah, that's the comment I was making above as well.

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeClonerContext.cs
MikeStall: This is the existing resolved, moved over from AttributeCloner class. I'm just preserving the existing logic. 

---
In reply to: [114160947](https://github.com/Azure/azure-webjobs-sdk/pull/1124#discussion_r114160947) [](ancestors = 114160947)

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
MikeStall: what is that?

---
In reply to: [114161428](https://github.com/Azure/azure-webjobs-sdk/pull/1124#discussion_r114161428) [](ancestors = 114161428)

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
MikeStall: we assign to it on line 60. 

---
In reply to: [114161355](https://github.com/Azure/azure-webjobs-sdk/pull/1124#discussion_r114161355) [](ancestors = 114161355)

src/Microsoft.Azure.WebJobs/AutoResolveValue.cs
MikeStall: Behaviorally, it makes sense. AppSetting specifies a string; AutoResolve specifies a fixed policy. 
So maybe AutoResolve.Default needs a better name. 



---
In reply to: [114159831](https://github.com/Azure/azure-webjobs-sdk/pull/1124#discussion_r114159831) [](ancestors = 114159831)

src/Microsoft.Azure.WebJobs.Host/WebJobs.Host.csproj
```diff
@@ -489,10 +489,11 @@
     <Compile Include="Bindings\BindingProviders\Rule.cs" />
     <Compile Include="Bindings\BindingProviders\FluentBindingProvider.cs" />
     <Compile Include="Bindings\BindingProviders\FilteringBindingProvider.cs" />
-    <Compile Include="Bindings\BindingProviders\FluidBindingProviderExtensions.cs" />
+    <Compile Include="Bindings\BindingProviders\FluentBindingProviderExtensions.cs" />
```
mathewc: Note that this change will break the Bot binding I believe since they're currently doing reflection. You'll have to stage this right :)

src/Microsoft.Azure.WebJobs/AutoResolveAttribute.cs
```diff
@@ -25,5 +26,10 @@ public AutoResolveAttribute()
         /// in the Microsoft.Azure.WebJobs.Host assembly.
         /// </summary>
         public Type ResolutionPolicyType { get; set; }
+
+        /// <summary>
+        /// If this is empty, then specify a default.
```
mathewc: nit: "Gets or sets the value to use if the target property doesn't contain a value."?

src/Microsoft.Azure.WebJobs.Host/WebJobs.Host.csproj
MikeStall: Right. I have a correpsonding BOT change. Staging would be:
- update SDK
- update BOT 
- update SCript with new SDK + new Bot 

---
In reply to: [114213832](https://github.com/Azure/azure-webjobs-sdk/pull/1124#discussion_r114213832) [](ancestors = 114213832)

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
mamaso: nvmd, default parameters must be compile time constants. Carry on :)

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
mamaso: ah nvmd on this - its usage is nested in member functions (admittedly these are only called in the constructor)

MikeStall:
Superseded by new PR, https://github.com/Azure/azure-webjobs-sdk/pull/1148

Issue #944, opened 2016-12-16
failures in QueueListener should not be fatal. 


MikeStall:
Failure to renew the queue message is critical, but failures in the Queue Listener (such as checking queue existence or calling GetMessage) should not be fatal.  Swallow the exception and retry. 

This happens in high load when a network call fails. Currently this brings down the process unnecessarily; instead, we can be a bit more resilient. 

[...]

PR #1127, opened 2017-04-27
Make QueueListener more resilient to transient errors (#944)

mathewc:
Fix for https://github.com/Azure/azure-webjobs-sdk/issues/944. This came up recently in a customer CRI (ICM #36139927). In that case, network latencies and other errors were causing the queue existence check to fail with a transient 503. I'm moving that check into our existing exception handling block.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Queues/Listeners/QueueListener.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Queues/Listeners/QueueListener.cs
@@ -142,15 +142,15 @@ namespace Microsoft.Azure.WebJobs.Host.Queues.Listeners
                 _stopWaitingTaskSource = new TaskCompletionSource<object>();
             }
 
-            if (!await _queue.ExistsAsync(cancellationToken))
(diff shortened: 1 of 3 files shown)


src/Microsoft.Azure.WebJobs.Host/Queues/Listeners/QueueListener.cs
```diff
@@ -142,15 +142,15 @@ public void Dispose()
...
-            if (!await _queue.ExistsAsync(cancellationToken))
-            {
-                // Back off when no message is available.
-                return CreateBackoffResult();
-            }
-
             IEnumerable<IStorageQueueMessage> batch;
             try
             {
+                if (!await _queue.ExistsAsync(cancellationToken))
```
mathewc: Actually looking at this code more, I don't see why we even have the overhead of a secondary existence check here. We have handling below for "queue not found" so I think we can just rely on that handling (optimistic) and save a network round trip.

mathewc:
Merged

PR #1132, opened 2017-05-01
[AppInsights]: improving Dispose() paths; simplifying ITelemetryClientFactory

brettsam:


src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ApplicationInsights/ApplicationInsightsLoggerProvider.cs
```diff
@@ -13,22 +13,47 @@ internal class ApplicationInsightsLoggerProvider : ILoggerProvider
...
+
+                if (_client != null)
+                {
+                    try
+                    {
+                        _client.Flush();
+
+                        // This sleep isn't ideal, but the flush is async so it's the best we have right now. This is
+                        // being tracked at https://github.com/Microsoft/ApplicationInsights-dotnet/issues/407
+                        Thread.Sleep(2000);
```
brettsam: Thoughts on this?

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ApplicationInsights/DefaultTelemetryClientFactory.cs
```diff
@@ -59,31 +78,50 @@ internal static string GetAssemblyFileVersion(Assembly assembly)
...
             return config;
         }
 
+        /// <summary>
+        /// Creates the <see cref="ITelemetryChannel"/> to be used by the <see cref="TelemetryClient"/>. If this channel
+        /// implements <see cref="ITelemetryModule"/> as well, <see cref="ITelemetryModule.Initialize(TelemetryConfiguration)"/> will
+        /// automatically be called.
+        /// </summary>
+        /// <returns>The <see cref="ITelemetryChannel"/></returns>
+        protected virtual ITelemetryChannel CreateTelemetryChannel()
```
brettsam: FYI -- this will be used by Script.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ApplicationInsights/DefaultTelemetryClientFactory.cs
```diff
@@ -92,5 +130,31 @@ internal static void AddInitializers(TelemetryConfiguration config)
...
+        /// Disposes the instance.
+        /// </summary>
+        /// <param name="disposing"></param>
+        protected virtual void Dispose(bool disposing)
+        {
+            if (disposing && !_disposed)
+            {
+                // TelemetryConfiguration.Dispose will dispose the Channel and the TelemetryProcessors
+                // registered with the TelemetryProcessorChainBuilder.
+                _config.Dispose();
```
mamaso: make these `?.Dispose` to help avoid issues with derived classes (these should override dispose anyway but..)

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ApplicationInsights/ApplicationInsightsLoggerProvider.cs
mamaso: Do we definitely have data loss without the sleep? Have you tried different values?

brettsam: Well, in testing E2E I see data loss without it. The TelemetryChannel we're using caches to the local file system so that anything it missed should eventually be sent. However, it'd be nicer if we didn't have to restart the process to pick it back up.

The Channel fires up a separate thread to send the data and we never see it. The suggestion to guarantee delivery was to wait a full minute, but that's not really reasonable in a shutdown path like this. In my experience, I was seeing my data come through with a two-second wait, but a one-second wait was not enough. I'd also be willing to go longer, but it'll apply to every host shutdown and still not even be guaranteed.

As you can see from the linked issue, others have this problem as well. As soon as there's a fix available, we can swap to use it.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ApplicationInsights/DefaultTelemetryClientFactory.cs
brettsam: Yeah good call.

Issue #1129, opened 2017-04-27
Extensions shouldn't need to depend on all the AppInsights dependencies


MikeStall:
With AppInsights, the SDK needs the ILogger interface, but that shouldn't pull in the rest of the AppInsights dependencies:
The problem is that these new dependencies get pushed to every single extension. 

We should minimize SDK dependencies. In this case, the AI dependencies could be in the existing WebJobs.Logging  package or in Script. 

  <package id="Microsoft.ApplicationInsights.Agent.Intercept" version="2.0.7" targetFramework="net46" />
  <package id="Microsoft.ApplicationInsights.DependencyCollector" version="2.3.0" targetFramework="net46" />
  <package id="Microsoft.ApplicationInsights.PerfCounterCollector" version="2.3.0" targetFramework="net46" />
  <package id="Microsoft.ApplicationInsights.WindowsServer" version="2.3.0" targetFramework="net46" />
  <package id="Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel" version="2.3.0" targetFramework="net46" />

brettsam:
I'm working on this now -- I'm pulling it into it's own new package `WebJobs.Logging.ApplicationInsights`. I first attempted to do this with the existing Logging package, but since it requires ILogger, all of those dependencies flowed to the Dashboard as well.

PR #1133, opened 2017-05-02
[AppInsights] Moving to separate package

brettsam:
Moving all Application Insights usage to a new package: `Microsoft.Azure.WebJobs.Logging.ApplicationInsights`. This removes the direct dependency on AI, which means Extensions don't need to know about it.

Addresses #1129.


src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Constants/LoggingKeys.cs
```diff
@@ -1,32 +1,101 @@
...
+using Microsoft.Extensions.Logging;
+
 namespace Microsoft.Azure.WebJobs.Host.Loggers
 {
-    internal static class LoggingKeys
+    /// <summary>
+    /// Keys used by the <see cref="ILogger"/> infrastructure.
+    /// </summary>
+    public static class LoggingKeys
     {
```
brettsam: Had to make these public. These could be useful to other ILoggers as well that plug into us. 

However, I could also just share the code file between the two and keep these internal. Thoughts?

FYI -- I'll fill in the descriptions.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Constants/ScopeKeys.cs
```diff
@@ -3,19 +3,19 @@
 
 namespace Microsoft.Azure.WebJobs.Host.Loggers
```
mathewc: Seems like the namespace should be Microsoft.Azure.WebJobs.Logging. That's the default namespace for the logging assembly

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Constants/ApplicationInsightsScopeKeys.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Host.Loggers
```
mathewc: Can you check/fix all the namespaces? I think you want to use the default namespace for your package: Microsoft.Azure.WebJobs.Logging.ApplicationInsights

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Constants/LoggingKeys.cs
```diff
@@ -1,32 +1,101 @@
...
-        // These are publicly visible as property names or prefixes
+        /// <summary>
+        /// </summary>
         public const string FullName = "FullName";
+
+        /// <summary>
```
fabiocav: Can we add the required suppressions so we don't have these empty summaries?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights.NuGet/WebJobs.Logging.ApplicationInsights.nuspec
```diff
@@ -0,0 +1,21 @@
...
+    <copyright>&#169; Microsoft Corporation. All rights reserved.</copyright>
+    <summary>Microsoft.Azure.WebJobs.Logging.ApplicationInsights is a library for writing WebJobs logs with ILogger and Application Insights.</summary>
+    <description>This package contains the runtime assemblies for Microsoft.Azure.WebJobs.Logging.ApplicationInsights. For more information, please visit http://go.microsoft.com/fwlink/?LinkID=320971</description>
+    <language>en-US</language>
+    <projectUrl>http://go.microsoft.com/fwlink/?LinkID=320972</projectUrl>
+    <licenseUrl>$PackageEULA$</licenseUrl>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <tags>Microsoft Azure WebJobs Jobs Logging Application Insights windowsazureofficial Web</tags>
+    <dependencies>
+      <dependency id="Microsoft.Azure.WebJobs" version="[$WebJobsPackageVersion$]" />
```
fabiocav: Are these all the nugets this depends on here? I guess `Microsoft.ApplicationInsights.WindowsServer` will pull the other AI packages, but if there are explicit dependencies, it would probably be good to have them here (JSON.NET, other AI packages?)

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,31 @@
...
+//      Major Version
+//      Minor Version
+//      Build Number
+//      Revision
+//
+// You can specify all the values or you can default the Build and Revision Numbers
+// by using the '*' as shown below:
+// [assembly: AssemblyVersion("1.0.*")]
+[assembly: AssemblyVersion("2.0.0.0")]
+[assembly: AssemblyFileVersion("2.0.0.0")]
```
fabiocav: Do we want this assembly to start at 2.0?

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Constants/LoggingKeys.cs
fabiocav: Or, you know, document them :)

brettsam: I commented above that they were coming -- but then I forgot them :-) 

I've already got a change coming with namespace updates recommended by @mathewc and I've filled these in as well.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Properties/AssemblyInfo.cs
brettsam: I thought it may be confusing to have one assembly starting at 1.0.0.0 -- but I'd be open to it. @mathewc, any thoughts?

brettsam: ... to expand on that thought. Whenever I'm grabbing a 1.x of an assembly that relies on a 2.x of an assembly within the same package namespace, I tend to think I'm missing an update somewhere. It feels nicer to me (even though it's not accurate) to have them match when they're coming from the same build.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights.NuGet/WebJobs.Logging.ApplicationInsights.nuspec
brettsam: All of our AI references come as dependencies from that one reference. All the others -- JSON, etc all come as a dependency on WebJobs.

PR #1141, opened 2017-05-10
fixing issues with Storage SDK 8.x

### LINKED ISSUE #985: WebJobs V1.1.2 fails to remove poison messages from queue with V8.0.1 of Storage
When using V8.0.1 of WindowsAzure.Storage V1.1.2 of WebJobs fail to remove poison messages from the queue. It copies the message to the poison queue but leaves the original message in the queue. The bad message gets processed again after the sleep timeout and generates another poison message. This goes on indefinitely.

This does not happen with Storage <= 7.1.2 for V1.1.2 of WebJobs.

#### Repro steps

Create a WebJob project using V1.1.2 of WebJobs and V7.1.2 of WindowsAzure.Storage

1. Create a queue handling function that raises an error when processing a message:

            public static async Task ProcessQueueMessage1([QueueTrigger("queue1")] string message,
            [Queue("queue2")] IAsyncCollector<string> queue2,
            TextWriter log)
            {
                try
                {
                    Task.Delay(10000).Wait();

                    log.WriteLine($"Queue1: {message}");

                    await queue2.AddAsync("Message for queue2");
                    throw new InvalidOperationException("Queue1 error.");
                }
                catch (Exception ex)
                {
                    log.WriteLine(ex.Message);
                    throw;
                }
            }

2. Add any message to a Queue named **queue1**.

3. The message will be added to the **queue1-poison** queue and removed from queue1.

4. Update the WindowsAzure.Storage package to 8.0.1

5. Republish the WebJob project

6. Add a message to **queue1**.

7. The message will be added to the **queue1-poison** queue and remain hidden in **queue1**.

8. After 10 minutes (the sleep timeout) a new poison message will be added and the message will remain in **queue1**. This goes on indefinitely.

#### Expected behavior

The poison message should be removed from **queue1** after it has been initially added to **queue1-poison** queue.

#### Actual behavior

The poison message is never removed from **queue1**

#### Known workarounds

Use version 7.1.2 or less of WindowsAzure.Storage.

brettsam:
Resolves #985

This re-sets the changed properties on a `CloudQueueMessage` after we pass it to a method that may be out of our control.

I ran all tests on 7.2.1 and 8.1.1 and they passed on both.


src/Microsoft.Azure.WebJobs.Host/PropertyHelper.cs
```diff
@@ -1,230 +1,526 @@
...
+using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
 
 namespace Microsoft.Azure.WebJobs.Host
 {
-    /// <summary>
-    /// Class used to facilitate reflection operations.
-    /// </summary>
+    // Original code here: https://github.com/aspnet/Common/blob/dev/shared/Microsoft.Extensions.PropertyHelper.Sources/PropertyHelper.cs
```
brettsam: I've refreshed this file from the source. The version we had didn't have a way to set values.

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Queues/QueueListenerTests.cs
```diff
@@ -0,0 +1,233 @@
...
+    {
+        public QueueListenerTests(TestFixture fixture)
+        {
+            Fixture = fixture;
+        }
+
+        public TestFixture Fixture { get; set; }
+
+        [Fact]
+        public async Task UpdatedQueueMessage_RetainsOriginalProperties()
```
brettsam: A big test that makes sure we pass the original values to each overloaded method.

src/Microsoft.Azure.WebJobs.Host/Extensions/CloudQueueMessageExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        /// <summary>
+        /// Overwrites the message's properties with the specified values.
+        /// </summary>
+        public static void SetProperties(this CloudQueueMessage message, string id, string popReceipt, DateTimeOffset? insertionTime,
+            DateTimeOffset? nextVisibleTime, DateTimeOffset? expirationTime)
+        {
+            IReadOnlyList<PropertyHelper> messageProperties = PropertyHelper.GetProperties(typeof(CloudQueueMessage));
+
+            PropertyHelper idProp = messageProperties.Single(p => p.Name == nameof(CloudQueueMessage.Id));
```
mathewc: Can these be initialized/cached statically?

src/Microsoft.Azure.WebJobs.Host/PropertyHelper.cs
mathewc: I had already pulled a copy into the Script repo and added set capabilities because they were lacking. Likely that version from script would work here as well. Your call. Benefit would be that we're using the same code, and perhaps could end up making it public in SDK so we can share (TBD).

src/Microsoft.Azure.WebJobs.Host/Queues/Listeners/QueueListener.cs
```diff
@@ -244,13 +244,23 @@ private TaskSeriesCommandResult CreateSucceededResult()
...
+            DateTimeOffset? expirationTime = message.ExpirationTime;
+
             try
             {
                 if (!await _queueProcessor.BeginProcessingMessageAsync(message.SdkObject, cancellationToken))
                 {
                     return;
                 }
 
+                message.SdkObject.SetProperties(id, popReceipt, insertionTime, nextVisibleTime, expirationTime);
```
mathewc: I wonder if we shouldn't just scope this to CopyMessageToPoisonQueue below to minimize impact. I know you've done this for completeness, but I think we'd be fine just scoping the fix to that?

src/Microsoft.Azure.WebJobs.Host/Queues/QueueProcessor.cs
```diff
@@ -119,6 +127,9 @@ public QueueProcessor(QueueProcessorFactoryContext context)
                 if (message.DequeueCount >= MaxDequeueCount)
                 {
                     await CopyMessageToPoisonQueueAsync(message, _poisonQueue, cancellationToken);
+
+                    message.SetProperties(id, popReceipt, insertionTime, nextVisibleTime, expirationTime);
```
mathewc: Can we put a simple conditional around this so we only do these updates if the pop receipt (and/or other values) have changed? That would keep this scoped to only scenarios where people are using 8.x, minimizing the chances of breaking changes or odd behaviors. Another possibility could be to make it conditional on the Azure SDK version.

mathewc: Also, can you comment the code with a TEMP that points to an issue in our repo tracking addressing this once we get Clone and move to 8.x (our next major version)?

src/Microsoft.Azure.WebJobs.Host/Extensions/CloudQueueMessageExtensions.cs
brettsam: Yeah good call. In the next update.

src/Microsoft.Azure.WebJobs.Host/Queues/QueueProcessor.cs
brettsam: I've built the conditional into SetProperties in my next update. It'll only update values if they've changed.

src/Microsoft.Azure.WebJobs.Host/PropertyHelper.cs
brettsam: I've grabbed the one from Script now.

src/Microsoft.Azure.WebJobs.Host/Queues/Listeners/QueueListener.cs
brettsam: It was difference between 1 and 3 updates, and I decided to go for completeness. I don't know what people may be doing from within their custom QueueProcessors and I'd rather not break things when we have a way to keep the existing behavior with 2 extra lines of code. I find it hard to explain "your Id and PopReceipt may have changed in these calls, but we'll reset it in this one".

src/Microsoft.Azure.WebJobs.Host/Extensions/CloudQueueMessageExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+            _popReceiptProp = messageProperties.Single(p => p.Name == nameof(CloudQueueMessage.PopReceipt));
+            _insertionTimeProp = messageProperties.SingleOrDefault(p => p.Name == nameof(CloudQueueMessage.InsertionTime));
+            _nextVisibleTimeProp = messageProperties.SingleOrDefault(p => p.Name == nameof(CloudQueueMessage.NextVisibleTime));
+            _expirationTimeProp = messageProperties.SingleOrDefault(p => p.Name == nameof(CloudQueueMessage.ExpirationTime));
+        }
+
+        /// <summary>
+        /// Overwrites the message's properties with the specified values, if different.
+        /// </summary>
+        public static void SetProperties(this CloudQueueMessage message, string id, string popReceipt, DateTimeOffset? insertionTime,
```
mathewc: nit: now that this is conditional, I'd rename to UpdateChangedProperties or something

mathewc (approved):

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/WebJobs.Host.FunctionalTests.csproj
```diff
@@ -168,7 +169,6 @@
     </Reference>
     <Reference Include="Microsoft.WindowsAzure.Storage, Version=7.2.1.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
       <HintPath>..\..\packages\WindowsAzure.Storage.7.2.1\lib\net40\Microsoft.WindowsAzure.Storage.dll</HintPath>
-      <Private>True</Private>
```
erikbra: I might be missing a bit of context, but since this fix is supposed to provide WindowsAzure.Storage 8.x compatibility, is it a good idea to test against WindowsAzure.Storage 7.2.1.0 with CopyLocal = true?

gruckion:
Azure webjob not appearing to respect MaxDequeueCount property #1045. Still having issues with this in 2019. This fix doesn't appear to have gone in. We are using Umbraco CMS version 7 which uses;

https://www.nuget.org/packages/UmbracoFileSystemProviders.Azure

So we cannot even downgrade to WindowsAzure.Storage version 7.2.1;

https://www.nuget.org/packages/WindowsAzure.Storage/

PR #1146, opened 2017-05-15
Allow optional/null parameter bindings for strings

mathewc:
Fix for https://github.com/Azure/azure-webjobs-sdk-extensions/issues/224. Related to a previous fix I did a while back for nullable types (PR https://github.com/Azure/azure-webjobs-sdk/pull/869). That handled structs, but not cases where non-struct types (e.g. strings) were null.


mathewc:
Merged

PR #1148, opened 2017-05-16
Default Name 

MikeStall:
Make AutoResolve.Default a string, and it can refer to a builtin 'sys' object that has a method name on it. This leverages the recent dot expression support. 
An alternative solution to https://github.com/Azure/azure-webjobs-sdk/pull/1124


src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -107,15 +103,46 @@ private BindingDataResolver GetResolver(PropertyInfo propInfo, INameResolver nam
...
             }
             // try to resolve with auto resolve ({...}, %...%)
-            if (autoResolveAttr != null && originalValue != null)
+            if (autoResolveAttr != null)
             {
-                _autoResolves[propInfo] = autoResolveAttr;
-                return GetTemplateResolver((string)originalValue, autoResolveAttr, nameResolver, propInfo, contract);
+                if (propInfo.PropertyType != typeof(string))
+                {
+                    throw new InvalidOperationException($"Property '{propInfo.Name}' with AutoResolve must be type string.");
```
mathewc: Can this new exception cause any regressions for existing apps?

mathewc: Also, recommend rewording to "AutoResolve property '{propInfo.Name}' must be of type string."

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -107,15 +103,46 @@ private BindingDataResolver GetResolver(PropertyInfo propInfo, INameResolver nam
...
+                }
+                var str = (string)originalValue;
```
mathewc: Now that you're requiring a string, I think you should move the GetValue call above down after you do the type validation, and define it as `string originalValue = propInfo.GetValue(_source) as string` so you don't need a second var

src/Microsoft.Azure.WebJobs.Host/Bindings/SysBindingData.cs
```diff
@@ -0,0 +1,84 @@
...
+    /// </summary>
+    /// <remarks>
+    /// It's expected this class is created and added to the binding data. 
+    /// </remarks>
+    internal class SysBindingData
+    {
+        // The name for this binding in the binding expressions. 
+        public const string Name = "sys";
+
+        // A name that can't be overwritten by the user that we cna place in the binding data dictionary
```
mathewc: nit: fix typos

src/Microsoft.Azure.WebJobs.Host/Bindings/SysBindingData.cs
```diff
@@ -0,0 +1,84 @@
...
+
+namespace Microsoft.Azure.WebJobs.Host.Bindings
+{
+    /// <summary>
+    /// Built in object for all binding expressions. 
+    /// </summary>
+    /// <remarks>
+    /// It's expected this class is created and added to the binding data. 
+    /// </remarks>
+    internal class SysBindingData
```
mathewc: nit: SystemBindingData?

src/Microsoft.Azure.WebJobs.Host/Bindings/SysBindingData.cs
```diff
@@ -0,0 +1,84 @@
...
+    {
+        // The name for this binding in the binding expressions. 
+        public const string Name = "sys";
+
+        // A name that can't be overwritten by the user that we cna place in the binding data dictionary
+        // and use to unambiguously retrieve this later. 
+        private const string InternalName = "$sys";
```
mathewc: nit: SystemBindingDataKey

mathewc: Also, I recommend we use MS_, like we do for all our system http request properties (e.g. see https://github.com/Azure/azure-webjobs-sdk-script/blob/0c2703b116456b4764d6082d74541e18a60ce04b/src/WebJobs.Script/ScriptConstants.cs#L11)

src/Microsoft.Azure.WebJobs.Host/Bindings/SysBindingData.cs
```diff
@@ -0,0 +1,84 @@
...
+        private static readonly IReadOnlyDictionary<string, Type> DefaultSysContract = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase)
+        {
+            { Name, typeof(SysBindingData) }
+        };
+
+        /// <summary>
+        /// The method name that the binding lives in. 
+        /// The method name can be override by the <see cref="FunctionNameAttribute"/> 
+        /// </summary>
+        public string MethodName { get; set; }
```
mathewc: Is the plan to expose our other built ins here as well? E.g. datetime, rand-guid, etc.?

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
MikeStall: We required a string before, see cast on line 124.  [https://github.com/Azure/azure-webjobs-sdk/issues/1110](https://github.com/Azure/azure-webjobs-sdk/issues/1110) 

---
In reply to: [116877366](https://github.com/Azure/azure-webjobs-sdk/pull/1148#discussion_r116877366) [](ancestors = 116877366)

src/Microsoft.Azure.WebJobs.Host/Bindings/SysBindingData.cs
MikeStall: The problem is that we need a unique name that can't conflict with user binding data (so that it doesn't get overwritten. I used '$' char for that. . 

---
In reply to: [116879586](https://github.com/Azure/azure-webjobs-sdk/pull/1148#discussion_r116879586) [](ancestors = 116879586)

src/Microsoft.Azure.WebJobs.Host/Bindings/SysBindingData.cs
MikeStall: Yes. However, they also require a formatting specifier, which was more work that deterred me. 

---
In reply to: [116881352](https://github.com/Azure/azure-webjobs-sdk/pull/1148#discussion_r116881352) [](ancestors = 116881352)

src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
```diff
@@ -22,11 +23,24 @@ public class FunctionBindingContext
...
+        public FunctionBindingContext(
+            Guid functionInstanceId, 
+            CancellationToken functionCancellationToken, 
+            TraceWriter trace,
+            FunctionDescriptor functionDescriptor = null)
         {
             _functionInstanceId = functionInstanceId;
             _functionCancellationToken = functionCancellationToken;
             _trace = trace;
+            if (functionDescriptor != null)
```
mamaso: nit: `this.MethodName = functionDescriptor?.Method?.Name;`

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingTemplateExtensions.cs
```diff
@@ -63,6 +63,10 @@ private static void ValidateContractCompatibility(IEnumerable<string> parameterN
             {
                 foreach (string parameterName in parameterNames)
                 {
+                    if (string.Equals(parameterName, SysBindingData.Name, StringComparison.OrdinalIgnoreCase))
```
mamaso: do we need to handle `SysBindingData.InternalName` as well?

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -107,15 +103,46 @@ private BindingDataResolver GetResolver(PropertyInfo propInfo, INameResolver nam
...
+
+                if (string.IsNullOrWhiteSpace(str))
+                {
+                    if (autoResolveAttr.Default != null)
```
mamaso: Agree with @mathewc - I'd love to see a refactoring pass over this to try to minimize the if statements here

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
MikeStall: The string requirement is only if you have a [AutoResolve] on it, so I still need the call on line 91. 

---
In reply to: [116878127](https://github.com/Azure/azure-webjobs-sdk/pull/1148#discussion_r116878127) [](ancestors = 116878127)

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingTemplateExtensions.cs
MikeStall: No. This validation is on public names. No user expression will refer to the internal name. 

---
In reply to: [116887647](https://github.com/Azure/azure-webjobs-sdk/pull/1148#discussion_r116887647) [](ancestors = 116887647)

src/Microsoft.Azure.WebJobs.Host/Bindings/SysBindingData.cs
mathewc: I can work to move those builtins over in a separate PR

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -145,6 +161,30 @@ internal static BindingDataResolver GetAppSettingResolver(string originalValue,
...
+        // These only have access to the {sys} builtin variable and don't get access to trigger binding data. 
+        internal static BindingDataResolver GetBuiltinTemplateResolver(string originalValue, INameResolver nameResolver)
+        {
+            string resolvedValue = nameResolver.ResolveWholeString(originalValue);
+
+            var template = BindingTemplate.FromString(resolvedValue);
+
+            SystemBindingData.ValidateStaticContract(template);
+
+            // Ignore trigger provided bindindData and use the sys binding data.
```
mathewc: We're not ignoring - we don't have access to it at all. Recommend clarifying or removing comment

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/FluentBindingProviderExtensions.cs
```diff
@@ -11,16 +11,14 @@ namespace Microsoft.Azure.WebJobs.Host
 {    
     // Internal Extension methods for setting backwards compatibility hooks on certain bindings. 
     // This keeps the hooks out of the public surface. 
-    internal static class FluidBindingProviderExtensions
+    internal static class FluentBindingProviderExtensions
```
mathewc: This is safe to do because you're going to send a corresponding PR removing the hook dependency from the Bot binding? Otherwise, this would break them.

mathewc: Why do we need this Type at all anymore?

src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
```diff
@@ -22,11 +23,18 @@ public class FunctionBindingContext
         /// <param name="functionInstanceId">The instance ID of the function being bound to.</param>
         /// <param name="functionCancellationToken">The <see cref="CancellationToken"/> to use.</param>
         /// <param name="trace">The trace writer.</param>
-        public FunctionBindingContext(Guid functionInstanceId, CancellationToken functionCancellationToken, TraceWriter trace)
+        /// <param name="functionDescriptor">Current function being executed. This will influence properties on the <see cref="SystemBindingData.MethodName"/>  </param>
```
mathewc: this comment doesn't read right

src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
```diff
@@ -22,11 +23,18 @@ public class FunctionBindingContext
...
+        public FunctionBindingContext(
+            Guid functionInstanceId,
```
mathewc: For this overload you can get the Id from the descriptor, so you don't need this separate Guid param. So:

```csharp
public FunctionBindingContext(FunctionDescriptor descriptor, TraceWriter trace, CancellationToken functionCancellationToken)
```

mathewc: Also, why is the descriptor optional?

src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
```diff
@@ -52,5 +60,10 @@ public TraceWriter Trace
         {
             get { return _trace; }
         }
+
+        /// <summary>
+        /// The short name of the current function. 
+        /// </summary>
+        public string MethodName { get; private set; }
```
mathewc: Thoughts on making the entire FunctionDescriptor accessible? Rather than hoisting various properties.

src/Microsoft.Azure.WebJobs.Host/Bindings/SystemBindingData.cs
```diff
@@ -0,0 +1,84 @@
...
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using Microsoft.Azure.WebJobs.Host.Bindings.Path;
+
+namespace Microsoft.Azure.WebJobs.Host.Bindings
+{
+    /// <summary>
+    /// Built in object for all binding expressions.
```
mathewc: nit: Class providing support for built in system binding expressions

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/FluentBindingProviderExtensions.cs
MikeStall: there's still the logging hook to maintain dashboard compat. 

---
In reply to: [117080395](https://github.com/Azure/azure-webjobs-sdk/pull/1148#discussion_r117080395) [](ancestors = 117080395)

src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
MikeStall: This is a public type ( :( )  so I was being safe by providing the existing signature. Missing a FuncDescriptor just means we're missing 'sys', which is fine in the majority case. 

---
In reply to: [117081641](https://github.com/Azure/azure-webjobs-sdk/pull/1148#discussion_r117081641) [](ancestors = 117081641)

src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
MikeStall: I'm just nervous about exposing the additional public surface area.  FuncDesc is non-public and has a poor shape. 

---
In reply to: [117082229](https://github.com/Azure/azure-webjobs-sdk/pull/1148#discussion_r117082229) [](ancestors = 117082229)

src/Microsoft.Azure.WebJobs.Host/Bindings/SysBindingData.cs
MikeStall: I'd love to do that, but it's going to be tricky because we still need to support the existing "{datetime:format}" expression and "{rand-guid}x{rand-guid}" semantics.  I didn't see how to do that without keeping the existing BuiltinResolver  classes. 


---
In reply to: [117077937](https://github.com/Azure/azure-webjobs-sdk/pull/1148#discussion_r117077937) [](ancestors = 117077937)

src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
mathewc: Well, it's already public

src/Microsoft.Azure.WebJobs.Host/Bindings/SystemBindingData.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        /// <summary>
+        /// The method name that the binding lives in. 
+        /// The method name can be override by the <see cref="FunctionNameAttribute"/> 
+        /// </summary>
+        public string MethodName { get; set; }
+
+        // Given a bindingData, extract just the sys binding data from it.
+        // This can be used when resolving default contracts that shouldn't be using an instance binding data. 
+        internal static IReadOnlyDictionary<string, object> GetSysBindingData(IReadOnlyDictionary<string, object> bindingData)
```
mathewc: nit: apply the Sys => System rename that you already started throughout

src/Microsoft.Azure.WebJobs.Host/Bindings/SystemBindingData.cs
```diff
@@ -0,0 +1,84 @@
...
+    /// It's expected this class is created and added to the binding data. 
+    /// </remarks>
+    internal class SystemBindingData
+    {
+        // The public name for this binding in the binding expressions. 
+        public const string Name = "sys";
+
+        // An internal name for this binding that uses characters that gaurantee it can't be overwritten by a user. 
+        // This ensures that we can always unambiguously retrieve this later. 
+        private const string InternalKeyName = "$sys";
```
mathewc: Not sure why you don't want to use the MS_ prefix for disambiguation, which is a precident we're already using in several places. E.g. bindings like http (request properties), in our trace event stuff (ScopeKeys), etc.

src/Microsoft.Azure.WebJobs.Host/Extensions/JobHostMetadataProvider.cs
```diff
@@ -40,6 +40,8 @@ internal void Init(IBindingProvider root)
             {
                 converter.AddAssemblies((type) => this.AddAssembly(type));
             }
+
+            AddTypesFromGraph(root as IRuleProvider);
```
mathewc: What is this change for - not related?

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/FluentBindingProviderExtensions.cs
MikeStall: Yes. Bot PR is: [https://github.com/Microsoft/BotBuilder-WebJobs-BotExtension/pull/6](https://github.com/Microsoft/BotBuilder-WebJobs-BotExtension/pull/6)

---
In reply to: [117080116](https://github.com/Azure/azure-webjobs-sdk/pull/1148#discussion_r117080116) [](ancestors = 117080116)

src/Microsoft.Azure.WebJobs.Host/Extensions/JobHostMetadataProvider.cs
MikeStall: It was for BOT - it came up when I was doing private testing for these changes with the BOT extension. 

---
In reply to: [117296439](https://github.com/Azure/azure-webjobs-sdk/pull/1148#discussion_r117296439) [](ancestors = 117296439)

src/Microsoft.Azure.WebJobs.Host/Bindings/SystemBindingData.cs
MikeStall: It is a correctness issue. I need a key that is guaranteed to not conflict with user values; so the $ character gives me that.  But a user could say "ms_prefix". 
Request property bag doesn't have this issue since all access is internal. But Binding data gets populated with public values. 
Ultimately, it shouldn't matter - this is a private name, only used in this class, not visible to the user, and not persisted. So it could be changed later (or even completely removed from the binding data if we came up with another means). 

---
In reply to: [117296062](https://github.com/Azure/azure-webjobs-sdk/pull/1148#discussion_r117296062) [](ancestors = 117296062)

src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
MikeStall: Shucks.  Looks like it went public with this change: 605f161fa1594bd61b076271287ebccb2d373f29 . I wish it was still internal.  
It's still got the wrong  shape.  It's got a FullName, ShortName, neither are which we want ... we need Method.Name.  I think exposing this class here would get confusing. 

---
In reply to: [117294645](https://github.com/Azure/azure-webjobs-sdk/pull/1148#discussion_r117294645) [](ancestors = 117294645)

MikeStall:
Merged.

Issue #881, opened 2016-10-24
Service Bus: Every new binding creates a new MessagingFactory; breaks IBinder scenarios


brettsam:
When you ask for an `IBinder` for every function invocation (which languages like node do by default in Script) and use it to create a new Service Bus binding, we end up creating a new TCP connection for each invocation. This quickly exhausts the 250 connection limit in Dynamic SKUs.

brettsam:
Note -- I've got one possible fix (not sure it's the best approach yet) with #877.

brettsam:
Looping in others for discussion: @fabiocav @mathewc

From this article: https://azure.microsoft.com/en-us/documentation/articles/service-bus-performance-improvements/. 

On re-using factories:
> Service Bus client objects, such as QueueClient or MessageSender, are created through a MessagingFactory object, which also provides internal management of connections. You should not close messaging factories or queue, topic, and subscription clients after you send a message, and then re-create them when you send the next message. Closing a messaging factory deletes the connection to the Service Bus service, and a new connection is established when recreating the factory. Establishing a connection is an expensive operation that you can avoid by re-using the same factory and client objects for multiple operations. You can safely use the QueueClient object for sending messages from concurrent asynchronous operations and multiple threads.

On not re-using factories:
> **Multiple factories:** all clients (senders in addition to receivers) that are created by the same factory share one TCP connection. The maximum message throughput is limited by the number of operations that can go through this TCP connection. The throughput that can be obtained with a single factory varies greatly with TCP round-trip times and message size. To obtain higher throughput rates, you should use multiple messaging factories.

Some possible solutions here:

**1. Cache `MessagingFactory` instances based on the SB connection string.**  (this is what I'm doing in #877)
All ServiceBus bindings that have the same connection string would share the same `MessagingFactory`.
**Pros**: Fixes the bug in a fairly contained way.
**Cons**: Multiple ServiceBus attributes in the same app would use the same underlying factory, and thus the same connection. This is a change from how things work today where every usage of the ServiceBus attribute would get its own factory. In practice, I'm not sure how much of an effect it would have on apps.

**2. Have `Binder` cache `Binding` instances internally, based on attributes**
Use something like a mapping of `Attribute[]` to `Binding` in the `Binder` implementation to cache bindings. If a function uses the exact same binding attributes over-and-over (as scripts do), you'll get the same instance.
**Pros**: Mostly addresses the con of *#*1 above. Bindings themselves don't have to know about this behavior.
[...]

PR #1150, opened 2017-05-17
ServiceBus MessageFactory caching (#881)

mathewc:
Fix based on the detailed investigations @brettsam did for https://github.com/Azure/azure-webjobs-sdk/issues/881

### DIFF
--- a/src/Microsoft.Azure.WebJobs.ServiceBus/MessagingProvider.cs
+++ b/src/Microsoft.Azure.WebJobs.ServiceBus/MessagingProvider.cs
@@ -2,8 +2,8 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Globalization;
-using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host;
 using Microsoft.ServiceBus;
 using Microsoft.ServiceBus.Messaging;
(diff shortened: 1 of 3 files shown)


test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ServiceBusEndToEndTests.cs
```diff
@@ -449,6 +473,26 @@ public class ServiceBusTestJobs : ServiceBusTestJobsBase
...
+            public static async Task ServiceBusBinderTest(
+                string message,
+                int numMessages,
+                Binder binder)
+            {
+                var attribute = new ServiceBusAttribute(BinderQueueName)
+                {
+                    EntityType = EntityType.Queue
+                };
+                var collector = await binder.BindAsync<IAsyncCollector<string>>(attribute);
```
mathewc: I verified that previously each call to BindAsync was resulting in a new MessagingFactory, and with these changes the factory for this connection is created only once

mamaso: how did you verify, manually? not seeing this in this test (though I see down below the caching test)

mathewc: Yeah, this test doesn't verify caching, but exercises the full end to end. I don't think we had Binder tests for SB. It also allowed me to debug through it all.

src/Microsoft.Azure.WebJobs.ServiceBus/MessagingProvider.cs
```diff
@@ -17,6 +18,7 @@ namespace Microsoft.Azure.WebJobs.ServiceBus
     public class MessagingProvider
     {
         private readonly ServiceBusConfiguration _config;
+        private readonly ConcurrentDictionary<string, MessagingFactory> _messagingFactoryCache = new ConcurrentDictionary<string, MessagingFactory>();
```
MikeStall: >string [](start = 46, length = 6)

Can you add a comment for what this string is?

MikeStall:
:shipit:

src/Microsoft.Azure.WebJobs.ServiceBus/MessagingProvider.cs
mathewc: Done

PR #1152, opened 2017-05-18
Add UtcNow, RandGuid support to SystemBindingData

MikeStall:


src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingParameterResolver.cs
```diff
@@ -107,16 +108,8 @@ public override string Name
...
-                {
-                    return DateTime.UtcNow.ToString(format, CultureInfo.InvariantCulture);
-                }
-                else
-                {
-                    // default to ISO 8601
-                    return DateTime.UtcNow.ToString("yyyy-MM-ddTHH-mm-ssK", CultureInfo.InvariantCulture);
-                }
+                var val = new SystemBindingData().UtcNow;
+                return BindingDataPathHelper.ConvertParameterValueToString(val, format);
```
fabiocav: Although I like the feature and the goal, I'm not a fan of moving the logic out of the specialized resolver types and having a single type (in this case, the `SystemBindingData`) that is now responsible for knowing how to format a bunch of different values/types.

With this approach, we're also allocating more, causing additional casts and unnecessary boxing.

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingParameterResolver.cs
```diff
@@ -8,6 +8,10 @@
 
 namespace Microsoft.Azure.WebJobs.Host.Bindings.Path
 {
+    /// <summary>
+    /// Support for adding built-in values to binding data. 
+    /// Don't add new resolvers here. Instead, add it to <see cref="SystemBindingData"/>
+    /// </summary>
     internal abstract class BindingParameterResolver
```
fabiocav: Can we flag this type as obsolete?

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingDataPathHelper.cs
```diff
@@ -45,10 +50,21 @@ public static string ConvertParameterValueToString(object parameterValue)
...
                         return ((SByte)parameterValue).ToString(CultureInfo.InvariantCulture);
+                    case TypeCode.DateTime:                        
+                        format = format ?? "yyyy-MM-ddTHH-mm-ssK"; // default to ISO 8601
+                        var dateTime = (DateTime)parameterValue;
+                        return dateTime.ToString(format, CultureInfo.InvariantCulture);
                     case TypeCode.Object:
                         if (parameterValue is Guid)
                         {
-                            return parameterValue.ToString();
+                            if (format == null)
```
fabiocav: Also, as we discussed, out of scope for this PR, but it would probably be good to have an issue tracking work to refactor this so this type is not handling all of this logic.

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingParameterResolver.cs
MikeStall: great idea

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingParameterResolver.cs
MikeStall: Discussed in person:
- we already have a centralized formatter.  This is just adding DateTime/Guid to it. This is actually a good thing since it enforces consistency 
- SystemBindingData does not do any formatting. It just exposes the builtin contract in one place.

MikeStall:
merged

mathewc:
I just sent a follow up PR with a few tweaks https://github.com/Azure/azure-webjobs-sdk/pull/1157

PR #1154, opened 2017-05-22
Changing MessageFactory caching to be per entity/connection

mathewc:
In PR https://github.com/Azure/azure-webjobs-sdk/pull/1150 I added caching per connection string. I'd like to make that more granular, so we're caching per entity/connection. Note that this more in line with the behavior before caching was added. Previously we'd create a new factory at index time per function. This makes the caching more granular, addressing some concerns Brett raised previously about overloading a single TCP connection.


brettsam:
This looks like a nice compromise to me. This closely maps to how things work currently when you don't use IBinder, right? (i.e. you get a single MessageFactory per use of the ServiceBusAttribute)

mathewc:
Merged

PR #1157, opened 2017-05-23
System binding data tweaks

mathewc:
A few proposed SystemBindingData renames. These renames change end user expressions to:

- `{sys.guid}`
- `{sys.datetime}`

This makes things consistent with the existing legacy built-ins supported:

- `{datetime}`
- `{guid}`  ** this one is new ***
- `{rand-guid}`

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingParameterResolver.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingParameterResolver.cs
@@ -21,6 +21,7 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings.Path
         {
             // create the static set of built in system resolvers
             _resolvers = new Collection<BindingParameterResolver>();
+            _resolvers.Add(new GuidResolver());
             _resolvers.Add(new RandGuidResolver());
             _resolvers.Add(new DateTimeResolver());
         }
@@ -74,8 +75,26 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings.Path
             return null;
         }
 
-        // This is an alias for 'sys.randguid'
-        private class RandGuidResolver : BindingParameterResolver
+        // This is an alias for 'sys.guid'
+        private class GuidResolver : BindingParameterResolver
+        {
+            public override string Name
+            {
+                get
+                {
+                    return "guid";
+                }
+            }
+
+            public override string Resolve(string value)
+            {
+                string format = GetFormatOrNull(value);
+                var val = new SystemBindingData().Guid;
+                return BindingDataPathHelper.ConvertParameterValueToString(val, format);                
+            }
+        }
+
+        private class RandGuidResolver : GuidResolver
         {
             public override string Name
             {
@@ -84,18 +103,8 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings.Path
                     return "rand-guid";
                 }
             }
-
-            public override string Resolve(string value)
-            {
-                string format = GetFormatOrNull(value);
-                var val = new SystemBindingData().RandGuid;
-                return BindingDataPathHelper.ConvertParameterValueToString(val, format);                
-            }
         }
 
-        // This can't be aliases to 'sys.UtcNow' because
-        // 'sys.UtcNow' always resolves to DateTime.UtcNow. 
-        // But 'datetime' may either resolve to user bidning data or to DateTime.UtcNow.        
         private class DateTimeResolver : BindingParameterResolver
         {
             public override string Name
@@ -109,7 +118,7 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings.Path
             public override string Resolve(string value)
             {
                 string format = GetFormatOrNull(value);
-                var val = new SystemBindingData().UtcNow;
+                var val = new SystemBindingData().DateTime;
                 return BindingDataPathHelper.ConvertParameterValueToString(val, format);
             }
         }
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/SystemBindingData.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/SystemBindingData.cs
@@ -19,7 +19,7 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
         // The public name for this binding in the binding expressions. 
         public const string Name = "sys";
 
-        // An internal name for this binding that uses characters that gaurantee it can't be overwritten by a user. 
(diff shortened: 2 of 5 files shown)


src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingParameterResolver.cs
```diff
@@ -74,28 +75,36 @@ protected string GetFormatOrNull(string value)
             return null;
         }
 
-        // This is an alias for 'sys.randguid'
-        private class RandGuidResolver : BindingParameterResolver
+        // This is an alias for 'sys.guid'
+        private class GuidResolver : BindingParameterResolver
```
mamaso: what's the difference between guid and randguid?

mathewc: Same thing. I regret the name "rand-guid" so I'm aliasing it

mamaso: Oh great, sounds good :)

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingParameterResolver.cs
```diff
@@ -74,28 +75,36 @@ protected string GetFormatOrNull(string value)
...
                 return BindingDataPathHelper.ConvertParameterValueToString(val, format);                
             }
         }
 
-        // This can't be aliases to 'sys.UtcNow' because
-        // 'sys.UtcNow' always resolves to DateTime.UtcNow. 
-        // But 'datetime' may either resolve to user bidning data or to DateTime.UtcNow.        
+        private class RandGuidResolver : GuidResolver
+        {
+            public override string Name
```
mamaso: nit: `public override string Name => "rand-guid"`

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingParameterResolver.cs
```diff
@@ -74,28 +75,36 @@ protected string GetFormatOrNull(string value)
             return null;
         }
 
-        // This is an alias for 'sys.randguid'
-        private class RandGuidResolver : BindingParameterResolver
+        // This is an alias for 'sys.guid'
+        private class GuidResolver : BindingParameterResolver
         {
             public override string Name
```
mamaso: nit: `=> "guid"`

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingParameterResolver.cs
MikeStall: >GuidResolver [](start = 22, length = 12)

I'd prefer we deprecate (ie, not expand) this entire path, not add any new BindingParameterResolver, and instead push people down  {sys} instead.

src/Microsoft.Azure.WebJobs.Host/Bindings/SystemBindingData.cs
```diff
@@ -36,14 +36,14 @@ internal class SystemBindingData
...
-        /// Get the current UTC date. 
+        /// Get the current UTC date. This returns the current value each time it is called.
         /// </summary>
-        public DateTime UtcNow => DateTime.UtcNow;
+        public DateTime DateTime => DateTime.UtcNow;
 
         /// <summary>
         /// Return a new random guid. This create a new guid each time it's called.
         /// </summary>
-        public Guid RandGuid => Guid.NewGuid();
```
MikeStall: >RandGuid [](start = 20, length = 8)

Ok. Although, I kind of prefer having 'RandGuid' since it's clearer about the source, and less ambiguous about which guid we're referring to.

src/Microsoft.Azure.WebJobs.Host/Bindings/SystemBindingData.cs
```diff
@@ -36,14 +36,14 @@ internal class SystemBindingData
         public string MethodName { get; set; }
 
         /// <summary>
-        /// Get the current UTC date. 
+        /// Get the current UTC date. This returns the current value each time it is called.
         /// </summary>
-        public DateTime UtcNow => DateTime.UtcNow;
+        public DateTime DateTime => DateTime.UtcNow;
```
MikeStall: >DateTime [](start = 24, length = 8)

We really should include the timezone, just like System.DateTime.UtcNow does. Avoids the confusion and accidentally mistaking this for local time. .

MikeStall:
:clock1:

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingParameterResolver.cs
mathewc: Well, we could say that you can't use "guid" but I think that would be inconsistent. Definitely going forward we won't add any new ones here. This instance is just special because I'm adding a new alias for an existing feature.

src/Microsoft.Azure.WebJobs.Host/Bindings/SystemBindingData.cs
mathewc: We could leave it RandGuid. I just always thought "rand" was strange - it's a new guid so its randomness is implicit. Also there is a symmetry with "DateTime" if we just use "Guid" - it's the type name. I could go either way on this.

src/Microsoft.Azure.WebJobs.Host/Bindings/SystemBindingData.cs
mathewc: I'm trying to keep these consistent with the existing built-ins. That's what I have now in this PR - the same identifiers work in both paths. I think it's fine that the fact that the DateTime is UTC is just implicit - that's what we have now for {datetime}.

src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingParameterResolver.cs
MikeStall: But we want to deprecate this altogether!   I never want somebody to write {guid}; only {sys.guid} . This is adding a codepath that we never want somebody to actually take. 



---
In reply to: [118123998](https://github.com/Azure/azure-webjobs-sdk/pull/1157#discussion_r118123998) [](ancestors = 118123998)

mathewc:
Chatted offline - we're going to keep sys.UtcNow and sys.RandGuid. We'll live with the mismatch between the old name {datetime}.

PR #1161, opened 2017-05-25
Removing Unused Performance Project

hhawkins:


WebJobs.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26403.7
+VisualStudioVersion = 15.0.26430.6
```
mathewc: You need to update your VS?

mathewc: My bad - you're actually on a later version. I'll update my VS :)

PR #1164, opened 2017-05-31
Improved AppSetting error

### LINKED ISSUE #1156: Improve error message for AppSetting resolution failure
Customers often add incorrect values for [AppSetting] attributed binding values. This happens often for connection/secret values. In such cases, even though we know we're trying to resolve an app setting, we don't give an error message that indicates that. In this example the bot secret value is given when an app setting is required:

```javascript
{
  "bindings": [
    {
      "name": "myQueueItem",
      "type": "queueTrigger",
      "direction": "in",
      "queueName": "myqueue",
      "connection": "my_STORAGE"
    },
    {
      "type": "bot",
      "name": "$return",
      "botId": "myBot",
      "secret": "jdiuidk984987dfkjdf.jjfdo7eijkkajdCldi88736",
      "direction": "out"
    }
  ]
}
```

This results in error message:

```
Function ($bot-reminder-function) Error: Microsoft.Azure.WebJobs.Host: Error indexing method 'Functions.bot-reminder-function'.
Microsoft.Azure.WebJobs.Host: Unable to resolve value for property 'BotAttribute.SecretSetting'.
```

For values that are attributed with [AppSetting] we should improve this to something like "Unable to resolve app setting for property 'BotAttribute.SecretSetting'. Make sure the app setting exists and has a valid value."

mamaso:
resolves #1156

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
@@ -156,7 +156,7 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
                 // It's important that we only log the attribute property name, not the actual value to ensure
                 // that in cases where users accidentally use a secret key *value* rather than indirect setting name
                 // that value doesn't get written to logs.
-                throw new InvalidOperationException($"Unable to resolve value for property '{propInfo.DeclaringType.Name}.{propInfo.Name}'.");
+                throw new InvalidOperationException($"Unable to resolve app setting for property '{propInfo.DeclaringType.Name}.{propInfo.Name}'. Make sure the app setting exists and has a valid value.");
             }
             return (newAttr, bindingData) => resolvedValue;
         }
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -156,7 +156,7 @@ internal static BindingDataResolver GetAppSettingResolver(string originalValue,
                 // It's important that we only log the attribute property name, not the actual value to ensure
                 // that in cases where users accidentally use a secret key *value* rather than indirect setting name
                 // that value doesn't get written to logs.
-                throw new InvalidOperationException($"Unable to resolve value for property '{propInfo.DeclaringType.Name}.{propInfo.Name}'.");
+                throw new InvalidOperationException($"Unable to resolve app setting for property '{propInfo.DeclaringType.Name}.{propInfo.Name}'. Make sure the app setting exists and has a valid value.");
```
brettsam: Are we able to log out the actual string that we're looking for? Or would that show up in our backend logs?

brettsam: Can't this be an environment variable as well as an app setting?

mamaso: Yep, we don't log the setting value because it shows up in backend logs.

mamaso: It can - the error is written in the context of web apps where these are essentially all Application Settings. I'm happy to improve the message if you have a suggestion.

PR #1165, opened 2017-05-31
[AppInsights] moving filtering further down the pipeline so QuickPulse always works

### LINKED ISSUE #1162: Application Insights should log Live Metrics, even if LogLevel is None
Live Metrics still give you a nice view into how your site is running -- I need to move the filtering to happen after it passes through the QuickPulse part of the pipeline.

Also should probably add a knob to allow people to turn off Live Metrics if they need to (for some reason).

brettsam:
Resolves #1162


brettsam:
We decided not to add a knob yet -- we can always do that in the future. Ultimately, @paulbatum and I decided that the people that wanted to enable App Insights yet disable Live Metrics (which is free) was so small that it wasn't worth it.

Filtering was breaking it b/c the Logger's job was previously to filter as the first thing it did, which meant that no Telemetry items made it to the App Insights pipeline. We have the pipeline configured with a series of Processors: QuickPulse -> AdaptiveStreaming -> Channel. So rather than filtering before it even gets to QuickPulse, I now filter after that with a new Processor that knows how to look at Category/LogLevel, which we store in the Telemetry item's properties. So now it goes QuickPulse -> Filter -> AdaptiveStreaming -> Channel

PR #1166, opened 2017-05-31
Improve doccomments for "connection" property.

lindydonna:
Customers often try to put in a connection string for the "connection" property and then get a poor error message. Maybe a better doccomment will help.

I removed use of "inheritdoc" since Visual Studio does not honor it when generating intellisense files. I found an old UserVoice suggestion from 2013 asking for this feature, so I'm going to assume it's not happening anytime soon: [Add IntelliSense support for the <inheritdoc /> tag in XML documentation comments](https://visualstudio.uservoice.com/forums/121579-visual-studio-ide/suggestions/3745102-add-intellisense-support-for-the-inheritdoc-ta).

### DIFF
--- a/src/Microsoft.Azure.WebJobs.ServiceBus/ServiceBusAccountAttribute.cs
+++ b/src/Microsoft.Azure.WebJobs.ServiceBus/ServiceBusAccountAttribute.cs
@@ -33,11 +33,13 @@ namespace Microsoft.Azure.WebJobs
         }
 
         /// <summary>
-        /// Gets the name of the ServiceBus connection string to use.
+        /// Gets or sets the name of the app setting that contains the Service Bus connection string.
         /// </summary>
         public string Account { get; private set; }
 
-        /// <inheritdoc />
+        /// <summary>
+        /// Gets or sets the app setting name that contains the Service Bus connection string.
+        /// </summary>
         string IConnectionProvider.Connection
         {
             get
--- a/src/Microsoft.Azure.WebJobs.ServiceBus/ServiceBusAttribute.cs
+++ b/src/Microsoft.Azure.WebJobs.ServiceBus/ServiceBusAttribute.cs
@@ -57,7 +57,9 @@ namespace Microsoft.Azure.WebJobs
         /// </summary>
         public string QueueOrTopicName { get; private set; }
 
-        /// <inheritdoc />
+        /// <summary>
+        /// Gets or sets the app setting name that contains the Service Bus connection string.
+        /// </summary>
         public string Connection { get; set; }
 
         /// <summary>
--- a/src/Microsoft.Azure.WebJobs.ServiceBus/ServiceBusTriggerAttribute.cs
+++ b/src/Microsoft.Azure.WebJobs.ServiceBus/ServiceBusTriggerAttribute.cs
@@ -77,7 +77,9 @@ namespace Microsoft.Azure.WebJobs
             Access = access;
         }
 
-        /// <inheritdoc />
+        /// <summary>
+        /// Gets or sets the app setting name that contains the Service Bus connection string.
+        /// </summary>
         public string Connection { get; set; }
 
         /// <summary>
--- a/src/Microsoft.Azure.WebJobs/BlobAttribute.cs
+++ b/src/Microsoft.Azure.WebJobs/BlobAttribute.cs
@@ -90,7 +90,9 @@ namespace Microsoft.Azure.WebJobs
             get { return _access; }
         }
 
-        /// <inheritdoc />
+        /// <summary>
+        /// Gets or sets the app setting name that contains the Azure Storage connection string.
+        /// </summary>
         public string Connection { get; set; }
     }
 }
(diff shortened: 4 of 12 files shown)


src/Microsoft.Azure.WebJobs.ServiceBus/ServiceBusAccountAttribute.cs
```diff
@@ -33,11 +33,13 @@ public ServiceBusAccountAttribute(string account)
         }
 
         /// <summary>
-        /// Gets the name of the ServiceBus connection string to use.
+        /// The name of the app setting containing the Service Bus connection string.
```
fabiocav: This would fail in script but we don't have the rule (SA1623) enabled here to help. For properties, we should maintain the gets/sets convention.

src/Microsoft.Azure.WebJobs/BlobAttribute.cs
```diff
@@ -90,7 +90,9 @@ public string BlobPath
             get { return _access; }
         }
 
-        /// <inheritdoc />
+        /// <summary>
+        /// The name of the app setting containing the Azure Storage connection string.
+        /// </summary>
         public string Connection { get; set; }
```
fabiocav: Gets or sets

src/Microsoft.Azure.WebJobs.ServiceBus/ServiceBusTriggerAttribute.cs
```diff
@@ -77,7 +77,9 @@ public ServiceBusTriggerAttribute(string topicName, string subscriptionName, Acc
             Access = access;
         }
 
-        /// <inheritdoc />
+        /// <summary>
+        /// The name of the app setting containing the Service Bus connection string.
+        /// </summary>
         public string Connection { get; set; }
```
fabiocav: Gets or sets

src/Microsoft.Azure.WebJobs.ServiceBus/ServiceBusAccountAttribute.cs
```diff
@@ -33,11 +33,13 @@ public ServiceBusAccountAttribute(string account)
...
         /// </summary>
         public string Account { get; private set; }
 
-        /// <inheritdoc />
+        /// <summary>
+        /// The name of the app setting containing the Service Bus connection string.
+        /// </summary>
```
fabiocav: Gets or sets

src/Microsoft.Azure.WebJobs.ServiceBus/ServiceBusAttribute.cs
```diff
@@ -57,7 +57,9 @@ public ServiceBusAttribute(string queueOrTopicName, AccessRights access)
         /// </summary>
         public string QueueOrTopicName { get; private set; }
 
-        /// <inheritdoc />
+        /// <summary>
+        /// The name of the app setting containing the Service Bus connection string.
+        /// </summary>
         public string Connection { get; set; }
```
fabiocav: Gets or sets

src/Microsoft.Azure.WebJobs/BlobTriggerAttribute.cs
```diff
@@ -46,7 +46,9 @@ public BlobTriggerAttribute(string blobPath)
             _blobPath = blobPath;
         }
 
-        /// <inheritdoc />
+        /// <summary>
+        /// The name of the app setting containing the Azure Storage connection string.
```
fabiocav: Same and with the other properties as appropriate :)

src/Microsoft.Azure.WebJobs/StorageAccountAttribute.cs
```diff
@@ -33,11 +33,13 @@ public StorageAccountAttribute(string account)
         }
 
         /// <summary>
-        /// Gets the name of the Azure Storage connection string to use.
+        /// The name of the app setting containing the Azure Storage connection string.
```
fabiocav: Gets

mathewc:
This PR is only addressing this for EventHub attributes. But we have lots of other attributes that implement `IConnectionProvider` and are also currently using `</inheritdoc>` for this Connection property. I think we should be consistent across all of the attributes.

That will mean that for 10+ attributes we'd need to duplicate all this text. I don't really think it's worth it.

lindydonna:
@mathewc Fair point, but I doubt we'll need to maintain these comments (i.e., functionality is unlikely to change). I'll address the review feedback.

lindydonna:
@fabiocav I addressed your feedback, could you sign off and merge?

lindydonna:
@fabiocav Ping.

fabiocav:
Sorry, started a reply on this and left it unfinished. There are two pending comments. I'm happy to go ahead and merge this, but they're easy changes, so if you want to do them now, we can wait. Just let me know and we can merge when ready.

lindydonna:
@fabiocav Oops, looks like I missed a couple in my last edit pass. Should all be addressed now.

lindydonna:
@mathewc Can you review and merge?

mathewc:
Merged, thanks.

Issue #1140, opened 2017-05-10
Support for controlling task awaiting 


MikeStall:
Some extensions, like durable tasks, need better control of the await semantics. 

Our current pipeline is like:
Pre() 
await UserFunction()
Post() // does cleanup logging, etc 

And we need to allow 
Pre()
Task t = UserFunction()
Post() 
// do some checking on T. 

This would be consumed by the extensions trigger binder. This is likely on ITriggeredFunctionExecutor.

MikeStall:
Mathew has a PR that supersedes 1167

mathewc:
Addressed by https://github.com/Azure/azure-webjobs-sdk/pull/1178

PR #1167, opened 2017-05-31
Provide a hook for invoke

MikeStall:
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1140

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/ITriggeredFunctionExecutorWithHook.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/ITriggeredFunctionExecutorWithHook.cs
@@ -1,0 +1,26 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Host.Executors
+{
+    /// <summary>
+    /// Interface defining the contract for executing a triggered function.
+    /// Allows a hook around the underlying execution.
+    /// This should only be used by extensions that need very specific control over the invocation. 
+    /// </summary>
+    public interface ITriggeredFunctionExecutorWithHook
+    {
+        /// <summary>
+        /// Try to invoke the triggered function using the values specified.
+        /// </summary>
+        /// <param name="input">The trigger invocation details.</param>
+        /// <param name="cancellationToken">The cancellation token</param>
+        /// <param name="hook">a hook that wraps the underlying invocation</param>
+        /// <returns>A <see cref="FunctionResult"/> describing the results of the invocation.</returns>
+        Task<FunctionResult> TryExecuteAsync(TriggeredFunctionData input, CancellationToken cancellationToken, Func<Func<Task>, Task> hook);
+    }
+}
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionExecutor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionExecutor.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host.Protocols;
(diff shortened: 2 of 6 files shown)


src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionExecutor.cs
```diff
@@ -30,16 +31,46 @@ public FunctionDescriptor Function
...
+            return TryExecuteAsync(input, cancellationToken, null);
+        }
+
+        public async Task<FunctionResult> TryExecuteAsync(TriggeredFunctionData input, CancellationToken cancellationToken, Func<Func<Task>, Task> hook)
         {
             IFunctionInstance instance = _instanceFactory.Create((TTriggerValue)input.TriggerValue, input.ParentId);
+            if (hook != null)
+            {
+                IFunctionInvoker invoker = new InvokeWrapper(instance.Invoker, hook);
+                instance = new FunctionInstance(instance.Id, instance.ParentId, instance.Reason, instance.BindingSource, invoker, instance.FunctionDescriptor);
```
mathewc: This is bypassing the instance factory. I don't think we should do that.

src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionExecutor.cs
```diff
@@ -2,14 +2,15 @@
...
+using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host.Protocols;
 using Microsoft.Azure.WebJobs.Host.Triggers;
 
 namespace Microsoft.Azure.WebJobs.Host.Executors
 {
-    internal class TriggeredFunctionExecutor<TTriggerValue> : ITriggeredFunctionExecutor
+    internal class TriggeredFunctionExecutor<TTriggerValue> : ITriggeredFunctionExecutor, ITriggeredFunctionExecutorWithHook
```
mathewc: I think we should find a more general solution to this, rather than hack in this hook. Let's give it some thought.

src/Microsoft.Azure.WebJobs.Host/Executors/ITriggeredFunctionExecutorWithHook.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// This should only be used by extensions that need very specific control over the invocation. 
+    /// </summary>
+    public interface ITriggeredFunctionExecutorWithHook
+    {
+        /// <summary>
+        /// Try to invoke the triggered function using the values specified.
+        /// </summary>
+        /// <param name="input">The trigger invocation details.</param>
+        /// <param name="cancellationToken">The cancellation token</param>
+        /// <param name="hook">a hook that wraps the underlying invocation</param>
```
mathewc: How does a custom trigger specify this hook?

MikeStall: The binding provides a listener which is in full control of making calls on ITriggerFunctionExecutor*.  So it's imperative calls; not registered.

src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionExecutor.cs
MikeStall: Why not? This doesn't need any factory services. What did you have in mind? An alternative approaches could be to make Invoker a mutable property  and just change it on the existing instance.

src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionExecutor.cs
MikeStall: I'm very open to alternatives. The criteria for me was the simplest solution that 
a) fulfilled some strict scenario requirements (it needs to be dispatched exactly at this spot in the pipeline); 
b) was relatively isolated and had a clear 'opt-in' model.  
This is a rocket-science case and I specifically didn't want it to get broad traction except from things that really needed it.  I didn't want to create a matrix of feature interactions.

mathewc: One think we might consider would be to extend the existing `TriggeredFunctionData` parameter by adding your callback on there. Easy for consumer to specify the callback, and below you just check for it and use it.

src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionExecutor.cs
mathewc: If we have a factory pattern, then all creation should go through it. You might sidestep this problem by having the factory interface take the TriggeredFunctionData<T> so it has access to the value as well as the callback. The interface is internal.

src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionExecutor.cs
MikeStall: I considered that, but it would violate #b - this is a very powerful hook and I _don't_ want to make it easy or tempting to get at it.

mathewc: You're adding a new public `ITriggeredFunctionExecutorWithHook` interface that we need to maintain going forward. It feels pretty hacky. If we really need to add this capability then I think we should add it in a clean way and commit to it.

MikeStall:
Closing, we'll go with https://github.com/Azure/azure-webjobs-sdk/pull/1178 instead

PR #1178, opened 2017-06-06
Invocation interception

mathewc:
Alternate proposal to https://github.com/Azure/azure-webjobs-sdk/pull/1167. In this case, a handler can be specified as part of the trigger data.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInstanceFactory.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInstanceFactory.cs
@@ -21,10 +21,10 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
             _descriptor = descriptor;
         }
 
-        public IFunctionInstance Create(Guid id, Guid? parentId, ExecutionReason reason, IDictionary<string, object> parameters)
+        public IFunctionInstance Create(FunctionInstanceFactoryContext context)
         {
-            IBindingSource bindingSource = new BindingSource(_binding, parameters);
-            return new FunctionInstance(id, parentId, reason, bindingSource, _invoker, _descriptor);
+            IBindingSource bindingSource = new BindingSource(_binding, context.Parameters);
+            return new FunctionInstance(context.Id, context.ParentId, context.ExecutionReason, bindingSource, _invoker, _descriptor);
         }
     }
 }
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionData.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionData.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Threading.Tasks;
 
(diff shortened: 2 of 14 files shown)


src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInstanceFactory.cs
```diff
@@ -21,10 +20,10 @@ public FunctionInstanceFactory(IFunctionBinding binding, IFunctionInvoker invoke
             _descriptor = descriptor;
         }
 
-        public IFunctionInstance Create(Guid id, Guid? parentId, ExecutionReason reason, IDictionary<string, object> parameters)
+        public IFunctionInstance Create(FunctionInstanceFactoryContext context)
```
mathewc: The main change is just an internal refactoring to use a context object here, which is a good change to make anyways

src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionData.cs
```diff
@@ -19,5 +20,10 @@ public class TriggeredFunctionData
         /// The trigger value for a specific triggered function invocation.
         /// </summary>
         public object TriggerValue { get; set; }
+
+        /// <summary>
+        /// Optional handler function for processing the invocation.
+        /// </summary>
+        public Func<Func<Task>, Task> InvokeHandler { get; set; }
```
mathewc: This is the key - the listener when it calls the executor can specify this handler

cgillum:
From the trigger author's perspective, it seems this would be very easy to use.  It solves my problem, so I'm all for it. :)

src/Microsoft.Azure.WebJobs.Host/Executors/IFunctionInvoker.cs
```diff
@@ -1,16 +1,23 @@
...
 using System.Threading.Tasks;
 
-namespace Microsoft.Azure.WebJobs.Host.Executors
+namespace Microsoft.Azure.WebJobs.Host
 {
-    internal interface IFunctionInvoker
+    /// <summary>
+    /// Provides an interface for invoking a job function.
+    /// </summary>
+    public interface IFunctionInvoker
```
mathewc: Making this public. We'll also be able to replace https://github.com/Azure/azure-webjobs-sdk-script/blob/1ab54a94916c8180105af1e43b24a483f070a53f/src/WebJobs.Script/Description/IFunctionInvoker.cs#L9 with this.

src/Microsoft.Azure.WebJobs.Host/Executors/IFunctionInvokerProvider.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// Defines an interface for the creation of <see cref="IFunctionInvoker"/> instances.
+    /// </summary>
+    public interface IFunctionInvokerProvider
```
mathewc: A custom binding can implement this interface to provide custom invokers.

PR #1181, opened 2017-06-07
removing c# 7 usage to fix build

brettsam:


fabiocav:
Should we update the project language version target to avoid unintended use?

brettsam:
Hopefully this is a very short-term problem -- maybe if we don't have a solution for Jenkins by EOD today, yeah.

fabiocav:
I didn't think we were actually building there, I thought it was just taking the build output and signing them. @mamaso can you confirm?

mamaso:
It does a full build. The build artifacts rely on getting intermediate signed assemblies to pack up the site extensions, etc, so it was tightly coupled.

fabiocav:
That's unfortunate. What's the long term plan? Are we moving away or upgrading that environment? I'm worried about .NET Core builds (which we'll need soon!). To we need to deploy our own agent for more flexibility?

mamaso:
I have access to the agent, so I've been upgrading as necessary. The VS 2017 install fails silently for some reason (nothing in eventvwr either). I'll look into it more when I get a bit more time.

Still, Di mentioned that the long term plan is to decommission zenkins (in a year or so), so eventually we'll need to move.

If we are going to be doing a significant amount of infrastructure work anyway, perhaps it would be a good time to evaluate the best long term solution for our build infra.

mamaso:
Changing the webjobs build so that it does only the signing on zenkins would be another option - workflow matches well with what the vs tooling bits need as well.

If all we do is signed prebuilt bits it should be easy to shift wherever we need. However, as I said it's currently tightly coupled in the webjobs build.

fabiocav:
Need to chat about this as we'll need changes for core (soon) and would be nice to have a plan way ahead of getting zenkins pulled from under us.

Issue #1183, opened 2017-06-07
Allow [Singleton] to run without a storage dependency 


MikeStall:
Make SingletonManager pluggable so that we can run the SDK without any storage dependency.

PR #1184, opened 2017-06-07
Decoupled SDK from Storage

MikeStall:
Decouple SDK from hard storage dependency. This is just Singleton manager.  Resolve https://github.com/Azure/azure-webjobs-sdk/issues/1183 

This is a replacement for https://github.com/Azure/azure-webjobs-sdk/pull/1113

Create a new IDistributedLockManager interface that SingletonManager layers on top of it. 
Create a default implementation based on the existing blob-lease code today. 

See new Singleton tests for how this could be swapped out.


src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLock.cs
```diff
@@ -0,0 +1,24 @@
...
+        /// <summary>
+        /// The Lock identity.  
+        /// </summary>
+        string LockId { get; set; }
+
+        /// <summary>
+        /// Task is set to completed if the lease is lost (after it's been acquired). 
+        /// This allows consumers to get a notification if the lock failed to be renewed. 
+        /// </summary>
+        Task LeaseLost { get; }
```
mathewc: What is the benefit of using a Task here v.s. a regular event. For example a LockStatusChanged event with an event args of type DistributedLockStatusChangedEventArgs with a DistributedLock status enum, etc.

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLockManager.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// 1. Account can be null or it may be a storage account name intended for <see cref="IStorageAccountProvider"/>. 
+    /// 2. lockId has the same naming restrictions as blobs. 
+    /// 3. For distributed locks, this implementation is responsible for renewing the locks and setting <see cref="IDistributedLock.LeaseLost"/> if the lock can't be renewed. 
+    /// </remarks>    
+    public interface IDistributedLockManager
+    {
+        /// <summary>
+        /// Try to acquire a lock specified by (account, lockId).                 
+        /// </summary>
+        /// <param name="account">a string specifying the account to use. LockIds are scoped to an account </param>
```
mathewc: It's not clear at all what an account is or what the string should contain.

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLockManager.cs
```diff
@@ -0,0 +1,60 @@
...
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// Manage distributed locks. A lock is specified by (account, lockId). 
+    /// </summary>
+    /// <remarks>
+    /// The default implementation of this is based on blob leases. 
+    /// 1. Account can be null or it may be a storage account name intended for <see cref="IStorageAccountProvider"/>. 
+    /// 2. lockId has the same naming restrictions as blobs. 
+    /// 3. For distributed locks, this implementation is responsible for renewing the locks and setting <see cref="IDistributedLock.LeaseLost"/> if the lock can't be renewed.
```
mathewc: I think we should separate renewal behavior out into its own interface/implementation since they're to independent concepts.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Singleton/SingletonEnd2EndTests.cs
```diff
@@ -0,0 +1,95 @@
...
+using Microsoft.Azure.WebJobs.Host.TestCommon;
+using System;
+using System.Threading;
+using Xunit;
+using System.Threading.Tasks;
+using System.Collections.Generic;
+
+namespace Microsoft.Azure.WebJobs.Host.UnitTests.Singleton
+{
+    public class SingletonEnd2EndTests
```
mathewc: There is already a SingletonEndToEndTests test suite. Let's add any new tests to that one

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLock.cs
MikeStall: One benefit is that it avoids races around completion.  With an event, you need to take an extra step to register; but if the event fires before you register, you miss it. (This is a lesson from JScript Promises)  

---
In reply to: [120950317](https://github.com/Azure/azure-webjobs-sdk/pull/1184#discussion_r120950317) [](ancestors = 120950317)

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLockManager.cs
MikeStall: see class-level comments.  The challenge is that [Singleton] lets you specify a storage account; so that's liability that needs to flow down here. 

---
In reply to: [120950837](https://github.com/Azure/azure-webjobs-sdk/pull/1184#discussion_r120950837) [](ancestors = 120950837)

test/Microsoft.Azure.WebJobs.Host.UnitTests/Singleton/SingletonEnd2EndTests.cs
MikeStall: But that uses external storage and is in the FunctionTests project. This has no dependency so can rightfully live in the UnitTest project.  but has the added benefit that it exercises the entire end-2-end [Singleton] machinery. 

---
In reply to: [120952102](https://github.com/Azure/azure-webjobs-sdk/pull/1184#discussion_r120952102) [](ancestors = 120952102)

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLockManager.cs
fabiocav: Agreed. Having this split would give us a lower level abstraction, which I think we need, and the additional functionality split up into another type, leading to simpler, more specialized implementations.

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLockManager.cs
```diff
@@ -0,0 +1,60 @@
...
+    {
+        /// <summary>
+        /// Try to acquire a lock specified by (account, lockId).                 
+        /// </summary>
+        /// <param name="account">a string specifying the account to use. LockIds are scoped to an account </param>
+        /// <param name="lockId">the name of the lock. </param>
+        /// <param name="lockOwnerId">a string hint specifying who owns this lock. </param>
+        /// <param name="lockPeriod">As this period nears expiry, the lock will be automatically renewed.</param>
+        /// <param name="cancellationToken"></param>
+        /// <returns>null if can't acquire the lock. This is common if somebody else holds it.</returns>
```
MikeStall: We can't combine Task and out like that. 

---
In reply to: [120949296](https://github.com/Azure/azure-webjobs-sdk/pull/1184#discussion_r120949296) [](ancestors = 120949296)

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLockManager.cs
```diff
@@ -0,0 +1,60 @@
...
+            CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Get the owner for a given lock or null if not held. 
+        /// </summary>
+        /// <param name="account"></param>
+        /// <param name="lockId"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        Task<string> GetLockOwnerAsync(
```
MikeStall: SingletonManager needed it, and that was my starting point.  I'll see if I can get rid of it.   

---
In reply to: [120945165](https://github.com/Azure/azure-webjobs-sdk/pull/1184#discussion_r120945165) [](ancestors = 120945165)

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLockManager.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// Try to acquire a lock specified by (account, lockId).                 
+        /// </summary>
+        /// <param name="account">a string specifying the account to use. LockIds are scoped to an account </param>
+        /// <param name="lockId">the name of the lock. </param>
+        /// <param name="lockOwnerId">a string hint specifying who owns this lock. </param>
+        /// <param name="lockPeriod">As this period nears expiry, the lock will be automatically renewed.</param>
+        /// <param name="cancellationToken"></param>
+        /// <returns>null if can't acquire the lock. This is common if somebody else holds it.</returns>
+        Task<IDistributedLock> TryLockAsync(
+            string account,
```
MikeStall: Yes, it's strange. The problem is that [Singleton] exposes an account property, so that paints us into a corner. This is a great example of new features creating liability.   Singleton.Account was easy to add when everything was tightly coupled to storage; but now that we're decoupling, it gets weird. 

---
In reply to: [120944841](https://github.com/Azure/azure-webjobs-sdk/pull/1184#discussion_r120944841) [](ancestors = 120944841)

src/Microsoft.Azure.WebJobs.Host/Singleton/BlobLeaseDistributedLockManager.cs
```diff
@@ -0,0 +1,460 @@
...
+            string leaseId = await TryAcquireLeaseAsync(lockBlob, lockPeriod, cancellationToken);
+
+            if (string.IsNullOrEmpty(leaseId))
+            {
+                return null;
+            }
+
+            if (!string.IsNullOrEmpty(lockOwnerId))
+            {
+                await WriteLeaseBlobMetadata(lockBlob, leaseId, lockOwnerId, cancellationToken);
```
MikeStall: Good catch. This is just inheriting the original implementation, so I don't think it's creating a new bug. But per your other comment, if we can get rid of owner altogether, this becomes moot.

---
In reply to: [120953083](https://github.com/Azure/azure-webjobs-sdk/pull/1184#discussion_r120953083) [](ancestors = 120953083)

src/Microsoft.Azure.WebJobs.Host/Singleton/BlobLeaseDistributedLockManager.cs
```diff
@@ -0,0 +1,460 @@
...
+        }
+
+        private static async Task<string> TryAcquireLeaseAsync(IStorageBlockBlob blob, TimeSpan leasePeriod, CancellationToken cancellationToken)
+        {
+            bool blobDoesNotExist = false;
+            try
+            {
+                // Optimistically try to acquire the lease. The blob may not yet
+                // exist. If it doesn't we handle the 404, create it, and retry below
+                return await blob.AcquireLeaseAsync(leasePeriod, null, cancellationToken);
```
MikeStall: Does TryLockAsync.LockId solve that? 

---
In reply to: [120949833](https://github.com/Azure/azure-webjobs-sdk/pull/1184#discussion_r120949833) [](ancestors = 120949833)

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLockManager.cs
fabiocav: Duh, yes... The `Try***` name completely threw me off and I overlooked/ignored that here (not like we don't deal with this issue in Functions on a daily basis :P). 

Just really don't like having to interpret null as a failure here. There are some options here (including returning a result type). Thoughts?

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLockManager.cs
MikeStall: It appears to be for diagnostics and consumed by SingletonValueProvider. 
I agree this interface is fundamentally misleading - the value could change right after we return, so a caller can really only use as an approximation for diagnostics.   
We need to keep it (existing behavior does use it), and the diagnostics are definitely valuable, but i can update the comments and make these issues more apparent. 

---
In reply to: [120969421](https://github.com/Azure/azure-webjobs-sdk/pull/1184#discussion_r120969421) [](ancestors = 120969421,120945165)

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLockManager.cs
```diff
@@ -0,0 +1,74 @@
...
```
shrishrirang: >string account, [](start = 12, length = 15)

Should this be connectionString to keep the implementation (Sql, Blob, etc) completely independent of how to get the connection string from an account name?

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLockManager.cs
shrishrirang: +1. It would be good to just expect the implementation to renew the lock when told to do so as per an appropriate renewal policy that is outside the scope of the specific IDistributedLockManager implementation. Something closer to https://github.com/shrishrirang/azure-webjobs-sdk/blob/builddev/src/Microsoft.Azure.WebJobs.Host/Lease/SqlLeaseProxy.cs#L103 would be good. In this implementation the Sql implementation just renews the lease when told to do so.

src/Microsoft.Azure.WebJobs.Host/Singleton/BlobLeaseDistributedLockManager.cs
```diff
@@ -0,0 +1,404 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.WindowsAzure.Storage;
+using Microsoft.WindowsAzure.Storage.Blob;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    // Provides a blob-leased based implementation 
+    internal class BlobLeaseDistributedLockManager : IDistributedLockManager
+    {
+        internal const string FunctionInstanceMetadataKey = "FunctionInstance";
```
shrishrirang: >internal const string FunctionInstanceMetadataKey = "FunctionInstance"; [](start = 8, length = 71)

Might be a good idea to keep the IDistributedLockManager implementations free of Functions-specific things like this one.

src/Microsoft.Azure.WebJobs.Host/Singleton/BlobLeaseDistributedLockManager.cs
```diff
@@ -0,0 +1,404 @@
...
+
+            IStorageBlobDirectory storageDirectory = null;
+            if (!_lockDirectoryMap.TryGetValue(accountName, out storageDirectory))
+            {
+                Task<IStorageAccount> task = _accountProvider.GetStorageAccountAsync(accountName, CancellationToken.None);
+                IStorageAccount storageAccount = task.Result;
+                // singleton requires block blobs, cannot be premium
+                storageAccount.AssertTypeOneOf(StorageAccountType.GeneralPurpose, StorageAccountType.BlobOnly);
+                IStorageBlobClient blobClient = storageAccount.CreateBlobClient();
+                storageDirectory = blobClient.GetContainerReference(HostContainerNames.Hosts)
```
shrishrirang: >HostContainerNames.Hosts [](start = 68, length = 24)

It would be better to pass the container, directory and blob names as parameters to the LockAsync() method instead of assuming names here. Besides, I think the BlobLeaseManager defines its own container name (azure-webjobs-hosts) and needs no directory.

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLockManager.cs
MikeStall: The comment is out of date. That's what this latest iteration does. See the Renew() api below.  


---
In reply to: [121541898](https://github.com/Azure/azure-webjobs-sdk/pull/1184#discussion_r121541898) [](ancestors = 121541898)

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLockManager.cs
MikeStall: It's not actually a connection string.  It will be null for all of your cases. 


---
In reply to: [121538582](https://github.com/Azure/azure-webjobs-sdk/pull/1184#discussion_r121538582) [](ancestors = 121538582)

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLockManager.cs
MikeStall: Latest iteration does this. 

---
In reply to: [120951100](https://github.com/Azure/azure-webjobs-sdk/pull/1184#discussion_r120951100) [](ancestors = 120951100)

src/Microsoft.Azure.WebJobs.Host/Singleton/BlobLeaseDistributedLockManager.cs
MikeStall: The nomenclature is confusing. This is not functions specific, it's the Function instance id (which is a core SDK concept). 

---
In reply to: [121542160](https://github.com/Azure/azure-webjobs-sdk/pull/1184#discussion_r121542160) [](ancestors = 121542160)

MikeStall:
Superseded by https://github.com/Azure/azure-webjobs-sdk/pull/1202, which addresses the concerns in this thread.

Issue #1169, opened 2017-06-01
Validate attribute for bindings


MikeStall:
Add a binding attribute "Validate", that goes on other binding properties (similar to AutoResolve, AppSeting) which enforces validation to a regex. 

```
        [AutoResolve]
        [Validate(Required = true, Regex = "^[A-Za-z][A-Za-z0-9]{2,62}$")]
        public string TableName
```
This would be enforced by AttributeCloner. 
This can also deal with the %% / { }   problem.  If a property is missing { }, we can validate statically (earlier). If it has { } , we must validate later. We have tests and existing expectations about validating early when possible. 

1. This lets us remove hacky code like this: 
```
        private static void ValidateAttribute(TableAttribute attribute, Type parameterType)
        {
            // Queue pre-existing  behavior: if there are { }in the path, then defer validation until runtime. 
            if (!attribute.TableName.Contains("{"))
            {
                TableClient.ValidateAzureTableName(attribute.TableName);
            }
        }
```
Many bindings (Blob, Table,Queue) have this sort of adhoc checking that can be replaced. 

2. Enables the portal to get the validation regex (https://github.com/Azure/azure-webjobs-sdk-script/issues/1416 ) 

3. Remove  BindingFactory.AddValidator -  https://github.com/Azure/azure-webjobs-sdk/issues/757

mathewc:
DataAnnotations: https://msdn.microsoft.com/en-us/library/system.componentmodel.dataannotations(v=vs.110).aspx

paulbatum:
Can we use the existing data annotation attributes instead of creating our own? There is already an ecosystem of tools and libraries that understand these.

MikeStall:
We could. Is it ok to take the new dependency on System.ComponentModel.DataAnnotations?

MikeStall:
Replacing Queue and Table can be tricky. a) there's a lot of existing cruft; b) they have some richer errors beyond just RegEx. 

Several phases here:
1. Expose this for extensions., 
2. Solve https://github.com/Azure/azure-webjobs-sdk/issues/757
3. Remove obvious hacks for Table + Queue (see comment above).
4. Fully apply to all internal types (Table, Blob, Queue)  and purge old-style validation logic. 

[...]

PR #1188, opened 2017-06-13
Add Validation attribute.

MikeStall:
Resolve https://github.com/Azure/azure-webjobs-sdk/issues/1169

Investigate if we can apply to more things internally in SDK (Table, Blob, Queue)

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
 using System.Linq;
 using System.Reflection;
 using System.Threading.Tasks;
@@ -16,6 +17,8 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
     // Func to transform Attribute,BindingData into value for cloned attribute property/constructor arg
     // Attribute is the new cloned attribute - null if constructor arg (new cloned attr not created yet)
     using BindingDataResolver = Func<Attribute, IReadOnlyDictionary<string, object>, object>;
+
+    using Validator = Action<object>;
 
     // Clone an attribute and resolve it.
     // This can be tricky since some read-only properties are set via the constructor.
@@ -88,12 +91,16 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
         // transforms binding data to appropriate resolver (appsetting, autoresolve, or originalValue)
         private BindingDataResolver GetResolver(PropertyInfo propInfo, INameResolver nameResolver, BindingDataContract contract)
         {
+            // Do the attribute lookups once upfront, and then cache them (via func closures) for subsequent runtime usage. 
             object originalValue = propInfo.GetValue(_source);
             AppSettingAttribute appSettingAttr = propInfo.GetCustomAttribute<AppSettingAttribute>();
             AutoResolveAttribute autoResolveAttr = propInfo.GetCustomAttribute<AutoResolveAttribute>();
+            Validator validator = GetValidatorFunc(propInfo, appSettingAttr != null);
 
             if (appSettingAttr == null && autoResolveAttr == null)
             {
+                validator(originalValue);
+
                 // No special attributes, treat as literal. 
                 return (newAttr, bindingData) => originalValue;
             }
@@ -113,37 +120,38 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
             // first try to resolve with app setting
             if (appSettingAttr != null)
             {
-                return GetAppSettingResolver(str, appSettingAttr, nameResolver, propInfo);
+                return GetAppSettingResolver(str, appSettingAttr, nameResolver, propInfo, validator);
             }
 
             // Must have an [AutoResolve]
             // try to resolve with auto resolve ({...}, %...%)
-            return GetAutoResolveResolver(str, autoResolveAttr, nameResolver, propInfo, contract);            
+            return GetAutoResolveResolver(str, autoResolveAttr, nameResolver, propInfo, contract, validator);            
         }
 
         // Apply AutoResolve attribute 
-        internal BindingDataResolver GetAutoResolveResolver(string originalValue, AutoResolveAttribute autoResolveAttr, INameResolver nameResolver, PropertyInfo propInfo, BindingDataContract contract)
+        internal BindingDataResolver GetAutoResolveResolver(string originalValue, AutoResolveAttribute autoResolveAttr, INameResolver nameResolver, PropertyInfo propInfo, BindingDataContract contract, Validator validator)
         {
             if (string.IsNullOrWhiteSpace(originalValue))
             {
                 if (autoResolveAttr.Default != null)
                 {
-                    return GetBuiltinTemplateResolver(autoResolveAttr.Default, nameResolver);
+                    return GetBuiltinTemplateResolver(autoResolveAttr.Default, nameResolver, validator);
                 }
                 else
                 {
+                    validator(originalValue);
                     return (newAttr, bindingData) => originalValue;
                 }
             }
             else
             {
                 _autoResolves[propInfo] = autoResolveAttr;
-                return GetTemplateResolver(originalValue, autoResolveAttr, nameResolver, propInfo, contract);
+                return GetTemplateResolver(originalValue, autoResolveAttr, nameResolver, propInfo, contract, validator);
             }
         }
 
         // Apply AppSetting attribute. 
-        internal static BindingDataResolver GetAppSettingResolver(string originalValue, AppSettingAttribute attr, INameResolver nameResolver, PropertyInfo propInfo)
+        internal static BindingDataResolver GetAppSettingResolver(string originalValue, AppSettingAttribute attr, INameResolver nameResolver, PropertyInfo propInfo, Validator validator)
         {
             string appSettingName = originalValue ?? attr.Default;
             string resolvedValue = string.IsNullOrEmpty(appSettingName) ?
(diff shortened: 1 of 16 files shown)


src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -161,25 +168,63 @@ internal static BindingDataResolver GetAppSettingResolver(string originalValue,
...
+
+            foreach (var attr in attrs)
+            {
+                try
+                {
+                    attr.Validate(value, propInfo.Name);
+                }
+                catch (Exception e)
+                {
+                    throw new InvalidOperationException($"'{propInfo.Name}' can't be '{value}': {e.Message}");
```
mamaso: We shouldn't log 'value' here, could be user secrets, etc.

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -161,25 +168,63 @@ internal static BindingDataResolver GetAppSettingResolver(string originalValue,
...
         // These only have access to the {sys} builtin variable and don't get access to trigger binding data. 
-        internal static BindingDataResolver GetBuiltinTemplateResolver(string originalValue, INameResolver nameResolver)
+        internal static BindingDataResolver GetBuiltinTemplateResolver(string originalValue, INameResolver nameResolver, PropertyInfo propInfo)
         {
             string resolvedValue = nameResolver.ResolveWholeString(originalValue);
 
             var template = BindingTemplate.FromString(resolvedValue);
+            if (!template.ParameterNames.Any())
+            {
+                // No { } tokens, bind eagerly up front.
```
mamaso: perhaps add convenience getters to BindingTemplate, something like

`bool IsRuntimeResolution => ParameterNames.Any()`

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -161,25 +168,63 @@ internal static BindingDataResolver GetAppSettingResolver(string originalValue,
...
+            return (newAttr, bindingData) =>
+            {
+                var newValue = template.Bind(SystemBindingData.GetSystemBindingData(bindingData));
+                RunValidateAttributes(newValue, propInfo);
+                return newValue;
+            };
         }
 
         // AutoResolve
         internal static BindingDataResolver GetTemplateResolver(string originalValue, AutoResolveAttribute attr, INameResolver nameResolver, PropertyInfo propInfo, BindingDataContract contract)
```
mamaso: can you remove this duplication between `GetTemplateResolver` and `GetBuiltinTemplateResolver`

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -292,10 +337,13 @@ private static string TemplateBind(IResolutionPolicy policy, PropertyInfo prop,
         {
             if (bindingData == null)
             {
+                // Skip validation if no binding data provided. We can't do the { } substitutions.
```
mamaso: what happens in this case? an error? we just use the pattern?

we might still want to validate

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
MikeStall: Which duplication?

---
In reply to: [123053610](https://github.com/Azure/azure-webjobs-sdk/pull/1188#discussion_r123053610) [](ancestors = 123053610)

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
MikeStall: We can't validate without the binding data. But this is to support GetNameResolvedAttribute(), in which case we don't want validation anyways. 

---
In reply to: [123054807](https://github.com/Azure/azure-webjobs-sdk/pull/1188#discussion_r123054807) [](ancestors = 123054807)

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
MikeStall: Our existing validation code (which this is replacing) includes the value name.  
[https://github.com/Azure/azure-webjobs-sdk/blob/663a508e8a851629c26a51e7de3af36629dfd120/src/Microsoft.Azure.WebJobs.Host/Tables/TableClient.cs#L107](https://github.com/Azure/azure-webjobs-sdk/blob/663a508e8a851629c26a51e7de3af36629dfd120/src/Microsoft.Azure.WebJobs.Host/Tables/TableClient.cs#L107) 

If we want to suppress secrets, we should have some way of annotating that a value could be secret (ie, SecureString) and then act on that. Refusing to log any user parameter is too broad a brush. 


---
In reply to: [123052472](https://github.com/Azure/azure-webjobs-sdk/pull/1188#discussion_r123052472) [](ancestors = 123052472)

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
mamaso: Ah nvmd, missed a line break in the github viewer, nvmd it's not clear duplication.

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -161,25 +168,63 @@ internal static BindingDataResolver GetAppSettingResolver(string originalValue,
             return (newAttr, bindingData) => resolvedValue;
         }
 
+        // Run validition. This needs to be run at different stages. 
+        // In general, run as early as possible. If there are { } tokens, then we can't run until runtime.
+        // But if there are no { }, we can run statically. 
+        // If there's no [AutoResolve], [AppSettings], then we can run immediately.
+        private static void RunValidateAttributes(object value, PropertyInfo propInfo)
+        {
+            var attrs = propInfo.GetCustomAttributes<ValidationAttribute>();
```
mamaso: These attrs should be cached to save so that we don't run GetCustomAttributes every execution.

IMO the logic is way too scattered.

I think we should change the BindingDataResolver alias to a class, which caches PropInfo attributes and differentiates between dynamic / static resolvers, i.e.:

```
//instead of return (newAttr, bindingData) => resolvedValue;
return new StaticResolver(propInfo, resolvedValue);  
````

This section:
https://github.com/Azure/azure-webjobs-sdk/pull/1188/files#diff-7146d5bc5e33487d5c81156c343746abR83

would then become something like
```
var resolver = GetResolver(prop, nameResolver, bindingDataContract);
resolver.Validate();
return (Action<TAttribute, BindingData>)((attr, data) => {
  object resolved = Resolver.Resolve(attr, data); // resolve validates if dynamic resolver
  prop.SetValue(attr, resolved);
});
```

File: src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs (hunk 1 above)
Comment (mamaso) [inline src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs]: This code enables more than just the existing table name validation code, it can be used by any extension author - unfortunately I think that the broad brush is what we need, as we can't ensure that they will use a [SecureString] attribute correctly.

Not logging the value if also annotated with [AppSetting] might be slightly better.

Comment (MikeStall) [inline src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs]: This "Not logging the value if also annotated with [AppSetting] might be slightly better." is a good compromise

---
In reply to: [123106464](https://github.com/Azure/azure-webjobs-sdk/pull/1188#discussion_r123106464) [](ancestors = 123106464)

File: src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs (hunk 2 above)
Comment (MikeStall) [inline src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs]: did that and applied throughout the code. 

---
In reply to: [123053060](https://github.com/Azure/azure-webjobs-sdk/pull/1188#discussion_r123053060) [](ancestors = 123053060)

File: src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs [hunk 6]
```diff
@@ -158,31 +166,84 @@ internal static BindingDataResolver GetAppSettingResolver(string originalValue,
...
+                {
+                    try
+                    {
+                        attr.Validate(value, propInfo.Name);
+                    }
+                    catch (Exception e)
+                    {
+                        if (dontLogValues)
+                        {
+                            throw new InvalidOperationException($"'{propInfo.Name}' is illegal: {e.Message}");
```
mathewc: nit: "Validation failed for property '{x}'. {message}"

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -158,31 +166,84 @@ internal static BindingDataResolver GetAppSettingResolver(string originalValue,
...
+                        }
+                        else
+                        {
+                            throw new InvalidOperationException($"'{propInfo.Name}' can't be '{value}': {e.Message}");
```
mathewc: nit: "Validation failed for property '{x}', value '{y}'. {message}"

src/Microsoft.Azure.WebJobs/TableAttribute.cs
```diff
@@ -40,6 +41,11 @@ public class TableAttribute : Attribute, IConnectionProvider
         private readonly string _partitionKey;
         private readonly string _rowKey;
 
+        // Table Name Rules are summerized in this page https://msdn.microsoft.com/en-us/library/azure/dd179338.aspx
+        // Table names may contain only alphanumeric characters, can not start with a numeric character and must be 
+        // 3 to 63 characters long
+        private const string TableNameRegex = "^[A-Za-z][A-Za-z0-9]{2,62}$";
```
mathewc: We have a different regex in our templates: https://github.com/Azure/azure-webjobs-sdk-templates/blob/dev/Functions.Templates/Bindings/bindings.json#L1177

test/Microsoft.Azure.WebJobs.Host.UnitTests/AttributeClonerTests.cs
```diff
@@ -140,6 +141,35 @@ public string TemplateBind(PropertyInfo propInfo, Attribute attribute, BindingTe
...
+            [AppSetting]
+            public string Value { get; set; }
+        }
+
+        // Validation with neither AppSetting nor AutoResolve
+        [Binding]
+        public class ValidationOnlyAttribute : Attribute
+        {
+            // Allow  { } that look like token substitution, but it's not since this isn't AutoResolve
+            [RegularExpression("^.{1,3}$")]
```
mathewc: I think we should also consider honoring these validation attributes when applied to poco bindings. Allowing declarative validation to be applied to function inputs.

src/Microsoft.Azure.WebJobs/TableAttribute.cs
MikeStall: Good catch. This here is the existing regex the runtime uses, so it's the definitive correct one. 
The one in Bindings.json looks like it's accounting for { } and % %tags - so it's actually incorrect.  Whereas this one is applied after the { } , %% substitutiosn. 



---
In reply to: [123387086](https://github.com/Azure/azure-webjobs-sdk/pull/1188#discussion_r123387086) [](ancestors = 123387086)

test/Microsoft.Azure.WebJobs.Host.UnitTests/AttributeClonerTests.cs
MikeStall: can you clarify? 

---
In reply to: [123387817](https://github.com/Azure/azure-webjobs-sdk/pull/1188#discussion_r123387817) [](ancestors = 123387817)

src/Microsoft.Azure.WebJobs/TableAttribute.cs
fabiocav: @MikeStall could you open a PR to fix that there as well?

fabiocav:
Sweet! :)

test/Microsoft.Azure.WebJobs.Host.UnitTests/AttributeClonerTests.cs
mathewc: We now support these declarative validation attributes for binding authors. Might be interesting to support them for end users as well on their model types. E.g. I mark up my queue trigger poco, and our pipeline runs the validation and fails the invoke if the validations don't pass. It'd be a new validation stage in our pipeline.

src/Microsoft.Azure.WebJobs/TableAttribute.cs
mathewc: I logged issue https://github.com/Azure/azure-webjobs-sdk-templates/issues/479

mathewc:
What about Blob, Queue, etc. as this issue mentions?

src/Microsoft.Azure.WebJobs/TableAttribute.cs
soninaren: The regex in binding.json is same as the one mentioned here, except for the { } and %% which accounts for variable name. How is the regex in binding.json incorrect?

MikeStall: I'll comment abut that on the new issue 479. 

---
In reply to: [123435508](https://github.com/Azure/azure-webjobs-sdk/pull/1188#discussion_r123435508) [](ancestors = 123435508)

test/Microsoft.Azure.WebJobs.Host.UnitTests/AttributeClonerTests.cs
MikeStall: It's out of scope for this PR; but I filed [https://github.com/Azure/azure-webjobs-sdk/issues/1199](https://github.com/Azure/azure-webjobs-sdk/issues/1199) to track it. 

---
In reply to: [123418394](https://github.com/Azure/azure-webjobs-sdk/pull/1188#discussion_r123418394) [](ancestors = 123418394)

MikeStall:
Out of scope here. I filed [https://github.com/Azure/azure-webjobs-sdk/issues/1200](https://github.com/Azure/azure-webjobs-sdk/issues/1200)  which also documents why that's tricky. 



---
In reply to: [310271399](https://github.com/Azure/azure-webjobs-sdk/pull/1188#issuecomment-310271399) [](ancestors = 310271399)

Issue #980, opened 2017-01-20
Invocation Filters


mathewc:
We should consider adding an Invoke Filter concept. This is very similar to WebAPI ActionFilters (the problem domain is basically the same). We've already had some related asks (see #734, #895). The idea is to provide a declarative programming model that allows people to attach filters to their job methods, classes, etc. similar to the way filters are specified in WebAPI. The invoke filter will receive a context object that contains everything needed to perform validations, do logging, etc. before any input/output bindings occur and before the function executes. Here’s an sketch:

```csharp
public static class Functions
{
    [MyInvokeFilter]
    public static void ProcessMessage([QueueTrigger(“test”)] string message) { … }
}

public class MyInvokeFilterAttribute : InvokeFilterAttribute
{
    public override void OnExecuting(FunctionExecutingContext context) 
    {
        // perform validations, etc.
    }

    public override void OnExecuted(FunctionExecutedContext context) 
    {
        // perform validations, etc.
    }
}
```

This attribute could be applied at the function or class level. We could also support host level as well (e.g. add to JobHostConfiguration service collection). Just like WebAPI ActionFilters, where they can be specified declaratively, as well as at the service config level.

abatishchev:
+1. Please. Otherwise extendability and traceability are actually very limited.

jamesbascle:
I'd like to make a note that, IMO, this should support async WebJobs as first class citizens - it was actually a rather complicated and touchy thing to get working with my interception strategy. Having to add ContinuesWith, manually setting cancel/exception/result, and more, is a lot to ask when the entire point behind WebJobs is to keep things as simple as possible.

mathewc:
@jamesbascle Of course we'd support async filters. Basically we'll be doing something like WebAPI [ActionFilters](https://msdn.microsoft.com/en-us/library/system.web.http.filters.actionfilterattribute(v=vs.118).aspx). As you can see, that model supports sync/async. If you're asking for something more, please share details.

VenkateshSrini:
My two cents modification to your code
`public static class Functions
{
    [MyInvokeFilter]
    public static void ProcessMessage([QueueTrigger(“test”)] string message) { … }
}

public class MyInvokeFilterAttribute : InvokeFilterAttribute
{
    public override **Task** OnExecuting(FunctionExecutingContext context) 
    {
        // perform validations, etc.
    }

    public override **Task** OnExecuted(FunctionExecutedContext context) 
    {
        // perform validations, etc.
[...]

PR #1189, opened 2017-06-16
WIP: Invocation Filters Feature

hhawkins:
For feature tracked here: https://github.com/Azure/azure-webjobs-sdk/issues/980. Sending out early to get some feedback.


src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -31,11 +31,12 @@ internal class FunctionExecutor : IFunctionExecutor
         private readonly IAsyncCollector<FunctionInstanceLogEntry> _functionEventCollector;
         private readonly ILogger _logger;
         private readonly ILogger _resultsLogger;
+        private readonly JobHost _host;
```
hhawkins: Included the host for the invoke functions attribute

src/Microsoft.Azure.WebJobs.Host/FunctionExecutingContext.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        /// <summary>
+        /// Constructor to set the context
+        /// </summary>
+        /// <param name="id"></param>
+        /// <param name="fullName"></param>
+        /// <param name="arguments"></param>
+        /// <param name="logger"></param>
+        /// <param name="host"></param>
+        public FunctionExecutingContext(Guid id, string fullName, IReadOnlyDictionary<string, object> arguments, ILogger logger, JobHost host) :
```
MikeStall: >, JobHost host [](start = 127, length = 14)

This break layering. JobHost is a very powerful object to be passing around. I'm guessing this is so we can get at JobHost.Call?   I'll suggest some alternatives in email.

src/Microsoft.Azure.WebJobs.Host/FunctionInvocationContext.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        /// <summary>
+        /// Constructor to set the context with the JobHost
+        /// </summary>
+        /// <param name="id"></param>
+        /// <param name="fullName"></param>
+        /// <param name="arguments"></param>
+        /// <param name="logger"></param>
+        /// <param name="host"></param>
+        protected FunctionInvocationContext(Guid id, string fullName, IReadOnlyDictionary<string, object> arguments, ILogger logger, JobHost host)
```
MikeStall: >JobHost host [](start = 133, length = 12)

same here.

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,96 @@
...
+                {
+                    executingContext.Logger.LogInformation("Executing function from filter...");
+
+                    var parameterName = methodInfo.GetParameters()[0].Name;
+                    Dictionary<string, object> parameters = new Dictionary<string, object>();
+                    parameters.Add(parameterName, executingContext);
+
+                    try
+                    {
+                        await host.CallAsync(methodInfo, parameters, cancellationToken);
```
MikeStall: >CallAsync [](start = 35, length = 9)

From a tracking perspective, we should be able to mark that this new instance is related to the outer instnace (TriggerReason, parent guid id, etc)

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,96 @@
...
+                    }
+                    catch (Exception e)
```
MikeStall: >Exception [](start = 27, length = 9)

Looks like we swallow and log the exception. Should an exception here cause an abort?

src/Microsoft.Azure.WebJobs.Host/JobHost.cs
```diff
@@ -85,6 +85,14 @@ public JobHost(JobHostConfiguration configuration)
             _stoppingTokenSource = CancellationTokenSource.CreateLinkedTokenSource(_shutdownTokenSource.Token);
         }
 
+        /// <summary>
+        /// Gets the JobHostConfiguration of this JobHost
+        /// </summary>
+        public JobHostConfiguration Configuration
```
MikeStall: > public JobHostConfiguration Configuration [](start = 7, length = 42)

Adding this also breaks layering. If we fix the other JobHost.Call issue, hopefully we can then avoid this.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -618,6 +625,48 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+                var cancellationToken = functionCancellationTokenSource.Token;
+                try
+                {
+                    await invokeFunction.OnExecutingAsync(functionContext, cancellationToken);
+                }
+                catch (Exception e)
+                {
+                    logger.LogInformation(e.ToDetails());
+                }
+            }
```
MikeStall: Can we avoid this? Since this attr is derived from InvocationFilterAttribute, can the above case cover this?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -618,6 +625,48 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+            if (invokeFilter != null)
+            {
+                var functionExecutingContext = new FunctionExecutingContext(instance.Id, instance.FunctionDescriptor.FullName, parameters, logger, host);
+                var cancellationToken = functionCancellationTokenSource.Token;
+                try
+                {
+                    await invokeFilter.OnExecutingAsync(functionExecutingContext, cancellationToken);
+                }
+                catch (Exception e)
+                {
```
MikeStall: >  { [](start = 14, length = 3)

This swallows. Should an error cause an abort?

src/Microsoft.Azure.WebJobs.Host/InvocationFilterAttribute.cs
```diff
@@ -0,0 +1,39 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// This is the first iteration for the Invocation Filter Attribute
+    /// </summary>
+    [CLSCompliant(false)]
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, Inherited = true, AllowMultiple = false)]
+    public abstract class InvocationFilterAttribute : Attribute
```
MikeStall: >InvocationFilterAttribute [](start = 26, length = 25)

What's the invariant around state here? Does each method instance get its own copy of this attribute? And such, can the OnExecuting set instance methods that are accessed in OnExecuted?

src/Microsoft.Azure.WebJobs.Host/FunctionExecutedContext.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        /// <summary>
+        /// Constructor to set the context
+        /// </summary>
+        /// <param name="id"></param>
+        /// <param name="fullName"></param>
+        /// <param name="arguments"></param>
+        /// <param name="logger"></param>
+        /// <param name="result"></param>
+        public FunctionExecutedContext(Guid id, string fullName, IReadOnlyDictionary<string, object> arguments, ILogger logger, FunctionResult result) :
```
MikeStall: >arguments [](start = 101, length = 9)

For out parameters, is this the initial result or the final result?

hhawkins: This would contain the final result after any before invocation filters and the actual function itself

src/Microsoft.Azure.WebJobs.Host/InvocationFilterAttribute.cs
hhawkins: The idea was that this was the base class for all invocation purposes. Currently it is possible to access instance methods that are able to be accessed in both OnExecuting and OnExecuted. The TestFailingFilter invocation test reflects this happening.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
hhawkins: I added a throw statement

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -31,11 +31,12 @@ internal class FunctionExecutor : IFunctionExecutor
         private readonly IAsyncCollector<FunctionInstanceLogEntry> _functionEventCollector;
         private readonly ILogger _logger;
         private readonly ILogger _resultsLogger;
+        private readonly JobHost _host;
 
         private HostOutputMessage _hostOutputMessage;
 
         public FunctionExecutor(IFunctionInstanceLogger functionInstanceLogger, IFunctionOutputLogger functionOutputLogger,
-                IWebJobsExceptionHandler exceptionHandler, TraceWriter trace, IAsyncCollector<FunctionInstanceLogEntry> functionEventCollector = null,
+                IWebJobsExceptionHandler exceptionHandler, TraceWriter trace, JobHost host = null, IAsyncCollector<FunctionInstanceLogEntry> functionEventCollector = null,
```
mathewc: As discussed, we should have a new IJobMethodInvoker interface that is passed in here. It must be specified, so you shouldn't have "= null" here.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -58,13 +59,19 @@ internal class FunctionExecutor : IFunctionExecutor
                 throw new ArgumentNullException("trace");
             }
 
+            if (host == null)
```
mathewc: remove this since the host/invoker is required

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
mathewc: Yes - this special block shouldn't be needed.

src/Microsoft.Azure.WebJobs.Host/InvocationFilterAttribute.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// This is the first iteration for the Invocation Filter Attribute
+    /// </summary>
+    [CLSCompliant(false)]
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, Inherited = true, AllowMultiple = false)]
```
mathewc: I think you can remove this AttributeUsage annotation, since this type is abstract. Only the AttributeUsage declared on concrete subclasses will actually have any effect I believe. You should verify that.

src/Microsoft.Azure.WebJobs.Host/InvocationFilterAttribute.cs
```diff
@@ -0,0 +1,39 @@
...
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// This is the first iteration for the Invocation Filter Attribute
```
mathewc: Put the proper code comments in place now. E.g. "Defines an interface for declarative invocation filters."

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,96 @@
...
+using System.Linq;
+using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// This is the first iteration of the InvokeFunctionAttribute
```
mathewc: Fix comments - here and everywhere else

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// </summary>
+        /// <returns></returns>
+        public override async Task OnExecutingAsync(FunctionExecutingContext executingContext, CancellationToken cancellationToken)
+        {
+            if (executingContext == null)
+            {
+                throw new ArgumentNullException("executingContext");
+            }
+            
+            // TODO: Implement invoke function logic
```
mathewc: Address and remove all TODO comments

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -618,6 +625,48 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
             //      a. If throwOnTimeout, we throw the FunctionTimeoutException.
             //      b. If !throwOnTimeout, wait for the task to complete.
 
+            // Where the invocation filter attribute functionality is currently temporarily placed
```
mathewc: Remove this comment? This isn't temporary anymore, right?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -618,6 +625,48 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+            var functionMethod = instance.FunctionDescriptor.Method;
+            var invokeFilter = functionMethod.GetCustomAttribute<InvocationFilterAttribute>();
```
mathewc: You need to support hierarchical attributes here, e.g. InvokeFunctionFilter attribute can be applied at both the Method and Class level. So you need to gather ALL attributes from all levels. You should also have some tests that verify you can apply attributes at various levels and they get invoked.

mathewc: Take a look at the existing TypeUtility.GetHierarchicalAttributeOrNull method, which walks up the hierarchy and returns the first attribute found.

You should add a new method to TypeUtility that does this walk and returns multiple attributes. E.g. name it GetHierarchicalAttributes. It walks up from the method and returns all Method/Class level attributes.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -618,6 +625,48 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+            
+            var parameters = new Dictionary<string, object>();
+            string[] pnames = parameterNames.ToArray<string>();
```
mathewc: you can remove this line - you don't need "pnames". You can just index into parameterNames directly below.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -618,6 +625,48 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+
+            for (var i = 0; i < invokeParameters.Length; i++)
+            {
+                parameters.Add(pnames[i], invokeParameters[i]);
+            }
+
+            if (invokeFilter != null)
```
mathewc: because you can have multiple filters, this will need to be a foreach loop

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
mathewc: The intent was to pass any exceptions thrown from filters or the function invocation on to any Executed filters. If we keep that design, you can't just rethrow here, instead you'd have to include all of this in the try/catch you have below.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -618,6 +625,48 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+                {
+                    await invokeFilter.OnExecutingAsync(functionExecutingContext, cancellationToken);
+                }
+                catch (Exception e)
+                {
+                    logger.LogInformation(e.ToDetails());
+                }
+            }
+
+            // Where invoke function is currently placed
```
mathewc: remove these extraneous comments throughout

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -633,7 +682,35 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
                 await TryHandleTimeoutAsync(invokeTask, CancellationToken.None, throwOnTimeout, timeoutTokenSource.Token, timerInterval, instance, null);
             }
 
-            await invokeTask;
+            FunctionResult functionResult = null;
+
+            try
+            {
+                await invokeTask;
```
mathewc: As mentioned above, you'll want to include invocation of executing filters within this try/catch so your below error handling gets any errors.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -633,7 +682,35 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+            }
+            catch (Exception ex)
+            {
+                functionResult = new FunctionResult(ex);
+                throw;
+            }
+            finally
+            {
+                // Where the invocation filter attribute functionality is currently placed
+                if (invokeFilter != null)
```
mathewc: same comment here, this will need to be a foreach over all attributes

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -633,7 +682,35 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+                {
+                    var functionExecutingContext = new FunctionExecutedContext(instance.Id, instance.FunctionDescriptor.FullName, parameters, logger, functionResult);
+                    var cancellationToken = functionCancellationTokenSource.Token;
+                    try
+                    {
+                        await invokeFilter.OnExecutedAsync(functionExecutingContext, cancellationToken);
+                    }
+                    catch (Exception e)
+                    {
+                        logger.LogInformation(e.ToDetails());
```
mathewc: errors should be logged via logger.LogError

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -618,6 +625,48 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+            string[] pnames = parameterNames.ToArray<string>();
+
+            for (var i = 0; i < invokeParameters.Length; i++)
+            {
+                parameters.Add(pnames[i], invokeParameters[i]);
+            }
+
+            if (invokeFilter != null)
+            {
+                var functionExecutingContext = new FunctionExecutingContext(instance.Id, instance.FunctionDescriptor.FullName, parameters, logger, host);
```
mathewc: It'd be good to factor out all this filter invoke code into an InvokeExecutingFilters/InvokeExecutedFilters method to keep the flow of this method easier to follow.

src/Microsoft.Azure.WebJobs.Host/FunctionExecutedContext.cs
```diff
@@ -0,0 +1,35 @@
...
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Azure.WebJobs.Host.Executors;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// The context for an executed function. This needs to be expanded on later.
```
mathewc: Let's get all the comments up to date in a somewhat finalized form, with no TODOs, etc.

src/Microsoft.Azure.WebJobs.Host/FunctionExecutedContext.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <param name="logger"></param>
+        /// <param name="result"></param>
+        public FunctionExecutedContext(Guid id, string fullName, IReadOnlyDictionary<string, object> arguments, ILogger logger, FunctionResult result) :
+            base(id, fullName, arguments, logger)
+        {
+            Result = result;
+        }
+
+        /// <summary>
+        /// Make sure the final result is logged
```
mathewc: Fix up comments throughout. E.g. for this one "Gets or sets the function result". See other public types for our practices

src/Microsoft.Azure.WebJobs.Host/FunctionExecutedContext.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// The context for an executed function. This needs to be expanded on later.
+    /// </summary>
+    [CLSCompliant(false)]
+    public class FunctionExecutedContext : FunctionInvocationContext
+    {
+        /// <summary>
+        /// Constructor to set the context
+        /// </summary>
+        /// <param name="id"></param>
+        /// <param name="fullName"></param>
```
mathewc: Let's rename "fullName" to "functionName"

src/Microsoft.Azure.WebJobs.Host/FunctionExecutingContext.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// The context for an executed function. This needs to be expanded on later.
+    /// </summary>
+    [CLSCompliant(false)]
+    public class FunctionExecutingContext : FunctionInvocationContext
+    {
+        /// <summary>
+        /// Constructor to set the context
+        /// </summary>
+        /// <param name="id"></param>
+        /// <param name="fullName"></param>
```
mathewc: Rename parameter to "name"

src/Microsoft.Azure.WebJobs.Host/FunctionInvocationContext.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// The context for an executed function. This needs to be expanded on later.
+    /// </summary>
+    [CLSCompliant(false)]
+    public abstract class FunctionInvocationContext
+    { 
+        /// <summary>
+        /// Constructor to set the context
+        /// </summary>
+        /// <param name="id"></param>
+        /// <param name="fullName"></param>
```
mathewc: Rename to "name"

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// <summary>
+        /// Executed filter
+        /// </summary>
+        public string ExecutedFilter { get; }
+
+        /// <summary>
+        /// Call the requested function
+        /// </summary>
+        /// <returns></returns>
+        public override async Task OnExecutingAsync(FunctionExecutingContext executingContext, CancellationToken cancellationToken)
```
mathewc: Need to implement OnExecuted as well

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,96 @@
...
+        {
+            if (executingContext == null)
+            {
+                throw new ArgumentNullException("executingContext");
+            }
+            
+            // TODO: Implement invoke function logic
+            if (!string.IsNullOrEmpty(ExecutingFilter))
+            {
+                IReadOnlyDictionary<string, object> paramaters = executingContext.Arguments;
```
mathewc: Factor out this function invoke code into a shared method that is used for both Executing and Executed (since most of the logic is the same)

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,96 @@
...
+                    {
+                        executingContext.Logger.LogInformation(e.ToDetails());
+                    }
+
+                    // function executed
+                    executingContext.Logger.LogInformation("Executed function: " + ExecutingFilter);
+                }
+                else
+                {
+                    // TODO: @Hamza add error handling
```
mathewc: Address todos

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,96 @@
...
+            // TODO: Implement invoke function logic
+            if (!string.IsNullOrEmpty(ExecutingFilter))
+            {
+                IReadOnlyDictionary<string, object> paramaters = executingContext.Arguments;
+                JobHost host = executingContext.Host;
+                MethodInfo methodInfo = null;
+
+                foreach (var type in host.Configuration.TypeLocator.GetTypes())
+                {
+                    methodInfo = type.GetMethods().SingleOrDefault(p => string.Compare(p.Name, ExecutingFilter, StringComparison.OrdinalIgnoreCase) == 0);
```
mathewc: We'll need some validation in place to verify the expected signature. E.g. for Executing methods, the first parameter must be an executing context. Similarly for Executed.

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,96 @@
...
+                }
+
+                if (methodInfo != null)
+                {
+                    executingContext.Logger.LogInformation("Executing function from filter...");
```
mathewc: Change to "Executing Function Filter '{ExecutingFilter}'"

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,96 @@
...
+                    {
+                        await host.CallAsync(methodInfo, parameters, cancellationToken);
+                    }
+                    catch (Exception e)
+                    {
+                        executingContext.Logger.LogInformation(e.ToDetails());
+                    }
+
+                    // function executed
+                    executingContext.Logger.LogInformation("Executed function: " + ExecutingFilter);
```
mathewc: Change to "Executed Function Filter '{ExecutingFilter}' (Result)", where "Result" logs whether it passed or failed.

src/Microsoft.Azure.WebJobs.Host/JobHost.cs
mathewc: Yes we can, since Typelocator will no longer be needed to find MethodInfos for filter functions. The new IJobMethodInvoker interface will take a string name. Internally it will do the string => MethodInfo conversion.

src/Microsoft.Azure.WebJobs.Host/FunctionInvocationContext.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// <param name="host"></param>
+        protected FunctionInvocationContext(Guid id, string fullName, IReadOnlyDictionary<string, object> arguments, ILogger logger, JobHost host)
+        {
+            Id = id;
+            FullName = fullName;
+            Arguments = arguments;
+            Logger = logger;
+            Host = host;
+        }
+
```
mathewc: We discussed also adding a property bag to this context object, to allow users to flow their own custom state through between their executing/executed filters. We should add:

```csharp
public IDictionary<string, object> Properties { get; }
```

Any properties added to the executing context are flowed through on the executed context.

hhawkins:
#1207 This is the new pull request to refer too. The pull request is coming from my feature branch instead of the dev branch.

PR #1190, opened 2017-06-19
Storage account caching and validation fix

mamaso:
Improvement for connection exhaustion.

For non storage/dashboard accounts, we were running the validation calls (blob endpoint & queue endpoint http calls) every time we called `TryGetAccountAsync`

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageAccountProvider.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageAccountProvider.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Globalization;
 using System.Threading;
 using System.Threading.Tasks;
@@ -15,6 +16,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
         private readonly IStorageCredentialsValidator _storageCredentialsValidator;
         private readonly IStorageAccountParser _storageAccountParser;
         private readonly IServiceProvider _services;
+        private readonly ConcurrentDictionary<string, IStorageAccount> _accounts = new ConcurrentDictionary<string, IStorageAccount>();
 
         private IStorageAccount _dashboardAccount;
         private bool _dashboardAccountSet;
@@ -150,7 +152,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
             }
         }
 
-        public async Task<IStorageAccount> TryGetAccountAsync(string connectionStringName, CancellationToken cancellationToken)
+        private async Task<IStorageAccount> CreateAndValidateAccountAsync(string connectionStringName, CancellationToken cancellationToken)
         {
             IStorageAccount account = null;
             var isPrimary = true;
@@ -175,7 +177,6 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
 
             if (account != null)
             {
-                // On the first attempt, this will make a network call to verify the credentials work.
                 await _storageCredentialsValidator.ValidateCredentialsAsync(account, cancellationToken);
 
                 if (isPrimary)
@@ -184,6 +185,18 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                 }
             }
 
+            return account;
+        }
+
+        public async Task<IStorageAccount> TryGetAccountAsync(string connectionStringName, CancellationToken cancellationToken)
+        {
+            IStorageAccount account;
+            if (!_accounts.TryGetValue(connectionStringName, out account))
+            {
+                // in rare cases createAndValidateAccountAsync could be called multiple times for the same account
+                account = await CreateAndValidateAccountAsync(connectionStringName, cancellationToken);
+                _accounts.AddOrUpdate(connectionStringName, (cs) => account, (cs, a) => account);
+            }
             return account;
         }
 
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
@@ -17,8 +17,6 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
 {
     internal class DefaultStorageCredentialsValidator : IStorageCredentialsValidator
     {
-        private readonly HashSet<StorageCredentials> _validatedCredentials = new HashSet<StorageCredentials>();
-
         public async Task ValidateCredentialsAsync(IStorageAccount account, CancellationToken cancellationToken)
         {
             if (account == null)
@@ -26,16 +24,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                 throw new ArgumentNullException("account");
             }
 
-            StorageCredentials credentials = account.Credentials;
-
-            // Avoid double-validating the same account and credentials.
-            if (_validatedCredentials.Contains(credentials))
-            {
-                return;
-            }
-
             await ValidateCredentialsAsyncCore(account, cancellationToken);
-            _validatedCredentials.Add(credentials);
         }
 
         // Test that the credentials are valid and classify the account.Type as one of StorageAccountTypes
@@ -58,9 +47,8 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
             catch (Exception e)
             {
                 var storageException = e as StorageException;
-                if (storageException != null && storageException.RequestInformation != null &&
-                    storageException.RequestInformation.HttpStatusCode == 400 &&
-                    storageException.RequestInformation.ExtendedErrorInformation.ErrorCode == "InvalidQueryParameterValue")
+                if (storageException?.RequestInformation?.HttpStatusCode == 400 &&
+                    storageException?.RequestInformation?.ExtendedErrorInformation?.ErrorCode == "InvalidQueryParameterValue")
                 {
                     // Premium storage accounts do not support the GetServicePropertiesAsync call, and respond with a 400 'InvalidQueryParameterValue'.
                     // If we see this error response classify the account as a premium account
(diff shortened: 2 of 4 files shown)


src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/QueueBindingProvider.cs
```diff
@@ -187,7 +187,7 @@ private IAsyncCollector<IStorageQueueMessage> BuildFromQueueAttribute(QueueAttri
 
             internal IStorageQueue GetQueue(QueueAttribute attrResolved)
             {
-                var account = Task.Run(() => this._accountProvider.GetStorageAccountAsync(attrResolved, CancellationToken.None)).GetAwaiter().GetResult();
```
paulbatum: Any idea why the old version of this code was doing a Task.Run(..).GetResult()? Seems like a weird pattern.

mamaso:
@paulbatum I'll have a think on E2E - it'd be more valuable to ensure that no binding is leaking connections as opposed to just this specific scenario

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/QueueBindingProvider.cs
mamaso: I don't think there's a good reason.

Task.Run is for cpu bound tasks to post to another thread, but as we're `.Result`ing it's a synchronous blocking call anyway.

paulbatum:
I seem to recall us discussing that our unbounded ServicePointManager.DefaultConnectionLimit might also be contributing to this issue.. did you investigate that any further?

mamaso:
@paulbatum the DefaultConnectionLimit isn't necessarily a bad thing - as long as we don't hit connection exhaustion. If we lower it, we'll get more request queueing but won't be as bursty with connections.

Good summary here, see the pipelining section: https://blogs.msdn.microsoft.com/adarshk/2005/01/02/understanding-system-net-connection-management-and-servicepointmanager/

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/QueueBindingProvider.cs
MikeStall: This is the correct way to run await an async task from an sync method and avoid deadlocks. 

---
In reply to: [122791012](https://github.com/Azure/azure-webjobs-sdk/pull/1190#discussion_r122791012) [](ancestors = 122791012)

MikeStall: Removing Task.Run is incorrect and can lead to deadlocks. 

---
In reply to: [122810016](https://github.com/Azure/azure-webjobs-sdk/pull/1190#discussion_r122810016) [](ancestors = 122810016)

MikeStall:
:clock1:

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageAccountProvider.cs
```diff
@@ -15,6 +16,7 @@ internal class DefaultStorageAccountProvider : IStorageAccountProvider
         private readonly IStorageCredentialsValidator _storageCredentialsValidator;
         private readonly IStorageAccountParser _storageAccountParser;
         private readonly IServiceProvider _services;
+        private readonly ConcurrentDictionary<string, Task<IStorageAccount>> _accounts = new ConcurrentDictionary<string, Task<IStorageAccount>>();
```
mathewc: Why is this a dictionary mapping strings to tasks? It should be strings to IStorageAccount.

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageAccountProvider.cs
```diff
@@ -187,6 +188,11 @@ private IStorageAccount StorageAccount
             return account;
         }
 
+        public Task<IStorageAccount> TryGetAccountAsync(string connectionStringName, CancellationToken cancellationToken)
```
mathewc: Recommend removing this new method, and just doing the normal pattern inside the above existing method of checking cache first and returning early if found.

mamaso: I split them because the name wasn't very informative as to what it was doing, and it seemed to better fit with separation of concerns

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/QueueBindingProvider.cs
mamaso: One issue with this is that this method is called extremely frequently - running all these tasks on the threadpool, most of which will be in a completed state, seems like it could be a negative.

Perhaps the answer is to move storage account validation out of the 'getaccount' method - it's the only reason it's an async call in the first place

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageAccountProvider.cs
mamaso: Intended as Task<IStorageAccount>, because it made for the cleanest change (Dict.GetOrAdd doesn't support async)

Do you see any negatives to Task<IStorageAccount>? Perhaps if you have a faulted task, but as we throw anyway if storage account isn't created I don't see that as an issue.

mathewc: I think caching the Task is incorrect - we want a **value** cache. That's what makes sense conceptually.

MikeStall: +1 for a value cache. Fewer things that can go wrong.

mamaso: The purpose is not to cache values, though that is a side effect - the purpose is to prevent the expensive validate call from occurring more than once. A value cache will not ensure that without extra locking code, which is currently abstracted away behind the ConcurrentDictionary. It's similar to in style to the function loader https://github.com/Azure/azure-webjobs-sdk-script/blob/dev/src/WebJobs.Script/Description/FunctionLoader.cs#L86

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageAccountProvider.cs
mathewc: Ok

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageAccountProvider.cs
mathewc: All right. While it is a little strange to have a cache of completed tasks, it should work ok. I would have just done something a bit more traditional, etc.:

```csharp
public async Task<IStorageAccount> TryGetAccountAsync(string connectionStringName, CancellationToken cancellationToken)
{
    IStorageAccount account;
    if (!_accounts.TryGetValue(connectionStringName, out account))
    {   
        account = await CreateAndValidateAccountAsync(connectionStringName, cancellationToken);
        _accounts.AddOrUpdate(connectionStringName, (cs) => account, (cs, a) => account);
    }
    return account;
}
```

From time to time we might end up validating an account twice with that code, but the value is cached quickly and this solves the issue.

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
```diff
@@ -17,25 +17,14 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
...
 
-            StorageCredentials credentials = account.Credentials;
-
-            // Avoid double-validating the same account and credentials.
-            if (_validatedCredentials.Contains(credentials))
-            {
-                return;
-            }
-
             await ValidateCredentialsAsyncCore(account, cancellationToken);
```
mathewc: now that you're no longer performing additional logic in this method, it seems we can get rid of the "Core" method and move the logic back into this method?

mamaso: Yeah, good point - I wanted to get rid of the validator entirely, but it's a useful layer in testing.

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/QueueBindingProvider.cs
mathewc: While I agree that it might make sense for this to be a sync interface, it has been async from virtually the beginning of the SDK. I think we should track optimizing these Task.Runs away in a separate PR. It's not required for your fix, right?

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageAccountProvider.cs
mamaso: Thanks for the alternative code, very clear as well.

I'd love to have a think over moving the validation out of this code path anyway in a later PR, so we can remove all the async Task.Run in Blob Table & Queue bindings.

I don't know that the validation needs to occur at the first storage call - we could fire it off and correlate back in the event of errors, operating under the default assumption that everything is correct.

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/QueueBindingProvider.cs
mamaso: Yep, out of scope for this one :)

mamaso:
@mathewc @MikeStall Updated, replaced Task.Run call & changed to a value cache. Created issue to investigate moving validation code elsewhere https://github.com/Azure/azure-webjobs-sdk/issues/1197

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/QueueBindingProvider.cs
MikeStall: +1. I think the root of the problem here is that there's some wacky layering problems around when things get validated. 

---
In reply to: [123315937](https://github.com/Azure/azure-webjobs-sdk/pull/1190#discussion_r123315937) [](ancestors = 123315937)

PR #1191, opened 2017-06-19
fixing queue visibility renewals

brettsam:
Reducing the number of places that we do our property updates to get 8.x working. I missed a scenario where we do a background message update, which changes the pop receipt. That caused future deletes to break.

Instead, taking @mathewc's advice and reducing the number of places we replace -- ensuring that the common scenario of poison-queue-handling works in 8.x as expected. If you are adding existing messages to queues from within your function in 8.x, we'll no longer be able to successfully delete that message, but that is manageable for the end-user.

### DIFF
--- a/test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Queues/QueueListenerTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Queues/QueueListenerTests.cs
@@ -4,12 +4,10 @@
 
 using System;
 using System.Diagnostics;
-using System.Runtime.ExceptionServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host.Executors;
 using Microsoft.Azure.WebJobs.Host.Listeners;
-using Microsoft.Azure.WebJobs.Host.Queues;
 using Microsoft.Azure.WebJobs.Host.Queues.Listeners;
 using Microsoft.Azure.WebJobs.Host.Storage.Queue;
 using Microsoft.Azure.WebJobs.Host.TestCommon;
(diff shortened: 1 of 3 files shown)


test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Queues/QueueListenerTests.cs
```diff
@@ -57,26 +57,23 @@ public QueueListenerTests(TestFixture fixture)
...
             CloudQueueMessage messageFromCloud = await queue.GetMessageAsync();
 
-            TestQueueProcessor processor = new TestQueueProcessor(context);
-            processor.Original = messageFromCloud;
-
             QueueListener listener = new QueueListener(storageQueue, storagePoisonQueue, mockTriggerExecutor.Object, mockExceptionDispatcher.Object, trace,
-                null, null, queuesConfig, processor);
+                null, null, queuesConfig);
 
             // Set up a function that will put the message in another queue.
```
mathewc: this comment no longer applies?

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Queues/QueueListenerTests.cs
```diff
@@ -86,99 +83,65 @@ public QueueListenerTests(TestFixture fixture)
...
-
             await listener.ProcessMessageAsync(new StorageQueueMessage(messageFromCloud), TimeSpan.FromMinutes(10), CancellationToken.None);
-        }
 
-        private class TestQueueProcessor : QueueProcessor
-        {
-            private CloudQueue _queue;
-            private CloudQueue _poisonQueue;
-            private CloudQueueMessage _original;
+            // Make sure the message was processed and deleted.
```
mathewc: Won't this be the case w/o your changes? I thought the issue only happened when users force upgraded storage versions?

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Queues/QueueListenerTests.cs
```diff
@@ -86,99 +83,65 @@ public QueueListenerTests(TestFixture fixture)
...
-                await base.CopyMessageToPoisonQueueAsync(message, poisonQueue, cancellationToken);
-            }
+            QueueListener listener = new QueueListener(storageQueue, null, mockTriggerExecutor.Object, mockExceptionDispatcher.Object, trace,
+                null, null, queuesConfig);
 
-            protected override async Task ReleaseMessageAsync(CloudQueueMessage message, FunctionResult result, TimeSpan visibilityTimeout, CancellationToken cancellationToken)
-            {
-                ValidateMessage(_original, message);
-                UpdateMessage(message);
+            // Set up a function that waits 8 seconds (to allow renewal)
```
mathewc: Hmmm, ideally we wouldn't have a test that sleeps for 8 seconds. Maybe we don't even need this test. It seems below like you're mocking too much out anyways with the visibility renewal timer.

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Queues/QueueListenerTests.cs
brettsam: Oops, yeah.

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Queues/QueueListenerTests.cs
brettsam: Yes, for this test, that worked before as well. I just wanted a better validation so added it. For the other test (the renewal) -- it affects people using 7.x. We should try to get a release out as soon as this is in.

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Queues/QueueListenerTests.cs
brettsam: I was fixing this up while you reviewed -- it still sleeps (for 4 secs) but I don't mock the renewal anymore. I let the real thing run. I had to expose a new property for testing. The SingletonManager has a similar pattern.

brettsam:
Fixes https://github.com/Azure/azure-webjobs-sdk-script/issues/1600

Issue #1177, opened 2017-06-05
.NET Core migration


fabiocav:
Opening this issue to track the work and discussions around the .NET Core/.NET Standard migration.

The current plan is for the WebJobs SDK to target **.NET Standard 2.0 only.** 

Work is currently in progress and will move through the various repositories, ultimately moving into the Azure Functions runtime (WebJobs Script): WebJobs SDK -> WebJobs extensions -> WebJobs Script.

fabiocav:
One open issue with Service Bus is the fact that the management capabilities have been removed from the SDK and the recommendation is to use ARM instead. Because of the added complexity this introduces (service principal configuration, etc.), the plan will likely be to drop support for automatic entity creation (auto creation of queues, topics, etc.)

/cc @mathewc @MikeStall

fabiocav:
Another item worth mentioning is that we'll have a gap on the supported binding types for table storage as their latest SDK does not support `IQueryable` and we'll wait for them to bring that support back instead of going down the path of closing that gap here.

funkyOne:
Hi guys, thanks for this, really happy the work is being done! Do you have any idea when a beta/preview version will be available on nuget? At least to understand time scale - month, months?

This is the last library holding us on full framework.

fabiocav:
The initial migration work tracked by this issue has been completed. Extensions and Functions Runtime will be tracked in their respective repositories.

Beta packages targeting .NET Standard 2.0 have been released on NuGet (version 3.x) for those wishing to run the new bits.

chris31389:
Hi, now that .net standard has been released is there an estimated time until 3.0 is released fully? Is there any examples you can point to?

fabiocav:
We'll release bits targeting the .NET Standard 2.0 RTM bits, but don't have an ETA on a stable 3.0 release at this point. There is some work that would need to go into 3.0 that is currently in progress.

@christopheranderson , can we put something up (label, milestone, project a tracking issue...) to help folks track the progress on this?

atrauzzi:
Is there a sample project using the `3.0.0-beta` release that you guys might be able to put up somewhere?

We're looking to get started with webjobs in a .net Core project and obviously would like to have it cross platform.

fabiocav:
We have the basic sample here: https://github.com/Azure/azure-webjobs-sdk/tree/dev/sample/SampleHost

Most of the concepts coming from 2.x still apply, so the existing documentation is relevant.

[...]

PR #1193, opened 2017-06-20
Initial migration to .NET Standard 2.0 (.NET Core support)

fabiocav:
Initial work to migrate to .NET Standard 2.0 ( part of #1177 )

Opening this to start iterating and gathering feedback.

Targeting the `netstandard2` branch so we have this merged sooner and iterate on the branch.

This work is **still in progress**, the Service Bus project has not been migrated (as there were significant API changes with loss of functionality).

The dashboard project has been removed from the main solution and will *not be* retargeted/migrated

Items to complete:

- [x] - Complete build setup (AppVeyor)
- [x] - Address flagged issues marked with `// TODO:`
- [ ] - ~~Ensure signing infrastructure works with the new process/bits~~ (created issue #1223 to track this separately.)


src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/DefaultTelemetryClientFactory.cs
```diff
@@ -128,23 +144,19 @@ internal TelemetryConfiguration InitializeConfiguration()
...
-
-        /// <summary>
-        /// Creates the <see cref="QuickPulseTelemetryModule"/> to be used by the <see cref="TelemetryClient"/>.
-        /// </summary>
-        /// <returns>The <see cref="QuickPulseTelemetryModule"/>.</returns>
-        protected virtual QuickPulseTelemetryModule CreateQuickPulseTelemetryModule()
-        {
-            return new QuickPulseTelemetryModule();
+#else
+            throw new NotSupportedException();
```
brettsam: Assuming it exists, you should return an `InMemoryChannel` here, which is the default.

src/Microsoft.Azure.WebJobs.Logging/WebJobs.Logging.csproj
```diff
@@ -1,150 +1,29 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
brettsam: I *think* this Logging project is mostly for the Dashboard. I wonder if we can get rid of it? Probably not necessary in this first wave, but something to look into.

MikeStall:
> #if !NETSTANDARD2_0

Are the #ifdefs temporary or do we expect to produce multiple builds going forward?

---
Refers to: src/Microsoft.Azure.WebJobs.Host/AzureStorageDeploymentValidator.cs:39 in 0dea24c. [](commit_id = 0dea24cd22e2a71b75366993ced1660ca83d1c31, deletion_comment = False)

MikeStall:
>                 TableResult retrievedResult = await instanceTable.SafeExecuteAsync(retrieveOperation);

Why are you switching from await to continueWith?  This has different error semantics. 

---
Refers to: src/Microsoft.Azure.WebJobs.Logging/Internal/LogReader.cs:109 in 0dea24c. [](commit_id = 0dea24cd22e2a71b75366993ced1660ca83d1c31, deletion_comment = True)

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/DefaultTelemetryClientFactory.cs
fabiocav: Good to know. The AI stuff is the key area I was hoping you'd get your eyes on. I'll update this.

fabiocav:
@MikeStall, I believe you're using CodeFlow, so I can't reply to your comments inline:

1. Yes, for the most part, I was keeping the conditional symbols because I was running some tests targeting net46. That's no longer the case and I'll be removing those
2. Fixing this. Not really sure why I ended up with this version as it shouldn't have been the case.

Does CF support source comments? Would be easier to track those.

src/Microsoft.Azure.WebJobs.Logging/WebJobs.Logging.csproj
fabiocav: I wondered. I can try to remove it, but thought some of the other projects had dependencies on this.

wind-rider:
Great to see that porting started!

Question - can you detail a bit on the reason(s) why the dashboard won't be ported and what the implications of that are for users?

fabiocav:
@wind-rider the existing dashboard will likely be moved to its own repository and was not in scope for this work. It will continue to exist and support current functionality for the foreseeable future, but future investments in logging and monitoring will be focused on the integration with Application Insights.

fabiocav:
@mamaso we're at a point where we need to test the signing process. I'm publishing the artifacts, but we need to finish the loop.

src/Dashboard/ApiControllers/DiagnosticsController.cs
```diff
@@ -1,117 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the MIT License. See License.txt in the project root for license information.
-
-using System;
-using System.Web.Http;
```
mathewc: So you're deleting all this code? Why not instead just remove it from the sln and leave in this repo? Until we decide to move it out to a different repo (if we do).

src/Microsoft.Azure.WebJobs.Host/Tables/TableExtension.cs
```diff
@@ -306,43 +305,6 @@ public JObjectBuilder(TableExtension bindingProvider)
             }
         }
 
-        // IStorageTable --> IQueryable<T>
-        // ConverterManager's pattern matcher will figure out TElement. 
-        private class TableToIQueryableConverter<TElement> :
```
mathewc: Can you explain again why we can no longer support IQueryable? What if we wanted to retain this capability - how hard would it be?

src/Microsoft.Azure.WebJobs.Storage/Table/StorageTable.cs
```diff
@@ -69,17 +69,6 @@ public IStorageTableOperation CreateInsertOrReplaceOperation(ITableEntity entity
         }
 
         /// <inheritdoc />
-        public IQueryable<TElement> CreateQuery<TElement>() where TElement : ITableEntity, new()
-        {
-            // TODO: FACAVAL - Find a suitable replacement
```
mathewc: So there is no suitable replacement?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -321,7 +321,7 @@ internal static string GetIpAddress(HttpRequestMessage httpRequest)
             HttpContextBase context = httpRequest.Properties.GetValueOrDefault<HttpContextBase>(ApplicationInsightsScopeKeys.HttpContext);
             return context?.Request?.UserHostAddress ?? LoggingConstants.ZeroIpAddress;
 #else
-            // TODO: FACAVAL
+            // TODO: FACAVAL - Need a proper implementation here.
```
mathewc: So what is the timeline/plan for addressing all the TODOs throughout the code?

src/Microsoft.Azure.WebJobs.Protocols/HeartbeatMonitor.cs
```diff
@@ -16,14 +17,13 @@ public class HeartbeatMonitor : IHeartbeatMonitor
...
         /// <param name="client">A blob client for the storage account in which to monitor heartbeats.</param>
-        
         public HeartbeatMonitor(CloudBlobClient client)
         {
             _client = client;
         }
 
         /// <inheritdoc />
-        public DateTimeOffset? GetSharedHeartbeatExpiration(string sharedContainerName, string sharedDirectoryName,
+        public async Task<DateTimeOffset?> GetSharedHeartbeatExpirationAsync(string sharedContainerName, string sharedDirectoryName,
```
mathewc: The Dashboard references these protocol shared types. If we've removed all the Dashboard code it will now diverge. Since we need to continue to support the Dashboard for the foreseeable future, how can we handle this?

src/Dashboard/ApiControllers/DiagnosticsController.cs
fabiocav: I could, but this would also mean backporting issues over to the 2.x branch. This may become tricky to do in the current branch because of how the code dependency is setup, which will become more challenging as the code diverges.

src/Microsoft.Azure.WebJobs.Host/Tables/TableExtension.cs
fabiocav: The storage SDK does not support `IQueryable` on the version used here. If we wanted to support it, we'd have to provide our own implementation, which I would advise against (that's also the feedback I got from them and other teams). I believe this is a _point in time_ issue and `IQueryable` support will eventually land there.

src/Microsoft.Azure.WebJobs.Storage/Table/StorageTable.cs
fabiocav: No. Delete the core with recent commits.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
fabiocav: Working on that. Have addressed the majority of them and should have the remaining ones done today.

src/Microsoft.Azure.WebJobs.Protocols/HeartbeatMonitor.cs
fabiocav: This would be done in the 2.x branch.

fabiocav:
@mamaso should I create an issue for item 3 (signing infrastructure) so we don't block on it?

mamaso:
@fabiocav sure, sounds good.

Issue #1201, opened 2017-06-22
JobHostConfiguration initialization 


MikeStall:
Fix JobHostConfiguration initialization. 

We need a means to get static services from the JobHostConfig, like so:

```
            JobHostConfiguration config = new JobHostConfiguration();
            var host = new JobHost(config);
            var lockManager = (IDistributedLockManager)host.GetService(typeof(IDistributedLockManager));
```

This is needed for https://github.com/Azure/azure-webjobs-sdk/issues/1183 
Bring-your-own-binding also needs this so that it can access the binding graph without actually spinning up the listeners.

MikeStall:
Added documentation for this: https://github.com/Azure/azure-webjobs-sdk/wiki/Initialization

MikeStall:
As part of this, we should move CreateMetadataProvider() from JobHostConfiguration to JobHost.

PR #1202, opened 2017-06-22
DistributedLockManager Service

MikeStall:
Fix JobHostConfiguration problem, needed so that Script can get access to the new  DistributedLockManager

Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1201

Also includes commit for resolving https://github.com/Azure/azure-webjobs-sdk/issues/1183


src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
```diff
@@ -31,12 +31,9 @@ public sealed class JobHostConfiguration : IServiceProvider
         private readonly JobHostBlobsConfiguration _blobsConfiguration = new JobHostBlobsConfiguration();
         private readonly JobHostTraceConfiguration _traceConfiguration = new JobHostTraceConfiguration();
         private readonly ConcurrentDictionary<Type, object> _services = new ConcurrentDictionary<Type, object>();
-        private readonly JobHostMetadataProvider _metadataProvider;
 
         private string _hostId;
 
-        private ServiceProviderWrapper _partialInitServices;
```
MikeStall: >ServiceProviderWrapper [](start = 16, length = 22)

Removing this field is a major benefit of this change.

src/Microsoft.Azure.WebJobs.Host/JobHost.cs
```diff
@@ -329,38 +335,104 @@ private static IFunctionDefinition ResolveFunctionDefinition(MethodInfo method,
...
+            catch (Exception e)
+            {
+                initializationTask.SetException(e);
+            }
         }
 
-        private void ThrowIfDisposed()
+        // Ensure the static services are initialized. 
+        // These are derived from the underlying JobHostConfiguration. 
+        private void PopulateStaticServices()
```
brettsam: It looks like this can be called from several places -- do we need to lock around this to make sure it's thread-safe?

src/Microsoft.Azure.WebJobs.Host/JobHost.cs
```diff
@@ -329,38 +335,104 @@ private static IFunctionDefinition ResolveFunctionDefinition(MethodInfo method,
...
+            var services = this._config.CreateStaticServices();
+
+            _services = services;
+        }
+
+        /// <summary>
+        /// Creates a <see cref="IJobHostMetadataProvider"/> for this configuration.
+        /// </summary>
+        /// <returns>The <see cref="IJobHostMetadataProvider"/>.</returns>
+        public IJobHostMetadataProvider CreateMetadataProvider()
```
brettsam: Nit: Should this be `GetMetadataProvider()`? To me, 'Create' implies to me that you'll get a new one each time you call it.

src/Microsoft.Azure.WebJobs.Host/JobHost.cs
```diff
@@ -23,7 +24,7 @@ namespace Microsoft.Azure.WebJobs
     /// A <see cref="JobHost"/> is the execution container for jobs. Once started, the
     /// <see cref="JobHost"/> will manage and run job functions when they are triggered.
     /// </summary>
-    public class JobHost : IDisposable
+    public class JobHost : IDisposable, IServiceProvider
```
fabiocav: Why are we making the `JobHost` an `IServiceProvider`? Based on our chat, this shouldn't be required to address the existing requirements for on-prem, and conflicts with the DI design plans we've discussed. 
`JobHost` is already doing more than it should as-is, it would be nice to avoid taking on this additional responsibility here. 

Can we leverage the existing model/support until we have the proper DI implementation in place?

mathewc: Yeah, we definitely shouldn't do this - it commits us to a path that we don't want to go down, and we'll be changing relatively soon anyways when we rework DI on core. We don't want 2 service providers (JobHostConfiguration) is already a SP.

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
```diff
@@ -103,7 +90,9 @@ public static ServiceProviderWrapper CreateStaticServices(this JobHostConfigurat
             services.AddService<TraceWriter>(trace);
 
             // Add built-in extensions 
-            config.AddAttributesFromAssembly(typeof(TableAttribute).Assembly);
+            var metadataProvider = new JobHostMetadataProvider();
```
mathewc: Can these 2 metadata provider lines be moved down below where the metadata provider service is added? Or do they need to be here?

src/Microsoft.Azure.WebJobs.Host/JobHost.cs
```diff
@@ -35,13 +36,18 @@ public class JobHost : IDisposable
...
         private JobHostContext _context;
         private IListener _listener;
-        private object _contextLock = new object();
+
+        // Null if we haven't yet started runtime initialization.
+        // Points to an incomplete task during initialization. 
+        // Points to a completed task after initialization. 
+        private Task _initializationRunning = null;
+
+        // These are serives that are accessible without starting the execution container.
```
mathewc: "services"

src/Microsoft.Azure.WebJobs.Host/JobHost.cs
```diff
@@ -329,37 +335,92 @@ private static IFunctionDefinition ResolveFunctionDefinition(MethodInfo method,
...
+        {
+            if (_context != null)
             {
-                if (_context == null)
+                return Task.CompletedTask;
+            }
+
+            TaskCompletionSource<bool> tsc = null;
+
+            lock (_stopTaskLock)
```
mathewc: This lock object is for the "stop" operation/function. Why are you locking on it here?

src/Microsoft.Azure.WebJobs.Host/JobHost.cs
```diff
@@ -329,37 +335,92 @@ private static IFunctionDefinition ResolveFunctionDefinition(MethodInfo method,
...
+        private void PopulateStaticServices()
         {
-            if (_disposed)
+            if (this._services != null)
             {
-                throw new ObjectDisposedException(null);
+                return; // already Created 
+            }
+
+            // var services = new ServiceProviderWrapper(this._config);
```
mathewc: remove commented out code

src/Microsoft.Azure.WebJobs.Host/JobHost.cs
```diff
@@ -329,37 +335,92 @@ private static IFunctionDefinition ResolveFunctionDefinition(MethodInfo method,
...
 
-            return _context;
+            return _initializationRunning;
         }
 
-        private Task EnsureHostStartedAsync(CancellationToken cancellationToken)
+        // Caller gaurantees this is single-threaded. 
+        // Set initializationTask when complete, many threads can wait on that. 
+        // When complete, the fields should be initialized to allow runtime usage. 
+        private async Task RuntimeInitAsync(CancellationToken cancellationToken, TaskCompletionSource<bool> initializationTask)
```
mathewc: nit: InitializeServicesAsync?

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLock.cs
```diff
@@ -0,0 +1,21 @@
...
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// Handle for a lock returned by <see cref="IDistributedLockManager"/>
+    /// The SDK will call call <see cref="IDistributedLockManager.RenewAsync(IDistributedLock, CancellationToken)"/> to renew the release.
```
mathewc: Please remove the second comment line from this summary block. Those details don't belong in this general interface description.

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLockManager.cs
```diff
@@ -0,0 +1,76 @@
...
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Executors;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// Manage distributed lock. A lock is specified by (account, lockId).  This implementation should cooperate with <see cref="SingletonConfiguration"/>
```
mathewc: Please fixup these summary comments. E.g. you're referring to "this implementation" but this is an interface, not an implementation. Also, shouldn't reference SingletonConfiguration here.

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLockManager.cs
```diff
@@ -0,0 +1,76 @@
...
+    /// </summary>
+    /// <remarks>
+    /// The default implementation of this is based on blob leases.
```
mathewc: I wouldn't reference blobs, or any restrictions here. this is just the interface. Implementation details are owned by implementors.

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLockManager.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// <param name="account">optional. A string specifying the account to use. LockIds are scoped to an account </param>
+        /// <param name="lockId">the name of the lock. </param>
+        /// <param name="lockOwnerId">a string hint specifying who owns this lock. Only used for diagnostics. </param>
+        /// <param name="proposedLeaseId">optional. This can allow the caller to immediately assume the lease.</param>
+        /// <param name="lockPeriod">The length of the lease to acquire. The caller is responsible for Renewing the lease well before this time is up. 
+        /// The exact value here is restricted based on the underlying implementation.  </param>
+        /// <param name="cancellationToken"></param>
+        /// <returns>null if can't acquire the lock. This is common if somebody else holds it.</returns>
+        Task<IDistributedLock> TryLockAsync(
+            string account,
```
mathewc: Account is an implementation specific thing, mapping to a storage account, or sql account?

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLockManager.cs
```diff
@@ -0,0 +1,76 @@
...
+    /// 1. Account can be null or it may be a storage account name intended for <see cref="IStorageAccountProvider"/>. 
+    /// 2. lockId has the same naming restrictions as blobs.     
+    /// </remarks>    
+    public interface IDistributedLockManager
+    {        
+        /// <summary>
+        /// Try to acquire a lock specified by (account, lockId).                 
+        /// </summary>
+        /// <param name="account">optional. A string specifying the account to use. LockIds are scoped to an account </param>
+        /// <param name="lockId">the name of the lock. </param>
```
mathewc: Nit: fix formatting/capitalization throughout

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLockManager.cs
```diff
@@ -0,0 +1,76 @@
...
+        Task<IDistributedLock> TryLockAsync(
+            string account,
+            string lockId, 
+            string lockOwnerId,
+            string proposedLeaseId,
+            TimeSpan lockPeriod,  
+            CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Called by the client to renew the lease. The timing internals here are determined by <see cref="SingletonConfiguration"/>
```
mathewc: Remove references to specific implementation SingletonConfiguration, etc. throughout

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLockManager.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// </summary>
+        /// <param name="lockHandle"></param>
+        /// <param name="cancellationToken"></param>
+        /// <returns>A hint used for determining the next time delay in calling Renew.
```
mathewc: Can you simplify the contract here? E.g. return true if the lease was renewed successfully, false otherwise? Having a bool return that indicates a delay hint is strange.

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLockManager.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// <returns>A hint used for determining the next time delay in calling Renew. 
+        /// True means the next execution should occur at a normal delay. False means the next execution should occur quickly; use this in network error cases.   </returns>
+        /// <remarks>
+        /// If this throws an exception, the lease is cancelled. 
+        /// </remarks>
+        Task<bool> RenewAsync(IDistributedLock lockHandle, CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Get the owner for a given lock or null if not held. 
+        /// This is used for diagnostics. The lock owner can change immediately after this function returned, so callers
```
mathewc: fix sentence

src/Microsoft.Azure.WebJobs.Host/Singleton/LockHandle.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.WebJobs.Host.Timers;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// Internal lock handle returned by <see cref="SingletonManager"/>
```
mathewc: Should this be named SingletonLockHandle then if it is specific to singleton?

src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonConfiguration.cs
```diff
@@ -8,6 +8,7 @@ namespace Microsoft.Azure.WebJobs.Host
 {
     /// <summary>
     /// Configuration options governing the lock functionality of <see cref="SingletonAttribute"/>.
+    /// These results are passed to <see cref="IDistributedLockManager"/>
```
mathewc: Can you remove this comment - it doesn't make any sense to me.

src/Microsoft.Azure.WebJobs.Host/JobHost.cs
MikeStall: Moved it to a property.

---
In reply to: [124319474](https://github.com/Azure/azure-webjobs-sdk/pull/1202#discussion_r124319474) [](ancestors = 124319474)

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLockManager.cs
MikeStall: I don't think I can - I needed to preserve the existing behavior, which involves a non-trivial backoff strategy.  And there's not a good way to factor this. 

---
In reply to: [126262306](https://github.com/Azure/azure-webjobs-sdk/pull/1202#discussion_r126262306) [](ancestors = 126262306)

src/Microsoft.Azure.WebJobs.Host/JobHost.cs
MikeStall: I'll just remove it entirely and get from services. 

---
In reply to: [124302433](https://github.com/Azure/azure-webjobs-sdk/pull/1202#discussion_r124302433) [](ancestors = 124302433)

src/Microsoft.Azure.WebJobs.Host/JobHost.cs
```diff
@@ -35,13 +36,18 @@ public class JobHost : IDisposable
         private readonly WebJobsShutdownWatcher _shutdownWatcher;
         private readonly CancellationTokenSource _stoppingTokenSource;
 
-        private Task<JobHostContext> _contextTask;
-        private bool _contextTaskInitialized;
-        private object _contextTaskLock = new object();
```
MikeStall: These 3 fields (and the one below)  get deleted as part of the simplicaiton. .

src/Microsoft.Azure.WebJobs.Host/JobHost.cs
MikeStall: This class should only have 1 lock.  I'll rename the field.  
And the logic before was pretty convoluted - I remove 4 state fields 

---
In reply to: [126255344](https://github.com/Azure/azure-webjobs-sdk/pull/1202#discussion_r126255344) [](ancestors = 126255344)

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLockManager.cs
MikeStall: I updated the comment. 

---
In reply to: [126262028](https://github.com/Azure/azure-webjobs-sdk/pull/1202#discussion_r126262028) [](ancestors = 126262028)

src/Microsoft.Azure.WebJobs.Host/Singleton/IDistributedLockManager.cs
MikeStall: I think this is important information for the implementor. The implementation needs to cooperate with SingletonConfiguration. 

---
In reply to: [126262129](https://github.com/Azure/azure-webjobs-sdk/pull/1202#discussion_r126262129) [](ancestors = 126262129)

src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonConfiguration.cs
MikeStall: I'll clarify it. 

---
In reply to: [126262845](https://github.com/Azure/azure-webjobs-sdk/pull/1202#discussion_r126262845) [](ancestors = 126262845)

src/Microsoft.Azure.WebJobs.Host/Singleton/LockHandle.cs
MikeStall: I'll rename. It's internal. 

---
In reply to: [126262728](https://github.com/Azure/azure-webjobs-sdk/pull/1202#discussion_r126262728) [](ancestors = 126262728)

src/Microsoft.Azure.WebJobs.Host/JobHost.cs
MikeStall: We have  2 categories of services: I want to distinguish between the phase 1 (Static) vs. phase 2(runtime). 

---
In reply to: [126260074](https://github.com/Azure/azure-webjobs-sdk/pull/1202#discussion_r126260074) [](ancestors = 126260074)

src/Microsoft.Azure.WebJobs.Host/JobHost.cs
MikeStall: the caller must ensure this is single-threaded. 

---
In reply to: [124301870](https://github.com/Azure/azure-webjobs-sdk/pull/1202#discussion_r124301870) [](ancestors = 124301870)

Issue #1170, opened 2017-06-02
SDK should Honor FunctionName attribute 


MikeStall:
We added a FunctionName attribute https://github.com/Azure/azure-webjobs-sdk/pull/1075 
It's only a placeholder used by VS tooling, but SDK should honor this attribute too and plumb it through. 

1. Function indexing 
2. {sys.MethodName} 
3. JobHost.Call?    (maybe moot since that uses MethodInfo, but double check) 
4. Dashboard 
5. New InvokeFunctionFilterAttribute attribute that specifies function names

paulbatum:
sys.MethodName should be the actual name of the method, it shouldn't be overridden by the functionname attribute

MikeStall:
@paulbatum  -  I'd disagree.  The actual C# method name should be viewed as an implementation detail - we don't want things taking a direct dependency on that. VS Tooling includes [FunctionName] attributes by default and that's what they show up in Script.  For example, Bot uses {sys.MethodName} and expects to get the Script function name which would be [FunctionName].

fabiocav:
I agree that {sys.MethodName} should reflect the method name as opposed to the logical function name. If we want to expose a way to get the function name, we should probably introduce a {sys.FunctionName} instead, as passing the function name when binding to the method would be surprising/confusing.

MikeStall:
@fabiocav  - please see my comments above. Exposing the reflection name would actually be incorrect; and there's not a good reason to add that low-level functionality.

fabiocav:
I agree, but I'd argue that, given its name today, that's what people would expect. We don't refer to the function name as a method name anywhere. Introducing a FunctionName and perhaps deprecating MethodName would probably be the right approach.

MikeStall:
This is recently new surface area. If it's a naming issue, we can do a full rename MethodName --> FunctionName, and we just need to update BOT.  But we shouldn't have anything handing out the underlying Reflection name.

paulbatum:
I'm fine with that, assuming we never disclosed sys.methodname publicly and therefore not a breaking change

MikeStall:
It was added here: https://github.com/Azure/azure-webjobs-sdk/commit/962dfa6ff28fd14f960cc948bc3a40e420f5713e

PR #1207, opened 2017-06-27
Function Invocation Filters

hhawkins:
For feature tracked here: #980. Sending out another pull request after making updates and changes from feedback from the first pull request #1189


src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -608,7 +613,8 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
         }
 
         internal static async Task InvokeAsync(IFunctionInvoker invoker, object[] invokeParameters, CancellationTokenSource timeoutTokenSource,
-            CancellationTokenSource functionCancellationTokenSource, bool throwOnTimeout, TimeSpan timerInterval, IFunctionInstance instance)
+            CancellationTokenSource functionCancellationTokenSource, bool throwOnTimeout, TimeSpan timerInterval, IFunctionInstance instance, ILogger logger = null,
+            JobHost jobHost = null, JobHostConfiguration config = null, IReadOnlyList<string> parameterNames = null)
```
mathewc: nit: can you reorder the parameterNames parameter next to the invokeParameters param, before it?

src/Microsoft.Azure.WebJobs.Host/FunctionInvocationContext.cs
```diff
@@ -0,0 +1,63 @@
...
+        internal FunctionInvocationContext(Guid id, string name, IReadOnlyDictionary<string, object> arguments, ILogger logger)
+        {
+            Id = id;
+            Name = name;
+            Arguments = arguments;
+            Logger = logger;
+        }
+
+        /// <summary>Gets or sets the ID of the function.</summary>
+        public Guid Id { get; set; }
```
MikeStall: >Id [](start = 20, length = 2)

Is there a more descriptive term than just Id? (InstanceId? FunctionInstanceId?)

src/Microsoft.Azure.WebJobs.Host/FunctionInvocationContext.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// <summary>Gets or sets the name of the function.</summary>
+        public string Name { get; set; }
+        
+        /// <summary>
+        /// Arguments from the function
+        /// </summary>
+        public IReadOnlyDictionary<string, object> Arguments { get; set; }
+
+        /// <summary>
+        /// User properties
```
MikeStall: >/// User properties [](start = 8, length = 19)

What's the difference between Properties and Arguments?

src/Microsoft.Azure.WebJobs.Host/FunctionInvocationContext.cs
```diff
@@ -0,0 +1,63 @@
...
+            Id = id;
+            Name = name;
+            Arguments = arguments;
+            Logger = logger;
+        }
+
+        /// <summary>Gets or sets the ID of the function.</summary>
+        public Guid Id { get; set; }
+
+        /// <summary>Gets or sets the name of the function.</summary>
```
MikeStall: > or sets [](start = 25, length = 8)

set?

src/Microsoft.Azure.WebJobs.Host/FunctionInvocationContext.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// <param name="logger"></param>
+        internal FunctionInvocationContext(Guid id, string name, IReadOnlyDictionary<string, object> arguments, ILogger logger)
+        {
+            Id = id;
+            Name = name;
+            Arguments = arguments;
+            Logger = logger;
+        }
+
+        /// <summary>Gets or sets the ID of the function.</summary>
```
MikeStall: >or sets [](start = 26, length = 7)

set?

src/Microsoft.Azure.WebJobs.Host/FunctionExecutingContext.cs
```diff
@@ -0,0 +1,30 @@
...
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using Microsoft.Azure.WebJobs.Host.Executors;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// The context for an executed function
```
MikeStall: > The context for an executed function [](start = 7, length = 37)

The context describing a function that's about to be executed.

src/Microsoft.Azure.WebJobs.Host/FunctionExecutingContext.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// </summary>
+    [CLSCompliant(false)]
```
MikeStall: >[CLSCompliant(false)] [](start = 4, length = 21)

What are these for?

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,122 @@
...
+            if (!string.IsNullOrEmpty(ExecutingFilter))
+            {
+                executingContext.Logger.LogInformation("Executing Function Filter '" + ExecutingFilter + "'");
+
+                try
+                {
+                    await InvokeJobFunctionAsync(executingContext, cancellationToken);
+                }
+                catch (Exception e)
+                {
```
MikeStall: >{ [](start = 16, length = 1)

We swallow the error - should this abort the call?

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,122 @@
...
+            MethodInfo methodInfo = null;
+
+            foreach (var type in context.Config.TypeLocator.GetTypes())
+            {
+                methodInfo = type.GetMethod(ExecutingFilter);
+                if (methodInfo != null)
+                {
+                    break;
+                }
+            }
```
MikeStall: This is related to our [FunctionName] support in [https://github.com/Azure/azure-webjobs-sdk/issues/1170](https://github.com/Azure/azure-webjobs-sdk/issues/1170) 
This will need to be fixed when address that bug.

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,122 @@
...
+        public InvokeFunctionFilterAttribute(string executingFilter = null, string executedFilter = null)
+        {
+            ExecutingFilter = executingFilter;
+            ExecutedFilter = executedFilter;
+        }
+
+        /// <summary>
+        /// Executing filter
+        /// </summary>
+        public string ExecutingFilter { get; }
```
MikeStall: >ExecutingFilter [](start = 22, length = 15)

What is the format for this function?  We should think about this in context of other function name usages (See  [https://github.com/Azure/azure-webjobs-sdk/issues/1170](https://github.com/Azure/azure-webjobs-sdk/issues/1170) )

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,122 @@
...
+                methodInfo = type.GetMethod(ExecutingFilter);
+                if (methodInfo != null)
+                {
+                    break;
+                }
+            }
+
+            IDictionary<string, object> invokeArguments = new Dictionary<string, object>();
+            string parameterName = methodInfo.GetParameters().SingleOrDefault(p => p.ParameterType == typeof(TContext)).Name;
+            invokeArguments.Add(parameterName, context);
```
MikeStall: >invokeArguments.Add(parameterName, context); [](start = 12, length = 44)

need a null check on parameterNAme. The target function may not have  a context.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -633,7 +641,105 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+                parameters.Add(parameterNames.ToArray<string>()[i], invokeParameters[i]);
+            }
+
+            if (invokeFilters != null)
+            {
+                foreach (var filter in invokeFilters)
+                {
+                    try
+                    {
+                        if (filter.GetType() == typeof(InvokeFunctionFilterAttribute))
```
mathewc: Shouldn't need this special type casing here. Since the JobHost/Config properties are internal, I think it's fine for them to always be set, which would allow you to compress this code.

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,122 @@
...
+            if (!string.IsNullOrEmpty(ExecutedFilter))
+            {
+                executedContext.Logger.LogInformation("Executing Function Filter '" + ExecutedFilter + "'");
+
+                try
+                {
+                    await InvokeJobFunctionAsync(executedContext, cancellationToken);
+                }
+                catch (Exception e)
+                {
```
MikeStall: >{ [](start = 16, length = 1)

swallow?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -633,7 +641,105 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+        {
+            MethodInfo functionMethod = instance.FunctionDescriptor.Method;
+            CancellationToken cancellationToken = functionCancellationTokenSource.Token;
+            var invokeFilters = functionMethod.GetCustomAttributes().OfType<InvocationFilterAttribute>();
+
+            var parameters = new Dictionary<string, object>();
+
+            for (var i = 0; i < invokeParameters.Length; i++)
+            {
+                parameters.Add(parameterNames.ToArray<string>()[i], invokeParameters[i]);
```
mathewc: Don't do the ToArray in the loop like this - it gets executed on every iteration. You can just do parameterNames[i]

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,122 @@
...
+            }
+        }
+
+        internal async Task InvokeJobFunctionAsync<TContext>(TContext context, CancellationToken cancellationToken) where TContext : FunctionInvocationContext
+        {
+            MethodInfo methodInfo = null;
+
+            foreach (var type in context.Config.TypeLocator.GetTypes())
+            {
+                methodInfo = type.GetMethod(ExecutingFilter);
```
MikeStall: >ExecutingFilter [](start = 44, length = 15)

Does this need to be a parameter passed in?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
hhawkins: The reason behind putting the "= null" was because when I would build it, certain functions in other files would call the function and not provide the new parameters thus causing errors. I can remove the "= null" but then I'd have to pass the new parameters in other places in the project where it does call it. I'd also have to modify other tests.

src/Microsoft.Azure.WebJobs.Host/JobHost.cs
```diff
@@ -5,6 +5,7 @@
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
+using System.Linq;
```
MikeStall: remove?

src/Microsoft.Azure.WebJobs.Host/Executors/IJobMethodInvoker.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Host.Executors
+{
+    internal interface IJobMethodInvoker
+    {
+        Task InvokeAsync(MethodInfo methodInfo, IReadOnlyDictionary<string, object> parameters, CancellationToken cancellationToken);
+
+        MethodInfo GetMethodInfoForInvoke(string methodName);
```
MikeStall: >MethodInfo GetMethodInfoForInvoke(string methodName); [](start = 8, length = 53)

Is this used?

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
```diff
@@ -238,7 +238,7 @@ public static ServiceProviderWrapper CreateStaticServices(this JobHostConfigurat
 
                 if (functionExecutor == null)
                 {
-                    functionExecutor = new FunctionExecutor(functionInstanceLogger, functionOutputLogger, exceptionHandler, trace, functionEventCollector, loggerFactory);
+                    functionExecutor = new FunctionExecutor(functionInstanceLogger, functionOutputLogger, exceptionHandler, trace, host, config, functionEventCollector, loggerFactory);
```
MikeStall: >config [](start = 137, length = 6)

Host has the config; so can we just pass host?

src/Microsoft.Azure.WebJobs.Host/FunctionInvocationContext.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// </summary>
+        public ILogger Logger { get; set; }
+
+        /// <summary>
+        /// Gets or sets the method invoker of the JobHost
+        /// </summary>
+        internal JobHost JobHost { get; set; }
+
+        /// <summary>
+        /// Gets or sets the method invoker of the JobHost
```
MikeStall: >the JobHost [](start = 47, length = 11)

typo?

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,122 @@
...
+                    executedContext.Logger.LogInformation(e.ToDetails());
+                }
+            }
+            else
+            {
+                await base.OnExecutedAsync(executedContext, cancellationToken);
+            }
+        }
+
+        internal async Task InvokeJobFunctionAsync<TContext>(TContext context, CancellationToken cancellationToken) where TContext : FunctionInvocationContext
```
MikeStall: >InvokeJobFunctionAsync [](start = 28, length = 22)

Could we move this to FunctionInvocationContext? 
And then FunctionInvocationContext.config, host could be private. (these are dangerous things to be passing around and we want to limit their scope.)

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
MikeStall: One workaround is to have 2 signatures. One with a required parameter (what Mathew wanted), and then one without the parameter that just passes null. The benefit of that is then we know statically which is which. 

---
In reply to: [124422411](https://github.com/Azure/azure-webjobs-sdk/pull/1207#discussion_r124422411) [](ancestors = 124422411)

src/Microsoft.Azure.WebJobs.Host/FunctionInvocationContext.cs
hhawkins: It is the FunctionInstanceId. I will change that.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -633,7 +641,105 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+            }
+
+            if (invokeFilters != null)
```
mathewc: I don't think this will ever be null - please test it out. You might have test holes. If there are no custom attributes of the right type, an empty collection/array is returned I believe, not null. You can then change this to `if (invokeFilters.Any())`

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -633,7 +641,105 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+            }
+        }
+
+        private static async Task InvokeExecutedFilters(IFunctionInvoker invoker, object[] invokeParameters, CancellationTokenSource functionCancellationTokenSource,
+            IFunctionInstance instance, ILogger logger, JobHost jobHost, JobHostConfiguration config,
+            FunctionResult functionResult, IReadOnlyList<string> parameterNames)
+        {
+            MethodInfo functionMethod = instance.FunctionDescriptor.Method;
+            CancellationToken cancellationToken = functionCancellationTokenSource.Token;
+            var invokeFilters = functionMethod.GetCustomAttributes().OfType<InvocationFilterAttribute>();
```
mathewc: Many of the same comments I made above apply in this method as well.

src/Microsoft.Azure.WebJobs.Host/FunctionExecutingContext.cs
hhawkins: Without that identifier, the compiler complains about the function not being CLS compliant

src/Microsoft.Azure.WebJobs.Host/Executors/IJobMethodInvoker.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Host.Executors
```
mathewc: This file can be removed now, right?

src/Microsoft.Azure.WebJobs.Host/FunctionExecutedContext.cs
```diff
@@ -0,0 +1,37 @@
...
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// The context for an executed function
+    /// </summary>
+    [CLSCompliant(false)]
+    public class FunctionExecutedContext : FunctionInvocationContext
+    {
+        /// <summary>
+        /// Constructor to set the context
```
mathewc: Need to fill all this param documentation out, here and througout

src/Microsoft.Azure.WebJobs.Host/FunctionExecutingContext.cs
mathewc: Apply similar wording to your other context classes as well

src/Microsoft.Azure.WebJobs.Host/FunctionExecutingContext.cs
mathewc: This assembly is CLS compliant, but these types are using non CLS compliant constructs, or at least they were. Hamza, if you remove these, what does the compiler say the issue is? I can't remember. Maybe they're not needed anymore.

src/Microsoft.Azure.WebJobs.Host/FunctionInvocationContext.cs
mathewc: Your question is an indication that this summary comment could be better :) This is a property bag that allows users to flow their own custom state through between their executing/executed filters. Standard pattern.

mathewc: However, you only have a getter, and the value will always be null? Likely you don't have any tests yet using this - please add some :)

src/Microsoft.Azure.WebJobs.Host/FunctionInvocationContext.cs
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        /// <summary>Gets or sets the ID of the function.</summary>
+        public Guid Id { get; set; }
+
+        /// <summary>Gets or sets the name of the function.</summary>
+        public string Name { get; set; }
+        
+        /// <summary>
+        /// Arguments from the function
```
mathewc: Gets or sets the function arguments

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
hhawkins: I do understand where you're coming from. The reason why it's separated like that is because the InvokeFunctionFiterAttribute is separate from a normal InocationFilterAttribute in terms of behavior and accessing the JobHost and the Config. A normal InvocationFilterAttribute can be set by the user who's putting a custom filter together, the InvokeFunctionFilterAttribute has a specific job to invoke a function and so the user doesn't have control over how the next function is called.

src/Microsoft.Azure.WebJobs.Host/FunctionExecutingContext.cs
hhawkins: The property type ILogger is non CLS compliant according to the compiler in the main FunctionInvocationContext class

src/Microsoft.Azure.WebJobs.Host/FunctionInvocationContext.cs
```diff
@@ -0,0 +1,63 @@
...
+        /// </summary>
+        public IDictionary<string, object> Properties { get;  }
+
+        /// <summary>
+        /// Gets or sets the logger of the function
+        /// </summary>
+        public ILogger Logger { get; set; }
+
+        /// <summary>
+        /// Gets or sets the method invoker of the JobHost
```
mathewc: Gets or sets the JobHost

src/Microsoft.Azure.WebJobs.Host/FunctionInvocationContext.cs
mathewc: Gets or sets the JobHostConfiguration

src/Microsoft.Azure.WebJobs.Host/InvocationFilterAttribute.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// This is the definition for the function filter attribute
+    /// </summary>
+    [CLSCompliant(false)]
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, Inherited = true, AllowMultiple = false)]
```
mathewc: I don't think this AttributeUsage on the base abstract class has any effect? If not, we can remove it. I think its the usage on subclasses that is honored by the reflection apis

src/Microsoft.Azure.WebJobs.Host/InvocationFilterAttribute.cs
```diff
@@ -0,0 +1,40 @@
...
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// This is the definition for the function filter attribute
```
mathewc: Base class for declarative function invocation filters

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
mathewc: Yes, I was thinking the same thing. Right now he's bee using just simple short names, but in cases where there may be multiple job classes with the same name (collisions), we'll need namespacing. I was thinking we should support both short and long names - e.g. if the name is "Foo", we return the first Foo function we find. If the name is "Bar.Foo" we look for type "Bar" in the type locator, and return its "Foo" function.

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        /// <summary>
+        /// Call the requested job method before the main function call
+        /// </summary>
+        /// <returns></returns>
+        public override async Task OnExecutingAsync(FunctionExecutingContext executingContext, CancellationToken cancellationToken)
+        {
+            if (executingContext == null)
+            {
+                throw new ArgumentNullException("executingContext");
```
mathewc: nit: here and throughout, replace these string literal param names with use of the nameof operator, e.g. nameof(executingContext). That helps keep the code in sync in the face of renames

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
mathewc: We shouldn't be swallowing errors. All of the error paths for filter invocations need to be reviewed. I think the behavior should be:

- if any executing filters throw, skip the rest of the executing filters, skip the user method, and invoke executing filters
- if any executed filter throws, skip the rest of the executed filters

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,122 @@
...
+        public override async Task OnExecutedAsync(FunctionExecutedContext executedContext, CancellationToken cancellationToken)
+        {
+            if (executedContext == null)
+            {
+                throw new ArgumentNullException("executingContext");
+            }
+
+            if (!string.IsNullOrEmpty(ExecutedFilter))
+            {
+                executedContext.Logger.LogInformation("Executing Function Filter '" + ExecutedFilter + "'");
```
mathewc: nit: string interpolation: $"Executing Function Filter '{ExecutedFilter}'"

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,122 @@
...
+            {
+                await base.OnExecutedAsync(executedContext, cancellationToken);
+            }
+        }
+
+        internal async Task InvokeJobFunctionAsync<TContext>(TContext context, CancellationToken cancellationToken) where TContext : FunctionInvocationContext
+        {
+            MethodInfo methodInfo = null;
+
+            foreach (var type in context.Config.TypeLocator.GetTypes())
```
mathewc: Based on what we decide on method name format, if we support long names (e.g. Foo.Bar), then we'll have to update this code to search for the type first, then method.

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,122 @@
...
+            {
+                methodInfo = type.GetMethod(ExecutingFilter);
+                if (methodInfo != null)
+                {
+                    break;
+                }
+            }
+
+            IDictionary<string, object> invokeArguments = new Dictionary<string, object>();
+            string parameterName = methodInfo.GetParameters().SingleOrDefault(p => p.ParameterType == typeof(TContext)).Name;
```
mathewc: Need to split the ParameterInfo access to its own line, and add validation for if no parameter of the right type exists. In that case, we should throw an exception with a descriptive error message.

src/Microsoft.Azure.WebJobs.Host/JobHost.cs
mathewc: Should be able to revert all change to this file at this point

src/Microsoft.Azure.WebJobs.Host/WebJobs.Host.csproj
```diff
@@ -89,6 +89,57 @@
...
+    <Reference Include="NuGet.Client, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\NuGet.Client.4.0.0\lib\net45\NuGet.Client.dll</HintPath>
+    </Reference>
+    <Reference Include="NuGet.Commands, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\NuGet.Commands.4.0.0\lib\net45\NuGet.Commands.dll</HintPath>
+    </Reference>
+    <Reference Include="NuGet.Common, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\NuGet.Common.4.0.0\lib\net45\NuGet.Common.dll</HintPath>
+    </Reference>
+    <Reference Include="NuGet.Configuration, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
```
mathewc: What are all these new references coming from? The changes you've made shouldn't be bringing in any new dependencies

src/Microsoft.Azure.WebJobs/packages.config
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="StyleCop.MSBuild" version="4.7.55.0" targetFramework="net45" developmentDependency="true" />
-  <package id="xunit.runner.visualstudio" version="2.1.0" targetFramework="net45" />
+  <package id="xunit.runner.visualstudio" version="2.2.0" targetFramework="net46" developmentDependency="true" />
```
mathewc: What are these changes?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -633,7 +641,105 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+                await InvokeExecutedFilters(invoker, invokeParameters, functionCancellationTokenSource, instance, logger, jobHost, config, functionResult, parameterNames);
+            }
+        }
+
+        private static async Task InvokeExecutingFilters(IFunctionInvoker invoker, object[] invokeParameters, CancellationTokenSource functionCancellationTokenSource,
+            IFunctionInstance instance, ILogger logger, JobHost jobHost, JobHostConfiguration config, IReadOnlyList<string> parameterNames)
+        {
+            MethodInfo functionMethod = instance.FunctionDescriptor.Method;
+            CancellationToken cancellationToken = functionCancellationTokenSource.Token;
+            var invokeFilters = functionMethod.GetCustomAttributes().OfType<InvocationFilterAttribute>();
```
mathewc: You're still not supporting class level attributes. See my comment on your first PR about this.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -633,7 +641,105 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+                functionResult = new FunctionResult(ex);
+                throw;
+            }
+            finally
+            {
+                await InvokeExecutedFilters(invoker, invokeParameters, functionCancellationTokenSource, instance, logger, jobHost, config, functionResult, parameterNames);
+            }
+        }
+
+        private static async Task InvokeExecutingFilters(IFunctionInvoker invoker, object[] invokeParameters, CancellationTokenSource functionCancellationTokenSource,
```
MikeStall: >IFunctionInvoker invoker [](start = 57, length = 24)

Is this param used?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -633,7 +641,105 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+
+            if (invokeFilters != null)
+            {
+                foreach (var filter in invokeFilters)
+                {
+                    try
+                    {
+                        if (filter.GetType() == typeof(InvokeFunctionFilterAttribute))
+                        {
+                            InvokeFunctionFilterAttribute invokeFunctionFilter = (InvokeFunctionFilterAttribute)filter;
```
MikeStall: nit: 
InvokeFunctionFilterAttribute invokeFunctionFilter = filter as invokeFunctionFilter;
if (invokeFunctionFilter != null)  { .. }

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -633,7 +641,105 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+                            context.Config = config;
+                            await invokeFunctionFilter.OnExecutingAsync(context, cancellationToken);
+                        }
+                        else
+                        {
+                            await filter.OnExecutingAsync(new FunctionExecutingContext(instance.Id, instance.FunctionDescriptor.FullName, parameters, logger), cancellationToken);
+                        }
+                    }
+                    catch (Exception e)
+                    {
```
MikeStall: >{ [](start = 20, length = 1)

swallow?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -633,7 +641,105 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+                foreach (var filter in invokeFilters)
+                {
+                    try
+                    {
+                        if (filter.GetType() == typeof(InvokeFunctionFilterAttribute))
+                        {
+                            InvokeFunctionFilterAttribute invokeFunctionFilter = (InvokeFunctionFilterAttribute)filter;
+                            FunctionExecutingContext context = new FunctionExecutingContext(instance.Id, instance.FunctionDescriptor.FullName, parameters, logger);
+                            context.JobHost = jobHost;
+                            context.Config = config;
```
MikeStall: Get rid of the If altogether and just always set these.  We shouldn't need to special case derived attributes.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -633,7 +641,105 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+                functionResult = new FunctionResult(true);
+            }
+            catch (Exception ex)
+            {
+                functionResult = new FunctionResult(ex);
+                throw;
+            }
+            finally
+            {
+                await InvokeExecutedFilters(invoker, invokeParameters, functionCancellationTokenSource, instance, logger, jobHost, config, functionResult, parameterNames);
```
MikeStall: >InvokeExecutedFilters [](start = 22, length = 21)

Should this filter execution be covered by the timeout protection above?

src/Microsoft.Azure.WebJobs.Host/FunctionInvocationContext.cs
MikeStall: But we don't actually carry these over from pre to post. 

---
In reply to: [124420331](https://github.com/Azure/azure-webjobs-sdk/pull/1207#discussion_r124420331) [](ancestors = 124420331)

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
mathewc: Yes, all those other locations need to be fixed up :) These are required parameters now. I see that all those places are in the tests - they just need to be updated.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
mathewc: I still don't understand. It seems to me you're calling these the same way in both cases, so if you just always set host/config on the context, the same code will work for both?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
mathewc: Yeah, that's what I said above :)

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
mathewc: We had removed the new public property we added (JobHost.Configuration) to avoid controversy :) If you like the new property we could add it back - I think it makes sense. Just removed it because it was new surface area that wasn't strictly needed.

src/Microsoft.Azure.WebJobs.Host/FunctionExecutingContext.cs
mathewc: Ah, I see, thanks. We might separately investigate just removing our CLS compliance, since this is just a headache that doesn't really add value, and forces us to sprinkle these attributes all over everywhere.

src/Microsoft.Azure.WebJobs.Host/FunctionInvocationContext.cs
mathewc: Yes, that's a hole that needs to be fixed, and tests added for it :)

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
hhawkins: If you mean ExecutingFilter here, then yes, this is the name of the filter that we would call.

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
```diff
@@ -29,13 +30,16 @@
 using Microsoft.Azure.WebJobs.Logging;
 using Microsoft.Extensions.Logging;
 
-namespace Microsoft.Azure.WebJobs.Host.Executors
+namespace Microsoft.Azure.WebJobs.Host
 {
-    internal static class JobHostConfigurationExtensions
```
mathewc: You should revert all the changes to this file - no longer needed

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -632,8 +638,100 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
             }
+            
+            // Get the list of filters to invoke if there are any
+            List<IFunctionInvocationFilter> filters = new List<IFunctionInvocationFilter>();
+            FunctionResult functionResult = null;
+            filters = GetFilters(instance, invoker);
+
+            try
+            {
+                await InvokeExecutingFilters(filters, parameterNames, invokeParameters, functionCancellationTokenSource, instance, logger, jobHost, config);
```
mathewc: As Mike had pointed out in a previous comment, we need all of this filter stuff to be within the Timeout logic.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -632,8 +638,100 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+                functionResult = new FunctionResult(ex);
+                throw;
+            }
+            finally
+            {
+                await InvokeExecutedFilters(filters, parameterNames, invokeParameters, functionCancellationTokenSource, instance, logger, jobHost, config, functionResult);
+            }
+        }
+
+        private static List<IFunctionInvocationFilter> GetFilters(IFunctionInstance instance, IFunctionInvoker invoker)
```
mathewc: Change the parameter types here to GetFilters(MethodInfo, object) and just pass in the instance.FunctionDescriptor.Method and invoker.Instance values.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -632,8 +638,100 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+            }
+        }
+
+        private static async Task InvokeExecutedFilters(List<IFunctionInvocationFilter> filters, IReadOnlyList<string> invokeParameterNames, object[] invokeParameters, CancellationTokenSource functionCancellationTokenSource,
+            IFunctionInstance instance, ILogger logger, JobHost jobHost, JobHostConfiguration config, FunctionResult functionResult)
+        {
+            if (filters.Any())
+            {
+                // Setup the parameters
+                var parameters = new Dictionary<string, object>();
```
mathewc: It's all duplicated code between these two functions. We can compress into a single function, since the only thing that varies between the two is which method is called (Executing/Executed). You can just have a single InvokeFiltersAsync method that takes a Func<IFunctionInvocationFilter, Task> param to do the invoke.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -522,7 +526,9 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
             IReadOnlyDictionary<string, IValueProvider> parameters,
             TraceWriter traceWriter,
             ILogger logger,
-            CancellationTokenSource functionCancellationTokenSource)
+            CancellationTokenSource functionCancellationTokenSource,
+            JobHost jobHost = null,
```
mathewc: These parameters shouldn't be optional

src/Microsoft.Azure.WebJobs.Host/FunctionExecutedContext.cs
```diff
@@ -0,0 +1,36 @@
...
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using Microsoft.Azure.WebJobs.Host.Executors;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// The context for an executed function
```
mathewc: Context class for for <see cref="IFunctionInvocationFilter"/>

src/Microsoft.Azure.WebJobs.Host/FunctionExecutedContext.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// </summary>
+    public class FunctionExecutedContext : FunctionInvocationContext
+    {
+        /// <summary>
+        /// Constructor to set the context
+        /// </summary>
+        /// <param name="id"></param>
```
mathewc: Need to fill out all these param descriptions

src/Microsoft.Azure.WebJobs.Host/FunctionInvocationContext.cs
mathewc: Also update the ctor param so it matches as well

src/Microsoft.Azure.WebJobs.Host/FunctionExecutingContext.cs
```diff
@@ -0,0 +1,29 @@
...
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using Microsoft.Azure.WebJobs.Host.Executors;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// The context describing a function that's about to be executed
```
mathewc: Context class for for <see cref="IFunctionInvocationFilter.OnExecuting"/>

src/Microsoft.Azure.WebJobs.Host/FunctionExecutedContext.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// <param name="name"></param>
+        /// <param name="arguments"></param>
+        /// <param name="logger"></param>
+        /// <param name="result"></param>
+        internal FunctionExecutedContext(Guid id, string name, IReadOnlyDictionary<string, object> arguments, ILogger logger, FunctionResult result) :
```
mathewc: Rename "id" to "functionInstanceId" as you did in the other class

src/Microsoft.Azure.WebJobs.Host/FunctionExecutedContext.cs
mathewc: Use a <<see cref

src/Microsoft.Azure.WebJobs.Host/FunctionExecutingContext.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// </summary>
+    public class FunctionExecutingContext : FunctionInvocationContext
+    {
+        /// <summary>
+        /// Constructor to set the context
+        /// </summary>
+        /// <param name="functionInstanceId"></param>
```
mathewc: Fill these out

src/Microsoft.Azure.WebJobs.Host/FunctionInvocationContext.cs
```diff
@@ -0,0 +1,62 @@
...
+            Name = name;
+            Arguments = arguments;
+            Logger = logger;
+        }
+
+        /// <summary>Gets or sets the ID of the function.</summary>
+        public Guid Id { get; set; }
+
+        /// <summary>Gets or sets the name of the function.</summary>
+        public string Name { get; set; }
```
mathewc: Let's rename this to FunctionName (also update all ctor params). This will make it match FunctionInstanceId better.

src/Microsoft.Azure.WebJobs.Host/IFunctionInvocationFilter.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// An interface for invoking function filters
```
mathewc: Defines a filter that will be called as part of the function invocation pipeline.

src/Microsoft.Azure.WebJobs.Host/IFunctionInvocationFilter.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// </summary>
+    public interface IFunctionInvocationFilter
+    {
+        /// <summary>
+        /// Called before the main function is called
```
mathewc: Reapply the summary block so parameter reference doc is added

src/Microsoft.Azure.WebJobs.Host/IFunctionInvocationFilter.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// </summary>
+        /// <returns></returns>
+        Task OnExecutingAsync(FunctionExecutingContext executingContext, CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Called after the main function is called
```
mathewc: Method invoked after the target function is called

src/Microsoft.Azure.WebJobs.Host/FunctionInvocationContext.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Collections.Generic;
+using Microsoft.Azure.WebJobs.Host.Executors;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// The context for an executed function. This needs to be expanded on later.
+    /// </summary>
+    public abstract class FunctionInvocationContext
```
mathewc: I don't see that you've updated PublicSurfaceTests.cs with all the new types you're adding. You'll need to run all the tests and get a clean run.

src/Microsoft.Azure.WebJobs.Host/FunctionInvocationContext.cs
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        /// <summary>Gets or sets the ID of the function.</summary>
+        public Guid Id { get; set; }
+
+        /// <summary>Gets or sets the name of the function.</summary>
+        public string Name { get; set; }
+
+        /// <summary>
+        /// Arguments from the function
```
mathewc: Gets or sets the function arguments

src/Microsoft.Azure.WebJobs.Host/FunctionInvocationContext.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// </summary>
+        public IReadOnlyDictionary<string, object> Arguments { get; set; }
+
+        /// <summary>
+        /// User properties
+        /// </summary>
+        public IDictionary<string, object> Properties { get; }
+
+        /// <summary>
+        /// Gets or sets the logger of the function
```
mathewc: Gets or sets the function logger

src/Microsoft.Azure.WebJobs.Host/InvocationFilterAttribute.cs
```diff
@@ -0,0 +1,39 @@
...
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, Inherited = true, AllowMultiple = false)]
+    public abstract class InvocationFilterAttribute : Attribute, IFunctionInvocationFilter
+    {
+        /// <summary>
+        /// Tasks here should execute before the actual function is executed
```
mathewc: Can use <<inheritdoc here

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,128 @@
...
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// An invocation filter that invokes job methods
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, Inherited = true, AllowMultiple = true)]
+    public sealed class InvokeFunctionFilterAttribute : InvocationFilterAttribute, IFunctionInvocationFilter
```
mathewc: Don't need the ", IFunctionInvocationFilter"

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,128 @@
...
+    {
+        /// <summary>
+        /// When this attribute is used, take the name of the function to invoke
```
mathewc: Constructs a new instance

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,128 @@
...
+        /// </summary>
+        /// <param name="executingFilter"></param>
```
mathewc: "The name of the function to execute before the target function is called."

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,128 @@
...
+        /// <param name="executedFilter"></param>
```
mathewc: "The name of the function to execute after the target function is called."

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,128 @@
...
+        public InvokeFunctionFilterAttribute(string executingFilter = null, string executedFilter = null)
+        {
+            ExecutingFilter = executingFilter;
+            ExecutedFilter = executedFilter;
+        }
+
+        /// <summary>
+        /// Executing filter
```
mathewc: Use the same verbiage here as in the ctor params

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,128 @@
...
+        /// </summary>
+        public string ExecutingFilter { get; }
+
+        /// <summary>
+        /// Executed filter
+        /// </summary>
+        public string ExecutedFilter { get; }
+
+        /// <summary>
+        /// Call the requested job method before the main function call
```
mathewc: Should be able to use <<inheritdoc

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,128 @@
...
+                }
+
+                if (methodInfo != null)
+                {
+                    break;
+                }
+            }
+
+            // If it is null, this means that we need to look for the methodinfo from the class
+            // TODO: Implement class level methodinfo onExecuting/onExecuted invocation
```
mathewc: Need to address this TODO. Recommend throwing an exception of type InvalidOperationException, with text "Filter method {methodName} not found."

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,128 @@
...
+
+            IDictionary<string, object> invokeArguments = new Dictionary<string, object>();
+            string parameterName = methodInfo.GetParameters().SingleOrDefault(p => p.ParameterType == typeof(TContext)).Name;
```
mathewc: Need proper error handling here.

mathewc: Also, the parameter type check needs to be changed to p.ParameterType.IsAssignableFrom(typeof(TContext))

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,128 @@
...
+            if (executingContext == null)
+            {
+                throw new ArgumentNullException("executingContext");
+            }
+            
+            if (!string.IsNullOrEmpty(ExecutingFilter))
+            {
+                executingContext.Logger.LogInformation("Executing Function Filter '" + ExecutingFilter + "'");
+
+                try
```
mathewc: Remove this try/catch block (don't swallow exceptions)

src/Microsoft.Azure.WebJobs/packages.config
hhawkins: I updated visual studio when I ran into errors, I don't think this is necessary though

src/Microsoft.Azure.WebJobs/packages.config
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="StyleCop.MSBuild" version="4.7.55.0" targetFramework="net45" developmentDependency="true" />
-  <package id="xunit.runner.visualstudio" version="2.1.0" targetFramework="net45" />
+  <package id="xunit.runner.visualstudio" version="2.1.0" targetFramework="net45"/>
```
hhawkins: Fixed this in upcoming PR

src/Dashboard/Web.config
```diff
@@ -1,17 +1,17 @@
-﻿<?xml version="1.0"?>
```
hhawkins: Some of the changes, like in this file for example, are things that VS Code changed on it's own. I didn't do this manually and I'm not sure why it made these weird changes anyway. I've gone through the review to revert most of these random changes but I don't know why it did it in the first place. It might have been some things it did when I had build issues.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -371,8 +375,7 @@ public HostOutputMessage HostOutputMessage
 
                 timer.Elapsed += (o, e) =>
                 {
-                    OnFunctionTimeout(timer, method, instance.Id, timeout.Value, attribute.TimeoutWhileDebugging, trace, logger, cancellationTokenSource,
-                        () => Debugger.IsAttached);
+                    OnFunctionTimeout(timer, method, instance.Id, timeout.Value, attribute.TimeoutWhileDebugging, trace, logger, cancellationTokenSource,                        () => Debugger.IsAttached);
```
mathewc: fix spacing

sample/SampleHost/Functions.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
```
mathewc: revert file

src/Dashboard/Web.config
mathewc: just revert this entire file

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -700,7 +788,22 @@ public ParameterHelper(IReadOnlyList<string> parameterNames)
...
+
+            // Convert the parameters and their names to a dictionary
+            public Dictionary<string, object> GetParametersAsDictionary()
+            {
+                Dictionary<string, object> parametersAsDictionary = new Dictionary<string, object>();
+
+                int counter = 0;
+                foreach (var name in _parameterNames)
+                {
+                    parametersAsDictionary[name] = InvokeParameters[counter];
```
mathewc: nit: InvokeParameters[i++]

src/Microsoft.Azure.WebJobs.Host/FunctionExecutingContext.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    /// <summary>
+    /// Context class for IFunctionInvocationFilter.OnExecutingAsync <see cref="IFunctionInvocationFilter"/>>
+    /// </summary>
+    public class FunctionExecutingContext : FunctionInvocationContext
+    {
+        /// <summary>
+        /// Constructor to set the context
+        /// </summary>
+        /// <param name="functionInstanceId"><see cref="FunctionInvocationContext"/></param>
```
mathewc: Fill these out, rather than use <see crefs

src/Microsoft.Azure.WebJobs.Host/IFunctionInvocationFilter.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// <summary>
+    /// Defines a filter that will be called as part of the function invocation pipeline.
+    /// </summary>
+    public interface IFunctionInvocationFilter
+    {
+        /// <summary>
+        /// Method invoked before the target function is called
+        /// </summary>
+        /// <param name="executingContext">The context</param>
+        /// <param name="cancellationToken">The cancellations token</param>
```
mathewc: spelling

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,119 @@
...
+                }
+
+                if (methodInfo != null)
+                {
+                    break;
+                }
+            }
+
+            // If it is null, this means that we need to look for the methodinfo from the class
+            // TODO: Implement class level methodinfo onExecuting/onExecuted invocation
```
mathewc: Remove this TODO now, since you've done this

mathewc: Actually remove the comment above this as well since it no longer applies

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            if (methodInfo == null)
+            {
+                throw new InvalidOperationException("Incorrect method signature for invocation filter '{methodName}'.");
```
mathewc: Change to "Invocation filter method '{methodName}' not found."

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,119 @@
...
+            }
+
+            IDictionary<string, object> invokeArguments = new Dictionary<string, object>();
+            string parameterName = methodInfo.GetParameters().SingleOrDefault(p => p.ParameterType.IsAssignableFrom(typeof(TContext))).Name;
```
mathewc: Need to break this up into two lines. First validate that you can get the parameter, and if it is null throw "Incorrect method signature for invocation filter '{methodName}'."

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/InvocationFilterEndToEndTests.cs
```diff
@@ -0,0 +1,635 @@
...
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Host.EndToEndTests
+{
+    public class InvocationFilterEndToEndTests : IClassFixture<TestFixture>
+    {
+        private JobHostConfiguration _config;
+        private JobHost _host;
+
+        private const string TestArtifactPrefix = "e2etestmultiaccount";
```
mathewc: remove all this stuff since it isn't used

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/InvocationFilterEndToEndTests.cs
```diff
@@ -0,0 +1,635 @@
...
+            JobHostConfiguration hostConfiguration = new JobHostConfiguration()
+            {
+                NameResolver = nameResolver,
+                TypeLocator = new FakeTypeLocator(typeof(MultipleStorageAccountsEndToEndTests)),
+            };
+            Config = hostConfiguration;
+
+            var account = CloudStorageAccount.Parse(hostConfiguration.StorageConnectionString);
+
+            CloudBlobClient blobClient1 = account.CreateCloudBlobClient();
```
mathewc: nit: blobClient

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/InvocationFilterEndToEndTests.cs
```diff
@@ -0,0 +1,635 @@
...
+        private readonly TestLoggerProvider _loggerProvider = new TestLoggerProvider();
+
+        public InvocationFilterEndToEndTests(TestFixture fixture)
+        {
+            _config = new JobHostConfiguration()
+            {
+                TypeLocator = new FakeTypeLocator(typeof(AsyncChainEndToEndTests))
+            };
+
+            ILoggerFactory loggerFactory = new LoggerFactory();
```
mathewc: Move all this into your TestFixture class below, so the fixture Host property returns this host.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/InvocationFilterEndToEndTests.cs
```diff
@@ -0,0 +1,635 @@
...
+            }
+            return base.Resolve(name);
+        }
+    }
+
+    public class TestFixture : IDisposable
+    {
+        public TestFixture()
+        {
+            RandomNameResolver nameResolver = new TestNameResolver();
```
mathewc: Remove this and the reference to MultipleStorageAccountsEndToEndTests which is from a different test class.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/InvocationFilterEndToEndTests.cs
```diff
@@ -0,0 +1,635 @@
...
+        public override Task OnExecutedAsync(FunctionExecutedContext executedContext, CancellationToken cancellationToken)
+        {
+            executedContext.Logger.LogInformation("The function failed!");
+            executedContext.Logger.LogInformation(executedContext.Result.Exception.ToString());
+
+            return Task.CompletedTask;
+        }
+    }
+
+    internal class TestNameResolver : RandomNameResolver
```
mathewc: Remove this

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/InvocationFilterEndToEndTests.cs
```diff
@@ -0,0 +1,635 @@
...
+        }
+
+        [NoAutomaticTrigger]
+        public static void AnotherFunction(FunctionExecutingContext executingContext)
+        {
+            executingContext.Logger.LogInformation("AnotherFunction invoked!");
+        }
+    }
+
+    public static class ClassWithFunctionToTest
```
mathewc: rename to TestFunctions2

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/InvocationFilterEndToEndTests.cs
```diff
@@ -0,0 +1,635 @@
...
+        [NoAutomaticTrigger]
+        public static void AddToPropertyLast(FunctionExecutedContext executedContext)
+        {
+
+            executedContext.Logger.LogInformation((string)executedContext.Properties["fakeproperty"] + "2");
+            executedContext.Logger.LogInformation("Property from context was edited");
+        }
+    }
+
+    public static class TestFunctionsInClass
```
mathewc: rename to TestFunctions

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/InvocationFilterEndToEndTests.cs
```diff
@@ -0,0 +1,635 @@
...
+    public static class ClassWithFunctionToTest
+    {
+        [NoAutomaticTrigger]
+        public static void AnotherFunction(FunctionExecutingContext executingContext)
+        {
+            executingContext.Logger.LogInformation("AnotherFunction invoked!");
+        }
+    }
+
+    public class TestClassInterface : IFunctionInvocationFilter
```
mathewc: rename to TestFunctions3

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/InvocationFilterEndToEndTests.cs
```diff
@@ -0,0 +1,635 @@
...
+                executingContext.Logger.LogInformation("This is an unauthorized user!");
+                throw new Exception("Not Allowing Unauthorized Users!");
+            }
+
+            executingContext.Logger.LogInformation("This is an authorized user!");
+            return Task.CompletedTask;
+        }
+    }
+
+    public class HTTPRequestFilter : InvocationFilterAttribute
```
mathewc: nit: HttpRequestFilter

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/InvocationFilterEndToEndTests.cs
```diff
@@ -0,0 +1,635 @@
...
+    {
+        public HttpRequestMessage HttpRequestToValidate { get; set; }
```
mathewc: not used - remove

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/InvocationFilterEndToEndTests.cs
```diff
@@ -0,0 +1,635 @@
...
+            executingContext.Logger.LogInformation("Test executing!");
+
+            IReadOnlyDictionary<string, object> arguments = executingContext.Arguments;
+            var testValues = string.Join(",", arguments.Values.ToArray());
+
+            // Check headers (I'm sure there's a better way to check this
+            if (testValues.Contains("Headers"))
+            {
+                executingContext.Logger.LogInformation("Found the header!");
+                // Perform a validation on the headers
```
mathewc: finish this code

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/InvocationFilterEndToEndTests.cs
```diff
@@ -0,0 +1,635 @@
...
+
+    public class HTTPRequestFilter : InvocationFilterAttribute
+    {
+        public HttpRequestMessage HttpRequestToValidate { get; set; }
+
+        public override Task OnExecutingAsync(FunctionExecutingContext executingContext, CancellationToken cancellationToken)
+        {
+            executingContext.Logger.LogInformation("Test executing!");
+
+            IReadOnlyDictionary<string, object> arguments = executingContext.Arguments;
```
mathewc: Fix this code. E.g. try/cast the first argument value to an HttpRequestMessage, then you can party on the query params/headers as needed

src/Dashboard/Web.config
hhawkins: Will do in the next update

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
hhawkins: I'm not sure I understand this comment

hhawkins:
There's another pull request open on #1244 due to this branch being out of sync

PR #1209, opened 2017-06-29
Add an internal helper class, ParameterHelper, for dealing with the v…

MikeStall:
…arious phases of parameters.

- lets us move some floating static methods onto a common instance class.
- enforces a strict contract for handling the various parameter dictionaries.
- this aids in formalizing the various stages of parameters.

No functional change.
Delete some dead code, ValueWatcher


src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -239,16 +239,16 @@ public HostOutputMessage HostOutputMessage
...
                     var valueBindingContext = new ValueBindingContext(functionContext, cancellationToken);
-                    var parameters = await instance.BindingSource.BindAsync(valueBindingContext);
-
+                    await parameterHelper.BindAsync(instance.BindingSource, valueBindingContext);
+                    
                     Exception invocationException = null;
                     ExceptionDispatchInfo exceptionInfo = null;
                     string startedMessageId = null;
-                    using (ValueProviderDisposable.Create(parameters))
+                    using (parameterHelper.GetDisposables())
```
mathewc: GetDisposables is a little strange. Can't the entire ParameterHelper be IDisposable, so you can just do `using(parameterHelper)` here?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -556,7 +531,7 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
                 TimeSpan timerInterval = timer == null ? TimeSpan.MinValue : TimeSpan.FromMilliseconds(timer.Interval);
                 try
                 {
-                    await InvokeAsync(invoker, invokeParameters, timeoutTokenSource, functionCancellationTokenSource,
+                    await InvokeAsync(invoker, parameterHelper, timeoutTokenSource, functionCancellationTokenSource,
```
mathewc: For Hamza's PR which drove your changes, the helper will provide access to both parameter names and values now?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -769,90 +670,268 @@ private static FunctionCompletedMessage CreateCompletedMessage(FunctionStartedMe
...
-            return arguments;
-        }
+            public IDictionary<string, ParameterLog> ParameterLogCollector => _parameterLogCollector;
 
-        private static int GetParameterIndex(IReadOnlyList<string> parameterNames, string name)
-        {
-            for (int index = 0; index < parameterNames.Count; index++)
+            public object[] InvokeParameters => _invokeParameters;
+
+            internal static ParameterHelper New(object[] args)
```
mathewc: It's confusing that this construction mechanism takes an array of arg values, while your other constructor just takes param names. Which should be used when?

src/Microsoft.Azure.WebJobs.Host/Loggers/ValueWatcher.cs
```diff
@@ -15,120 +15,4 @@
 
 namespace Microsoft.Azure.WebJobs.Host.Executors
 {
-    internal class ValueWatcher
```
mathewc: How can you remove this? This is what handles periodic updates of live parameter logs in the Dashboard for long running functions, based on a background timer. It looks now like you only update parameters once on Flush.

I'm nervous about messing around with this code too much, since for Dashboard stuff our main line of defense is our Dashboard UI tests. If you're changing this stuff you should run them.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
MikeStall: Correct. @hhawkins  may need to add a new method on the helper to expose exactly what he needs, but the key information is all in one place now. 


---
In reply to: [125711434](https://github.com/Azure/azure-webjobs-sdk/pull/1209#discussion_r125711434) [](ancestors = 125711434)

src/Microsoft.Azure.WebJobs.Host/Loggers/ValueWatcher.cs
MikeStall: It's appears to be dead code and not used. Dashboard builds fine without it.  Only thing that's used is the static call to AddLogs, which got moved to parameter helper.  Do you see a usage somewhere I'm missing?

---
In reply to: [125713772](https://github.com/Azure/azure-webjobs-sdk/pull/1209#discussion_r125713772) [](ancestors = 125713772)

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
MikeStall: good point. I added this for testing; but it is confusing. I'll simplify. 

---
In reply to: [125712803](https://github.com/Azure/azure-webjobs-sdk/pull/1209#discussion_r125712803) [](ancestors = 125712803)

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
MikeStall: Yes.  I did this since it was closer to the original code; but I'll take it the next step. 

---
In reply to: [125710090](https://github.com/Azure/azure-webjobs-sdk/pull/1209#discussion_r125710090) [](ancestors = 125710090)

src/Microsoft.Azure.WebJobs.Host/Loggers/ValueWatcher.cs
MikeStall: I filed [https://github.com/Azure/azure-webjobs-sdk/issues/1217](https://github.com/Azure/azure-webjobs-sdk/issues/1217) to track down whether the references got deleted erroneously. 

---
In reply to: [126188615](https://github.com/Azure/azure-webjobs-sdk/pull/1209#discussion_r126188615) [](ancestors = 126188615,125713772)

PR #1211, opened 2017-06-29
Removing CLS Compliance

hhawkins:
While developing the Invocation Filter feature, the CLS Compliance warnings were a nuisance.


mathewc:
Thanks for fixing this.

PR #1218, opened 2017-07-07
Add a pre-bind logging instrumentation end point.

MikeStall:
Add an aggressive unit test to verify ordering, including in failure situations.
This is SDK support for Script bug (https://github.com/Azure/azure-webjobs-sdk-script/issues/578 , PR  https://github.com/Azure/azure-webjobs-sdk-script/pull/1660 ) , which will consume the event.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
@@ -88,10 +88,12 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                 ParentId = functionStartedMessage.ParentId,
                 FunctionName = functionStartedMessage.Function.ShortName,
                 TriggerReason = functionStartedMessage.ReasonDetails,
-                StartTime = functionStartedMessage.StartTime.DateTime
+                StartTime = functionStartedMessage.StartTime.DateTime,
+                Properties = new Dictionary<string, object>()
             };
 
-            Stopwatch sw = Stopwatch.StartNew();
+            await this.NotifyPreBindAsync(fastItem);
+                        
             try
             {
                 using (_logger?.BeginFunctionScope(functionInstance))
@@ -135,29 +137,11 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                 logCompletedCancellationToken = cancellationToken;
             }
 
-            // log result
-            sw.Stop();
-            fastItem.EndTime = DateTime.UtcNow;
-            fastItem.Duration = sw.Elapsed;
-            fastItem.Arguments = functionCompletedMessage.Arguments;
-
-            if (_functionEventCollector != null)
-            {
-                // Log completed
-                if (exceptionInfo != null)
-                {
-                    var ex = exceptionInfo.SourceException;
-                    if (ex.InnerException != null)
-                    {
-                        ex = ex.InnerException;
-                    }
-                    fastItem.ErrorDetails = ex.Message;
-                }
-                await _functionEventCollector.AddAsync(fastItem);
-            }
+            await NotifyCompleteAsync(fastItem, functionCompletedMessage.Arguments, exceptionInfo);
+          
             using (_resultsLogger?.BeginFunctionScope(functionInstance))
             {
-                _resultsLogger?.LogFunctionResult(functionInstance.FunctionDescriptor.Method.Name, fastItem, sw.Elapsed, exceptionInfo?.SourceException);
+                _resultsLogger?.LogFunctionResult(functionInstance.FunctionDescriptor.Method.Name, fastItem, fastItem.LiveTimer.Elapsed, exceptionInfo?.SourceException);
             }
 
             if (functionCompletedMessage != null &&
@@ -254,8 +238,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                         if (_functionEventCollector != null)
                         {
                             // Log started
-                            fastItem.Arguments = message.Arguments;
-                            await _functionEventCollector.AddAsync(fastItem);
+                            await NotifyPostBindAsync(fastItem, message.Arguments);
                         }
 
                         try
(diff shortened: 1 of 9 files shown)


src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -670,7 +653,61 @@ private static FunctionCompletedMessage CreateCompletedMessage(FunctionStartedMe
...
+        {
+            fastItem.LiveTimer = Stopwatch.StartNew();
+
+            if (_functionEventCollector == null)
+            {
+                return Task.CompletedTask;
+            }            
+
+            // Log pre-bind event. 
+            return _functionEventCollector.AddAsync(fastItem);
```
brettsam: Would it make sense to have an explicit `FunctionExecutionState`  as part of the fast item? Something that the collector could confidently key off of -- like `PreBind`, `PostBind`, `Complete`...

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -670,7 +653,61 @@ private static FunctionCompletedMessage CreateCompletedMessage(FunctionStartedMe
...
+        }
+
+        // Called before function body is executed; after arguments are bound. 
+        private Task NotifyPostBindAsync(FunctionInstanceLogEntry fastItem, IDictionary<string, string> arguments)
```
brettsam: Rather than pass the arguments as a parameter, can you add them to the fastItem before calling this? It'd simplify the sig.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
brettsam: It feels like this `_functionEventCollector` really isn't much of an `AsyncCollector` anymore. Now, if you want to implement this, you decipher the current state of the `FunctionInstanceLogEntry` to determine where in the invocation life cycle you are. `AddAsync` and `FlushAsync` don't really seem to be enough to describe what's happening.

Did you think about making these `Notify` methods virtual so that we could override them in Script to know for sure what event was being fired?

MikeStall: The problem with making them virtual is all these classes are internal. So it would be exposing significant new surface area for script to get at them.  

---
In reply to: [126788620](https://github.com/Azure/azure-webjobs-sdk/pull/1218#discussion_r126788620) [](ancestors = 126788620)

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -88,10 +88,12 @@ public HostOutputMessage HostOutputMessage
                 ParentId = functionStartedMessage.ParentId,
                 FunctionName = functionStartedMessage.Function.ShortName,
                 TriggerReason = functionStartedMessage.ReasonDetails,
-                StartTime = functionStartedMessage.StartTime.DateTime
+                StartTime = functionStartedMessage.StartTime.DateTime,
+                Properties = new Dictionary<string, object>()
             };
 
-            Stopwatch sw = Stopwatch.StartNew();
+            await this.NotifyPreBindAsync(fastItem);
```
MikeStall: Yes. This is the key part. Subtle but critical. SDK already exposes an instrumentation stream, and this is adding another notification. 

---
In reply to: [126777925](https://github.com/Azure/azure-webjobs-sdk/pull/1218#discussion_r126777925) [](ancestors = 126777925)

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
MikeStall: I'll add accessor properties. 

---
In reply to: [126782080](https://github.com/Azure/azure-webjobs-sdk/pull/1218#discussion_r126782080) [](ancestors = 126782080)

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
MikeStall: this was intentional - I wanted the sigs here to enforce some basic invariants; like requiring an arguments was passed in. 

---
In reply to: [126782972](https://github.com/Azure/azure-webjobs-sdk/pull/1218#discussion_r126782972) [](ancestors = 126782972)

src/Microsoft.Azure.WebJobs.Host/Loggers/FunctionInstanceLogEntry.cs
```diff
@@ -49,12 +51,41 @@ public class FunctionInstanceLogEntry
...
+
+        /// <summary>
+        /// Function has just started. This is before arguments are bound. 
+        /// </summary>
+        public bool IsStart => this.Arguments == null;
+
+        /// <summary>
+        /// Function has bound the arguments but the the body is not yet invoked. 
+        /// </summary>
+        public bool IsPostBind => !this.EndTime.HasValue && this.Arguments != null;
```
brettsam: Will Arguments always be set, even if there aren't any (if thats possible)?

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/ILoggerTests.cs
```diff
@@ -124,7 +130,7 @@ public void Aggregator_Runs_WhenEnabled_AndFlushes_OnStop()
 
             // Add will be called 5 times. The default aggregator will ingore the 
             // 'Function started' calls.
```
brettsam: Now that you've changed this to ignore in the test, you can remove this comment.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Loggers/FunctionExecutorLogOrdering.cs
```diff
@@ -0,0 +1,161 @@
...
+            {
+                if (item.IsStart)
+                {
+                    Assert.Null(item.EndTime);
+                    _log.Append(PreBindLog);
+
+                    Assert.False(item.Properties.ContainsKey(key));
+                    item.Properties[key] = 1;
+                }
+                else
```
brettsam: nit -- you can make these `else if (item.IsPostBind)` / `else if (item.IsComplete)` now, and probably throw at the end if it doesn't match any?

src/Microsoft.Azure.WebJobs.Host/Loggers/FunctionInstanceLogEntry.cs
```diff
@@ -49,12 +51,41 @@ public class FunctionInstanceLogEntry
...
+        /// <summary>
+        /// An in-memory timer to measure the duration during execution. Use <see cref="Duration"/> for persisted cases.
+        /// </summary>
+        [JsonIgnore]
+        public Stopwatch LiveTimer { get; set; }
+
+        /// <summary>
+        /// Function has just started. This is before arguments are bound. 
+        /// </summary>
+        public bool IsStart => this.Arguments == null;
```
brettsam: One and only one of these `IsStart`/`IsPostBind`/`IsCompleted` should be true at any time, right? Is there any way to guarantee that, or is this good enough?

MikeStall: Yes. I'll make it more obvious. 

---
In reply to: [127044405](https://github.com/Azure/azure-webjobs-sdk/pull/1218#discussion_r127044405) [](ancestors = 127044405)

src/Microsoft.Azure.WebJobs.Host/Loggers/FunctionInstanceLogEntry.cs
MikeStall: Yes, we'll still have a non-null zero-size dictionary. 

---
In reply to: [127043423](https://github.com/Azure/azure-webjobs-sdk/pull/1218#discussion_r127043423) [](ancestors = 127043423)

Issue #1208, opened 2017-06-28
[AppInsights] Add Dependency tracking


brettsam:

Eli-Black-Work:
Would this be as simple as adding `services.AddApplicationInsightsTelemetryWorkerService();` inside `AddApplicationInsightsWebJobs()`?

Eli-Black-Work:
@brettsam Sorry to bother, but could you please post a quick update on this? 🙂  Does AppInsights on WebJobs now support dependency tracking, or is this something that's still awaiting implementation?

PR #1220, opened 2017-07-11
Raise Microsoft.ApplicationInsights version to latest GA package (2.4.0)

TheeJamesLee:
To take advantage of the latest Microsoft.ApplicationInsights features provided in 2.4.0.

Working in Azure Functions, wondered how we may be able to utilise the x-ms-request-id style headers that Application Insights introduced for correlating the parent request id's automatically within Azure Functions to work with Application Insight's direction for visualizing micro services.  re Azure/Azure-Functions#384

Found it is likely to be as simple as raising the package version.  Application Insights introduced it in 2.4.0-beta3 [(blog announcement)](https://azure.microsoft.com/en-gb/blog/app-insights-microservices/).

(For some reason test pack isn't running on my local machine - going to be an issue on my local machine, but being open I haven't run them locally)

### DIFF
--- a/sample/SampleHost/App.config
+++ b/sample/SampleHost/App.config
@@ -25,6 +25,10 @@
         <assemblyIdentity name="System.Diagnostics.DiagnosticSource" publicKeyToken="cc7b13ffcd2ddd51" culture="neutral" />
         <bindingRedirect oldVersion="0.0.0.0-4.0.1.0" newVersion="4.0.1.0" />
       </dependentAssembly>
+      <dependentAssembly>
+        <assemblyIdentity name="System.Net.Http" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-4.1.1.0" newVersion="4.1.1.0" />
+      </dependentAssembly>
     </assemblyBinding>
   </runtime>
 </configuration>
--- a/sample/SampleHost/SampleHost.csproj
+++ b/sample/SampleHost/SampleHost.csproj
@@ -77,9 +77,8 @@
       <Private>True</Private>
     </Reference>
     <Reference Include="System.Core" />
-    <Reference Include="System.Diagnostics.DiagnosticSource, Version=4.0.1.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\System.Diagnostics.DiagnosticSource.4.3.0\lib\net46\System.Diagnostics.DiagnosticSource.dll</HintPath>
-      <Private>True</Private>
+    <Reference Include="System.Diagnostics.DiagnosticSource, Version=4.0.2.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\System.Diagnostics.DiagnosticSource.4.4.0\lib\net46\System.Diagnostics.DiagnosticSource.dll</HintPath>
     </Reference>
     <Reference Include="System.Globalization.Calendars, Version=4.0.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
       <HintPath>..\..\packages\System.Globalization.Calendars.4.3.0\lib\net46\System.Globalization.Calendars.dll</HintPath>
@@ -152,7 +151,9 @@
   </ItemGroup>
   <ItemGroup>
     <None Include="App.config" />
-    <None Include="packages.config" />
+    <None Include="packages.config">
+      <SubType>Designer</SubType>
+    </None>
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\..\src\Microsoft.Azure.WebJobs.Host\WebJobs.Host.csproj">
--- a/sample/SampleHost/SampleHost.csproj
+++ b/sample/SampleHost/SampleHost.csproj
@@ -79,6 +79,7 @@
     <Reference Include="System.Core" />
     <Reference Include="System.Diagnostics.DiagnosticSource, Version=4.0.2.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51, processorArchitecture=MSIL">
       <HintPath>..\..\packages\System.Diagnostics.DiagnosticSource.4.4.0\lib\net46\System.Diagnostics.DiagnosticSource.dll</HintPath>
+      <Private>True</Private>
     </Reference>
     <Reference Include="System.Globalization.Calendars, Version=4.0.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
       <HintPath>..\..\packages\System.Globalization.Calendars.4.3.0\lib\net46\System.Globalization.Calendars.dll</HintPath>
@@ -151,9 +152,7 @@
   </ItemGroup>
   <ItemGroup>
     <None Include="App.config" />
-    <None Include="packages.config">
-      <SubType>Designer</SubType>
-    </None>
+    <None Include="packages.config" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\..\src\Microsoft.Azure.WebJobs.Host\WebJobs.Host.csproj">
--- a/sample/SampleHost/packages.config
+++ b/sample/SampleHost/packages.config
@@ -16,7 +16,7 @@
   <package id="System.ComponentModel" version="4.3.0" targetFramework="net46" />
   <package id="System.Console" version="4.3.0" targetFramework="net46" />
   <package id="System.Diagnostics.Debug" version="4.3.0" targetFramework="net46" />
-  <package id="System.Diagnostics.DiagnosticSource" version="4.3.0" targetFramework="net46" />
+  <package id="System.Diagnostics.DiagnosticSource" version="4.4.0" targetFramework="net46" />
   <package id="System.Diagnostics.Tools" version="4.3.0" targetFramework="net46" />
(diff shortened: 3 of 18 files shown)


TheeJamesLee:
For reference, CI test run shows 3 failed;
2x Queue listener because 'Storage' connection string missing or empty.
1x Servicebus connection string missing or empty.

brettsam:
@TheeJamesLee -- are you specifically looking for the auto-header inclusion so you can link multiple Function apps together? I'd love to hear your scenario -- I *think* this would only really work for functions that use HttpTriggers, is that right?

We're very careful about moving any dependency versions forward as they can cause issues with existing function apps (we've had to roll back in the past). For that reason, I'm going to sit on this until we can do some more thorough testing.

TheeJamesLee:
@brettsam the use case is another web app calling a HttpTrigger.  Not necessarily 'linking' functions as I'm not too worried around a functions dependencies (at this point in time..) you've covered that here Azure/azure-webjobs-sdk#1208, however **if Application Insights have worked around the correlation story I think it is critical that we up this version to build upon anything further that has been gained for free**. 

The link in the App Insight data is when it applies the Parent Id (off the top of my head) and this is what Application Insights is now looking like it is doing under the hood.  Otherwise I need to (some how) manually extract all the headers (can do, but in the function run = remembering to do this each time (bad)) then apply the values to the 'core' Telemetry Client the items I need (Can't see how)... all other examples show that you can bring your own Telemetry Client, but need to avoid sending certain calls because they would be duplicates of the internal client code... the examples only show how to bring it in line.

I had a slight problem within Azure Functions where looking at the steps for the App Insight preview inclusion, it was simply to add your Application insights api key - and it would work automagically - I really prefer pulling in an explicit nuget.  In the nuget list there is no obvious inclusion of application insights (because it is compiled into the webjob sdk) so it was a big struggle to pot exactly what was going on... more importantly what version of dependency was being used Azure/Azure-Functions#384.  (Unsure if it feels right or wrong in this case that versioning is being controlled by a sub component for something that links to a very external resource...  so if I update the webjobs sdk in the future, it may possible include an updated Application Insights package!? hmmm).

This now does cause this problem because I want (need) to increment the app insight version, however it has been strongly coupled within the WebJobs sdk (unless I am missing something obvious?) I have tried to see if just adding the Application Insights nuget to the FunctionApp 'overrides' the package version, but, as expected, no luck there.

I completely agree with your risk management across moving dependency versions forwards - however, it is a bigger risk that with this being so tightly coupled that when this discussion happens again in the future we're at the pace of this repo.
I would suggest that maybe we look at that sooner rather than later?  Should ApplicationInsights as a dependency of WebJobs Sdk be moved to a Logging.Extensions.ApplicationInsights (for example) and have a style of common interface between the two?

Cons:
- Lose the nice story of just adding the app insight key and it works. But you just need to add the Microsoft.Azure.WebJobs.Logging.Extensions.ApplicationInsights package (At least you know exactly which version it is)
-->Mitigation, just auto include the nuget in the functions tooling (and other start templates?)

Pros:
- If someone wants to write their own output adapter for logging (Seq, etc) they can.
- Version changes of these dependencies are more independent of this repo (I'm viewing as core).
- That repo can have it's own 'nightly builds' and stable version, in line with the moving application insights version.

As a thought?

It doesn't solve the more immediate issue that I'd like to up this version sooner rather than later though :) ...

TheeJamesLee:
@brettsam just curious, is it possible to build this pull request onto a feed so that I could at least consume it? I'm not really a nuget ninja, so really open to some ideas of how I can move forward if we are going to sit on this (and I can test this package in the mean time from a good source and feed back :) - think everyone wins then for a while!)

TheeJamesLee:
@brettsam 1) sorry for sending so much directly.. 2) I have had a thought writing on another thread around this:

> are you specifically looking for the auto-header inclusion so you can link multiple Function apps together? I'd love to hear your scenario -- I think this would only really work for functions that use HttpTriggers, is that right?

Yes, the above will only pick up Http incoming called, because AI detects the headers.

Other bindings;

I think we need to look at each binding specifically and cater for how we pass the meta data in a non-breaking way.  Some that I thought of;

- ServiceBus
--- Apply / read these headers ourselves in the binding
- Queues
--- I don't think off the top of my head we can add metadata to a queue message - it's only a payload. So probably the best way would be on the binding itself to allow the auto resolve of a property on the POCO object... if it's null, let App Insights do it's thing.

etc.  (Original thought from Azure/Azure-Functions#245)

I think this gets us very close to the dependency stuff with these headers with not too much work and avoiding breaking changes leaning on the expectation that these elements will be null allowing developers to upgrade at their own pace.

TheeJamesLee:
@brettsam regarding building, I've (finally) sorted building my fork onto my own nuget feed and replacing the webjobs package with my custom build and deploying it into an environment for now. On first glance, the up version doesn't seem to have broken anything, however, the link I was hoping webjobs could get for free doesn't seem to occur - auto correlation via the injected headers by an upstream AI wired web app.

brettsam:
No need to apologize, we like thoughtful feedback :-) 

It looks like you're bringing up a few things:

 **Allowing customers to independently move a dependency forward:**
We've had a lot of discussion internally about people wanting newer nuget dependencies (and some on github: [here is one of the main issues](https://github.com/Azure/azure-webjobs-sdk-script/issues/992) -- and you can see a bunch of linked issues in there). The issue most commonly comes up with Azure Storage, but trying to update any dependency that the host requires is an issue.

For App Insights, we've opted for the 'set a setting and it turns on' approach to make this a lot easier for the majority of customers to get started. I understand it's not quite as customizable as some folks would like. We're continuing to discuss this based on the issue I linked above. 

Your approach above about removing the nuget package from our build and allowing customers to add one themselves is interesting from the logging perspective as all of our logging is decoupled (everything goes through ILogger and knows nothing about App Insights). In theory, I think it could work, but we'd have to weigh the desire for a 'single setting' approach vs 'multiple step' approach. I've filed https://github.com/Azure/azure-webjobs-sdk-script/issues/1668 to track this.

Also, I'll tell you that we *will* be moving to App Insights 2.4 in the near future (some things like filtering the Live Stream in the portal require it), so ultimately it'd really only get you a ~month head-start. I'll leave this issue here to track that.

**Getting parent-child correlation working:** 
I'm calling the multi-call parent-child relationship 'correlation' because using the word 'dependency' conflicts with what App Insights calls a 'dependency' (which is a call out to a service that you don't own). I think you're talking about having multiple calls between services that you own, and wanting them to appear linked together in App Insights. As you've mentioned, there's a lot of different pieces to get this working everywhere -- just working with HttpTriggers is one thing.

This is tracked by https://github.com/Azure/Azure-Functions/issues/245 -- but as you noted, it can be tough to get this working with every kind of Trigger.

**Testing a newer build yourself:**
Testing this in Functions is possible (see: https://github.com/Azure/azure-webjobs-sdk-script/wiki/Deploying-the-Functions-runtime-as-a-private-site-extension). You should be able to pull the https://github.com/Azure/azure-webjobs-sdk-script repo, build it locally yourself (moving App Insights to 2.4), and deploy it following those steps. 

**Trying 2.4:**
Were you able to test this and make sure the correlation was working with your HttpTrigger calls when you used 2.4? We track things differently than a normal web app, so I'm not entirely sure if the correllation would work by default. We may need to add some code to make sure the header gets extracted and applied to the RequestTelemetry correctly. If that's the case, I'll take care of that when we move to 2.4.

TheeJamesLee:
Yeah, just a few things. Thanks for the comprehensive reply back!

**Allowing customers to independently move a dependency forward:**
Yes, I was trying to just focus with AI because it didn't seem like a core dependency (like storage would be), hence moving it into more of an extension pattern style.

> For App Insights, we've opted for the 'set a setting and it turns on' approach to make this a lot easier for the majority of customers to get started. 

And I think that's great! Most people are starting from the VS 15.3 preview functions tooling for example. Why not just add the initial nuget in the start up templates as it fundamentally achieves the same thing. I'd assume anyone building up from scratch would be able to find what they were after through nuget package names.

**Trying 2.4**
I took this pull request, build it CD style on a VSTS agent and onto a personal nuget feed (So I had the WebJobs, etc packages).
I then repointed one of our solutions to use that personal build of the webjobs packages, worked fine locally and put it out on an instance and it ran with no problems.
However, it did not automatically pick up the headers locally or deployed (I had a thought it might be the case). So I'm just looking to see if I can do a POC for myself of exactly what you've written (extracting the headers, applying to the telemetryClient context) just to see how the data comes through at the moment, because as you say, we will need to touch it for each trigger ideally so it may be worth to take it manually in a sense.

If I have a chance later on I'll try with changing the functions runtime to see incase the above method wouldn't have worked because of the functions package runtime, but either way it sounds like it'll be attempted in a few weeks anyway by you guys :)

TheeJamesLee:
**Regarding the parent-child correlations:**

Not wanting to turn this thread into that discussion, however, I am guessing that you mean AI define dependency as something you 'have no control over' (is that what you mean by don't own?).  However the way they seem to log does not fit that...

AI's auto instrumentation of a webapp means that all Http calls out are logged as a 'dependency' - to me that makes sense, there is a transport mechanism that is required to 'fulfil' the original request ... so in the case where I have a WebApp calling a FunctionApp with a http trigger; (Both separate deployments and 'services' that I own with the same AI key)
In the context of the WebApp, the FunctionApp is a dependency.  Because the 2 deployments have the same AI key, we also see the FunctionApp's request 'linked' because of the webapp's operation Id generation (through the operationId/parentId/requestId, etc) being passed through to the function app (headers, payload, doesn't matter) to write to the TelemetryClient.  Which it may make more sense for it to be written.

As you said, this is tracked here Azure/Azure-Functions#245 -- maybe it might be worth to refine the title as it is actually about "Correlation" in a more general sense utilizing AI than just setting the OperationId (but I think more needs to be set than that - maybe) in some way - the other part is the 'how' it gets set.

brettsam:
@TheeJamesLee -- I've moved our repo up to 2.4 with this PR #1312. I'll close this one out -- thanks for submitting it and bearing with me! :-)

Issue #1104, opened 2017-04-07
Return value support for BYOB


MikeStall:
BYOB is all about Script consuming SDK extensions directly - with no additional Script-specific glue code. 
Script has some support for return values (IResultProcessingBinding), but SDK does not support return value binding. 

There are 2 ways we use return value:
1. an implicit 'out' parameter.  This is just binding sugar and can work with existing IAsyncCollector patterns where:
```
T Foo()  { 
return T;
}
```
becomes
```
void Foo2(ICollector<T> temp)
{
  T x = Foo();
  temp.Add(x);
}
```

2. applying something to the trigger (ie, HttpResponseMessage)

This is primarily for case #2.  Need some means for SDK extensions to signal about return value binding.

MikeStall:
FYI, existing bug for case 1: https://github.com/Azure/azure-webjobs-sdk/issues/656

cgillum:
FYI we need this for Durable Functions: https://github.com/Azure/azure-functions-durable-extension/issues/4

This is required both for handling return values and for binding to input parameters - thus is really important for our experience.

lindydonna:
Visual Studio tooling depends on this, see https://github.com/Azure/azure-webjobs-sdk-script/issues/1508

MikeStall:
As a design point, An explicit binding on the return attribute should take place over an implicit trigger binding. 

Case | Example |  What does return value do?
--- | --- | ---
1 | void Foo()   | none 
2 | [Q] string Foo()   |  [Q] 
3 | string Foo(Request)  | error, no bindings at all 
4 | string Foo([Http] Request)  |  [Http], return is implicit given the trigger binding 
5 |  [Q] string Foo([Http] Request)  |  [Q], takes precedence.

PR #1224, opened 2017-07-13
Return value support for BYOB

cgillum:
This is to resolve https://github.com/Azure/azure-webjobs-sdk/issues/1104.

Phase 1 simply allows functions in WebJobs to have a return value. Previously we supported `void` and `Task`. With this change, we also support `TReturnValue` and `Task<TReturnValue>`.

Example functions I'm using for testing:

```csharp
public static int QueueTriggerWithReturnValue(
    [QueueTrigger("test")] string message)
{
    Console.WriteLine("Processed message: " + message);
    return 5;
}

public static Task<int> QueueTriggerWithReturnValueAsync(
    [QueueTrigger("test")] string message)
{
    Console.WriteLine("Processed message: " + message);
    return Task.FromResult(5);
}
```

Phase 2 (TBD) is to connect the return value back to the trigger binding.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInvokerFactory.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInvokerFactory.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
+using System.Threading.Tasks;
 
 namespace Microsoft.Azure.WebJobs.Host.Executors
 {
@@ -27,7 +28,14 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
             MethodInfo genericMethodDefinition = typeof(FunctionInvokerFactory).GetMethod("CreateGeneric",
                 BindingFlags.NonPublic | BindingFlags.Static);
             Debug.Assert(genericMethodDefinition != null);
-            MethodInfo genericMethod = genericMethodDefinition.MakeGenericMethod(reflectedType);
+
+            Type returnType;
+            if (!TypeUtility.TryGetReturnType(method, out returnType))
+            {
+                returnType = typeof(object);
+            }
+
+            MethodInfo genericMethod = genericMethodDefinition.MakeGenericMethod(reflectedType, returnType);
             Debug.Assert(genericMethod != null);
             Func<MethodInfo, IJobActivator, IFunctionInvoker> lambda =
                 (Func<MethodInfo, IJobActivator, IFunctionInvoker>)Delegate.CreateDelegate(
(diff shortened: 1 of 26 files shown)


src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInvokerFactory.cs
```diff
@@ -27,25 +28,42 @@ public static IFunctionInvoker Create(MethodInfo method, IJobActivator activator
...
+                returnType = typeof(object);
+            }
+            else if (returnType == typeof(Task))
+            {
+                returnType = typeof(object);
+            }
+            else if (typeof(Task).IsAssignableFrom(returnType))
+            {
+                returnType = returnType.GetGenericArguments()[0];
+            }
```
MikeStall: Nit: can this logic be in TypeUtility next to your related helper?

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
```diff
@@ -198,6 +205,34 @@ private bool HasJobAttribute(CustomAttributeData attributeData)
             bool hasParameterBindingAttribute = false;
             Exception invalidInvokeBindingException = null;
 
+            bool triggerHasReturnBinding = false;
+            if (TypeUtility.TryGetReturnType(method, out Type methodReturnType))
+            {
+                if (bindingDataContract != null && bindingDataContract.TryGetValue(ReturnParamName, out Type triggerReturnType))
+                {
+                    // CONSIDER: Special casing for JObject and related types for cross-language support?
+                    if (!triggerReturnType.IsAssignableFrom(methodReturnType))
```
MikeStall: >if (!triggerReturnType.IsAssignableFrom(methodReturnType)) [](start = 20, length = 58)

We probably need to go through the converter managers here, which  would handle this.   For now, the trigger can just provide a contract of type  Object, which makes this moot.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInvokerFactory.cs
cgillum: Looks like I can replace this logic with TypeUtility.TryGetReturnType.

src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionData.cs
```diff
@@ -24,6 +24,6 @@ public class TriggeredFunctionData
         /// <summary>
         /// Optional handler function for processing the invocation.
         /// </summary>
-        public Func<Func<Task>, Task> InvokeHandler { get; set; }
+        public Func<Func<Task<object>>, Task<object>> InvokeHandler { get; set; }
```
fabiocav: Since we intend to have these changes in 2.x, we can't make public API breaking changes.

cgillum:
I've rebased the PR on top of the .NET Core changes in the dev branch. As a result, I had to do a force push on the PR branch which unfortunately invalidated previous review comments.

@fabiocav I have reverted the breaking change you had asked about previously. The result is we simply don't support return values when going through the InvokeHandler path. Durable Functions actually doesn't need this in it's InvokeHandler case, so it's not a problem to leave out this support (for now).

MikeStall:
:shipit:

PR #1241, opened 2017-07-18
SDK honors [FunctionName] attribute

MikeStall:
SDK honors [FunctionName] attribute
Resolve https://github.com/Azure/azure-webjobs-sdk/issues/1170
ADd new JobHost.CallAsync(string) endpoint which can honor the method name.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndex.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndex.cs
@@ -46,6 +46,20 @@ namespace Microsoft.Azure.WebJobs.Host.Indexers
             return _functionsById[functionId];
         }
 
+        public IFunctionDefinition LookupByName(string name)
+        {
+            foreach (var items in _functionDescriptors)
+            {
+                if (string.Equals(items.ShortName, name, StringComparison.OrdinalIgnoreCase))
+                {
+                    var id = items.Id;
+                    return Lookup(id);
+                }
+            }
+            // Not found.
+            return null;
+        }        
+
         public IFunctionDefinition Lookup(MethodInfo method)
         {
             if (!_functionsByMethod.ContainsKey(method))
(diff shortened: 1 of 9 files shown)


src/Microsoft.Azure.WebJobs.Protocols/FunctionDescriptor.cs
```diff
@@ -34,10 +34,14 @@ public class FunctionDescriptor
...
 #if PUBLICPROTOCOL
 #else
-        /// <summary>Gets or sets the name used for logging. This is 'Method'. </summary>
+        /// <summary>Gets or sets the name used for logging. This is 'Method' or the value overwritten by [FunctionName] </summary>
         [JsonIgnore] 
         internal string LogName { get; set; }
 
+        /// <summary>Gets or sets the name used for logging. This is ShortName or the value overwitten by [FunctionName]. </summary>
+        [JsonIgnore] 
+        internal string ShortNameOverride { get; set; }
```
mamaso: this is a pretty confusing property: 

1. should it be formatted as `Class.UserOverride`?
2. do we need two separate `ShortName` properties? I think that people will choose incorrectly

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndex.cs
```diff
@@ -46,6 +46,20 @@ public IFunctionDefinition Lookup(string functionId)
             return _functionsById[functionId];
         }
 
+        public IFunctionDefinition LookupByName(string name)
```
mamaso: lookup by which name type? Can we make it more explicit from the function name or definition?

src/Microsoft.Azure.WebJobs.Protocols/FunctionDescriptor.cs
MikeStall: It's confusing ... But I can't find a correct alternative that actually complies with our maze of backwards compat requirements.

mamaso: what are the other uses for ShortName?

PR #1244, opened 2017-07-20
Function Invocation Filters

hhawkins:
For feature tracked here: #980. Sending out another pull request after making updates and changes from feedback from the first pull request #1189 and second pull request #1207

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
@@ -32,11 +32,14 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
         private readonly IAsyncCollector<FunctionInstanceLogEntry> _functionEventCollector;
         private readonly ILogger _logger;
         private readonly ILogger _resultsLogger;
+        private readonly JobHost _jobHost;
+        private readonly JobHostConfiguration _config;
 
         private HostOutputMessage _hostOutputMessage;
 
         public FunctionExecutor(IFunctionInstanceLogger functionInstanceLogger, IFunctionOutputLogger functionOutputLogger,
-                IWebJobsExceptionHandler exceptionHandler, TraceWriter trace, IAsyncCollector<FunctionInstanceLogEntry> functionEventCollector = null,
+                IWebJobsExceptionHandler exceptionHandler, TraceWriter trace, JobHost jobHost, JobHostConfiguration config,
+                IAsyncCollector<FunctionInstanceLogEntry> functionEventCollector = null,
                 ILoggerFactory loggerFactory = null)
         {
             if (functionInstanceLogger == null)
@@ -63,6 +66,8 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
             _functionOutputLogger = functionOutputLogger;
             _exceptionHandler = exceptionHandler;
             _trace = trace;
+            _jobHost = jobHost;
+            _config = config;
             _functionEventCollector = functionEventCollector;
             _logger = loggerFactory?.CreateLogger(LogCategories.Executor);
             _resultsLogger = loggerFactory?.CreateLogger(LogCategories.Results);
@@ -448,7 +453,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
 
             try
             {
-                await ExecuteWithWatchersAsync(instance, parameterHelper, trace, logger, functionCancellationTokenSource);
+                await ExecuteWithWatchersAsync(instance, parameterHelper, trace, logger, functionCancellationTokenSource, _jobHost, _config);
 
                 if (updateParameterLogTimer != null)
                 {
@@ -474,7 +479,9 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
             ParameterHelper parameterHelper,
             TraceWriter traceWriter,
             ILogger logger,
-            CancellationTokenSource functionCancellationTokenSource)
+            CancellationTokenSource functionCancellationTokenSource,
+            JobHost jobHost,
+            JobHostConfiguration config)
         {
             IFunctionInvoker invoker = instance.Invoker;
 
@@ -494,24 +501,46 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                 await singleton.AcquireAsync(functionCancellationTokenSource.Token);
             }
 
-            // Create a source specifically for timeouts
-            using (CancellationTokenSource timeoutTokenSource = new CancellationTokenSource())
+            // Check to see if there are any filters
+            var jobInstance = invoker.CreateInstance();
+            parameterHelper.ObjectInstance = jobInstance;
+            using (jobInstance as IDisposable)
             {
-                TimeoutAttribute timeoutAttribute = instance.FunctionDescriptor.TimeoutAttribute;
-                bool throwOnTimeout = timeoutAttribute == null ? false : timeoutAttribute.ThrowOnTimeout;
-                var timer = StartFunctionTimeout(instance, timeoutAttribute, timeoutTokenSource, traceWriter, logger);
-                TimeSpan timerInterval = timer == null ? TimeSpan.MinValue : TimeSpan.FromMilliseconds(timer.Interval);
-                try
+                List<IFunctionInvocationFilter> filters = GetFilters(instance.FunctionDescriptor, jobInstance);
+
+                // Create a source specifically for timeouts
+                using (CancellationTokenSource timeoutTokenSource = new CancellationTokenSource())
                 {
-                    await InvokeAsync(invoker, parameterHelper, timeoutTokenSource, functionCancellationTokenSource,
-                    throwOnTimeout, timerInterval, instance);
-                }
-                finally
-                {
-                    if (timer != null)
+                    TimeoutAttribute timeoutAttribute = instance.FunctionDescriptor.TimeoutAttribute;
+                    bool throwOnTimeout = timeoutAttribute == null ? false : timeoutAttribute.ThrowOnTimeout;
+                    var timer = StartFunctionTimeout(instance, timeoutAttribute, timeoutTokenSource, traceWriter, logger);
+                    TimeSpan timerInterval = timer == null ? TimeSpan.MinValue : TimeSpan.FromMilliseconds(timer.Interval);
+                    try
                     {
-                        timer.Stop();
-                        timer.Dispose();
+                        // Create a dictionary that will pass the user properties from the Executing filter to the Executed filter
+                        Dictionary<string, object> properties = new Dictionary<string, object>();
+
+                        // Create the context objects for the filter execution
+                        FunctionExecutingContext executingContext = new FunctionExecutingContext(instance.Id, instance.FunctionDescriptor.FullName, parameterHelper.GetParametersAsDictionary(), properties, logger);
+                        executingContext.JobHost = jobHost;
+                        executingContext.Config = config;
+
+                        FunctionExecutedContext executedContext = new FunctionExecutedContext(instance.Id, instance.FunctionDescriptor.FullName, parameterHelper.GetParametersAsDictionary(), properties, logger);
+                        executedContext.JobHost = jobHost;
+                        executedContext.Config = config;
+
+                        invoker = FunctionWithFilterInvoker.ApplyFilters(invoker, filters, executingContext, executedContext);
+
+                        await InvokeAsync(invoker, parameterHelper, timeoutTokenSource, functionCancellationTokenSource,
+                            throwOnTimeout, timerInterval, instance);
+                    }
+                    finally
+                    {
+                        if (timer != null)
+                        {
+                            timer.Stop();
+                            timer.Dispose();
+                        }
                     }
                 }
             }
@@ -538,28 +567,50 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
             //      b. If !throwOnTimeout, wait for the task to complete.
 
             // Start the invokeTask.
-            var objInstance = invoker.CreateInstance();
+            var objInstance = parameterHelper.ObjectInstance;
 
-            using (objInstance as IDisposable)
+            Task<object> invokeTask = invoker.InvokeAsync(objInstance, invokeParameters);
+
+            // Combine #1 and #2 with a timeout task (handled by this method).
+            // functionCancellationTokenSource.Token is passed to each function that requests it, so we need to call Cancel() on it
+            // if there is a timeout.
+            bool isTimeout = await TryHandleTimeoutAsync(invokeTask, functionCancellationTokenSource.Token, throwOnTimeout, timeoutTokenSource.Token,
+                timerInterval, instance, () => functionCancellationTokenSource.Cancel());
+
+            // #2 occurred. If we're going to throwOnTimeout, watch for a timeout while we wait for invokeTask to complete.
+            if (throwOnTimeout && !isTimeout && functionCancellationTokenSource.IsCancellationRequested)
             {
-                Task<object> invokeTask = invoker.InvokeAsync(objInstance, invokeParameters);
-
-                // Combine #1 and #2 with a timeout task (handled by this method).
-                // functionCancellationTokenSource.Token is passed to each function that requests it, so we need to call Cancel() on it
-                // if there is a timeout.
-                bool isTimeout = await TryHandleTimeoutAsync(invokeTask, functionCancellationTokenSource.Token, throwOnTimeout, timeoutTokenSource.Token,
-                    timerInterval, instance, () => functionCancellationTokenSource.Cancel());
-
-                // #2 occurred. If we're going to throwOnTimeout, watch for a timeout while we wait for invokeTask to complete.
-                if (throwOnTimeout && !isTimeout && functionCancellationTokenSource.IsCancellationRequested)
-                {
-                    await TryHandleTimeoutAsync(invokeTask, CancellationToken.None, throwOnTimeout, timeoutTokenSource.Token, timerInterval, instance, null);
-                }
-
-                object returnValue = await invokeTask;
-
-                parameterHelper.SetReturnValue(returnValue);
+                await TryHandleTimeoutAsync(invokeTask, CancellationToken.None, throwOnTimeout, timeoutTokenSource.Token, timerInterval, instance, null);
             }
+
+            object returnValue = await invokeTask;
+
+            parameterHelper.SetReturnValue(returnValue);
+        }
+
+        // Return a list of filters in order that they should be executed. 
+        private static List<IFunctionInvocationFilter> GetFilters(FunctionDescriptor functionDescriptor, object instance)
+        {
+            List<IFunctionInvocationFilter> filters = new List<IFunctionInvocationFilter>();
+
+            if (instance is IFunctionInvocationFilter instanceFilter)
+            {
+                filters.Add(instanceFilter);
+            }
+
+            // Check for class filters
+            if (functionDescriptor.DeclaringClassFilters != null)
+            {
+                filters.AddRange(functionDescriptor.DeclaringClassFilters);
+            }
+
+            // Check for method filters
+            if (functionDescriptor.DeclaringMethodFilters != null)
+            {
+                filters.AddRange(functionDescriptor.DeclaringMethodFilters);
+            }
+
+            return filters;
         }
 
         /// <summary>
@@ -573,7 +624,8 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
         /// <param name="instance">The function instance. Used only in the exceptionMessage</param>
         /// <param name="onTimeout">A callback to be executed if a timeout occurs.</param>
         /// <returns>True if a timeout occurred. Otherwise, false.</returns>
-        private static async Task<bool> TryHandleTimeoutAsync(Task invokeTask, CancellationToken shutdownToken, bool throwOnTimeout, CancellationToken timeoutToken,
+        private static async Task<bool> TryHandleTimeoutAsync(Task invokeTask, 
+            CancellationToken shutdownToken, bool throwOnTimeout, CancellationToken timeoutToken,
             TimeSpan timeoutInterval, IFunctionInstance instance, Action onTimeout)
         {
             Task timeoutTask = Task.Delay(-1, timeoutToken);
@@ -751,9 +803,26 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
             // These are produced by executing the binders. 
             public object[] InvokeParameters { get; internal set; }
 
+            public object ObjectInstance { get; set; }
+
             public IDictionary<string, ParameterLog> ParameterLogCollector => _parameterLogCollector;
 
             public object ReturnValue => _returnValue;
+
+            // Convert the parameters and their names to a dictionary
+            public Dictionary<string, object> GetParametersAsDictionary()
+            {
+                Dictionary<string, object> parametersAsDictionary = new Dictionary<string, object>();
+
+                int counter = 0;
+                foreach (var name in _parameterNames)
+                {
+                    parametersAsDictionary[name] = InvokeParameters[counter];
+                    counter++;
+                }
+
+                return parametersAsDictionary;
+            }
 
             public IReadOnlyDictionary<string, IWatcher> CreateParameterWatchers()
             {
@@ -994,6 +1063,96 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                     }
 
                     return orderedBinder.StepOrder;
+                }
+            }
+        }
+
+        public class FunctionWithFilterInvoker : IFunctionInvoker
+        {
+            private List<IFunctionInvocationFilter> _filters;
+
+            private IFunctionInvoker _body;
+
+            private FunctionExecutingContext _pre;
+
+            private FunctionExecutedContext _post;
+
+            public IReadOnlyList<string> ParameterNames => _body.ParameterNames;
+
+            public static IFunctionInvoker ApplyFilters(IFunctionInvoker inner, List<IFunctionInvocationFilter> filters, FunctionExecutingContext pre, FunctionExecutedContext post)
+            {
+                if (filters.Count == 0)
+                {
+                    return inner;
+                }
+
+                return new FunctionWithFilterInvoker
+                {
+                    _body = inner,
+                    _filters = filters,
+                    _pre = pre, 
+                    _post = post
+                };
+            }
+
+            public object CreateInstance()
+            {
+                return _body.CreateInstance();
+            }
+
+            public async Task<object> InvokeAsync(object instance, object[] arguments)
+            {
+                // Invoke the filter pipeline.
+                // Basic rules:
+                // - Iff a Pre filter runs, then run the corresponding post. 
+                // - if a pre-filter fails, short circuit the pipeline. Skip the remaining pre-filters and body.
+
+                CancellationToken token;
+
+                int len = _filters.Count;
+                int highestSuccessfulFilter = -1;
+
+                try
+                {
+                    for (int i = 0; i < len; i++)
+                    {
+                        var filter = _filters[i];
+
+                        await filter.OnExecutingAsync(_pre, token);
+
+                        highestSuccessfulFilter = i;
+                    }
+
+                    var result = await _body.InvokeAsync(instance, arguments);
+                    return result;
+                }
+                finally
+                {
+                    Exception exception = null;
+
+                    // Run post filters in reverse order. 
+                    // Only run the post if the corresponding pre executed successfully. 
+                    for (int i = highestSuccessfulFilter; i >= 0; i--)
+                    {
+                        var filter = _filters[i];
+
+                        try
+                        {
+                            await filter.OnExecutedAsync(_post, token);
+                        }
+                        catch (Exception e)
+                        {
+                            exception = e;
+                        }
+                    }
+
+                    // last exception wins. 
+                    // If the body threw, then the finally will automatically rethrow that. 
+                    // If a post-filter throws, capture that 
+                    if (exception != null)
+                    {
+                        throw exception;
+                    }
                 }
             }
         }
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
@@ -521,13 +521,8 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                         Dictionary<string, object> properties = new Dictionary<string, object>();
 
                         // Create the context objects for the filter execution
-                        FunctionExecutingContext executingContext = new FunctionExecutingContext(instance.Id, instance.FunctionDescriptor.FullName, parameterHelper.GetParametersAsDictionary(), properties, logger);
-                        executingContext.JobHost = jobHost;
-                        executingContext.Config = config;
-
-                        FunctionExecutedContext executedContext = new FunctionExecutedContext(instance.Id, instance.FunctionDescriptor.FullName, parameterHelper.GetParametersAsDictionary(), properties, logger);
-                        executedContext.JobHost = jobHost;
-                        executedContext.Config = config;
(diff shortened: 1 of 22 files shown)


src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
```diff
@@ -416,6 +416,8 @@ private bool HasJobAttribute(CustomAttributeData attributeData)
                         
             descr.Parameters = parameters;
             descr.TriggerParameterDescriptor = parameters.OfType<TriggerParameterDescriptor>().FirstOrDefault();
+            descr.FilterAttributes = method.GetCustomAttributes().OfType<IFunctionInvocationFilter>();
```
MikeStall: >descr.FilterAttributes =  [](start = 12, length = 25)

Can you move these to the FromMethod() helper above? That's called by this func.

src/Microsoft.Azure.WebJobs.Protocols/FunctionDescriptor.cs
```diff
@@ -69,6 +69,16 @@ public class FunctionDescriptor
         /// </summary>
         [JsonIgnore]
         internal IEnumerable<SingletonAttribute> SingletonAttributes { get; set; }
+
+        /// <summary>
+        /// Gets any <see cref="IFunctionInvocationFilter"/>s for this function
+        /// </summary>
+        internal IEnumerable<IFunctionInvocationFilter> FilterAttributes { get; set; }
```
MikeStall: Add [JsonIgnore] like the properties above.

hhawkins: Done

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
hhawkins: I see, I moved it

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -552,9 +576,119 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+            finally
+            {
+                await InvokeExecutedFilters(filters, parameterHelper, properties, functionCancellationTokenSource, instance, logger, jobHost, config, functionResult);
+                if (invokeException != null)
+                {
+                    throw invokeException;
+                }
+            }
+
+            return invokeTask;
```
mamaso: I believe this is incorrect - we will return a Task<Task<object>>

Do we want to wait for all filters to execute, or execute the filters outside of the async context? This question doesn't even make sense to me, so maybe we should chat it out

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -552,9 +576,119 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+                invokeException = ex;
+                functionResult = new FunctionResult(ex);
+                throw;
+            }
+            finally
+            {
+                await InvokeExecutedFilters(filters, parameterHelper, properties, functionCancellationTokenSource, instance, logger, jobHost, config, functionResult);
+                if (invokeException != null)
+                {
+                    throw invokeException;
```
mamaso: In order to maintain stack traces and context, I think we'll want to capture the exception info - see other examples of

`exceptionInfo = ExceptionDispatchInfo.Capture(exception);`

This will make it appear that the exception originates from this line, when we have better info.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -537,8 +545,24 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+            List<IFunctionInvocationFilter> filters = new List<IFunctionInvocationFilter>();
+            filters = GetFilters(instance.FunctionDescriptor, invoker.Instance);
+
+            // Prepare the invokeTask result
+            Task<object> invokeTask = null;
+
+            if (filters.Any())
+            {
+                // Because there are filters, invoke the function with the filters
+                invokeTask = InvokeWithFiltersAsync(filters, invoker, parameterHelper, invokeParameters, functionCancellationTokenSource, instance,
```
mamaso: instead of splitting the invoke path, I'd suggest always running the invoke filters logic - however, if there are no filters, we don't run anything extra.

Basically, move the logic in `InvokeWithFiltersAsync` to this method

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -750,6 +884,21 @@ public ParameterHelper(IReadOnlyList<string> parameterNames)
 
             public object ReturnValue => _returnValue;
 
+            // Convert the parameters and their names to a dictionary
+            public Dictionary<string, object> GetParametersAsDictionary()
+            {
+                Dictionary<string, object> parametersAsDictionary = new Dictionary<string, object>();
```
mamaso: nit:

```
return _parameterNames.Zip(InvokeParameters, (name, param) => new KeyValuePair(name, param))
  .ToDictionary();
```

I like working with enumerables instead of dealing with indexes or counters if possible, saves some off by one bugs etc

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInvoker.cs
```diff
@@ -12,6 +12,7 @@ internal class FunctionInvoker<TReflected, TReturnValue> : IFunctionInvoker
         private readonly IReadOnlyList<string> _parameterNames;
         private readonly IFactory<TReflected> _instanceFactory;
         private readonly IMethodInvoker<TReflected, TReturnValue> _methodInvoker;
+        private TReflected _instance;
```
mamaso: what is TReflected and Instance supposed to represent?

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,124 @@
...
+
+            if (!string.IsNullOrEmpty(ExecutingFilter))
+            {
+                executingContext.Logger.LogInformation("Executing Function Filter '" + ExecutingFilter + "'");
+
+                await InvokeJobFunctionAsync(ExecutingFilter, executingContext, cancellationToken);
+            }
+            else
+            {
+                await base.OnExecutingAsync(executingContext, cancellationToken);
```
mamaso: these base class calls don't seem to be necessary

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
```diff
@@ -0,0 +1,124 @@
...
+            {
+                await base.OnExecutedAsync(executedContext, cancellationToken);
+            }
+        }
+
+        internal static async Task InvokeJobFunctionAsync<TContext>(string methodName, TContext context, CancellationToken cancellationToken) where TContext : FunctionInvocationContext
+        {
+            MethodInfo methodInfo = null;
+
+            foreach (var type in context.Config.TypeLocator.GetTypes())
```
mamaso: Can you make this an extension method on TypeLocator, something like

`static MethodInfo GetFunctionByName(this ITypeLocator typeLocator, string methodName)`

Also, assuming that the `methodName` is constant, I'd recommend caching the MethodInfo, ParameterInfo[], etc in the filter itself so that we don't run expensive reflection calls every execution.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -552,9 +576,119 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+                {
+                    throw invokeException;
+                }
+            }
+
+            return invokeTask;
+        }
+
+
+        private static List<IFunctionInvocationFilter> GetFilters(FunctionDescriptor functionDescriptor, object instance)
```
mamaso: Also, move this to be an extension method for FunctionDescriptors

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -552,9 +576,119 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+            IFunctionInstance instance, ILogger logger, JobHost jobHost, JobHostConfiguration config)
+        {
+            if (filters.Any())
+            {
+                // Create the context
+                FunctionExecutingContext context = new FunctionExecutingContext(instance.Id, instance.FunctionDescriptor.FullName, parameterHelper.GetParametersAsDictionary(), properties, logger);
+                context.JobHost = jobHost;
+                context.Config = config;
+
+                foreach (var filter in filters)
```
mamaso: how is the order of filter execution controlled?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -552,9 +576,119 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+
+            // Create a dictionary that will pass the user properties from the executing filter to the executed filter
+            Dictionary<string, object> properties = new Dictionary<string, object>();
+
+            Task<object> invokeTask = null;
+            try
+            {
+                await InvokeExecutingFilters(filters, parameterHelper, properties, functionCancellationTokenSource, instance, logger, jobHost, config);
+                invokeTask = invoker.InvokeAsync(invokeParameters);
+                functionResult = new FunctionResult(true);
```
mamaso: `FunctionResult(true)`? should this be `FunctionResult(await invokeTask)`?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
MikeStall: >invokeParameters [](start = 87, length = 16)

We shouldn't need to pass this if we're already passing parameterHElper

src/Microsoft.Azure.WebJobs.Host/JobHost.cs
```diff
@@ -95,6 +95,14 @@ public JobHost(JobHostConfiguration configuration)
             _stoppingTokenSource = CancellationTokenSource.CreateLinkedTokenSource(_shutdownTokenSource.Token);
         }
 
+        /// <summary>
+        /// Gets the JobHostConfiguration of this JobHost
+        /// </summary>
+        public JobHostConfiguration Configuration
```
mamaso: where is this property used?

src/Microsoft.Azure.WebJobs.Host/Triggers/TriggeredFunctionInstanceFactory.cs
```diff
@@ -66,6 +66,15 @@ public InvokeWrapper(IFunctionInvoker inner, Func<Func<Task<object>>, Task<objec
                 _inner = inner;
                 _handler = handler;
             }
+
+            public object Instance
```
mamaso: where is this property used?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -552,9 +576,119 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
                 await TryHandleTimeoutAsync(invokeTask, CancellationToken.None, throwOnTimeout, timeoutTokenSource.Token, timerInterval, instance, null);
             }
 
-            object returnValue = await invokeTask;
+            if (invokeTask != null)
```
MikeStall: >if (invokeTask != null) [](start = 12, length = 23)

do we need this?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -552,9 +576,119 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+        }
+
+        private static async Task InvokeExecutedFilters(List<IFunctionInvocationFilter> filters, ParameterHelper parameterHelper,
+            Dictionary<string, object> properties, CancellationTokenSource functionCancellationTokenSource,
+            IFunctionInstance instance, ILogger logger, JobHost jobHost, JobHostConfiguration config, FunctionResult functionResult)
+        {
+            if (filters.Any())
+            {
+                // Create the context
+                FunctionExecutedContext context = new FunctionExecutedContext(instance.Id, instance.FunctionDescriptor.FullName, parameterHelper.GetParametersAsDictionary(), properties, logger, functionResult);
```
MikeStall: yes, this should be ShortName. 

---
In reply to: [128640456](https://github.com/Azure/azure-webjobs-sdk/pull/1244#discussion_r128640456) [](ancestors = 128640456)

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
MikeStall: Actually, this gets even easier. There's not a JobHost.Call(string) overload you can use; which will handle the typelocator for you altogether.  Just use that. 


---
In reply to: [128647339](https://github.com/Azure/azure-webjobs-sdk/pull/1244#discussion_r128647339) [](ancestors = 128647339)

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInvoker.cs
```diff
@@ -48,16 +49,26 @@ public IReadOnlyList<string> ParameterNames
...
+        {
+            get
+            {
+                if (_instance == null)
+                {
+                    _instance = _instanceFactory.Create();
+                }
+                return _instance;
+            }
+        }
```
MikeStall: Why is this being cached?  Before, each call to InvokeAsync() gives a new instance. Now, they'll share.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/InvocationFilterEndToEndTests.cs
```diff
@@ -0,0 +1,637 @@
...
+using Microsoft.Azure.WebJobs.Host.TestCommon;
+using Microsoft.Azure.WebJobs.Logging;
+using Microsoft.Extensions.Logging;
+using Microsoft.WindowsAzure.Storage;
+using Microsoft.WindowsAzure.Storage.Blob;
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Host.EndToEndTests
+{
+    public class InvocationFilterEndToEndTests : IClassFixture<TestFixture>
```
MikeStall: >InvocationFilterEndToEndTests [](start = 17, length = 29)

We should be able to move these to unit tests and decouple from external things like storage. I can show you how.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -552,9 +576,119 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+
+        private static async Task InvokeExecutingFilters(List<IFunctionInvocationFilter> filters, ParameterHelper parameterHelper,
+            Dictionary<string, object> properties, CancellationTokenSource functionCancellationTokenSource, 
+            IFunctionInstance instance, ILogger logger, JobHost jobHost, JobHostConfiguration config)
+        {
+            if (filters.Any())
+            {
+                // Create the context
+                FunctionExecutingContext context = new FunctionExecutingContext(instance.Id, instance.FunctionDescriptor.FullName, parameterHelper.GetParametersAsDictionary(), properties, logger);
+                context.JobHost = jobHost;
```
MikeStall: no, they're internal properties that aren't publicly setable. 

---
In reply to: [128639757](https://github.com/Azure/azure-webjobs-sdk/pull/1244#discussion_r128639757) [](ancestors = 128639757)

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -504,7 +511,7 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
                 try
                 {
                     await InvokeAsync(invoker, parameterHelper, timeoutTokenSource, functionCancellationTokenSource,
-                    throwOnTimeout, timerInterval, instance);
+                    throwOnTimeout, timerInterval, instance, logger, jobHost, config);
```
MikeStall: >, logger, jobHost, config [](start = 59, length = 25)

Let's leave the InvokeAsync() signature untouched and instead handled Filters here. That will also minimize test churn in FunctionExecutorTests.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
hhawkins: It's utilized for the original implementation of invoking the function, InvokeAsync on the invoker inside that InvokeWithFiltersAsync method

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
mamaso: In that case, how about an internal constructor?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
hhawkins: The test Microsoft.Azure.WebJobs.Host.EndToEndTests.AsyncChainEndToEndTests.TraceWriterLogging doesn't log properly and the test fails when the check for null is removed. This is something we can further discuss.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInvoker.cs
hhawkins: It was utilized to get the filters from the instance

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -448,7 +453,7 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
 
             try
             {
-                await ExecuteWithWatchersAsync(instance, parameterHelper, trace, logger, functionCancellationTokenSource);
+                await ExecuteWithWatchersAsync(instance, parameterHelper, trace, logger, functionCancellationTokenSource, _jobHost, _config);
```
hhawkins: The one issue with changing all of these to instance is the InvokeAsync method. Other tests and functions are expecting InvokeAsync to be static.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
hhawkins: Whenever you're free let's talk about this

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/InvocationFilterEndToEndTests.cs
```diff
@@ -0,0 +1,460 @@
...
+            // Setup the test in memory
+            var activator = new FakeActivator();
+            var loggerFactory = new LoggerFactory();
+            var logger = new MyLogger();
+            activator.Add(new FilterOrderTestClass());
+            var host = TestHelpers.NewJobHost<FilterOrderTestClass>(activator, logger);
+            loggerFactory.AddProvider(_loggerProvider);
+            host.Configuration.LoggerFactory = loggerFactory;
+            host.Configuration.TypeLocator = new FakeTypeLocator(typeof(FilterOrderTestClass), typeof(FilterOrderTestClass.SampleClassForFilterOrderTest));
+            var method = typeof(FilterOrderTestClass).GetMethod("Main", BindingFlags.Public | BindingFlags.Instance);
```
MikeStall: > var method = typeof(FilterOrderTestClass).GetMethod("Main", BindingFlags.Public | BindingFlags.Instance); [](start = 11, length = 106)

You should be able to use CAllAsync(string) and pass in the method name directly. No MethodInfo needed!

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/InvocationFilterEndToEndTests.cs
```diff
@@ -0,0 +1,460 @@
...
+                await host.CallAsync(method);
+            }
+            catch (Exception e)
+            {
+                var message = e.Message;
+            }
+            
+            var loggerToTest = _loggerProvider.CreatedLoggers.Where(l => l.Category == LogCategories.Executor).Single();
+            Assert.NotNull(loggerToTest.LogMessages.SingleOrDefault(p => p.FormattedMessage.Contains("A.Both")));
+            //Assert.NotNull(loggerToTest.LogMessages.SingleOrDefault(p => p.FormattedMessage.Contains("B.Both")));
```
MikeStall: > //Assert.NotNull(loggerToTest.LogMessages.SingleOrDefault(p => p.FormattedMessage.Contains("B.Both"))); [](start = 11, length = 104)

dead code?

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/InvocationFilterEndToEndTests.cs
```diff
@@ -0,0 +1,460 @@
...
+            Assert.NotNull(loggerToTest.LogMessages.SingleOrDefault(p => p.FormattedMessage.Contains("C.Both")));
```
MikeStall: What is this testing?

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/InvocationFilterEndToEndTests.cs
```diff
@@ -0,0 +1,460 @@
...
+using Microsoft.Azure.WebJobs.Host.Loggers;
+
+namespace Microsoft.Azure.WebJobs.Host.EndToEndTests
+{
+    public class InvocationFilterEndToEndTests
+    {
+        private readonly TestLoggerProvider _loggerProvider = new TestLoggerProvider();
+        
+        [Fact]
+        public async Task TestFilterOrdering()
```
MikeStall: >TestFilterOrdering [](start = 26, length = 18)

Can you add a comment to each test clearly stating specifically what it is testing?

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/InvocationFilterEndToEndTests.cs
```diff
@@ -0,0 +1,460 @@
...
+        {
+            // Setup the test in memory
+            var activator = new FakeActivator();
+            var loggerFactory = new LoggerFactory();
+            var logger = new MyLogger();
+            activator.Add(new StandardFilterTests());
+            var host = TestHelpers.NewJobHost<StandardFilterTests>(activator, logger);
+            loggerFactory.AddProvider(_loggerProvider);
+            host.Configuration.LoggerFactory = loggerFactory;
+            host.Configuration.TypeLocator = new FakeTypeLocator(typeof(StandardFilterTests));
```
MikeStall: > host.Configuration.TypeLocator = new FakeTypeLocator(typeof(StandardFilterTests)) [](start = 11, length = 82)

This shouldn't be necessary. The helper above already sets the TypeLocator

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
hhawkins: The only thing the methodinfo is used for now is getting the correct parameter name for the context

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/InvocationFilterEndToEndTests.cs
hhawkins: For some of the tests, the TypeLocator information is still required or it won't be able to find the method

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInvoker.cs
MikeStall: >private TReflected _instance; [](start = 8, length = 29)

We should not be caching this.  It breaks semantics about how many times IJobActivator is called

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -503,8 +520,60 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+                    Func<Task> filteredInvocation = filters?.Aggregate(invokeAsync, wrapFunctionWithFilter);
+
+                    await filteredInvocation();
+
+                    // Throw for the sake of error tracing
+                    if (executedContext.Result != null)
+                    {
+                        if (!executedContext.Result.Succeeded)
+                        {
+                            throw executedContext.Result.Exception;
```
mamaso: Instead of re-throwing here which destroys the call stack, use an [ExceptionDispatchInfo](https://msdn.microsoft.com/en-us/library/system.runtime.exceptionservices.exceptiondispatchinfo.capture(v=vs.110).aspx) to capture the exception inside the filteredFunction lambda and then throw (which maintains call stack)

This may change depending on the AggregateException behavior above

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -503,8 +520,60 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+                            try
+                            {
+                                await filter.OnExecutingAsync(executingContext, functionCancellationTokenSource.Token);
+                                await innerFunction();
+                                await filter.OnExecutedAsync(executedContext, functionCancellationTokenSource.Token);
+                            }
+                            catch (Exception e)
+                            {
+                                // How should the exception results be cached?
+                                executedContext.Result = new FunctionResult(e);
```
mamaso: The FunctionResult should probably include a list of exceptions thrown (by invocations, filters, etc) in the event that there are multiple exceptions at different stages (particularly invocation & OnExecuted).

When we've finished invocation we can wrap these in an aggregate exception and throw.

@mathewc @MikeStall thoughts?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
mathewc: Yeah, let's avoid refactoring all this to make it instance, which will result in a bunch of cascading changes. Hamza is just following the existing pattern here.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -494,6 +504,13 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
                 await singleton.AcquireAsync(functionCancellationTokenSource.Token);
             }
 
+            // Check to see if there are any filters to decide which invoke method to call
+            // Note: Only do this if filters weren't checked for on the original invocation
+            List<IFunctionInvocationFilter> filters = new List<IFunctionInvocationFilter>();
+
+            // Get the filters to invoke
+            filters = GetFilters(instance.FunctionDescriptor, invoker.Instance);
```
mathewc: Merge this statement with the one above

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -503,8 +520,60 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+
+                    // Create a dictionary that will pass the user properties from the Executing filter to the Executed filter
+                    Dictionary<string, object> properties = new Dictionary<string, object>();
+
+                    // Create the context objects for the filter execution
+                    FunctionExecutingContext executingContext = new FunctionExecutingContext(instance.Id, instance.FunctionDescriptor.FullName, parameterHelper.GetParametersAsDictionary(), properties, logger);
+                    executingContext.JobHost = jobHost;
+                    executingContext.Config = config;
+
+                    FunctionExecutedContext executedContext = new FunctionExecutedContext(instance.Id, instance.FunctionDescriptor.FullName, parameterHelper.GetParametersAsDictionary(), properties, logger, functionResult);
```
mathewc: You're assigning null to functionResult a few lines above, which means you're passing this null into this constructor. That is incorrect - the constructor shouldn't allow null and should throw ArgumentNullException if a null is passed in.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
mamaso: Update: aspnet doesn't wrap multiple exceptions in AggregateException, thanks @fabiocav for testing that.

In that case, we want to go with my below comment of capturing the last exception in an ExceptionDispatchInfo on the filter ExecutedContext

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
hhawkins: Done

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -524,51 +516,77 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
-                        ExecutionLogicHandler executionLogicHandler = async (filtersToInvoke, functionToInvoke) =>
+                    // Lambda function that wraps filters around the function
+                    Func<Func<Task>, IFunctionInvocationFilter, Func<Task>> wrapFunctionWithFilter = (innerFunction, filter) =>
+                    {
+                        Func<Task> filteredFunction = async () =>
                         {
-                            List<ExecutionLogic> executionTree = new List<ExecutionLogic>();
-
-                            foreach (IFunctionInvocationFilter filter in filtersToInvoke)
+                            try
```
mamaso: I'm pretty sure we can get the desired behavior without the nested try catch. I also don't see a need for multiple FunctionResults - there should only be one, attached to the `executedContext`, which is overwritten any time a new exception is caught

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -1128,5 +1060,92 @@ private static BindStepOrder GetStepOrder(IValueProvider provider)
...
+                            exception = e;
+                        }
+                    }
+
+                    // last exception wins. 
+                    // If the body threw, then the finally will automatically rethrow that. 
+                    // If a post-filter throws, capture that 
+                    if (exception != null)
+                    {
+                        throw exception;
```
mamaso: think we'll want to use an `ExceptionDispatchInfo`

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
MikeStall: I fixed this in /dev and rebased on top of it.

src/Microsoft.Azure.WebJobs.Host/InvokeFunctionFilterAttribute.cs
MikeStall: I added a FilterBindingProvider to address that.

mathewc:
Merged

PR #1250, opened 2017-07-21
Adding appsettings so sample shows how to configure storage

damienpontifex:
The sample doesn't have appsettings configuration and appropriate project file copy task. So adding it to have a more complete sample project

### DIFF
--- a/sample/SampleHost/appsettings.json
+++ b/sample/SampleHost/appsettings.json
@@ -1,0 +1,6 @@
+{
+    "ConnectionStrings": {
+        "AzureWebJobsDashboard": "Your connection string here",
+        "AzureWebJobsStorage": "Your connection string here"
+    }
+}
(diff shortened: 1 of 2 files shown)


sample/SampleHost/appsettings.json
```diff
@@ -0,0 +1,6 @@
+{
+    "ConnectionStrings": {
+        "AzureWebJobsDashboard": "UseDevelopmentStorage=true",
+        "AzureWebJobsStorage": "UseDevelopmentStorage=true"
```
mamaso: wary of `UseDevelopmentStorage=true` as we don't fully support storage emulators + I doubt that it would work cross plat

damienpontifex: Agreed. It was just to demonstrate usage and replacement. Would it be better to put some sort of description there? Like `"AzureWebJobsDashboard": "Your connection string here",`

mamaso: Yep, I think that's reasonable - lets make that change and get this in :)

damienpontifex: @mamaso updated setting values and also rebased from dev to get it up to date with source

PR #1263, opened 2017-07-27
[AppInsights] adding additional e2e tests

brettsam:


fabiocav:
Do we need to make sure these changes are ported to `dev`? is this testing surface we've removed?

PR #1265, opened 2017-07-27
[AppInsights] using start/stop on RequestTelemetry to support Profiler

### LINKED ISSUE #1147: [AppInsights] Use Start() and Stop() with RequestTelemetry for profiler
App Insights profiler requires events emitted from Start() and Stop() calls on RequestTelemetry. Details on Profiler: https://docs.microsoft.com/en-us/azure/application-insights/app-insights-profiler

brettsam:
Fixes #1147 

In order to support the App Insights Profiler, we need to emit Start/StopOperation(). This change now starts the operation just before the function begins, stores the operation in the current scope, and then stops the operation when we log the final result. The end result should be the same as before this change, but now the App Insights Profiler can work as expected.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
@@ -3,11 +3,13 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.Linq;
 using System.Net.Http;
 using System.Web;
 using Microsoft.ApplicationInsights;
 using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.ApplicationInsights.Extensibility;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
@@ -18,6 +20,7 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
         private readonly string _categoryName;
         private const string DefaultCategoryName = "Default";
         private const string DateTimeFormatString = "yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'fffK";
+        private const string OperationContext = "MS_OperationContext";
 
         public ApplicationInsightsLogger(TelemetryClient client, string categoryName)
         {
@@ -44,8 +47,8 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
             // for all telemetry.
             using (BeginScope(new Dictionary<string, object>
             {
-                [LoggingKeys.CategoryName] = _categoryName,
-                [LoggingKeys.LogLevel] = (LogLevel?)logLevel
+                [LogConstants.CategoryNameKey] = _categoryName,
+                [LogConstants.LogLevelKey] = (LogLevel?)logLevel
             }))
             {
                 // Log a function result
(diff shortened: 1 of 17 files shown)


src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -301,7 +314,32 @@ public IDisposable BeginScope<TState>(TState state)
...
+            string functionInvocationId = stateValues.GetValueOrDefault<string>(ScopeKeys.FunctionInvocationId);
+            string eventName = stateValues.GetValueOrDefault<string>(ScopeKeys.Event);
+
+            // If we have the invocation id, function name, and event, we know it's a new function. That means
+            // that we want to start a new operation and let App Insights track it for us.
+            if (!string.IsNullOrEmpty(functionName) &&
+                !string.IsNullOrEmpty(functionInvocationId) &&
+                eventName == LogConstants.FunctionStartEvent)
+            {
+                RequestTelemetry request = new RequestTelemetry();
```
mamaso: Instead of the conditional check on state, I think I'd prefer to have a separate `BeginInvocationScope` method that adds the function invocation state, starts the operation, and delegates the rest to normal `BeginScope`. Is that possible?

brettsam: Might be -- let me give it a try.

fabiocav:
Should this be targeting `dev` instead?

brettsam:
@fabiocav -- It really should be in both `dev` and `v2.x`. Is there a smooth way to do that, or just create two PRs? I'll wait to get my test changes into `dev` (#1266) before I submit this, as I want to make sure the tests continue passing as expected.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
brettsam: Actually, since we're dealing with the `ILogger` interface here, there's really no way to jump in and do something specific like this. However, in `BeginScope`, I did separate things out in the last commit so maybe that makes it a little easier to follow?

Note -- this complication is exactly the reason I didn't implement it this way in the first place. It's easier in the `ILogger` world to handle this as a 1-part operation and create the telemetry ourselves after the invocation is done. Breaking it into multiple parts complicates things, as you can see. But the benefits of getting 'free' Profiler support are worth it.

mamaso: Ah yep, thanks for the clarification. Agreed that the free Profiler support is worthwhile

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsightsEndToEndTests.cs
```diff
@@ -25,6 +28,8 @@ public class ApplicationInsightsEndToEndTests
     {
         private const string _mockApplicationInsightsUrl = "http://localhost:4005/v2/track/";
         private const string _mockQuickPulseUrl = "http://localhost:4005/QuickPulseService.svc/";
+
+        private readonly TestTelemetryChannel _channel = new TestTelemetryChannel();
```
brettsam: Most of the changes here are to get the e2e tests (except the QuickPulse) using a test Channel, rather than an HttpListener. This is more reliable going forward and has already been changed in `dev`.

PR #1271, opened 2017-08-02
Caching assembly name resolution

fabiocav:
These are changes proposed by Kranthi based on results from his performance tests.

The `Assembly.GetName()` calls throughout the code are fairly expensive, and caching the result has a measurable impact.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/AssemblyNameCache.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/AssemblyNameCache.cs
@@ -1,0 +1,31 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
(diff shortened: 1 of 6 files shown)


src/Microsoft.Azure.WebJobs.Host/AssemblyNameCache.cs
```diff
@@ -0,0 +1,31 @@
...
+    public static class AssemblyNameCache
+    {
+        private static readonly ConcurrentDictionary<Assembly, AssemblyName> AssemblyToNameCache = new ConcurrentDictionary<Assembly, AssemblyName>();
+
+        /// <summary>
+        /// Returns a cached copy of the given Assembly's AssemblyName if available, otherwise, retrieves the AssemblyName, caches it and returns it.
+        /// </summary>
+        /// <param name="assembly">The Assembly</param>
+        /// <returns>The AssemblyName</returns>
+        public static AssemblyName GetAssemblyName(this Assembly assembly)
```
MikeStall: Can we give this a more intuitive name, like GetCachedAssemblyName(). Especially useful as an extension method.

PR #1274, opened 2017-08-02
[AppInsights] using start/stop on RequestTelemetry to support Profiler

brettsam:
This is the `dev` version of #1265.
 
Also pulling in the change @MikeStall made to fix logger scope (#1248). That change only went into `v2.x`.


brettsam:
Nothing massive -- some slight changes here and there due to missing functionality in Core. Also the verification changed a little b/c some stuff in AppInsights 2.4 changed.

PR #1283, opened 2017-08-04
[AppInsights] adding a LogMetric ILogger extension; attach custom scope properties

### LINKED ISSUE #1264: [AppInsights] DictionaryLoggerScope needs to handle FormattedLogValues correctly
If someone does this in their function: 

    using (logger.BeginScope("{MyKey}", "SomeScope")) 
    {
        // Do some logging. 
    }
We should be attaching MyKey : SomeScope as custom properties to any logging they do inside that scope. But we're not -- we only look for `IDictionary<string, object>` while this produces an `IEnumerable<KeyValuePair<string, object>>`.

### LINKED ISSUE #1281: Add a LogMetric extension method to ILogger
You should be able to log metrics from within your function and have all of the proper correlation that you do with traces. Right now it's a pain as you need to construct your telemetry in a very specific way. We can make this agnostic of App Insights and just add an extension method on ILogger that does the right thing for App Insights.

### LINKED ISSUE #1282: [AppInsights] attach custom scope properties to traces
Spawned from: https://github.com/Azure/azure-webjobs-sdk-script/issues/1675.
Also related to #1264 

Function authors should be able to use scope like this and have those properties attached to all of their traces inside the scope:

```csharp
using(logger.BeginScope(new Dictionary<string, object>
{
    ["CustomKey"] = "Custom Value"
}))
{
   //do a bunch of logging
}
```

brettsam:
Fixes #1281 
Fixes #1282
Fixes #1264

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.Globalization;
 using System.Linq;
 using Microsoft.ApplicationInsights;
 using Microsoft.ApplicationInsights.DataContracts;
@@ -19,9 +20,26 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
     {
         private readonly TelemetryClient _telemetryClient;
         private readonly string _categoryName;
+
         private const string DefaultCategoryName = "Default";
         private const string DateTimeFormatString = "yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'fffK";
         private const string OperationContext = "MS_OperationContext";
+                
+        internal const string MetricCountKey = "count";
+        internal const string MetricMinKey = "min";
+        internal const string MetricMaxKey = "max";
+        internal const string MetricStandardDeviationKey = "standarddeviation";
+
+        private static readonly string[] SystemScopeKeys =
+            {
+                LogConstants.CategoryNameKey,
+                LogConstants.LogLevelKey,
+                LogConstants.OriginalFormatKey,
+                ScopeKeys.Event,
+                ScopeKeys.FunctionInvocationId,
+                ScopeKeys.FunctionName,
+                OperationContext
+            };
 
         public ApplicationInsightsLogger(TelemetryClient client, string categoryName)
         {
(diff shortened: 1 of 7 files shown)


brettsam:
Also now fixes #1264.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -78,13 +95,78 @@ public ApplicationInsightsLogger(TelemetryClient client, string categoryName)
...
+                }
+
+                // Name and Value are not lower-case so check for them explicitly first and move to the
+                // next entry if found.
+                switch (entry.Key)
+                {
+                    case LogConstants.NameKey:
+                        telemetry.Name = entry.Value.ToString();
+                        continue;
+                    case LogConstants.MetricValueKey:
```
mamaso: I think we should also accept `Sum` key here, instead of only `Value`

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -78,13 +95,78 @@ public ApplicationInsightsLogger(TelemetryClient client, string categoryName)
...
+            _telemetryClient.TrackMetric(telemetry);
+        }       
+
+        // Applies custom scope properties; does not apply 'system' used properties
+        private static void ApplyCustomScopeProperties(ISupportProperties telemetry)
+        {
+            var scopeProperties = DictionaryLoggerScope.GetMergedStateDictionary()
+                .Where(p => !SystemScopeKeys.Contains(p.Key, StringComparer.Ordinal));
+
+            foreach (var scopeProperty in scopeProperties)
```
mamaso: just use the ApplyProperties overload here

brettsam: Oh yeah, duh. :-)

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
brettsam: That really wouldn't be usable with `LogMetric()` as you'd have to ignore the `value` parameter and add a `sum` entry to the properties. And we'd need to say `sum` wins if there's a conflict between `sum` and `value`. It feels more confusing that way. I will have to document the App Insights behavior -- I'll do that in the wiki once this is all in.

PR #1285, opened 2017-08-08
Fix unable to bind custom CancellationTokens

Thnslj:
Fix being unable to call an instance of the `JobHost `with a custom `CancellationToken`.

```
CancellationToken customCancellationToken = cancellationTokenSource.Token;
_jobHost.Call(testFunction, new {customCancellationToken});
```
This currently gives an `Unable to convert value to CancellationToken.` error.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/Cancellation/CancellationTokenBinding.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/Cancellation/CancellationTokenBinding.cs
@@ -33,7 +33,7 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings.Cancellation
 
         public Task<IValueProvider> BindAsync(object value, ValueBindingContext context)
         {
-            if (value is CancellationToken)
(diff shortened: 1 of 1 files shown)


src/Microsoft.Azure.WebJobs.Host/Bindings/Cancellation/CancellationTokenBinding.cs
```diff
@@ -33,7 +33,7 @@ private Task<IValueProvider> BindAsync(CancellationToken token, ValueBindingCont
 
         public Task<IValueProvider> BindAsync(object value, ValueBindingContext context)
         {
-            if (value is CancellationToken)
```
mamaso: nit: 
```
CancellationToken token = value as CancellationToken
if (token == null) {
//throw
}
`

Thnslj: It dosen't look like we can use the `as` operator on a struct:
image

PR #1286, opened 2017-08-08
updating StyleCop to v5.0.0

brettsam:
There was some C# 7 syntax that we weren't able to use b/c StyleCop didn't like it. The new version came out recently and appears to fix those issues.

Edit: @mamaso reminded me that we actually can't use C# 7 in this repo due to the build server. However, we should still move StyleCop forward so I changed this to *also* restrict our build to C# 6 syntax. Once that restriction is lifted, we'll be all set with new StyleCop too.

Edit 2: Now that the C# 6 restriction is lifted in the internal build, I've also removed it here. So now this is **only a StyleCop update**


brettsam:
Note this is what @fabiocav recommended in #1181 -- restrict to C# 6. But @mamaso is going to try to get us unblocked here. Pulling code changes between `dev` and `v2.x` where one can use C# 7 and one can't is rough :-)

fabiocav:
This will be rough with the `dev` -> `v2.x` flow and cause a lot of headaches. We need to solve our CI issue and remove the C# 7.0 restriction.

mamaso:
if we get in #1294 we're unblocked for C# 7

brettsam:
Removed restriction to C# 6. Now this is just a StyleCop update, which will let us use some of the newly-added syntax without StyleCop issuing an error.

PR #1289, opened 2017-08-09
Event hub batching - can only send in a single partition.

MikeStall:
EventHub ASync collector must sort by partition key.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubAsyncCollector.cs
+++ b/src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubAsyncCollector.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.ServiceBus.Messaging;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubAsyncCollector.cs
```diff
@@ -112,7 +112,35 @@ public EventHubAsyncCollector(EventHubClient client)
         /// Send the batch of events.
         /// </summary>
         /// <param name="batch">the set of events to send</param>
-        protected virtual async Task SendBatchAsync(EventData[] batch)
+        protected async Task SendMultipleBatchAsync(EventData[] batch)
+        {
+            // Must sort by partition key. 
+            // All items in a call to Send() must be in the same partition. 
+            Dictionary<string, List<EventData>> buckets = new Dictionary<string, List<EventData>>();
```
mamaso: thoughts on using a ConcurrentDictionary instead of the locked _list in the `AddAsync` call to build these partition buckets, rather than after the fact? Seems like we're adding an extra step here, unless if you think this limitation may change in the future

mamaso (commented):
Where is this limitation documented / do you have an issue for it?

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubAsyncCollector.cs
```diff
@@ -18,16 +19,13 @@ public class EventHubAsyncCollector : IAsyncCollector<EventData>
     {
         private readonly EventHubClient _client;
 
-        private List<EventData> _list = new List<EventData>();
-
-        // total size of bytes in _list that we'll be sending in this batch. 
-        private int _currentByteSize = 0;
-
+        private readonly Dictionary<string, PartitionCollector> _partitions = new Dictionary<string, PartitionCollector>();
```
mamaso: I'd prefer to see ConcurrentDictionary - is there any compelling reason we're managing the locking ourselves?

MikeStall: author preference. Plus usage at line 81 where we remove the first item.

PR #1292, opened 2017-08-11
Function Filter changes

mathewc:
Taking care of some loose ends with the new Function Invocation Filters feature. Main changes in this review:

- adding a new filter type for errors/exceptions (see FunctionExceptionFilterAttribute and related classes)
- temporarily removing some of the stuff that was added for Azure Functions integration. We'll re-add that back when we address https://github.com/Azure/azure-webjobs-sdk-script/issues/1761
- bug fixes
- some attribute/interface renames


src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInstance.cs
```diff
@@ -51,6 +52,18 @@ public IFunctionInvoker Invoker
             get { return _invoker; }
         }
 
+        public object JobInstance
```
mathewc: The job instance was needed earlier in the pipeline for exception filters, so I'm moving to a create/cache model so other places in the pipeline can access the cached instance

src/Microsoft.Azure.WebJobs.Host/Filters/FunctionExceptionContext.cs
```diff
@@ -0,0 +1,32 @@
...
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Host.Filters
+{
+    /// <summary>
+    /// Base context class for <see cref="IFunctionExceptionFilter.OnExceptionAsync(FunctionExceptionContext, System.Threading.CancellationToken)"/>.
+    /// </summary>
+    public class FunctionExceptionContext : FunctionFilterContext
```
mathewc: Because exception filters can run before any input arguments are bound, the base class isn't FunctionInvocationContext (i.e. no Arguments dictionary). We might try to conditionally expose the arguments if the exception happened after they were bound. Thoughts?

src/Microsoft.Azure.WebJobs.Host/Filters/FunctionExecutedContext.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
```
mathewc: I moved all the Filters stuff under a new Host/Filters directory. That's why many of these show as additions instead of modifications.

src/Microsoft.Azure.WebJobs.Host/Filters/FunctionInvocationContext.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        /// <summary>
+        /// Gets the function arguments.
+        /// </summary>
+        public IReadOnlyDictionary<string, object> Arguments { get; }
+
+        /// <summary>
+        /// Gets the property bag that can be used to pass information between filters.
+        /// </summary>
+        public IDictionary<string, object> Properties { get; }
```
mathewc: I'm thinking I need to move this to the base FunctionFilterContext so exception filters also have access to this property bag. I think all filters should have access to the shared bag for the invocation scope. In ASP.NET, this can be done via ActionContext.HttpContext.Items which is available to all filter types

src/Microsoft.Azure.WebJobs.Host/Filters/FunctionInvocationFilterAttribute.cs
```diff
@@ -5,12 +5,13 @@
...
 
-namespace Microsoft.Azure.WebJobs.Host
+namespace Microsoft.Azure.WebJobs.Host.Filters
 {
     /// <summary>
     /// Base class for declarative function invocation filters.
     /// </summary>
-    public abstract class InvocationFilterAttribute : Attribute, IFunctionInvocationFilter
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = true, Inherited = true)]
+    public abstract class FunctionInvocationFilterAttribute : Attribute, IFunctionInvocationFilter
```
mathewc: Renaming so attribute and interface names are better in sync. Other filters follow the same pattern

src/Microsoft.Azure.WebJobs.Host/FunctionExecutedContext.cs
```diff
@@ -1,18 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the MIT License. See License.txt in the project root for license information.
-
-using Microsoft.Azure.WebJobs.Host.Executors;
```
mathewc: These files all moved to the Filters directory

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/InvocationFilterEndToEndTests.cs
```diff
@@ -1,598 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the MIT License. See License.txt in the project root for license information.
-
-using System;
-using System.Collections.Generic;
```
mathewc: All of these test scenarios are now covered via the unit test suite more concisely

test/Microsoft.Azure.WebJobs.Host.UnitTests/Common/BindToGenericItemTests.cs
```diff
@@ -35,15 +35,8 @@ public void TestConcreteTypeNoConverter()
             TestWorker<ConfigConcreteTypeNoConverter>();
         }
         
-        public class ConfigConcreteTypeNoConverter : IExtensionConfigProvider, ITest<ConfigConcreteTypeNoConverter>
+        public class ConfigConcreteTypeNoConverter : BindingPathAttribute.Extension, ITest<ConfigConcreteTypeNoConverter>
```
mathewc: Merged in tests that Mike added in Hamza's branch

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInstance.cs
```diff
@@ -51,6 +52,18 @@ public IFunctionInvoker Invoker
...
+        {
+            get
+            {
+                if (_jobInstance == null)
+                {
+                    _jobInstance = Invoker.CreateInstance();
```
MikeStall: >_jobInstance = Invoker.CreateInstance(); [](start = 20, length = 40)

Can we just create this upfront in the ctor?
1. Ensures no races. 
2. lets the field be read-only like the others?

src/Microsoft.Azure.WebJobs.Host/Filters/FunctionInvocationContext.cs
MikeStall: Agreed - I think this should be on the base class. 

---
In reply to: [132727087](https://github.com/Azure/azure-webjobs-sdk/pull/1292#discussion_r132727087) [](ancestors = 132727087)

src/Microsoft.Azure.WebJobs.Host/Filters/FunctionInvocationContext.cs
```diff
@@ -0,0 +1,40 @@
...
+            : base(functionInstanceId, functionName, logger)
+        {
+            Arguments = arguments;
+            Properties = properties;
+        }
+
+        /// <summary>
+        /// Gets the function arguments.
+        /// </summary>
+        public IReadOnlyDictionary<string, object> Arguments { get; }
```
MikeStall: >string [](start = 35, length = 6)

Nit - can we comment that the key is the  parameter name?

src/Microsoft.Azure.WebJobs.Host/Filters/FunctionExceptionContext.cs
```diff
@@ -0,0 +1,32 @@
...
+        public FunctionExceptionContext(Guid functionInstanceId, string functionName, ILogger logger, Exception exception)
+            : base(functionInstanceId, functionName, logger)
+        {
+            Exception = exception;
+        }
+
+        /// <summary>
+        /// Gets the <see cref="Exception"/> that caused the function invocation to fail.
+        /// </summary>
+        public Exception Exception { get; }
```
fabiocav: We need to capture the `ExceptionDispatchInfo` in this context.

src/Microsoft.Azure.WebJobs.Host/Filters/FunctionExceptionContext.cs
MikeStall: I think this is good. We can always add more later.  It's safer to not have arguments yet - people may come to rely on them and break when they get an early exception. 

---
In reply to: [132725208](https://github.com/Azure/azure-webjobs-sdk/pull/1292#discussion_r132725208) [](ancestors = 132725208)

src/Microsoft.Azure.WebJobs.Host/Filters/IFunctionInvocationFilter.cs
```diff
@@ -4,23 +4,24 @@
 using System.Threading;
 using System.Threading.Tasks;
 
-namespace Microsoft.Azure.WebJobs.Host
+namespace Microsoft.Azure.WebJobs.Host.Filters
```
MikeStall: >Filters [](start = 39, length = 7)

I'd vote against the sub namespace here. Function authors will be directly consuming this interface and itd be good to be top-level to be more discoverable.

src/Microsoft.Azure.WebJobs.Host/Indexers/DefaultBindingProvider.cs
```diff
@@ -54,7 +54,6 @@ internal static class DefaultBindingProvider
             ContextAccessor<IBindingProvider> bindingProviderAccessor = new ContextAccessor<IBindingProvider>();
             innerProviders.Add(new RuntimeBindingProvider(bindingProviderAccessor));
             innerProviders.Add(new DataBindingProvider());
-            innerProviders.Add(new FunctionFilterBindingProvider());
```
MikeStall: >FunctionFilterBindingProvider [](start = 35, length = 29)

Where did this go?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -567,27 +580,41 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
+        /// <summary>
+        /// Returns the list of filters in the order they should be executed in.
+        /// Filter order is decided by the scope at which the filter is declared.
+        /// The scopes (in order) are: Instance, Global, Class, Method.
+        /// The execution model is "Russian Doll" - Instance filters surround Global filters
+        /// which surround Class filters which surround Method filters. As a result of this nesting,
+        /// for filters with pre/post methods (e.g. <see cref="IFunctionInvocationFilter"/>) the executing portion
+        /// of the filters runs in the reverse order.
+        /// </summary>
+        private static List<TFilter> GetFilters<TFilter>(IExtensionRegistry extensionRegistry, FunctionDescriptor functionDescriptor, object instance) where TFilter : class
```
MikeStall: >class [](start = 167, length = 5)

Could this be more specific. IFunctionFilter?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -153,6 +159,19 @@ public HostOutputMessage HostOutputMessage
             return exceptionInfo != null ? new ExceptionDispatchInfoDelayedException(exceptionInfo) : null;
         }
 
+        private async Task InvokeExceptionFiltersAsync(Exception exception, IFunctionInstance functionInstance, CancellationToken cancellationToken)
+        {
+            var exceptionFilters = GetFilters<IFunctionExceptionFilter>(_extensionRegistry, functionInstance.FunctionDescriptor, functionInstance.JobInstance);
```
MikeStall: >functionInstance.JobInstance [](start = 129, length = 28)

currently, JobInstance is lazy; so this could call IJobActivator which could throw; and then we'd miss the exception.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -31,12 +31,12 @@ internal class FunctionExecutor : IFunctionExecutor
         private readonly IAsyncCollector<FunctionInstanceLogEntry> _functionEventCollector;
         private readonly ILogger _logger;
         private readonly ILogger _resultsLogger;
-        private readonly IJobInvoker _jobInvoker;
+        private readonly IExtensionRegistry _extensionRegistry;
```
MikeStall: >IExtensionRegistry [](start = 25, length = 18)

We shouldn't be passing this object around - too high level (same objection I had to passing JobHostConfig or even JobHost).

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -153,6 +159,19 @@ public HostOutputMessage HostOutputMessage
...
 
+        private async Task InvokeExceptionFiltersAsync(Exception exception, IFunctionInstance functionInstance, CancellationToken cancellationToken)
+        {
+            var exceptionFilters = GetFilters<IFunctionExceptionFilter>(_extensionRegistry, functionInstance.FunctionDescriptor, functionInstance.JobInstance);
+            if (exceptionFilters.Any())
+            {
+                var exceptionContext = new FunctionExceptionContext(functionInstance.Id, functionInstance.FunctionDescriptor.ShortName, _logger, exception);
+                foreach (var exceptionFilter in exceptionFilters)
+                {
+                    await exceptionFilter.OnExceptionAsync(exceptionContext, cancellationToken);
```
MikeStall: >OnExceptionAsync [](start = 42, length = 16)

This is user code - what should we do if it throws?   
1. Keep executing other eh filters?
2. Does that become the new FunctionResult? 
Throwing from an EH filter probably ought to be similar to throwing from a Post filter.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInstance.cs
MikeStall: Actually, maybe just remove this whole property and put JobInstance on the ParameterHelper (where the other instantiated arguments live). That will solve of these issues. 

---
In reply to: [132730667](https://github.com/Azure/azure-webjobs-sdk/pull/1292#discussion_r132730667) [](ancestors = 132730667)

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -130,8 +136,8 @@ public HostOutputMessage HostOutputMessage
                     logCompletedCancellationToken = cancellationToken;
                 }
 
-                await NotifyCompleteAsync(fastItem, functionCompletedMessage.Arguments, exceptionInfo);
-                _resultsLogger?.LogFunctionResult(fastItem);
+                await NotifyCompleteAsync(instanceLogEntry, functionCompletedMessage.Arguments, exceptionInfo);
```
MikeStall: >NotifyCompleteAsync [](start = 22, length = 19)

This is the "Stop Billing" event. An exception coming from the new call to InvokeExceptionFiltersAsync (which is feasible since it calls user code)  might skip this.  The call to InvokeExceptionFiltersAsync should probably be wrapped in a try/catch.

fabiocav:
Why are we targeting v2.x for this work? I thought the original work was based on `dev` and we'd backport.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -494,23 +508,22 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
             {
-                List<IFunctionInvocationFilter> filters = GetFilters(instance.FunctionDescriptor, jobInstance);
-
                 // Create a source specifically for timeouts
                 using (CancellationTokenSource timeoutTokenSource = new CancellationTokenSource())
                 {
                     TimeoutAttribute timeoutAttribute = instance.FunctionDescriptor.TimeoutAttribute;
                     bool throwOnTimeout = timeoutAttribute == null ? false : timeoutAttribute.ThrowOnTimeout;
-                    var timer = StartFunctionTimeout(instance, timeoutAttribute, timeoutTokenSource, traceWriter, logger);
+                    var timer = StartFunctionTimeout(instance, timeoutAttribute, timeoutTokenSource, traceWriter, _logger);
```
MikeStall: >StartFunctionTimeout [](start = 32, length = 20)

The new EH Filter won't be covered by this timeout. (Although neither is parameter binding  - so maybe this timeout is just too low in the stack).

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInstance.cs
mathewc: Good idea, done.

src/Microsoft.Azure.WebJobs.Host/Filters/FunctionInvocationContext.cs
mathewc: Moved to base class and added test

src/Microsoft.Azure.WebJobs.Host/Filters/FunctionInvocationContext.cs
mathewc: done

src/Microsoft.Azure.WebJobs.Host/Filters/FunctionExceptionContext.cs
mathewc: Will make both properties available as in ASP.NET

src/Microsoft.Azure.WebJobs.Host/Filters/IFunctionInvocationFilter.cs
mathewc: Moved all back into the Host namespace

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
mathewc: Done

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
mathewc: No longer referencing extension registry in executor now. Passing in collection of global filters.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
mathewc: I'll track the timeout issue in a bug. I have a list of some of the other remaining items we want to address, and I'll include this

mathewc:
@fabiocav I originally targeted this branch because I was doing the Azure Functions integration at the same time, and AF isn't on Core yet. Anyhow, I'll get all this into Core as well.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
mathewc: JobActivator is no longer lazy - I'm creating it up front early. I'll make sure it is encapsulated in the invocation try/catch so errors are handled, and I'll add a test.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
mathewc: Now handling this the same way as we do for invocation filters. We keep track of the last exception, keep executing filters, and the last exception wins

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
mathewc: InvokeExceptionFiltersAsync is now handling exceptions. That said, this end event should probably have been in a finally block to begin with...

mathewc:
Merged

PR #1293, opened 2017-08-11
Updating package references

fabiocav:


fabiocav:
Revving packages. Will address any concerns if needed.

PR #1294, opened 2017-08-11
Remove jenkins build step for signing

mamaso:


mamaso:
@mathewc @fabiocav @brettsam ready for review

Removed dependency on build on jenkins / azure files / AzureWebJobsInternal repo.

Current steps:
1. Build src binaries
2. Drop dlls in bin/sign folder, zip, upload to blob
3. Add `command;container;blob` to a queue & then build & run tests
4. Jenkins polls queue, grabs zip, signs bit
5. Pushes bits back to `container-signed/blob.zip`
6. Appveyor finishes tests, polls for blob in step 5
7. Copies signed bits back to project bin folders
8. Run Packages.csproj
9. Upload artifacts from packages

I should be able to clean up the project files, but I'd rather wait until this is established before tearing down the old system.

appveyor.yml
```diff
@@ -7,83 +7,11 @@ platform: x86
...
-        "signed build onenote:   https://microsoft.sharepoint.com/teams/AzureWebjobs/_layouts/OneNote.aspx?id=%2Fteams%2FAzureWebjobs%2FSiteAssets%2FAzureWebjobs%20Notebook&wd=target%28Signed%20Sdk%20Build.one%7C0C569AE2-E052-428D-B3C1-8421ED9B1CF8%2F%29"
-    }
 build_script:
-- cmd: msbuild "WebJobs.proj" /t:BuildTestBinaries /logger:"C:\Program Files\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll" /property:Configuration=Release;Platform=AnyCPU;OutputPath=%APPVEYOR_BUILD_FOLDER%\bin
-test_script:
-- cmd: "vstest.console /logger:Appveyor /TestAdapterPath:bin bin/Microsoft.Azure.WebJobs.Host.UnitTests.dll bin/Microsoft.Azure.WebJobs.Host.FunctionalTests.dll bin/Microsoft.Azure.WebJobs.ServiceBus.UnitTests.dll bin/Dashboard.UnitTests.dll \n\nvstest.console /logger:Appveyor /TestAdapterPath:bin bin/Microsoft.Azure.WebJobs.Host.EndToEndTests.dll"
-artifacts:
-- path: bin\Binaries\*.*
-  name: Binaries
+- cmd: msbuild "WebJobs.proj" /t:GetBitsToSign;BuildTestBinaries /logger:"C:\Program Files\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll" /property:Configuration=Release;Platform=AnyCPU
```
brettsam: Does this mean we're signing even PRs now? Is that what we want?

appveyor.yml
```diff
@@ -7,83 +7,11 @@ platform: x86
...
+test: off
```
brettsam: Are tests meant to be off? Are they kicked off elsewhere?

appveyor.yml
mamaso: It does - we can disable this if we choose.

I enabled it so I could test that this PR gets green artifacts

appveyor.yml
mamaso: Tests are off when I was ironing out the last few kinks (they were just slowing down the iterative process), I need to re-enable them before merging

appveyor.yml
brettsam: It makes sense for testing here, but it feels like we should disable it for regular PRs. I may be paranoid, but allowing someone to submit a PR that contains assemblies that will get passed to our signing server made me nervous.

mamaso:
@brettsam disabled for PRs, reenabled tests

brettsam:
I just remembered we do the `full_ci` thing for non-team-members. So, to refresh my memory -- that means that this appveyor.yml isn't in play for them, right? And they'll just get the restricted non-e2e (and no environment variables) test run?

mamaso:
Correct, so this only applies to the main team job

PR #1305, opened 2017-08-17
expanding range for QuickPulse test

brettsam:


mathewc:
Yeah, I was seeing this test fail alot.

PR #1308, opened 2017-08-21
Exposing basic configuration extensibility

fabiocav:
This is a small change to better enable the CLI (and other clients) to more cleanly pass configuration settings.


src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
```diff
@@ -38,18 +39,45 @@ public sealed class JobHostConfiguration : IServiceProvider
...
         /// <param name="dashboardAndStorageConnectionString">The Azure Storage connection string to use.
         /// </param>
         public JobHostConfiguration(string dashboardAndStorageConnectionString)
+            : this(dashboardAndStorageConnectionString, null)
         {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="JobHostConfiguration"/> class, using the
+        /// specified connection string for both reading and writing data as well as Dashboard logging.
```
brettsam: nit: this description doesn't really makes sense for this constructor.

Issue #1310, opened 2017-08-22
Safely short circuit function invocation from invocation filter


kevbite:
I'm trying to create a `IFunctionInvocationFilter` but I want to be able to safely short circuit the invocation of the function from a filter and pass back a return value.

It's currently possible to short circuit it by throwing an exception within the `OnExecutingAsync` method of the filter but sequential filters wont be called and the exception is bubbled up.

It would be useful if within the Invocation filters to be able to set a property that allows skipping of the function execution (similar to the MVC framework)

I could throw something together and submit a PR of an example of what I mean?

kevbite:
It might be useful if the result was also passed in and could be changed by the `OnExecutedAsync` on the filters too.

mamaso:
Looks like there's some value here - asp.net action filters appear to have a similar feature short-circuiting if result is set in a filter, we should try to mirror that behavior.

https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Filters/ActionFilterAttribute.cs#L49

mathewc:
We're tracking some remaining work items for Filters in this issue https://github.com/Azure/azure-webjobs-sdk/issues/1284. For the initial set of functionality we kept things scoped to get feedback. We have it on our TODO list to look at return values from filters, and I updated the issue to also mention short circuiting so we can explore that.

kevbite:
That's great, I look forward to the features! Thanks

sinapis:
Can you please provide status update?

joshua-hayes:
No progress on this halfway through 2019 now?

PR #1311, opened 2017-08-22
Short circuiting inner invoker 

kevbite:
Small PR to show example of short circuiting function for Issue #1310

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
@@ -1171,7 +1171,8 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                         highestSuccessfulFilter = i;
                     }
 
-                    var result = await _innerInvoker.InvokeAsync(instance, arguments);
+                    var result = executingContext.Result ?? await _innerInvoker.InvokeAsync(instance, arguments);
+
                     return result;
                 }
                 catch (Exception ex)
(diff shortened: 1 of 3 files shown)


kevbite:
Seems we also need to set the result value as well

```
parameterHelper.SetReturnValue(returnValue);
```

Then also call `ProcessOutputParameters` to set all the values back on the binders, actually looks like more work than I originally thought.

mamaso:
Looks like there's some value here - asp.net action filters appear to have a similar feature short-circuiting if result is set in a filter, we should try to mirror that behavior.

https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Filters/ActionFilterAttribute.cs#L49

mamaso:
Perhaps we can move that inner invoker logic to outside the invoke filters in order to light this up

kevbite:
Would need to remove this bit of logic in the static `Create` method so it always get executed

```
if (filters.Count == 0)
 {
     return innerInvoker;
}
```

Just seems a bit hacky, would be more sensibly to just try remove that logic away

mathewc:
Earlier versions of the code actually had a Result capability but we scaled back and removed it because it requires more design. I created #1314 to track this work.

kevbite:
Yeah looked simple at first but then after another look it opened a bag of worms! Thanks for information I'll track the progress on  #1314

mathewc:
Thank for sending the sketch PR though. You're right, it is a bag of worms and we have several related scenarios we want to enable with this, so we need the right design in place.

PR #1312, opened 2017-08-22
updating AppInsights to 2.4

brettsam:


src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/app.config
```diff
@@ -3,10 +3,6 @@
   <runtime>
     <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
       <dependentAssembly>
-        <assemblyIdentity name="System.Diagnostics.DiagnosticSource" publicKeyToken="cc7b13ffcd2ddd51" culture="neutral" />
```
mathewc: Wonder why this was removed instead of updated?

brettsam: I pulled this one out manually -- it didn't auto-update, which was confusing, and it wasn't being used anyway.

brettsam:
@mamaso -- core already uses 2.4 (https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/WebJobs.Logging.ApplicationInsights.csproj#L19)

PR #1315, opened 2017-08-24
Bloburl

watashiSHUN:
“function.json”

```
{
  "bindings": [
    {
      "type": "eventGridTrigger",
      "name": "eventGridEvent",
      "direction": "in"
    },
    {
      "type":"blob",
      "name":"myBlob",
      "direction": "in",
      "connection": "ShunTestConnectionString",
      "path": "{data.fileUrl}"
    }
  ],
  "disabled": false
}

```
“run.csx”
```
#r "Microsoft.Azure.WebJobs.Extensions.EventGrid"
using Microsoft.Azure.WebJobs.Extensions.EventGrid;

public static void Run(EventGridEvent eventGridEvent, Stream myBlob, TraceWriter log)
{
    log.Info(eventGridEvent.ToString());
    StreamReader sr = new StreamReader(myBlob);
    log.Info(sr.ReadToEnd());
}

```

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Blobs/BindableBlobPath.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Blobs/BindableBlobPath.cs
@@ -12,19 +12,16 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs
         public static IBindableBlobPath Create(string pattern, bool isContainerBinding = false)
         {
             BlobPath parsedPattern = null;
-            string containerBlob = null;
-            // if pattern is a blob url, we extract the container/blob
-            if (!BlobPath.TryConvertAbsUrlToContainerBlob(pattern, out containerBlob))
+            // try blob/container pattern from url or string
+            if (!BlobPath.TryParseAbsUrl(pattern, out parsedPattern) &&
+                !BlobPath.TryParse(pattern, isContainerBinding, out parsedPattern))
             {
-                containerBlob = pattern;
-            }
-
-            if (!BlobPath.TryParse(containerBlob, isContainerBinding, out parsedPattern))
-            {
+                // if failed => not blob/container pattern
+                // we try pattern url
                 BindingTemplate urlTemplate = BindingTemplate.FromString(pattern);
                 if (urlTemplate.ParameterNames.Count() == 1)
                 {
-                    return new ParameterizedBlobPath(urlTemplate);
+                    return new ParameterizedBlobUrl(urlTemplate);
                 }
                 throw new FormatException($"Invalid blob path '{pattern}'. Paths must be in the format 'container/blob' or 'blob Url'.");
             }
--- a/src/Microsoft.Azure.WebJobs.Host/Blobs/BindableBlobPath.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Blobs/BindableBlobPath.cs
@@ -11,18 +11,15 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs
     {
         public static IBindableBlobPath Create(string pattern, bool isContainerBinding = false)
         {
-            var containerBlob = BlobPath.TryConvertAbsUrlToContainerBlob(pattern);
             BlobPath parsedPattern = null;
-            if (!string.IsNullOrEmpty(containerBlob))
+            string containerBlob = null;
+            // if pattern is a blob url, we extract the container/blob
+            if (!BlobPath.TryConvertAbsUrlToContainerBlob(pattern, out containerBlob))
             {
-                parsedPattern = BlobPath.Parse(containerBlob, isContainerBinding);
-            }
-            else
-            {
-                parsedPattern = BlobPath.Parse(pattern, isContainerBinding);
+                containerBlob = pattern;
             }
 
-            if (parsedPattern == null)
+            if (!BlobPath.TryParse(containerBlob, isContainerBinding, out parsedPattern))
             {
                 BindingTemplate urlTemplate = BindingTemplate.FromString(pattern);
                 if (urlTemplate.ParameterNames.Count() == 1)
--- a/src/Microsoft.Azure.WebJobs.Host/Blobs/BindableBlobPath.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Blobs/BindableBlobPath.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using System;
 using System.Linq;
 using Microsoft.Azure.WebJobs.Host.Bindings.Path;
 
@@ -10,7 +11,27 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs
     {
         public static IBindableBlobPath Create(string pattern, bool isContainerBinding = false)
         {
-            BlobPath parsedPattern = BlobPath.Parse(pattern, isContainerBinding);
+            var containerBlob = BlobPath.TryConvertAbsUrlToContainerBlob(pattern);
+            BlobPath parsedPattern = null;
+            if (!string.IsNullOrEmpty(containerBlob))
+            {
+                parsedPattern = BlobPath.Parse(containerBlob, isContainerBinding);
+            }
+            else
+            {
+                parsedPattern = BlobPath.Parse(pattern, isContainerBinding);
+            }
+
+            if (parsedPattern == null)
+            {
+                BindingTemplate urlTemplate = BindingTemplate.FromString(pattern);
+                if (urlTemplate.ParameterNames.Count() == 1)
+                {
+                    return new ParameterizedBlobPath(urlTemplate);
+                }
+                throw new FormatException($"Invalid blob path '{pattern}'. Paths must be in the format 'container/blob' or 'blob Url'.");
+            }
+
             BindingTemplate containerNameTemplate = BindingTemplate.FromString(parsedPattern.ContainerName);
             BindingTemplate blobNameTemplate = BindingTemplate.FromString(parsedPattern.BlobName);
 
--- a/src/Microsoft.Azure.WebJobs.Host/Blobs/BlobPath.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Blobs/BlobPath.cs
@@ -55,39 +55,32 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs
             return path;
         }
 
-        public static string TryConvertAbsUrlToContainerBlob(string value)
+        // throws exception if failed
+        public static string ConvertAbsUrlToContainerBlob(string blobUrl)
         {
             try
             {
-                Uri uri = new Uri(value); // default to UriKind.Absolute
+                Uri uri = new Uri(blobUrl); // default to UriKind.Absolute
                 return uri.LocalPath.Substring(1); // [0] is slash, same logic as ToBlobPath() in StorageAnalyticsLogEntry.cs
             }
             catch (Exception)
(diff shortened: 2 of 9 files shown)


src/Packages/Packages.csproj
```diff
@@ -5,7 +5,7 @@
     <Configuration Condition=" '$(Configuration)' == '' ">Release</Configuration>
     <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
     <SkipCopyBuildProduct>true</SkipCopyBuildProduct>
-    <ProjectGuid>{34AB8F63-18DE-4E0D-B21C-15E33B091634}</ProjectGuid>
+    <ProjectGuid>{064F4B52-09DB-4882-9EE4-16B1AEFB4F9B}</ProjectGuid>
```
fabiocav: Why is this changing? Was the project removed/re-added?

src/Microsoft.Azure.WebJobs.Host/Blobs/ParameterizedBlobPath.cs
```diff
@@ -21,16 +31,33 @@ public ParameterizedBlobPath(BindingTemplate containerNameTemplate, BindingTempl
...
 
         public string ContainerNamePattern
         {
-            get { return _containerNameTemplate.Pattern; }
+            get
+            {
+                if (_urlBindingTemplate != null)
+                {
+                    // return template string for container
+                    return _urlBindingTemplate.Pattern + ".getContainer()";
```
fabiocav: This isn't clear. What is `.getContainer()` here?

src/Microsoft.Azure.WebJobs.Host/Blobs/BlobPath.cs
```diff
@@ -55,16 +55,39 @@ public static BlobPath ParseAndValidate(string value, bool isContainerBinding =
...
-            if (!TryParse(value, isContainerBinding, out path))
+        // Does not valid containerName and blobName
+        public static void ExtractContainerNameAndBlobName(string src, out string containerName, out string blobName)
+        {
+            if (String.IsNullOrEmpty(src))
             {
-                throw new FormatException($"Invalid blob path '{value}'. Paths must be in the format 'container/blob'.");
+                throw new FormatException($"can not get container name and blob name from {src}");
             }
+            int slashIndex = src.IndexOf('/');
```
watashiSHUN: we already use our own way of retrieving containerName + blobName from blobUrl
https://github.com/Azure/azure-webjobs-sdk/blob/e692a1ff0db6decf1ba216752c7d86414d4489ca/src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/StorageAnalyticsLogEntry.cs#L134-L135
https://github.com/Azure/azure-webjobs-sdk/blob/26e3e7aaca039d9676f5aca41e7bb088619bb68a/src/Microsoft.Azure.WebJobs.Host/Blobs/BlobPath.cs#L92-L93

I can either call the existing method, or change it to

`var blob = new CloudBlob(new Uri(blobUrl))`
`containerName = blob.Container.Name`
`blobName = blob.Name`

src/Microsoft.Azure.WebJobs.Host/Blobs/ParameterizedBlobPath.cs
watashiSHUN: string representation of a `urlTemplate`is something like `{data.url}`, so a string representation for `containerTemplate` can be `{data.url}.getContainer()`

src/Microsoft.Azure.WebJobs.Host/Blobs/BlobPath.cs
```diff
@@ -55,16 +55,32 @@ public static BlobPath ParseAndValidate(string value, bool isContainerBinding =
             return path;
         }
 
-        public static BlobPath Parse(string value, bool isContainerBinding)
+        // throws exception if failed
+        public static string ConvertAbsUrlToContainerBlob(string blobUrl)
```
MikeStall: >ConvertAbsUrlToContainerBlob [](start = 29, length = 28)

This approach feels wrong. We should create the CloudBlob around the URI directly. That would support sas urls, etc.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Blobs/BlobPathTests.cs
```diff
@@ -8,9 +8,24 @@
 
 namespace Microsoft.Azure.WebJobs.Host.UnitTests.Blobs
 {
-    public class BlobPathTests 
+    public class BlobPathTests
     {
         [Fact]
+        public void CreateBlobWithValidPattern()
```
mathewc: You should also add some E2E tests to the existing BlobBindingEndToEndTests suite. E.g. once SAS URLs and public URLs are supported, have an e2e test demonstrating both.

src/Microsoft.Azure.WebJobs.Host/Blobs/BlobPath.cs
```diff
@@ -55,16 +55,32 @@ public static BlobPath ParseAndValidate(string value, bool isContainerBinding =
...
         {
-            BlobPath path;
-
-            if (!TryParse(value, isContainerBinding, out path))
+            try
+            {
+                Uri uri = new Uri(blobUrl); // default to UriKind.Absolute
+                return uri.LocalPath.Substring(1); // [0] is slash, same logic as ToBlobPath() in StorageAnalyticsLogEntry.cs
+            }
+            catch (Exception)
```
MikeStall: >catch (Exception) [](start = 12, length = 17)

This will add an exception on the hotpath for every [Blob] binding.  Does Uri have a TryParse method or equivalent?

src/Microsoft.Azure.WebJobs.Host/Blobs/ParameterizedBlobPath.cs
MikeStall: I don't understand.  That's not a valid value for this method to return. 

---
In reply to: [135114358](https://github.com/Azure/azure-webjobs-sdk/pull/1315#discussion_r135114358) [](ancestors = 135114358)

src/Microsoft.Azure.WebJobs.Host/Blobs/ParameterizedBlobPath.cs
```diff
@@ -21,16 +31,33 @@ public ParameterizedBlobPath(BindingTemplate containerNameTemplate, BindingTempl
...
 
         public string BlobNamePattern
         {
-            get { return _blobNameTemplate.Pattern; }
+            get
+            {
+                if (_urlBindingTemplate != null)
+                {
+                    // return template string for blob
+                    return _urlBindingTemplate.Pattern + ".getBlob()";
```
MikeStall: >".getBlob()"; [](start = 57, length = 13)

same here.  - The caller will get very confused to receive this.

src/Microsoft.Azure.WebJobs.Host/Blobs/BlobPath.cs
MikeStall: I filed [https://github.com/Azure/azure-webjobs-sdk/issues/1320](https://github.com/Azure/azure-webjobs-sdk/issues/1320) to track this issue. 

---
In reply to: [135612149](https://github.com/Azure/azure-webjobs-sdk/pull/1315#discussion_r135612149) [](ancestors = 135612149)

src/Microsoft.Azure.WebJobs.Host/Blobs/BindableBlobPath.cs
```diff
@@ -10,7 +11,24 @@ internal static class BindableBlobPath
     {
         public static IBindableBlobPath Create(string pattern, bool isContainerBinding = false)
```
MikeStall: >pattern [](start = 54, length = 7)

could pattern include { } and be these? 
{data.url} -- a url 
{container}  - a container 
{container}/{blobname}

MikeStall:
>     internal class ParameterizedBlobPath : IBindableBlobPath

Every method here seems to do a "IsUri" check.  Can we have a new URiParameterizedBlobPath and leave ParameterizedBlobPath unchanged. 

---
Refers to: src/Microsoft.Azure.WebJobs.Host/Blobs/ParameterizedBlobPath.cs:12 in bc170a7. [](commit_id = bc170a7c2f2926fd7c09e446cfc4f526e5664d73, deletion_comment = False)

src/Microsoft.Azure.WebJobs.Host/Blobs/ParameterizedBlobPath.cs
MikeStall: I see  it looks like these results are only used for logging, so we could get away with returning anything. Can you update the comment to make that clear? 

---
In reply to: [135612801](https://github.com/Azure/azure-webjobs-sdk/pull/1315#discussion_r135612801) [](ancestors = 135612801,135114358)

watashiSHUN:
I wasn't able to run e2e test on my local machine...due to some azure subscription issue, does appveyor run these tests?

src/Microsoft.Azure.WebJobs.Host/Blobs/BlobPath.cs
```diff
@@ -55,16 +56,30 @@ public static BlobPath ParseAndValidate(string value, bool isContainerBinding =
             return path;
         }
 
-        public static BlobPath Parse(string value, bool isContainerBinding)
+        // throws exception if failed
+        public static BlobPath ParseAbsUrl(string blobUrl)
```
watashiSHUN: https://github.com/Azure/azure-webjobs-sdk/blob/e692a1ff0db6decf1ba216752c7d86414d4489ca/src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/StorageAnalyticsLogEntry.cs#L131-L136
can be replace by this to use storage sdk for parsing, but I didn't change it since it also parse the nonAbsoluteUrl..

src/Microsoft.Azure.WebJobs.Host/Blobs/ParameterizedBlobUrl.cs
```diff
@@ -0,0 +1,56 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Diagnostics;
+using Microsoft.Azure.WebJobs.Host.Bindings.Path;
+
+namespace Microsoft.Azure.WebJobs.Host.Blobs
+{
+    internal class ParameterizedBlobUrl : IBindableBlobPath
```
MikeStall: >ParameterizedBlobUrl [](start = 19, length = 20)

Can you add a comment like:
This handles values like "{data.url}" which are expected to resolve to a full blob URL.

src/Microsoft.Azure.WebJobs.Host/Blobs/ParameterizedBlobUrl.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        public ParameterizedBlobUrl(BindingTemplate urlBindingTemplate)
+        {
+            Debug.Assert(urlBindingTemplate != null);
+
+            _urlBindingTemplate = urlBindingTemplate;
+        }
+        public string ContainerNamePattern
+        {
+            get { return _urlBindingTemplate.Pattern + ".getContainer()"; }
```
MikeStall: >".getContainer()"; [](start = 55, length = 18)

Can you add a comment like:
"the return value here is just used for logging and does not need to be a valid binding expression"

watashiSHUN:
rebase and merged 3a8439158fdad38193ad8c5d0fbb375231d75682

PR #1316, opened 2017-08-24
[AppInsights] changing to use requestTelemetry.Start()

brettsam:
Moving to AppInsights 2.4 changed how request ids were set when you called `telemetryClient.StartOperation()`. It would give you a new id based on the one you had set, but it would not match exactly. This meant that even though all of the traces for a request were correlated in the App Insights portal UX, they weren't showing up as directly related to the request. Instead, they were showing as part of the root operation (slightly different than before). 

If we use `requestTelemetry.Start()` instead, the request id doesn't get altered and our other important behavior (Profiler support) still works. This changes the start/stop code to use this API instead.

Also cleaned up a few messages I was getting in VS.

This will be ported to `v2.x` as well.


brettsam:
@mamaso -- I'd still like a quick review on this, but I'm going to merge since it fixes an issue that I'd like to move through to Script today.

PR #1333, opened 2017-09-12
[AppInsights] reducing metrics from Aggregator

### LINKED ISSUE #1323: [AppInsights] Reduce duration metrics from Aggregator
Metrics are currently being logged with separate avg/max/min values. We can combine this into one.

brettsam:
Fixes #1323.  

App Insights has support for Max/Min/Sum on each of their metrics. Previously we were sending MaxDuration/MinDuration/AvgDuration as individual metrics. This change now sends a single Duration metric with Count/Min/Max/Sum all populated.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
@@ -21,6 +21,9 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
         private const string DateTimeFormatString = "yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'fffK";
         private const string OperationContext = "MS_OperationContext";
 
+        internal const string CountMetricName = "Count";
+        internal const string SuccessRateMetricName = "Success Rate";
+
         internal const string MetricCountKey = "count";
         internal const string MetricMinKey = "min";
         internal const string MetricMaxKey = "max";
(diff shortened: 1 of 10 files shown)


brettsam:
@mamaso -- this is the metric change you've been asking about for a while :-)

brettsam:
For reference, here's what the data for a single aggregation window used to look like:
image

And here's what it looks like now:
image

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -245,43 +248,79 @@ private static void ApplyProperties(ISupportProperties telemetry, IEnumerable<Ke
...
+            _telemetryClient.TrackMetric(new MetricTelemetry($"{functionName} {LogConstants.SuccessesKey}", successCount));
+            _telemetryClient.TrackMetric(new MetricTelemetry($"{functionName} {LogConstants.FailuresKey}", failureCount));
+            _telemetryClient.TrackMetric(new MetricTelemetry($"{functionName} {SuccessRateMetricName}", successRate));
+        }
+
+        private static double ConvertToDouble(object value)
+        {
+            double convertedValue = 0;
+
+            if (value is int || value is double)
```
mamaso: do we want to allow this for all types, i.e. a string "120"

Could wrap the whole convert switch above in a try catch to handle formatting.

brettsam: For now, we own this "contract" so we know what data to expect. There's no real need to support strings -- we could eventually though.

mamaso: In that case, why not directly use Convert.ToDouble?

brettsam: Being overly-defensive, I guess. I could -- it wouldn't break anything.

PR #1343, opened 2017-09-15
[AppInsights] removing some HTTP details from telemetry

brettsam:
The tests all failed until I added the `HttpRequest` key to the exclusion list. I'm also prepping a private site extension to verify it works.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
@@ -37,6 +37,7 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
                 ScopeKeys.Event,
                 ScopeKeys.FunctionInvocationId,
                 ScopeKeys.FunctionName,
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -37,6 +37,7 @@ internal class ApplicationInsightsLogger : ILogger
                 ScopeKeys.Event,
                 ScopeKeys.FunctionInvocationId,
                 ScopeKeys.FunctionName,
+                ApplicationInsightsScopeKeys.HttpRequest,
```
mamaso: Is there any other scope information that app insights adds which we might want to be cautious about?

Can we have a comment on why we are excluding this information?

brettsam: No, this is really the only thing we add from the Functions side of things. And really it's the only 'important' one to drop. I dropped all the other ones to save space because they were redundant.

Yeah, I'll add some details.

PR #1351, opened 2017-09-20
Make BindingProviders type match exactly to bind

ConnorMcMahon:
If you have binding attributes that are subtypes of another attribute with a binding provider, and the supertype's attribute happens to be higher in the order of binding provider's checked, then the binding provider for that attribute's exact type will never be hit.

This may not be an ideal fix, as you lose the ability to apply a rule on a super type and also have that work for it's subtypes. This goes into what the design should be for binding attributes that inherit from each other, and whether we want to even support that.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/GenericCompositeBindingProvider.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/GenericCompositeBindingProvider.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
         {
             var attr = context.Parameter.GetCustomAttribute<TAttribute>();
 
-            if (attr == null)
+            if (attr == null || attr.GetType() != typeof(TAttribute))
             {
                 return null;
             }
--- a/test/Microsoft.Azure.WebJobs.Host.UnitTests/ExtensionConfigContextTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.UnitTests/ExtensionConfigContextTests.cs
@@ -6,11 +6,14 @@ using System.IO;
 using Microsoft.Azure.WebJobs.Host.Config;
 using Microsoft.Azure.WebJobs.Description;
 using Xunit;
+using Microsoft.Azure.WebJobs.Host.Bindings;
+using System.Collections.Generic;
 
 namespace Microsoft.Azure.WebJobs.Host.UnitTests
 {
     public class ExtensionConfigContextTests
     {
+
         [Fact]
         public void BasicRules()
         {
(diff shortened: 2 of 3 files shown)


src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/GenericCompositeBindingProvider.cs
```diff
@@ -37,7 +37,7 @@ public GenericCompositeBindingProvider(IEnumerable<IBindingProvider> providers)
         {
             var attr = context.Parameter.GetCustomAttribute<TAttribute>();
 
-            if (attr == null)
+            if (attr == null || attr.GetType() != typeof(TAttribute))
```
MikeStall: > || attr.GetType() != typeof(TAttribute)) [](start = 28, length = 41)

we should add a unit test for this too.

ConnorMcMahon: True. I've been holding off on doing more work on this PR, as I am not convinced that this fix is the best way to approach inheritance in binding attributes.

test/Microsoft.Azure.WebJobs.Host.UnitTests/ExtensionConfigContextTests.cs
```diff
@@ -127,11 +130,62 @@ public void ErrorOnDanglingWhen()
...
         }
+
+        [Binding]
+        public class TestAttributeChild : TestAttribute { }
+
+        public class SampleAttributeUsage
+        {
+            public static void InexactAttribute([TestAttributeChild(Mode = "sample")] string input)
+            {
+                Console.Write(input);
```
MikeStall: >Console.Write(input); [](start = 16, length = 21)

Unit tests shouldn't write to Console.

test/Microsoft.Azure.WebJobs.Host.UnitTests/ExtensionConfigContextTests.cs
```diff
@@ -127,11 +130,62 @@ public void ErrorOnDanglingWhen()
...
+            var sampleMethods = typeof(SampleAttributeUsage);
+            {
+                var paramInfo = sampleMethods.GetMethod("InexactAttribute").GetParameters()[0];
+                Dictionary<string, Type> bindingContract = new Dictionary<string, Type>()
+                {
+                    {"input", typeof(string) }
+                };
+                var bindingProvider = typeRule.CreateBinding();
+
+                BindingProviderContext context = new BindingProviderContext(paramInfo, bindingContract, new System.Threading.CancellationToken());
```
MikeStall: >new System.Threading.CancellationToken() [](start = 104, length = 40)

CancellationToken.None

fabiocav:
This PR is pretty old. Doesn't look like the comments have been addressed. Is this still relevant?

ConnorMcMahon:
Having had a discussion with @MikeStall about this issue, it is clear that this is a relatively deep problem that would give us little gain for the amount of work it would take to properly solve this. In the meantime, the workaround is to avoid inheritance structures in binding attributes.

PR #1355, opened 2017-09-22
[AppInsights] Remove parameter logging for RequestTelemetry

### LINKED ISSUE #1352: [AppInsights] Remove parameter details from RequestTelemetry
For WebJobs, you can supply your own `TelemetryClientFactory` that customizes the `RequestTelemetry` we send, but that's not possible today with Functions. One or two simple switches (like `excludeFunctionParameters`) could help functions users customize their payloads until we have full support for rplacing this factory.

brettsam:
Fixes #1352 by removing all parameter logs. In order to log parameter details, a Function can do it from a Trace.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/LoggerExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/LoggerExtensions.cs
@@ -55,14 +55,6 @@ namespace Microsoft.Extensions.Logging
             properties.Add(LogConstants.DurationKey, logEntry.Duration);
             properties.Add(LogConstants.SucceededKey, succeeded);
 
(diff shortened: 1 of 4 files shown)


brettsam:
@paulbatum -- here's the change we talked about.

brettsam:
For now, you need to do it yourself as Traces. Event he opt-in method had limitations. Maybe it's only certain parameters they want logged. Or you want to format the data somehow. Or, since a lot of parameters are JSON objects, you just want to exclude/include a certain property from that object.

I'd really like to allow for a custom TelemetryClientFactory in Functions so all the control can be written in code rather than declarative settings like host.json, but that's a bigger change. This feels like the simplest fix that still allows that data to be logged (albeit manually via Traces).

PR #1362, opened 2017-09-29
Add a sharedQ for distribute the workload

watashiSHUN:
most of the logic is in `src/Microsoft.Azure.WebJobs.Host/Queues/SharedQueueHandler.cs`
the idea is to have a workload queue per host so that a single http request with a batch payload can be processed by multiple instances (in the case of [EventGrid](https://github.com/Azure/azure-functions-eventgrid-extension))

In order to use it, user needs to implement a `messageHandler`:
   ```
 public abstract class MessageHandler
    {
        /// <summary>
        /// Default implementation returns null
        /// </summary>
        public virtual CloudQueue PoisonQueue => null;

        /// <summary>
        /// from JObject to actual function execution
        /// if generic is used, object serialization error might happen at runtime
        /// </summary>
        public abstract Task<FunctionResult> TryExecuteAsync(JObject data, CancellationToken cancellationToken);
```
where they specify 
1, what to do with the message
2, where to put the poison message (after failures, null will store it in host poison queue)

this is capable of replacing the current logic we use for blobtrigger

### DIFF
--- a/test/Microsoft.Azure.WebJobs.Host.EndToEndTests/DispatchQueueEndToEndTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.EndToEndTests/DispatchQueueEndToEndTests.cs
@@ -46,7 +46,7 @@ namespace Microsoft.Azure.WebJobs.Host.EndToEndTests
                 return _total;
             }
 
-            public bool hasDuplicate()
+            public bool HasDuplicate()
             {
                 return _duplicate;
             }
@@ -65,7 +65,8 @@ namespace Microsoft.Azure.WebJobs.Host.EndToEndTests
         public async Task DispatchQueueBatchTriggerTest()
         {
             _hostConfiguration.TypeLocator = new FakeTypeLocator(typeof(SampleTrigger));
-            _hostConfiguration.HostId = "bttest"; // use test name
+            // each test will have a unique hostId so that consecutive test run will not be affected by clean up code
+            _hostConfiguration.HostId = "bttest";
 
             using (var host = new JobHost(_hostConfiguration))
             {
@@ -75,15 +76,13 @@ namespace Microsoft.Azure.WebJobs.Host.EndToEndTests
 
                 _stopwatch.Restart();
 
-                // it take 2secs to trigger the first function
-                // only one dequeue is happening if batchSize*2 < maxPollingInterval 
                 int twoFuncCount = DispatchQueueTestConfig.batchSize * 2;
-                await TestHelpers.Await(() => _funcInvokation.TotalAdd() >= twoFuncCount || _funcInvokation.hasDuplicate(),
-                                        6000, 2000);
+                await TestHelpers.Await(() => _funcInvokation.TotalAdd() >= twoFuncCount || _funcInvokation.HasDuplicate(),
+                                        7000, 1000);
 
                 // make sure each function is triggered once and only once
                 Assert.Equal(twoFuncCount, _funcInvokation.TotalAdd());
-                Assert.False(_funcInvokation.hasDuplicate());
+                Assert.False(_funcInvokation.HasDuplicate());
 
                 _stopwatch.Stop();
             }
@@ -106,10 +105,10 @@ namespace Microsoft.Azure.WebJobs.Host.EndToEndTests
                 // this test takes long since it does at least 5 dequeue on the poison message
                 // count retries caused by failures and poison queue function process
                 int funcWithExceptionCount = DispatchQueueTestConfig.batchSize + _hostConfiguration.Queues.MaxDequeueCount;
-                await TestHelpers.Await(() => _funcInvokation.TotalAdd() >= funcWithExceptionCount, 10000, 2000);
+                await TestHelpers.Await(() => _funcInvokation.TotalAdd() >= funcWithExceptionCount, 10000, 1000);
 
                 Assert.Equal(funcWithExceptionCount, _funcInvokation.TotalAdd());
-                Assert.True(_funcInvokation.hasDuplicate());
+                Assert.True(_funcInvokation.HasDuplicate());
                 Assert.True(SampleTriggerWithPoisonQueue.poisonQueueResult);
 
                 _stopwatch.Stop();
@@ -136,7 +135,7 @@ namespace Microsoft.Azure.WebJobs.Host.EndToEndTests
                 int order = json["order"].Value<int>();
                 string funcSignature = "PoisonQueueTrigger arg: " + order;
                 _funcInvokation.Add(funcSignature);
-                _output.WriteLine(funcSignature + " elapsed time: " + _stopwatch.ElapsedMilliseconds);
+                _output.WriteLine(funcSignature + " elapsed time: " + _stopwatch.ElapsedMilliseconds + " ms");
                 if (order == 0)
                 {
                     throw new Exception("Can't deal with zero :(");
@@ -152,7 +151,7 @@ namespace Microsoft.Azure.WebJobs.Host.EndToEndTests
                     poisonQueueResult = true;
                 }
                 _funcInvokation.Add("PosionQueueProcess");
-                _output.WriteLine("PoisonQueueProcess" + " elapsed time: " + _stopwatch.ElapsedMilliseconds);
+                _output.WriteLine("PoisonQueueProcess" + " elapsed time: " + _stopwatch.ElapsedMilliseconds + " ms");
             }
         }
 
@@ -162,14 +161,14 @@ namespace Microsoft.Azure.WebJobs.Host.EndToEndTests
             {
                 string funcSignature = "DispatchQueueTrigger arg: " + json["order"].Value<int>();
                 _funcInvokation.Add(funcSignature);
-                _output.WriteLine(funcSignature + " elapsed time: " + _stopwatch.ElapsedMilliseconds);
+                _output.WriteLine(funcSignature + " elapsed time: " + _stopwatch.ElapsedMilliseconds + " ms");
             }
 
             public void DispatchQueueTrigger2([DispatchQueueTrigger]JObject json)
             {
                 string funcSignature = "DispatchQueueTrigger2 arg: " + json["order"].Value<int>();
                 _funcInvokation.Add(funcSignature);
-                _output.WriteLine(funcSignature + " elapsed time: " + _stopwatch.ElapsedMilliseconds);
+                _output.WriteLine(funcSignature + " elapsed time: " + _stopwatch.ElapsedMilliseconds + " ms");
             }
         }
     }
--- a/src/Microsoft.Azure.WebJobs.Host/Dispatch/InMemoryDispatchQueueHandler.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Dispatch/InMemoryDispatchQueueHandler.cs
@@ -21,6 +21,7 @@ namespace Microsoft.Azure.WebJobs.Host.Dispatch
         {
             // start executing the function
             // ignore the function result, no retry or poisonQueue
+            // we don't wait on this task, exceptions will be logged by functionExecutor
             _messageHandler.TryExecuteAsync(message, cancellationToken);
             return Task.CompletedTask;
         }

### CODE REVIEW

File: src/Microsoft.Azure.WebJobs.Host/Queues/DispatchProcessor.cs [hunk 1]
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Host.Queues
```
MikeStall: >.Queues [](start = 38, length = 7)

This doesn't seem like the right namespace.   I think .Queue is for Queue  bindings; whereas the fact that this is implemented on a queue is a detail.

src/Microsoft.Azure.WebJobs.Host/Queues/DispatchProcessor.cs
```diff
@@ -0,0 +1,31 @@
...
+        internal DispatchProcessor(SharedQueueHandler sharedQueue, string functionId)
+        {
+            _sharedQueue = sharedQueue;
+            _functionId = functionId;
+        }
+
+        /// <summary>
+        /// enqueue
+        /// </summary>
+        public async Task ProcessAsync(JObject message, CancellationToken cancellationToken)
```
MikeStall: >ProcessAsync [](start = 26, length = 12)

Process implies we do the work right now. Can we rename to Enqueue?

src/Microsoft.Azure.WebJobs.Host/Queues/MessageHandler.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// <summary>
+    /// register function executor in the queue
+    /// </summary>
+    /// TODO generic
+    public abstract class MessageHandler
+    {
+        /// <summary>
+        /// Default implementation returns null
+        /// </summary>
+        public virtual CloudQueue PoisonQueue => null;
```
MikeStall: Can you comment the invariants on these new public classes:
- who is expected to implement this class?
- What does Poison Queue mean here. We've abstracted away the other queue mechanics, why is this exposed?

src/Microsoft.Azure.WebJobs.Host/Queues/SharedQueueHandler.cs
```diff
@@ -0,0 +1,272 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.WindowsAzure.Storage.Queue;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Host.Queues
+{
+    /// <summary>
+    /// can be accessed through listener context
+    /// not exposed to the user
```
MikeStall: Can you update the xml comment mention what is the purpose of this class?

src/Microsoft.Azure.WebJobs.Host/Listeners/ListenerFactoryContext.cs
```diff
@@ -13,21 +15,37 @@ namespace Microsoft.Azure.WebJobs.Host.Listeners
...
+        /// Shared Queue for dispatching
+        /// </summary>
+        public DispatchProcessorFactory DispatchProcessorFactory
+        {
+            get
+            {
+                if (_dispatchProcessorFactory == null)
+                {
+                    _dispatchProcessorFactory = new DispatchProcessorFactory(_sharedQueue, Descriptor.Id);
+                }
```
MikeStall: CAn this be in the constructor? And then the property is trivial.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/BlobListenerFactory.cs
```diff
@@ -12,56 +12,60 @@
 using Microsoft.Azure.WebJobs.Host.Storage.Blob;
 using Microsoft.Azure.WebJobs.Host.Storage.Queue;
 using Microsoft.Azure.WebJobs.Host.Timers;
-using Microsoft.Extensions.Logging;
+using Microsoft.WindowsAzure.Storage.Queue;
+using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
 {
+    // no longer takes argument related to the sharedQ
```
mathewc: ? remove this

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/BlobListenerFactory.cs
```diff
@@ -123,58 +127,49 @@ internal class BlobListenerFactory : IListenerFactory
...
         }
 
         public async Task<IListener> CreateAsync(CancellationToken cancellationToken)
         {
             // Note that these clients are intentionally for the storage account rather than for the dashboard account.
             // We use the storage, not dashboard, account for the blob receipt container and blob trigger queues.
-            IStorageQueueClient primaryQueueClient = _hostAccount.CreateQueueClient();
-            IStorageBlobClient primaryBlobClient = _hostAccount.CreateBlobClient();
+
+            IStorageQueueClient primaryQueueClient = _hostAccount.CreateQueueClient(); // use for default poisonqueue
```
mathewc: nit: remove these new comments on the various clients. My comment above already indicates what they're used for

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/BlobListenerFactory.cs
```diff
@@ -198,50 +197,103 @@ internal class BlobListenerFactory : IListenerFactory
                 // Create a singleton shared blob listener, since we only
                 // want a single instance of the blob poll/scan logic to be running
                 // across host instances
-                var singletonBlobListener = _singletonManager.CreateHostSingletonListener(
-                    new BlobListener(sharedBlobListener), SingletonBlobListenerScopeId);
+                var singletonBlobListener = _singletonManager.CreateHostSingletonListener(sharedBlobListener, SingletonBlobListenerScopeId);
                 _sharedContextProvider.SetValue(SingletonBlobListenerScopeId, true);
-
-                return new CompositeListener(singletonBlobListener, queueListener);
+                // since now that queuelistener is null, we no longer need CompositeListener
```
mathewc: remove comment - this is just a note to code reviewers?

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/BlobListenerFactory.cs
```diff
@@ -198,50 +197,103 @@ internal class BlobListenerFactory : IListenerFactory
...
             SharedBlobListener sharedBlobListener,
             IStorageBlobClient blobClient,
-            IStorageQueue hostBlobTriggerQueue,
-            IMessageEnqueuedWatcher messageEnqueuedWatcher,
+            DispatchProcessor dispatchProcessor,
             CancellationToken cancellationToken)
         {
             BlobTriggerExecutor triggerExecutor = new BlobTriggerExecutor(hostId, _functionId, _input,
-                BlobETagReader.Instance, new BlobReceiptManager(blobClient),
-                new BlobTriggerQueueWriter(hostBlobTriggerQueue, messageEnqueuedWatcher));
```
mathewc: The previous queue writer took the "messageEnqueuedWatcher" so that it could notify the queue listener quickly when messages were added. Are you retaining that capability? It allows the listener to break out quickly from its polling loop.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/BlobListenerFactory.cs
```diff
@@ -198,50 +197,103 @@ internal class BlobListenerFactory : IListenerFactory
...
+        private class BlobMessageHandler : MessageHandler
         {
-            BlobQueueRegistration registration = new BlobQueueRegistration
+            private readonly IStorageBlobClient _blobClient;
+            private readonly ITriggeredFunctionExecutor _executor;
+            private readonly IBlobWrittenWatcher _blobWrittenWatcher;
+            internal BlobMessageHandler(ITriggeredFunctionExecutor executor,
+                                        IStorageBlobClient blobClient,
+                                        IStorageQueueClient queueClient,
+                                        IBlobWrittenWatcher blobWrittenWatcher)
```
mathewc: I don't understand why your message handler is taking this watcher. The semantics before were that the blob scanner would use this to notify when it discovered a new blob. I don't think we need to change that.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/BlobListenerFactory.cs
```diff
@@ -198,50 +197,103 @@ internal class BlobListenerFactory : IListenerFactory
...
-            };
+                _executor = executor;
+                _blobClient = blobClient;
+                _blobWrittenWatcher = blobWrittenWatcher;
+                PoisonQueue = queueClient?.GetQueueReference(HostQueueNames.BlobTriggerPoisonQueue).SdkObject;
+            }
+            public override CloudQueue PoisonQueue { get; }
 
-            sharedBlobQueueListener.Register(_functionId, registration);
+            public override async Task<FunctionResult> TryExecuteAsync(JObject data, CancellationToken cancellationToken)
```
mathewc: This is basically the same logic that previously existed in BlobQueueTriggerExecutor, right? Where is the logic that the functionID is still valid (function still exists)?

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/BlobTriggerExecutor.cs
```diff
@@ -110,7 +128,16 @@ internal class BlobTriggerExecutor : ITriggerExecutor<IStorageBlob>
                     BlobName = value.Name,
                     ETag = possibleETag
                 };
-                await _queueWriter.EnqueueAsync(message, cancellationToken);
+                // BlobTriggerMessage to JObject
+                // we can save two functionid, not a big deal
```
mathewc: ?

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/BlobTriggerExecutor.cs
```diff
@@ -110,7 +128,16 @@ internal class BlobTriggerExecutor : ITriggerExecutor<IStorageBlob>
...
+                if (_queueWriter == null)
```
mathewc: I don't understand this conditional - I assume this is just temporary?

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/SharedBlobListener.cs
```diff
@@ -11,7 +11,8 @@
 
 namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
 {
-    internal sealed class SharedBlobListener : ISharedListener
+    // shun TODO remove ISharedListener
```
mathewc: I'm all for removing unnecessary types. Can you explain why no longer needed?

src/Microsoft.Azure.WebJobs.Host/Blobs/Triggers/BlobTriggerBinding.cs
```diff
@@ -266,10 +266,12 @@ public Task<IListener> CreateListenerAsync(ListenerFactoryContext context)
             }
 
             IStorageBlobContainer container = _blobClient.GetContainerReference(_path.ContainerNamePattern);
+            // blobtriggerbinding uses dispatch
```
mathewc: nit: remove unnecessary comment. Also the local declaration naming/casing is off.

src/Microsoft.Azure.WebJobs.Host/Listeners/ListenerFactoryListener.cs
```diff
@@ -6,21 +6,24 @@
...
         private readonly IListenerFactory _factory;
+        private readonly SharedQueueHandler _sharedQueue;
         private readonly CancellationTokenSource _cancellationSource;
 
         private IListener _listener;
         private CancellationTokenRegistration _cancellationRegistration;
         private bool _disposed;
 
-        public ListenerFactoryListener(IListenerFactory factory)
+        public ListenerFactoryListener(IListenerFactory factory, SharedQueueHandler sharedQueue)
```
mathewc: Shouldn't need a new parameter here. Rather, roll the underlying shared listener into the factory already being passed in.

src/Microsoft.Azure.WebJobs.Host/Queues/SharedQueueHandler.cs
mathewc: Yes - please put nice xml class comments on all your new types

src/Microsoft.Azure.WebJobs.Host/Queues/MessageHandler.cs
mathewc: Agree - we shouldn't have to expose any CloudQueueDetails.

src/Microsoft.Azure.WebJobs.Host/Queues/MessageHandler.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Executors;
+using Microsoft.WindowsAzure.Storage.Queue;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Host.Queues
+{
+    /// <summary>
+    /// register function executor in the queue
```
mathewc: Improve xml comments. These should be comments that make sense for extension authors

src/Microsoft.Azure.WebJobs.Host/Queues/DispatchProcessor.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Host.Queues
+{
+    /// <summary>
+    /// wrapper around shared queue
+    /// </summary>
+    public class DispatchProcessor
```
mathewc: For the types/concepts that are exposed via ListenerFactoryContext to users, I think we should consider better names. Rather than DispatchProcessor/DispatchProcessorFactory, how about MessageQueue/MessageQueueFactory? In the end, from the user's perspective, it really is a general message queue.

src/Microsoft.Azure.WebJobs.Host/Queues/DispatchProcessorFactory.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Host.Queues
+{
+    /// <summary>
+    /// factory that only creates a single instance
```
mathewc: ?

src/Microsoft.Azure.WebJobs.Host/Queues/DispatchProcessorFactory.cs
```diff
@@ -0,0 +1,36 @@
...
+        internal DispatchProcessorFactory(SharedQueueHandler sharedQueue, string functionId)
+        {
+            _functionId = functionId;
+            _instance = null;
+            _sharedQueue = sharedQueue;
+        }
+
+        /// <summary>
+        /// register once created
+        /// TODO register with a different handler???
```
mathewc: ?

src/Microsoft.Azure.WebJobs.Host/Queues/DispatchProcessorFactory.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    /// <summary>
+    /// factory that only creates a single instance
+    /// </summary>
+    public class DispatchProcessorFactory
+    {
+        private readonly string _functionId;
+        private readonly SharedQueueHandler _sharedQueue;
+        private DispatchProcessor _instance;
+        // created in listenerContext, cannot be created by user
```
mathewc: ?

src/Microsoft.Azure.WebJobs.Host/Queues/QueueProcessor.cs
```diff
@@ -238,6 +238,7 @@ protected internal virtual void OnMessageAddedToPoisonQueue(PoisonMessageEventAr
             MessageAddedToPoisonQueue?.Invoke(this, e);
         }
 
+        // TODO the same as StorageQueueExtensions
```
mathewc: ?

src/Microsoft.Azure.WebJobs.Host/Queues/DispatchProcessorFactory.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Host.Queues
+{
+    /// <summary>
+    /// factory that only creates a single instance
+    /// </summary>
+    public class DispatchProcessorFactory
```
mathewc: Naming suggestion: how about DispatchQueueFactory/DispatchQueue? Or perhaps ExecutionQueue or InvocationQueue.

src/Microsoft.Azure.WebJobs.Host/Queues/SharedQueueHandler.cs
```diff
@@ -0,0 +1,272 @@
...
+            {
+                _messageHandler = new ConcurrentDictionary<string, MessageHandler>();
+            }
+            // handle dequed message, execute the function
+            public async Task<FunctionResult> ExecuteAsync(IStorageQueueMessage value, CancellationToken cancellationToken)
+            {
+                QueueMessage message = JsonConvert.DeserializeObject<QueueMessage>(value.AsString, JsonSerialization.Settings);
+                if (message == null)
+                {
+                    throw new InvalidOperationException("Invalid blob trigger message.");
```
mathewc: copy paste error

cgillum:
What are the requirements on the scale controller to work with this shared queue?

src/Microsoft.Azure.WebJobs.Host/Listeners/ListenerFactoryContext.cs
watashiSHUN: It was in the constructor, but I realized that most users of `ListenerFactoryContext` will not use the `dispatchQueue`, so I moved it out to be more lazy

src/Microsoft.Azure.WebJobs.Host/Queues/MessageHandler.cs
watashiSHUN: this is added because blobtrigger has its own logic specifying which poison queue to use
https://github.com/Azure/azure-webjobs-sdk/blob/e648136df61a6f43ac7bc139739e24df9f2fbfef/src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/BlobListenerFactory.cs#L178-L188
Expose CloudQueue is probably too much, it can be replaced by some sort of `storage interface` (has a write method) so that if user of sharedQueue does not want to have their failed messaged stored in the default host poison queue, they can bring their own alternative

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/BlobListenerFactory.cs
watashiSHUN: yes, the logic is applied when we initialize the sharedQ
```
            SharedQueueWatcher sharedQueueWatcher = _sharedContextProvider.GetOrCreateInstance<SharedQueueWatcher>(
                                                        new SharedQueueWatcherFactory(_messageEnqueuedWatcherSetter));
            _sharedQueueWriter = new SharedQueueWriter(sharedQueue, sharedQueueWatcher);
```

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/BlobListenerFactory.cs
watashiSHUN: are you referring to this code
https://github.com/Azure/azure-webjobs-sdk/blob/e648136df61a6f43ac7bc139739e24df9f2fbfef/src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/BlobQueueTriggerExecutor.cs#L52-L70

it is moved to sharedQ, it will be applied for every message before dispatch to their handlers

watashiSHUN:
@cgillum we need the scale controller to monitor this newly added queue

src/Microsoft.Azure.WebJobs.Host/Queues/QueueProcessor.cs
watashiSHUN: https://github.com/Azure/azure-webjobs-sdk/blob/e648136df61a6f43ac7bc139739e24df9f2fbfef/src/Microsoft.Azure.WebJobs.Host/Queues/StorageQueueExtensions.cs#L16-L44

and it's twin brother

https://github.com/Azure/azure-webjobs-sdk/blob/e648136df61a6f43ac7bc139739e24df9f2fbfef/src/Microsoft.Azure.WebJobs.Host/Queues/QueueProcessor.cs#L241-L267

but it looks like the `poison queue` don't have an abstraction (take `CloudQueue` as argument)? my guess is that since `poison queue` is always associated with queuelistener, it only have end to end test

src/Microsoft.Azure.WebJobs.Host/Queues/DispatchProcessorFactory.cs
watashiSHUN: I believe I tried `DispatchQueue` before and got `CA1711	Rename type name 'DispatchQueue' so that it does not end in 'Queue'.` I cannot name it Shun**Queue** if it does not extend `System.Collections.Queue`

src/Microsoft.Azure.WebJobs.Host/Listeners/ListenerFactoryListener.cs
watashiSHUN: hmm the `IListenerFactory` is an interface that has a lot of implementation, adding field feels difficult
I only wants `sharedQueue` here so that I can control **it's start/stop time against other listeners**:
1. start after all other listeners started (this is not a serious concern, since I don't really know when exactly user of `sharedQueue` will register their handler, the best I can do is to start as late as possible, Technically speaking I can start after other listener's creation has finished)
2. stop after all other listeners have stopped (try to not leave unprocessed in the queue, also a best effort)

src/Microsoft.Azure.WebJobs.Host/Queues/DispatchProcessor.cs
watashiSHUN: changed it to "Microsoft.Azure.WebJobs.Host.Dispatch"

src/Microsoft.Azure.WebJobs.Host/Queues/DispatchProcessor.cs
watashiSHUN: changed to "EnqueueAsync"

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/BlobTriggerExecutor.cs
watashiSHUN: TODO change the unit test

src/Microsoft.Azure.WebJobs.Host/Dispatch/MessageHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+    public abstract class MessageHandler
+    {
+        /// <summary>
+        /// This is where we put the failed messages,
+        /// after trying to process couples times without success.
+        /// If you don't provide your own storage (for example, an azure queue),
+        /// by default we are going to put it in the host poison queue .
+        /// <see cref="HostQueueNames.GetHostSharedPoisonQueueName(string)"/>
+        /// </summary>
+        public virtual CloudQueue PoisonQueue => null;
```
MikeStall: This should not be publicly exposed.

test/Microsoft.Azure.WebJobs.Host.UnitTests/PublicSurfaceTests.cs
```diff
@@ -247,7 +247,10 @@ public void WebJobsHostPublicSurface_LimitedToSpecificTypes()
                 "IFunctionExceptionFilter",
                 "IFunctionFilter",
                 "IFunctionInvocationFilter",
-                "LoggerExtensions"
+                "LoggerExtensions",
+                "DispatchQueueFactory",
+                "DispatchQueueHandler",
```
MikeStall: Why are we publicly exposing these?

src/Microsoft.Azure.WebJobs.Host/Listeners/ListenerFactoryContext.cs
```diff
@@ -13,21 +15,39 @@ namespace Microsoft.Azure.WebJobs.Host.Listeners
...
+            }
             Descriptor = descriptor;
             Executor = executor;
+            _sharedQueue = sharedQueue;
             CancellationToken = cancellationToken;
         }
 
         /// <summary>
+        /// Shared Queue for dispatching
+        /// TODO technically we can return a DispatchQueueHandler by calling a function on ListenerFactoryContext,
```
MikeStall: >TODO [](start = 12, length = 4)

We shouldn't be checking in new 'TODO' comments.

src/Microsoft.Azure.WebJobs.Host/Dispatch/IMessageHandler.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Executors;
+using Microsoft.Azure.WebJobs.Host.Listeners;
```
MikeStall: nit: Can we remove these now?

src/Microsoft.Azure.WebJobs.Host/Dispatch/DispatchQueueHandler.cs
```diff
@@ -0,0 +1,36 @@
...
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Host.Dispatch
+{
+    /// <summary>
+    /// Queue where user can add their function triggering messages,
+    /// these messages will be distributed to multiple worker instance for processing
+    /// It is a wrapper around <see cref="SharedQueueHandler"/>
+    /// </summary>
+    public class DispatchQueueHandler
```
MikeStall: > public class DispatchQueueHandler [](start = 3, length = 34)

Nit: Can the exposes type be an interface? Then this specific class is just an internal impl of that interface. 
1. That's symmetric with ITriggeredFunctionExecutor 
2. It's easier to mock in testing (whereas this class can't be mocked)

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/DispatchQueueEndToEndTests.cs
```diff
@@ -0,0 +1,149 @@
...
+using Microsoft.WindowsAzure.Storage.Queue;
+using Newtonsoft.Json.Linq;
+using Newtonsoft.Json;
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Host.EndToEndTests
+{
+    public class DispatchQueueEndToEndTests : IDisposable
+    {
+        private static ConcurrentDictionary<string, int> _concurrent = new ConcurrentDictionary<string, int>();
```
MikeStall: >ConcurrentDictionary [](start = 23, length = 20)

For dictionaries, can you comment what it's mapping from and to?

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/DispatchQueueEndToEndTests.cs
```diff
@@ -0,0 +1,149 @@
...
+            _hostConfiguration.TypeLocator = new FakeTypeLocator(typeof(SampleTrigger));
+
+            using (var host = new JobHost(_hostConfiguration))
+            {
+                _concurrent.AddOrUpdate("DispatchQueueTrigger", 0, (k, v) => 0);
+                _concurrent.AddOrUpdate("DispatchQueueTrigger2", 0, (k, v) => 0);
+
+                host.Start();
+
+                await Task.Delay(15000);
```
MikeStall: > await Task.Delay(15000); [](start = 15, length = 25)

Woah! We should not have a 15 second delay in the test.  What exactly is this waiting on?

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/DispatchQueueEndToEndTests.cs
```diff
@@ -0,0 +1,149 @@
...
+
+        public DispatchQueueEndToEndTests()
+        {
+            _hostConfiguration = new JobHostConfiguration();
+            _hostConfiguration.AddExtension(new DispatchQueueTestConfig());
+            _hostConfiguration.HostId = HOSTID;
+            CloudStorageAccount sdkAccount = CloudStorageAccount.Parse(_hostConfiguration.StorageConnectionString);
+            CloudQueueClient client = sdkAccount.CreateCloudQueueClient();
+            _sharedQueue = client.GetQueueReference("azure-webjobs-shared-" + HOSTID);
+            _poisonQueue = client.GetQueueReference("azure-webjobs-poison-" + HOSTID);
```
MikeStall: Can we leverage the faked out storage account stuff to avoid a real storage dependency?

src/Microsoft.Azure.WebJobs.Host/Dispatch/IMessageHandler.cs
watashiSHUN: these xml comments require it
`<see cref="ListenerFactoryContext.GetDispatchQueue(IMessageHandler)"/>` --> how `IMessageHandler` is used
`<see cref="ITriggeredFunctionExecutor.TryExecuteAsync(TriggeredFunctionData, CancellationToken)"/>` --> its doppelganger

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ExtensionWithDispatchQueue.cs
```diff
@@ -0,0 +1,221 @@
...
+                public Task<object> GetValueAsync()
+                {
+                    return Task.FromResult(_value);
+                }
+
+                public string ToInvokeString()
+                {
+                    return _value.ToString();
+                }
+            }
```
MikeStall: We have this helper in several places. See ObjectValueProvider

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ExtensionWithDispatchQueue.cs
```diff
@@ -0,0 +1,221 @@
...
+                {
+                }
+
+                public void Cancel()
+                {
+                }
+
+                private async Task DelayEnqueue()
+                {
+                    await Task.Delay(500);
```
MikeStall: >await Task.Delay(500); [](start = 20, length = 22)

We shouldn't need this.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/DispatchQueueEndToEndTests.cs
watashiSHUN: but this is an end to end test, do we not want to test with actual storage queue at all?
the "phantom" message should be unlikely since I have a Dispose():
 ```
       public void Dispose()
        {
            _sharedQueue.DeleteIfExists();
            _poisonQueue.DeleteIfExists();
        }
```

File: test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ExtensionWithDispatchQueue.cs (hunk 33 above)
Comment (watashiSHUN) [reply test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ExtensionWithDispatchQueue.cs]: that's an internal class, I can't use it easily in end to end test

File: test/Microsoft.Azure.WebJobs.Host.FunctionalTests/DispatchQueueTests.cs [hunk 35]
```diff
@@ -0,0 +1,194 @@
...
+            var enqueue = new List<Task>();
+            for (int i = 0; i < max; i++)
+            {
+                JObject message = JObject.Parse("{count:" + i + "}");
+                enqueue.Add(_sharedQueue.EnqueueAsync(message, "mockFunction1", CancellationToken.None));
+            }
+
+            await Task.WhenAll(enqueue);
+            // wait for dequeue
+            await Task.Delay(1000);
```
MikeStall: >await Task.Delay(1000); [](start = 12, length = 23)

Why do we need this at all?

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/DispatchQueueEndToEndTests.cs
```diff
@@ -0,0 +1,186 @@
...
+                int timeOut = 6000;
+                // it take 2secs to trigger the first function
+                // only one dequeue is happening if batchSize*2 < maxPollingInterval 
+                int twoFuncCount = DispatchQueueTestConfig.batchSize * 2;
+                while (timeOut > 0 && !_funcInvokation.hasDuplicate() && _funcInvokation.TotalAdd() < twoFuncCount)
+                {
+                    // let it run
+                    await Task.Delay(2000);
+                    timeOut -= 2000;
+                }
```
MikeStall: Can this use the existing TestHelpers.Await() methods?

fabiocav:
Just making sure, we'll be porting these changes to `dev` as well, correct?

src/Microsoft.Azure.WebJobs.Host/Dispatch/DummyDispatchQueueHandler.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Host.Dispatch
+{
+    // this class is used for local testing
+    // an extension developer can test DispatchQueueHandler without 
+    // providing an azure storage account
+    internal class DummyDispatchQueueHandler : IDispatchQueueHandler
```
MikeStall: >DummyDispatchQueueHandler [](start = 19, length = 25)

This can actually be used in production. For example, the AFR will use it (since they don't have real blob storage). Can we rename  this, like "InMemoryDispatchQueueHandler"

src/Microsoft.Azure.WebJobs.Host/Dispatch/DummyDispatchQueueHandler.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        private IMessageHandler _messageHandler;
```
MikeStall: >private [](start = 8, length = 7)

nit: readonly

src/Microsoft.Azure.WebJobs.Host/Dispatch/DummyDispatchQueueHandler.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            _messageHandler = messageHandler;
+        }
+
+        // any messages enqueued will get processed in worker pool immediately
+        public Task EnqueueAsync(JObject message, CancellationToken cancellationToken)
+        {
+            // start executing the function
+            // ignore the function result, no retry or poisonQueue
+            _messageHandler.TryExecuteAsync(message, cancellationToken);
```
MikeStall: >_messageHandler [](start = 12, length = 15)

Nit: 
Task taskIngore = _messageHandler....

(I thought we issued a warning on non-awaited tasks) 
Makes it obvious that there is indeed a task, but we're not blocking on it.  
I think this has similar semantics to the live queue.

src/Microsoft.Azure.WebJobs.Host/Dispatch/SharedQueueHandler.cs
```diff
@@ -0,0 +1,261 @@
...
+                IStorageQueue sharedPoisonQueue = primaryQueueClient.GetQueueReference(HostQueueNames.GetHostSharedPoisonQueueName(hostId));
+
+                // queueWatcher will update queueListener's polling interval when queueWriter performes an enqueue operation
+                SharedQueueWatcher sharedQueueWatcher = _sharedContextProvider.GetOrCreateInstance<SharedQueueWatcher>(
+                                                            new SharedQueueWatcherFactory(_messageEnqueuedWatcherSetter));
+                _sharedQueueWriter = new SharedQueueWriter(sharedQueue, sharedQueueWatcher);
+
+                // use default poisonQueue setup
+                _sharedQueuelistener = new QueueListener(sharedQueue, sharedPoisonQueue, _triggerExecutor,
+                    _exceptionHandler, _trace, _loggerFactory, sharedQueueWatcher, _queueConfiguration);
```
MikeStall: So if there's no storage (AFR), then all this gets skipped. Is that ok ... will it cause problems later? For example, ListenerFactoryListener  looks like it calls Initialize() and then StartQueueAsync(); but even though Initialize() returns "Safely", StartQueueAsync() which will just throw.

MikeStall: I'd suggest having an explicit State, like "unavailable", and set that.  

---
In reply to: [147010623](https://github.com/Azure/azure-webjobs-sdk/pull/1362#discussion_r147010623) [](ancestors = 147010623)

src/Microsoft.Azure.WebJobs.Host/Dispatch/SharedQueueHandler.cs
```diff
@@ -0,0 +1,261 @@
...
+            {
+                throw new InvalidOperationException(ErrorMessage(State.Started, _state));
+            }
+            // the count value does not change once SharedQueue is started, see "RegisterHandler()"
+            if (_triggerExecutor.HasMessageHandlers())
+            {
+                await _sharedQueuelistener.StopAsync(cancellationToken);
+            }
+            // if there's no messageHandlers registed, stopQueue is a NOOP
+            _state++;
```
MikeStall: >_state++; [](start = 12, length = 9)

Can you avoid ++ to trasntiion between statemachine states. That's too brittle on enum ordering. Can you explicitly set it ("_state = State.Stopped")

PR #1372, opened 2017-10-10
Fixing test runs on macOS/Linux

fabiocav:


fabiocav:
This is pretty stale. Will reopen with updates.

PR #1379, opened 2017-10-13
Disable functions using configuration

alrod:
• Get disabled state from IConfiguration
	• Format of the setting: <FunctionName>.Disabled = true.
	• Function is enabled by default if there is no setting.
	• The <FunctionName>.Disabled setting overwrites [Disable] attribute only if function is disabled.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/ConfigurationUtility.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/ConfigurationUtility.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Globalization;
 using Microsoft.Extensions.Configuration;
 
 namespace Microsoft.Azure.WebJobs.Host
@@ -32,6 +33,21 @@ namespace Microsoft.Azure.WebJobs.Host
         {
             _configurationFactory = configurationRootFactory;
             Reset();
+        }
+
+        public static bool IsSettingEnabled(string settingName)
+        {
+            // check the target setting and return false (disabled) if the value exists
+            // and is "falsey"
+            string value = GetSetting(settingName);
+            if (!string.IsNullOrEmpty(value) &&
+                (string.Compare(value, "1", StringComparison.OrdinalIgnoreCase) == 0 ||
+                 string.Compare(value, "true", StringComparison.OrdinalIgnoreCase) == 0))
+            {
+                return true;
+            }
+
+            return false;
         }
 
         private static IConfigurationRoot BuildConfiguration()
--- a/test/Microsoft.Azure.WebJobs.Host.UnitTests/ConfigurationUtilityTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.UnitTests/ConfigurationUtilityTests.cs
@@ -60,5 +60,35 @@ namespace Microsoft.Azure.WebJobs.Host.UnitTests
 
             Environment.SetEnvironmentVariable("DisableSetting0", null);
         }
+
+        [Theory]
+        [InlineData("True", true)]
+        [InlineData("true", true)]
+        [InlineData("1", true)]
+        [InlineData("", false)]
+        [InlineData("0", false)]
+        [InlineData("false", false)]
+        [InlineData("test", false)]
+        public void IsSettingEnabled_ReturnsExpected(string settingValue, bool expectedResult)
+        {
+            // Arrange
+            ConfigurationUtility.Reset();
+            string settingName = "SettingEnabledTest";
+            Environment.SetEnvironmentVariable(settingName, settingValue);            
+
+            try
+            {
+                // Act
+                bool isDisabled = ConfigurationUtility.IsSettingEnabled(settingName);
+
+                // Assert
+                Assert.True(isDisabled == expectedResult);
+            }
+            finally
+            {
+                // Clear
+                Environment.SetEnvironmentVariable(settingName, null);
+            }
+        }
     }
 }
--- a/src/Microsoft.Azure.WebJobs.Host/Listeners/HostListenerFactory.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Listeners/HostListenerFactory.cs
@@ -79,31 +79,41 @@ namespace Microsoft.Azure.WebJobs.Host.Listeners
 
         internal static bool IsDisabled(MethodInfo method, INameResolver nameResolver, IJobActivator activator)
         {
-            ParameterInfo triggerParameter = method.GetParameters().FirstOrDefault();
-            if (triggerParameter != null)
+            // First try to resolve disabled state by setting
+            string settingName = string.Format(CultureInfo.InvariantCulture, "AzureWebJobs.{0}.Disabled", method.Name);
+            if (ConfigurationUtility.IsSettingEnabled(settingName))
             {
-                // look for the first DisableAttribute up the hierarchy
-                DisableAttribute disableAttribute = TypeUtility.GetHierarchicalAttributeOrNull<DisableAttribute>(triggerParameter);
-                if (disableAttribute != null)
+                return true;
+            }
+            else 
+            {
+                // Second try to resolve disabled state by attribute
+                ParameterInfo triggerParameter = method.GetParameters().FirstOrDefault();
+                if (triggerParameter != null)
                 {
-                    if (!string.IsNullOrEmpty(disableAttribute.SettingName))
+                    // look for the first DisableAttribute up the hierarchy
+                    DisableAttribute disableAttribute = TypeUtility.GetHierarchicalAttributeOrNull<DisableAttribute>(triggerParameter);
+                    if (disableAttribute != null)
                     {
-                        return IsDisabledBySetting(disableAttribute.SettingName, method, nameResolver);
-                    }
-                    else if (disableAttribute.ProviderType != null)
-                    {
-                        // a custom provider Type has been specified
-                        return IsDisabledByProvider(disableAttribute.ProviderType, method, activator);
-                    }
-                    else
-                    {
-                        // the default constructor was used
-                        return true;
+                        if (!string.IsNullOrEmpty(disableAttribute.SettingName))
+                        {
+                            return IsDisabledBySetting(disableAttribute.SettingName, method, nameResolver);
+                        }
+                        else if (disableAttribute.ProviderType != null)
+                        {
+                            // a custom provider Type has been specified
+                            return IsDisabledByProvider(disableAttribute.ProviderType, method, activator);
+                        }
+                        else
+                        {
+                            // the default constructor was used
+                            return true;
+                        }
                     }
                 }
-            }
 
-            return false;
+                return false;
+            }
         }
 
         internal static bool IsDisabledBySetting(string settingName, MethodInfo method, INameResolver nameResolver)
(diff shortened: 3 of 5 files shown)


src/Microsoft.Azure.WebJobs.Host/Settings/ISettingProvider.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// Defines settings provider.
+    /// </summary>
+    public interface ISettingProvider
+    {
+        /// <summary>
+        /// Gets disabled state for the method defined in the settings.
+        /// </summary>
+        /// <param name="methodName">Name of the method.</param>
+        /// <returns>Boolean value if defined in the settings otherwise returns null.</returns>
+        bool? Disabled (string methodName);
```
mamaso: why does this return a `bool?` instead of just a `bool`?

src/Microsoft.Azure.WebJobs.Host/Settings/SettingManager.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.WebJobs.Host.Indexers;
+
+namespace Microsoft.Azure.WebJobs.Host.Settings
+{
+    internal class SettingManager
```
mamaso: Instead of adding a new `SettingManager` (we already have a `SettingsManager`), I wonder if we can use an `IConfiguration` in order to handle function metadata (we could populate from `function.json`, environment, and command line via an `IConfigurationBuilder`. @fabiocav might have some more thoughts on this

src/Microsoft.Azure.WebJobs.Host/Settings/ConfigurationSettingProvider.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// </summary>
+    public class ConfigurationSettingProvider : ISettingProvider
+    {
+        /// <inheritdoc />
+        public bool? Disabled(string methodName)
+        {
+            var disabledSetting = ConfigurationUtility.GetSetting(string.Format(CultureInfo.InvariantCulture, "{0}.Disabled", methodName));
+            if (!string.IsNullOrEmpty(disabledSetting))
+            {
+                return string.Compare(disabledSetting, "false", StringComparison.OrdinalIgnoreCase) != 0;
```
mamaso: could we remove the `IsNullOrEmpty` and return the `Compare` expression for 'true'?

test/Microsoft.Azure.WebJobs.Host.UnitTests/PublicSurfaceTests.cs
```diff
@@ -247,7 +247,9 @@ public void WebJobsHostPublicSurface_LimitedToSpecificTypes()
                 "IFunctionExceptionFilter",
                 "IFunctionFilter",
                 "IFunctionInvocationFilter",
-                "LoggerExtensions"
+                "LoggerExtensions",
+                "ISettingProvider",
```
mamaso: Do these need to be public types? I think we can make do with internal.

fabiocav:
Had some notes for feedback and ended up losing them before sending, but to summarize:

- The goal is to avoid new abstractions here, in favor of `IConfiguration` as we discussed. The new types are just introducing another layer of indirection on top of that and shouldn't be required (`ISettingsProvider` and the `ConfigurationSettingsProvider`)
- The `SettingManager` shouldn't be required either

We should be able to just pass the `IConfiguration` instance (this will become a bit simpler after the DI work) to be used by the extension method or the indexer itself and simplify the implementation here a bit.

src/Microsoft.Azure.WebJobs.Host/Settings/ConfigurationSettingProvider.cs
MikeStall: we should be even stricter (since that would still allow many different values for false, like 'false', 'disabled', 'stopped', 'please_dont_run_this', etc)

---
In reply to: [144928487](https://github.com/Azure/azure-webjobs-sdk/pull/1379#discussion_r144928487) [](ancestors = 144928487)

MikeStall:
+1. I don't understand what adding a new ISettingProvider is accomplishing here.  If it's specifically for looking up disable status, we should give it a more specific name. 
And why is it public? 

---
In reply to: [337021131](https://github.com/Azure/azure-webjobs-sdk/pull/1379#issuecomment-337021131) [](ancestors = 337021131)

test/Microsoft.Azure.WebJobs.Host.UnitTests/ConfigurationUtilityTests.cs
```diff
@@ -60,5 +60,41 @@ public void GetSettingFromConfigOrEnvironment_ConfigAndEnvironment_ConfigWins()
...
+        [InlineData("True", true)]
+        [InlineData("true", true)]
+        [InlineData("", false)]
+        [InlineData("false", false)]
+        [InlineData("test", false)]
+        public void IsDisabledMethod_ReturnsExpected(string settingValue, bool expectedResult)
+        {
+            // Arrange
+            ConfigurationUtility.Reset();
+            Environment.SetEnvironmentVariable("TestMethod.Disabled", settingValue);
```
MikeStall: >Environment.SetEnvironmentVariable [](start = 12, length = 34)

Can we avoid setting an Env var? That can lead to test brittleness. 
I see we try to reset it below - at leas that should be in a finally block.

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
```diff
@@ -522,5 +523,17 @@ internal override void AddMetadata(IDictionary<string, string> metadata)
                 throw new NotSupportedException();
             }
         }
+
+        private static void ApplyConfigurationToFunctionIndex(IFunctionIndex functionIndex)
+        {
+            foreach (var functioneDescriptor in functionIndex.ReadAllDescriptors())
```
MikeStall: >functioneDescriptor [](start = 25, length = 19)

typo

test/Microsoft.Azure.WebJobs.Host.UnitTests/ConfigurationUtilityTests.cs
alrod: Finally block was added.

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
```diff
@@ -522,5 +523,17 @@ internal override void AddMetadata(IDictionary<string, string> metadata)
...
+
+        private static void ApplyConfigurationToFunctionIndex(IFunctionIndex functionIndex)
+        {
+            foreach (var functionDescriptor in functionIndex.ReadAllDescriptors())
+            {
+                var isDisabled = ConfigurationUtility.IsDisabledMethod(functionDescriptor.LogName);
+                if (isDisabled)
+                {
+                    functionDescriptor.IsDisabled = isDisabled;
+                }
```
brettsam: Can this be reduced to one line of  `functionDescriptor.IsDisabled = ConfigurationUtility.IsDisabledMethod(functionDescriptor.LogName)`?

test/Microsoft.Azure.WebJobs.Host.UnitTests/ConfigurationUtilityTests.cs
brettsam: I think one of the benefits of using `IConfiguration` is that we should be able to supply our own `IConfigurationBuilder` for testing -- you can use an in-memory dictionary for example -- and not worry about environment vars or files, etc. Do we need to refactor ConfigurationUtility so we can plug in a custom builder via tests? @fabiocav?

brettsam: Oh -- I see we have `SetConfigurationFactory()` already. Should probably take advantage of that instead of using Environment vars for unit tests. I also see that it uses a static factory Func without a way to reset it to the default behavior which may be confusing in tests -- so maybe there's some refactoring that still needs to be done. Should it be an instance class that you create rather than static like this?

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
alrod: no, we want to set isDisabled only if it's "true". We do not want to overwrite [Disabled] attribute value.

brettsam: Oh right -- the joy of having multiple ways to change the same setting :-) So if the attribute is there, it wins?

alrod: The app setting wins if is "true". If app setting is "false" (or another value) attribute wins.

src/Microsoft.Azure.WebJobs.Host/ConfigurationUtility.cs
```diff
@@ -34,6 +35,12 @@ public static void SetConfigurationFactory(Func<IConfiguration> configurationRoo
             Reset();
         }
 
+        public static bool IsDisabledMethod(string methodName)
+        {
+            var disabledSetting = GetSetting(string.Format(CultureInfo.InvariantCulture, "{0}.Disabled", methodName));
```
mathewc: All our other settings have the standard "AzureWebJobs" prefix (e.g. the Prefix defined in AmbientConnectionStringProvider). Should we do the same? E.g. `AzureWebJobs.MyFunction.Disabled`

src/Microsoft.Azure.WebJobs.Host/ConfigurationUtility.cs
```diff
@@ -34,6 +35,12 @@ public static void SetConfigurationFactory(Func<IConfiguration> configurationRoo
...
+            return string.Compare(disabledSetting, "true", StringComparison.OrdinalIgnoreCase) == 0;
```
mathewc: Consider also allowing 1/0 in addition to true/false. We already have this logic in HostListenerFactory.IsDisabledBySetting. Recommend just factoring that logic out into a helper on ConfigurationUtility that is shared

src/Microsoft.Azure.WebJobs.Host/ConfigurationUtility.cs
```diff
@@ -34,6 +35,12 @@ public static void SetConfigurationFactory(Func<IConfiguration> configurationRoo
             Reset();
         }
 
+        public static bool IsDisabledMethod(string methodName)
```
mathewc: nit: IsMethodDisabled?

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
```diff
@@ -276,6 +276,7 @@ public static ServiceProviderWrapper CreateStaticServices(this JobHostConfigurat
                 }
 
                 IFunctionIndex functions = await functionIndexProvider.GetAsync(combinedCancellationToken);
+                ApplyConfigurationToFunctionIndex(functions);
```
mathewc: All the current disable code we have is in FunctionIndexer.FromMethod. Why can't your new logic just slot in there? E.g. we already apply DisableAttribute and disabled settings coming from that attribute to the FunctionDescriptor there. Seems like your new additional setting check can be applied there?

mhoeger:
@alrod is this change reflected in documentation? ex: https://github.com/projectkudu/kudu/wiki/Configurable-settings#limit-the-scaling-of-function-apps

alrod:
@mhoeger, as I know we do not have documentation for this change yet

PR #1380, opened 2017-10-13
[AppInsights] improving Exception logging

### LINKED ISSUE #1365: ILogger - LogException does not log Exception Message property
When using ILogger with Application Insights integration, passing an Exception should result in a row added to the exceptions table in AI with various exception properties, including Exception Message. Currently there are a lot of helpful properties there, including outerMessage and innermostMessage, but no message from the Exception.Message property, which contains whatever message is passed via new Exception("my message string").

Repro steps:

1. Set your Azure Function App up to send data to AI.
2. From an Azure function, make a call to your ILogger Log method with an exception that has a message.
3. In AI, Find your Exception in the exceptions table
4. Notice that the text you passed in as a message is in none of the properties.

brettsam:
Fixes #1365.

The issue here is that the `ExceptionTelemetry.Message` property didn't behave as I expected. It overwrites the exception message of the most outer exception in the payload that gets sent to App Insights. If you had nested exceptions, the inner (most important) ones would still look correct, which is why our normal exception throwing still contains the relevant details.

This change moves any supplied log message to a `FormattedMessage` property, which will show up in `customDimensions` in Analytics. This way you can still add additional logging details to supplement the exception itself.

I've added tests that use reflection to double-check that we're not mangling the internal data. These both failed before the change and pass after.

This will also be ported to `dev`.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
@@ -53,8 +53,8 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
             string formattedMessage = formatter?.Invoke(state, exception);
             IEnumerable<KeyValuePair<string, object>> stateValues = state as IEnumerable<KeyValuePair<string, object>>;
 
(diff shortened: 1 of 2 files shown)


brettsam:
I've already submitted it -- #1381. Once it passes the tests, I'll merge.

Issue #1369, opened 2017-10-09
Incorrect WEBSITE_HOSTNAME when using slots


1N50MN14:
Please provide a succinct description of the issue.

#### Repro steps
Create three slots, `context.log(process.env['WEBSITE_HOSTNAME'])` on each

#### Expected behavior
Correct host name.

#### Actual behavior
Incorrect host name.

#### Known workarounds
None, this is causing headaches as I use the hostname to detect which environment the code is running in.

paraspatidar:
The Environment variable **WEBSITE_HOSTNAME** works as a sticky slot setting for Azure Web Apps and remains with that perticular slot even after the Slot Swap.
However Being Azure Function App **Slot** in Preview as of Dated comment , it may not be working.

The work around for the same is as follows any of the one below : 
1.	Use the variable “**WEBSITE_SLOT_NAME**”  , this is reliable and gives the slot name correct (like Production/Stage/Dev or whatever the name of slot), however it doesn’t give the url like WEBSITE_HOSTNAME gives mywebappname-slot.azurewebsites.net.

2.	Use the App setting’s Sticky setting , by marking them as sticky to slot , that will not swap with the slot , it will be stick to that slot only  (in function App , you will need to go to All setting --> App setting and check the slot setting check box) refer [:](https://docs.microsoft.com/en-us/azure/app-service/web-sites-staged-publishing)

3.Alternatively , while using WEBSITE_HOSTNAME , you can restart your function app after the slot swap , it might work.

paulbatum:
Its a general app service issue that you cannot use WEBSITE_HOSTNAME reliably once slots are in play. Use the hostname environment variable in combination with WEBSITE_SLOT_NAME.

MikeyF-Anovia:
As mentioned in https://github.com/Azure/azure-webjobs-sdk/issues/2121, WEBSITE_SLOT_NAME is not an option. I'm running into this issue as well. I have two slots, a continuous webjob runs on both. On the prod slot it runs normally. One the deployment slot it runs and exits because it sees '-deployment' in WEBSITE_HOSTNAME. Once I swap the two, the webjob that was previously in the deployment slot that is now in the prod slot still thinks its in the deployment slot. 

The webjob continuously logs, over and over, these sets of events:
[08/10/2023 20:56:04 > e7cfa4: SYS INFO] Status changed to PendingRestart
[...]

PR #1382, opened 2017-10-13
[AppInsights] changing RoleName calculation to work with Functions slots

### LINKED ISSUE #1366: Logging Property cloud_roleName gets and retains "-Staging" postfix indefinitely after swap
I have a deployment with 8 or so function apps running in production slots. I deployed new code (via an VSTS Build template) to the Staging slot of each function app. Within an hour or so, I swapped the slots and deleted the Staging slot.

In my ApplicationInsights logs, before my deployment, my functions app(s) had a cloud_roleName of "myfunctionapp." After deployment, the cloud_roleName for functions now running in my production slot were "myfunctionapp-staging." Some of my function apps don't run often, and those eventually went back to just "myfunctionapp." However, one of my functions, a trigger that runs on the minute, continued to have the "-staging" postfix until about 8 days later when it went back to "myfunctionapp".

My assumption is that when my other function app instances started back up after inactivity, they resolved to the correct name. But my one function instance that was active enough to not restart, was unable to resolve the name...until one day it was...

My expectations are that something running in the Staging slot would have the "-staging" postfix and something running in the production slot would not.

brettsam:
Fixes #1366.

Due to #1369, we cannot rely on WEBSITE_HOSTNAME for slots, which is what the `AzureWebAppRoleEnvironmentTelemetryInitializer`. So I took that class and changed it to calculate the name with stable environment variables.

I'm about to start testing this with a private site extension but wanted to get the PR out.

I'll move this to `dev` immediately after this is in for `v2.x`.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/DefaultTelemetryClientFactory.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/DefaultTelemetryClientFactory.cs
@@ -9,7 +9,6 @@ using Microsoft.ApplicationInsights.Extensibility;
 using Microsoft.ApplicationInsights.Extensibility.Implementation;
 using Microsoft.ApplicationInsights.Extensibility.PerfCounterCollector;
 using Microsoft.ApplicationInsights.Extensibility.PerfCounterCollector.QuickPulse;
-using Microsoft.ApplicationInsights.WindowsServer;
 using Microsoft.ApplicationInsights.WindowsServer.Channel.Implementation;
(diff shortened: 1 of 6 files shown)


brettsam:
I've been testing this with a site extension and it looks good. I was able to verify the broken old behavior and this fixes it.

brettsam:
@SergeyKanzhelev -- FYI that we had to change our behavior with setting RoleName and NodeName due to #1369. The behavior should be similar... but now correct for Functions.

SergeyKanzhelev:
@brettsam would you mind contributing this back to Application Insights? Issue tracking this bug on our side: https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/648

Issue #1360, opened 2017-09-29
Allow asynchronous creation of collectors in BindToCollector


ConnorMcMahon:
#### Problem

FluentBindingRule<TAttribute>.BindToCollector<T> currently has a signature for Func<TAttribute, IAsyncCollector<T>> that is a function that maps the attribute to it's collector.

However, sometimes these collectors depend on other calls that may be asynchronous, making the method passed to BindToCollector to wait on those calls.

#### Known workarounds

You can force the method to block on the asynchronous call, making the method synchronous, but this is very ugly and probably not desirable. See [here](https://github.com/Azure/azure-webjobs-sdk/blob/e915ee574136207cab3b01cac7139bf09b013ace/src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/QueueBindingProvider.cs#L190) for an example.

#### Proposed Solution
Add a signature for BindToCollector that allows for asynchronous creation of a collector from an attribute.

@MikeStall

MikeStall:
We can add an overload like:
```
 BindToCollector<TMessage>(IAsyncConverter<TAttribute, IAsyncCollector<TMessage>> buildFromAttribute)
```
But a pure `Func<TAttribute, Task<IAsyncCollector<TMessage>>>` overload is ambiguous with `Func<TAttribute, IAsyncCollector<TMessage>>` .

PR #1391, opened 2017-10-17
Add binding rule for async collectors:

MikeStall:
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1360
Apply to Queue.


src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
```diff
@@ -3,6 +3,7 @@
 
 using System;
 using System.Reflection;
+using System.Threading;
```
ConnorMcMahon: nit: don't need this using statement.

PR #1403, opened 2017-10-26
sanitizing app insights logs

brettsam:


src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Sanitizer.cs
```diff
@@ -0,0 +1,52 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs.Logging
+{
+    /// <summary>
+    /// Utility class for sanitizing logging strings.
+    /// </summary>
+    public static class Sanitizer
```
brettsam: Any objections to this being public? We'll need it in Script as well...

brettsam: FYI -- this came from internal code that's been used for a while.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Sanitizer.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        internal const string SecretReplacement = "!!!SECRET!!!";
```
brettsam: I don't like this as a public-facing replacement -- any other suggestions?

src/Microsoft.Azure.WebJobs.Host/Executors/StorageAccountParser.cs
```diff
@@ -70,6 +71,9 @@ public static StorageAccountParseResult TryParseAccount(string connectionString,
         /// <returns>Formatted error message with details about reason of the failure and possible ways of mitigation</returns>
         public static string FormatParseAccountErrorMessage(StorageAccountParseResult error, string connectionStringName)
         {
+            // Users may accidentally use their real connection strings here, so let's be safe before throwing.
+            connectionStringName = Sanitizer.Sanitize(connectionStringName);
```
brettsam: Exception messages inside of ExceptionTelemetry cannot be sanitized, so handling at the source.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Sanitizer.cs
MikeStall: This is high-liability code making a best-effort to implement an impossible contract.   Therefore, I prefer non-public. If Script really needs this; then copy it? 


---
In reply to: [147281384](https://github.com/Azure/azure-webjobs-sdk/pull/1403#discussion_r147281384) [](ancestors = 147281384)

brettsam:
@MikeStall this is ready for another look...

MikeStall:
>     public class PublicSurfaceTests

Can you revert this file? 

---
Refers to: test/Microsoft.Azure.WebJobs.Host.UnitTests/PublicSurfaceTests.cs:19 in 1d623c4. [](commit_id = 1d623c4452d636f4a339f58e3a348297428bedf1, deletion_comment = False)

MikeStall:
What's the WebJobs.Shared thing? Creating a new dll here seems pretty heavyweight  just to share a 20 line function.

brettsam:
It's not a DLL -- it's a nicer way to share a file between projects over 'Add as link'. https://stackoverflow.com/a/30638495/3516125

brettsam:
Also reverted that file.

MikeStall:
What is it being shared between?

---
In reply to: [340753466](https://github.com/Azure/azure-webjobs-sdk/pull/1403#issuecomment-340753466) [](ancestors = 340753466)

brettsam:
App Insights and Host in this repo. And it's been copied to Script.

App Insights can't sanitize exceptions: https://github.com/Microsoft/ApplicationInsights-dotnet/issues/498. So I decided to target the biggest culprit in the logs and sanitize it before we throw (see StorageAccountParser change). We could completely refactor that message, but it's pretty useful except for the case where someone pastes their connection string into the "Connection" property. Sanitizing there catches that scenario.

PR #1405, opened 2017-10-27
Use JobHostMetadataProvider to get disabled state

alrod:


src/Microsoft.Azure.WebJobs.Host/DisabledMethodHelper.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Globalization;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    public static class DisabledMethodHelper
+    {
+        public static bool IsMethodDisabled(string methodName)
```
MikeStall: This should not be static. It locks us into appsettings. Discussing offline

src/Microsoft.Azure.WebJobs.Host/Listeners/HostListenerFactory.cs
```diff
@@ -77,11 +77,10 @@ internal class HostListenerFactory : IListenerFactory
             return new CompositeListener(listeners);
         }
 
-        internal static bool IsDisabled(MethodInfo method, INameResolver nameResolver, IJobActivator activator)
+        internal static bool IsDisabled(MethodInfo method, INameResolver nameResolver, IJobActivator activator, IJobHostMetadataProvider metadataProvider)
```
MikeStall: I think the layering here is backwards. If there's a [Disabled] attribute, IJobHostMetadataProvider.IsDisabled should also return true.

src/Microsoft.Azure.WebJobs.Host/Extensions/IJobHostMetadataProvider.cs
```diff
@@ -60,5 +60,12 @@ public interface IJobHostMetadataProvider
         /// </summary>
         /// <param name="output">The <see cref="TextWriter"/> to write the graph to.</param>
         void DebugDumpGraph(TextWriter output);
+
+        /// <summary>
+        /// Gets function metadata.
+        /// </summary>
+        /// <param name="methodName">Method name.</param>
+        /// <returns></returns>
+        MethodMetadata GetMethodMetadata(string methodName);
```
MikeStall: >methodName [](start = 48, length = 10)

We should comment in the interface whether this methodName match is case sensitive or not.

src/Microsoft.Azure.WebJobs.Host/Extensions/MethodMetadata .cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// Represents method metadata.
+    /// </summary>
+    public class MethodMetadata
+    {
+        /// <summary>
+        /// Gets or sets whether this method is disabled.
```
MikeStall: >sets [](start = 20, length = 4)

This isn't true, right? Setting this to True does not disable the method.  We return a poco, but we ignore any mutations.  
The current behavior is fine, but we should update the comment.

src/Microsoft.Azure.WebJobs.Host/Extensions/IJobHostMetadataProvider.cs
MikeStall: Also, we should comment what we do if the method name doesn't exist. (throw? return a null MethodMetadata)  . The current impl looks like it always create an object. 

---
In reply to: [148619644](https://github.com/Azure/azure-webjobs-sdk/pull/1405#discussion_r148619644) [](ancestors = 148619644)

src/Microsoft.Azure.WebJobs.Host/Listeners/HostListenerFactory.cs
MikeStall: although I see that creates a complicated bootstrapping problem in script. So maybe this is unavoidable. 

---
In reply to: [148617945](https://github.com/Azure/azure-webjobs-sdk/pull/1405#discussion_r148617945) [](ancestors = 148617945)

src/Microsoft.Azure.WebJobs.Host/Extensions/IJobHostMetadataProvider.cs
```diff
@@ -60,5 +60,12 @@ public interface IJobHostMetadataProvider
         /// </summary>
         /// <param name="output">The <see cref="TextWriter"/> to write the graph to.</param>
         void DebugDumpGraph(TextWriter output);
+
+        /// <summary>
+        /// Gets function metadata.
```
mathewc: Rather than MethodMetadata, I think we should name the type FunctionMetadata - we use the "Function" terminology elsewhere. Note that you used "function" in your xml description as well because it is more natural :)

So You'd have `FunctionMetadata GetFunctionMetadata(string functionName)`

src/Microsoft.Azure.WebJobs.Host/Extensions/JobHostMetadataProvider.cs
```diff
@@ -273,5 +274,14 @@ private void AddTypesFromGraph(IBindingRuleProvider root)
                 }
             }
         }
+
+        public MethodMetadata GetMethodMetadata(string methodName)
+        {
+            string disabledSettingName = string.Format(CultureInfo.InvariantCulture, "AzureWebJobs.{0}.Disabled", methodName);
```
mathewc: For implementation, I think you should be passing in a deferred getter for IFunctionIndexProvider into the JobHostMetadataProvider constructor, which will allow you to use IFunctionIndex to get function metadata that has already been computed. Note that the index has already computed IsDisabled per function, so you won't need to recompute it again as you're doing now.  That would also set us up for the future for returning additional function metadata.

Also - Need to see if there are any timing issues with when the index is created, and when we the IJobHostMetadataProvider can be called to return function metadata.

Using IFunctionIndex to get FunctionDescriptors (which contain the IsDisabled property you need) does raise again the earlier question of whether we create a new type FunctionMetadata, or use the existing FunctionDescriptor. We decided to introduce a new, nicer type, though in this method you'll basically be doing a mechanical transformation from FunctionDescriptor to FunctionMetadata.

src/Microsoft.Azure.WebJobs.Host/Extensions/IJobHostMetadataProvider.cs
mathewc: Though I do see that in the functions runtime, we already have a public type "FunctionMetadata". Likely our usage of the new core SDK API won't cause any conflicts because it is constrained, and namespacing can resolve any conflicts.

src/Microsoft.Azure.WebJobs.Host/Listeners/HostListenerFactory.cs
mathewc: Yeah I think layering is off here as well. We already have logic that computes disabled state and other per function metadata at index time. That's the same metadata that we should be returning via this new IJobHostMetadataProvider API.

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
```diff
@@ -276,6 +276,7 @@ public static ServiceProviderWrapper CreateStaticServices(this JobHostConfigurat
                 }
 
                 IFunctionIndex functions = await functionIndexProvider.GetAsync(combinedCancellationToken);
+                IJobHostMetadataProvider metadataProvider = services.GetService<IJobHostMetadataProvider>();
```
mathewc: what is this line for? I don't see the new variable being used?

src/Microsoft.Azure.WebJobs.Host/Extensions/IJobHostMetadataProvider.cs
```diff
@@ -60,5 +63,12 @@ public interface IJobHostMetadataProvider
...
+        /// </summary>
+        /// <param name="methodName">Method name.</param>
+        /// <returns></returns>
+        Task<FunctionMetadata> GetFunctionMetadata(string methodName, CancellationToken cancellationToken = default(CancellationToken));
```
mathewc: nit: string functionName

src/Microsoft.Azure.WebJobs.Host/Extensions/IJobHostMetadataProvider.cs
```diff
@@ -60,5 +63,12 @@ public interface IJobHostMetadataProvider
         /// </summary>
         /// <param name="output">The <see cref="TextWriter"/> to write the graph to.</param>
         void DebugDumpGraph(TextWriter output);
+
+        /// <summary>
+        /// Gets function metadata.
+        /// </summary>
+        /// <param name="methodName">Method name.</param>
```
mathewc: nit: "Name of function to return metadata for."

src/Microsoft.Azure.WebJobs.Host/Extensions/IJobHostMetadataProvider.cs
mathewc: Shame that this needs to be async. Let's rename it to GetFunctionMetadataAsync though

src/Microsoft.Azure.WebJobs.Host/Extensions/JobHostMetadataProvider.cs
```diff
@@ -273,5 +283,24 @@ private void AddTypesFromGraph(IBindingRuleProvider root)
...
             }
         }
+
+        public FunctionMetadata GetFunctionMetadata(string functionName)
+        {
+            FunctionMetadata result = null;
+            var provider = _getFunctionIndexProvider.Invoke();
+            if (provider != null)
+            {
+                var index = provider.GetAsync(new CancellationToken()).GetAwaiter().GetResult();
```
mathewc: Rather than new up a token here, use CancellationToken.None

Issue #1406, opened 2017-10-30
Fixes to Converter Manager 


MikeStall:
For ConverterManager

1. Remove the [Obsolete] attributes so that we can have the interfaces for Function v2. 
2. Support Async 
3. Add (Type typeSource, Type typeDest) signatures rather than just <TSrc, TDest> so that loosely typed code can call it with System.Type. 
4. remove special casing for [Queue] and pocos. 
5. currently, exact type and open type have totally separate paths. But since exact type is just a subset of open type, this can lead to inconsistencies. Merge these to ensure correct semantics. 
4. pull in fixes from the VS Plugin example. 
5. Ability to remove rules.

Issue #757, opened 2016-06-23
Remove BindingFactory.AddValidator


brettsam:
We already have AddFilter which can act as both a filter and a validator. I don't see anyone actually using AddValidator anywhere and I'm not sure it adds much value over AddFilter.

mathewc:
What do we need to do here?

brettsam:
I proposed a bunch of extension API improvements -- this is one of them. Not needed ASAP.

lindydonna:
@brettsam Should we move this to the next milestone?

MikeStall:
Will get this as part of  https://github.com/Azure/azure-webjobs-sdk/issues/686

Issue #686, opened 2016-03-30
Finish BindingFactory and other Extension binding helpers


mathewc:
We need to:
- finalize the new APIs and public surface area
- add full unit tests for everything
- address all outstanding bugs

MikeStall:
We'll remove BindingFactory complete; and superseded by the rule-based scheme off ExtensionContext.
Fixing this in conjunction with https://github.com/Azure/azure-webjobs-sdk/issues/1406

PR #1413, opened 2017-11-02
Finalizing Binding interfaces for v2.

MikeStall:
This does some major plumbing fixes to ConverterManager and removes the [Obsolete] attributes. 
This has breaking changes that will need to be pushed through other extensions to be on a full BYOB pipeline. 

  https://github.com/Azure/azure-webjobs-sdk/issues/1406
  https://github.com/Azure/azure-webjobs-sdk/issues/686
  https://github.com/Azure/azure-webjobs-sdk/issues/757

1) Adding converters should be done via an extension and standard rules; and not via direct IConverterManager calls.
This lets us shrink public surface area:
- Remove most of BindingFactory and IConverterManager/Extesions.
- some renames to clarify between Exact types and OpenTypes

2) Async converters

3) Stronger invariants:
- Better asserts for misusing open converters
- Have QueueProvider now have an explicit rule for Poco --> QueueMessage (Rather than hackey). Other extensions can use this same Poco definition to get consistent behavior.
- Allow adding multiple rules to an attribute. This was important for factoring and composition.
- fix subtle bugs about the difference between Open Type and Exact Type

4) fixing issues in Converter manager:
- Loose signature using System.Type instead of requiring <T> everywhere. That lets callers avoid needing to MakeGenericType just to call .
- Async everywhere
- Stronger consistency between OpenType and System.Type


src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/QueueBindingProvider.cs
```diff
@@ -80,19 +81,30 @@ public void Initialize(ExtensionConfigContext context)
 
                 binding.SetPostResolveHook(ToReadWriteParameterDescriptorForCollector)
                         .BindToInput<CloudQueue>(builder);
+            }
+
+            private async Task<IStorageQueueMessage> ConvertPocoToCloudQueueMessage(object arg, Attribute attrResolved, ValueBindingContext context)
+            {
+                var attr = (QueueAttribute)attrResolved;
```
brettsam: Since this hangs off of a binding with known TAttribute, can we make it pass TAttribute to the func? Then you wouldn't have to cast.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
```diff
@@ -33,15 +33,15 @@ public BindingFactory(INameResolver nameResolver, IConverterManager converterMan
         /// <summary>
         /// Get the name resolver for resolving %% tokens. 
         /// </summary>
-        public INameResolver NameResolver
+        internal INameResolver NameResolver
```
brettsam: Why are these moved to internal and not just removed?

src/Microsoft.Azure.WebJobs.Host/Bindings/PatternMatcher.cs
```diff
@@ -268,6 +268,7 @@ private static bool CheckArg(Type openType, Type specificType, Dictionary<string
         // Helper for getting a converter func that invokes the converter interface on an object.
         private abstract class InvokerBase
         {
+            // This should get changed to a FuncAsyncConverter
```
brettsam: Can you make that change here?

src/Microsoft.Azure.WebJobs.Host/Config/FluentConverterRules.cs
```diff
@@ -16,7 +17,7 @@ namespace Microsoft.Azure.WebJobs.Host.Config
     public abstract class FluentConverterRules<TAttribute, TThis> where TAttribute : Attribute
     {
         // Access the converter manager that we're adding rules to. 
-        internal abstract IConverterManager Converters { get; }
+        internal abstract ConverterManager Converters { get; }
```
brettsam: Why not use the interface?

src/Microsoft.Azure.WebJobs.Host/Bindings/OpenType.cs
```diff
@@ -6,28 +6,328 @@
 
 namespace Microsoft.Azure.WebJobs.Host.Bindings
 {
+    // Ensures that open Types match consistently 
+    public class OpenTypeMatchContext
```
brettsam: Should this be public?

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs
MikeStall: They're still used by the Fluent rules (from ExtensionConfigContext).  Those rules are the public surface area that people should be using. 

---
In reply to: [148657335](https://github.com/Azure/azure-webjobs-sdk/pull/1413#discussion_r148657335) [](ancestors = 148657335)

src/Microsoft.Azure.WebJobs.Host/Config/FluentConverterRules.cs
MikeStall: The public interface is very constrained (just a Getter). Whereas the class is private, and it has additional setters. I'm trying to shirnk the public surface area as much as possible to reduce liability. 

---
In reply to: [148663519](https://github.com/Azure/azure-webjobs-sdk/pull/1413#discussion_r148663519) [](ancestors = 148663519)

src/Microsoft.Azure.WebJobs.Host/Bindings/OpenType.cs
MikeStall: Unfortunately, it needs to be since public abstract OpenType.IsMatch(Type type, OpenTypeMatchContext context); needs to be public. 
That method needs to be public since derived classes may implement it. But the only thing a derived class would do with this context is pass it into a IsMatch call on one of our builtin opentypes. 

---
In reply to: [148666948](https://github.com/Azure/azure-webjobs-sdk/pull/1413#discussion_r148666948) [](ancestors = 148666948)

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/QueueBindingProvider.cs
MikeStall: I'll try - the problem I'm running into is "delegate overload hell".  We have too many permutations, and it stresses C#'s overload rules. 

---
In reply to: [148656706](https://github.com/Azure/azure-webjobs-sdk/pull/1413#discussion_r148656706) [](ancestors = 148656706)

MikeStall:
This has been superseded by: https://github.com/Azure/azure-webjobs-sdk/pull/1422

MikeStall:
This has been superseded by: https://github.com/Azure/azure-webjobs-sdk/pull/1422

PR #1417, opened 2017-11-04
Fixes #1086 - handle two event hubs with the same name in different namespaces

### LINKED ISSUE #1086: Issue with two event hubs with the same name in different namespaces
The EventHubs binding does not correctly handle the case where a function app has two different event hub triggered functions, when the event hubs are in different namespaces but have the same name.

#### Repro steps

1. Create two event hub namespaces ns1 and ns2. 
2. Create an eventhub called 'test' in each namespace. 
3. Create a function app with two event hub triggered functions, func1 and func2. Configure func1 to point to ns1/test and configure func2 to point to ns2/test.
4.  Write data onto each event hub

#### Expected behavior

Each function triggers for the events on the appropriate hub

#### Actual behavior

func1 triggers for messages to ns2. func2 does not trigger

#### Known workarounds

Either rename one of the eventhubs or split the two functions into different function apps.

#### Related information 

Problem seems to be related to the code here:
https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs

The dictionaries are keyed off of the event hub name, but this is not a unique identifier.

paulbatum:
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1086 by updating how `EventHubConfiguration` generates lookup keys for its internal state - specifically, keying off a combination of endpoint and event hub name (instead of just the event hub name).

This was a somewhat tricky to change to make because `EventHubConfiguration` exposes public APIs that are simply difficult to make work correctly. For example, `AddEventProcessorHost(string eventHubName, EventProcessorHost listener)` is ambiguous since there is no namespace included in the parameters.

The approach I've taken is to do a best effort for maintaining these problematic APIs while trying to make sure that functions does not use them. The script PR for that is here:
https://github.com/Azure/azure-webjobs-sdk-script/pull/2103

I've tested this manually with a few different permutations:
- function using two event hub output bindings that only differ by namespace (same hub name)
- function app with two event hub triggered functions that only differ by namespace
- for above scenarios, tested both with direct load (`configurationSource:attributes`) and function.json 
- for above scenarios, tested the webjobs equivalent

Feedback is welcome on whether these changes warrant e2e/nightly test coverage.


paulbatum:
Looks like some legit test failures. Working on these.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
```diff
@@ -436,6 +475,60 @@ private static byte[] ConvertEventData2Bytes(EventData input)
...
+            return messagingFactory.Address;
+        }
+
+        internal static string GetLookupKey(Uri endpoint, string hubName)
+            => $"{endpoint.Host}:{hubName}";
+
+        internal static string GetLookupKey(EventHubClient client)
+            => GetLookupKey(GetEndpointFromEventHubClient(client), client.Path);
+
+        internal static string GetLookupKey(string eventHubName, string connectionString)
```
MikeStall: >GetLookupKey(string eventHubName [](start = 31, length = 32)

nit: it's very easy to have Uris represented as strings, so it's dangerous to have overloads on Uri vs. String that have different semantics.  Consider giving the overloads different names.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
```diff
@@ -436,6 +475,60 @@ private static byte[] ConvertEventData2Bytes(EventData input)
         private static EventData ConvertString2EventData(string input) 
             => ConvertBytes2EventData(Encoding.UTF8.GetBytes(input));
 
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA2204:Literals should be spelled correctly", MessageId = "MessagingFactory")]
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA2204:Literals should be spelled correctly", MessageId = "EventHubClient")]
```
MikeStall: nit: Can these go in the global suppression file (not source).

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
```diff
@@ -436,6 +475,60 @@ private static byte[] ConvertEventData2Bytes(EventData input)
...
+        internal static Uri GetEndpointFromEventHubClient(EventHubClient client)
+        {
+            var messagingFactoryProperty = typeof(EventHubClient).GetProperty("MessagingFactory", BindingFlags.Instance | BindingFlags.NonPublic);
+            if (messagingFactoryProperty == null)
+            {
+                throw new InvalidOperationException("Reflection call to obtain EventHubClient.MessagingFactory property info failed. Did a service bus package update occur?");
+            }
+
+            var messagingFactory = (MessagingFactory)messagingFactoryProperty.GetValue(client);
```
MikeStall: >var messagingFactory = (MessagingFactory [](start = 12, length = 40)

nit:  since the property is internal, it may change the return type and then this would become a cast exception.  Suggested: 
    var messagingFactory = messagingFactoryProperty.GetValue(client) as MessagingFactory;
Then your nicer exception will still fire instead of the cast.

paulbatum:
Addressed review feedback. Thanks!

Issue #1415, opened 2017-11-02
DefaultType not matching when using JavaScript with OpenType


brettsam:
This scenario came up with the CosmosDB output binding, but it may affect others. The `AsyncCollecto`r is registered as supporting `OpenType`: https://github.com/Azure/azure-webjobs-sdk-extensions/blob/dev/src/WebJobs.Extensions.CosmosDB/Config/CosmosDBConfiguration.cs#L68

In v2 we had our `DefaultType` support as `JObject`/`JArray`, which was the fallback when using JavaScript -- https://github.com/Azure/azure-webjobs-sdk-script/blob/v1.x/src/WebJobs.Script/Binding/DocumentDBScriptBindingProvider.cs#L84-L98.

In the new world without our own ScriptBindingProvider, this DefaultType when using a JavaScript output binding now comes back as `object`, which doesn't seem to hit the collector.

We did this to work around it: https://github.com/Azure/azure-webjobs-sdk-extensions/pull/324

What's the expected behavior here?

brettsam:
@fabiocav FYI

fabiocav:
Thank you, @brettsam !

MikeStall:
Execpation is that we still return IAsyncCollector<JObject>. Need to fix AsyncCollectorBindingProvider.GetDefaultType to be smarter here.

MikeStall:
I've made the fix as a commit in the Converter PR I have out: https://github.com/Azure/azure-webjobs-sdk/pull/1413/commits/7c059b07c1bb65db58f384e43d106d9d9a1e9757

Issue #1001, opened 2017-02-01
Rule for BindToStream


MikeStall:
Create a binding factory rule for BindToStream. This enables [Blob] and [File] to share the same code. 
This is the last remaining major rule that we need. 
There is a prototype of this in ApiHub's file binding.

tohling:
Can't wait to have this support!

MikeStall:
This requires https://github.com/Azure/azure-webjobs-sdk/issues/1406

daulet:
This issue is mentioned in [Creating custom input and output bindings](https://github.com/Azure/azure-webjobs-sdk/wiki/Creating-custom-input-and-output-bindings). Update the wiki with details how to use BindToStream.

AymanGaafar:
Is there another updated version of [Creating custom input and output bindings](https://github.com/Azure/azure-webjobs-sdk/wiki/Creating-custom-input-and-output-bindings) wiki page or is it not updated yet?

Issue #995, opened 2017-01-30
Support Blob POCO bindings


mathewc:
Remove ICloudBlobStreamBinder (replace with custom Converters). The blob binding had an existing extensibility mechanism [ICloudBlobStreamBinder](https://github.com/Azure/azure-webjobs-sdk/blob/663a508e8a851629c26a51e7de3af36629dfd120/src/Microsoft.Azure.WebJobs/ICloudBlobStreamBinder.cs) that customers can use to register custom blob stream => POCO types.

This also causes problems for people because they expect to be able to **bind blobs to POCO types directly** as they can with all other bindings. E.g. see SO question [here](https://github.com/Azure/azure-webjobs-sdk/issues/995).

When the blob binding is converted over to the new binding APIs, we can remove that old extensibility mechanism.

fabiocav:
relates to #1001

brettsam:
I'm pretty sure this isn't actually fixed. In moving to our new DI patterns, I removed the legacy ICloudBlobStreamBinder registration code and a bunch of tests started failing. I also don't see anywhere in the blob binding setup that indicates we support binding to POCOs. I'm going to re-enable this until we can take a deeper look and figure out what we need to do.

Also note that we have official docs on this that we'll need to update when we remove this interface: https://docs.microsoft.com/en-us/azure/visual-studio/vs-storage-webjobs-getting-started-blobs#getting-serialized-blob-content-by-using-icloudblobstreambinder

brettsam:
Note: Some tests pass due to this binder: https://github.com/Azure/azure-webjobs-sdk/blob/dev/test/Microsoft.Azure.WebJobs.Host.EndToEndTests/BlobToCustomObjectBinder.cs. It's likely that logic needs to move into the core binding wire-up.

Part of closing this issue should be removal of that interface -- assuming all scenarios can be covered without it.

fabiocav:
Running into issues because of this work item again. There's logic in script that explicitly calls the methods to support the legacy behavior, which, unfortunately, doesn't play well with the assembly isolation work we've put in place.

Let's confirm that we have functionality coverage and remove the legacy code (which should no longer be needed)

brettsam:
When doing the DI tests, I confirmed that we do *not* have support for Blob POCOs. We'll need to add it.

mathewc:
Right - we don't have POCO support. The old ICloudStreamBinder stuff was how you did POCOs. That's what this issue was created for.

[...]

PR #1422, opened 2017-11-08
Move Blob onto new extensibility interfaces

MikeStall:
This does some major overhaul to remove some [Obsolete] , and verify it by fully plumbing Blob onto the BYOB extensibility model. 

1. Fixes to Converter Manager. 
2. BindToStream rule 
3. BindToTrigger rule - which lets triggers leverage the non-trigger binding support so they don't need to create ValueProviders anymore. 

Plumb [Blob] and [BlobTrigger] onto this, and remove the old classes.


MikeStall:
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1415 
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1406
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1001

And makes significant progress on other Sprint 10 bugs.

MikeStall:
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1415 
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1406
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1001

And makes significant progress on other Sprint 10 bugs.

MikeStall:
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1415 
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1406
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1001

And makes significant progress on other Sprint 10 bugs.

MikeStall:
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1415 
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1406
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1001

And makes significant progress on other Sprint 10 bugs.

MikeStall:
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1415 
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1406
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1001

And makes significant progress on other Sprint 10 bugs.

MikeStall:
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1415 
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1406
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1001

And makes significant progress on other Sprint 10 bugs.

MikeStall:
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1415 
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1406
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1001

And makes significant progress on other Sprint 10 bugs.

src/Microsoft.Azure.WebJobs.Host/FuncConverter.cs
```diff
@@ -3,19 +3,21 @@
 
 using System;
 using Microsoft.Azure.WebJobs.Host.Bindings;
+using System.Threading.Tasks;
 
 namespace Microsoft.Azure.WebJobs
```
MikeStall: >WebJobs [](start = 26, length = 7)

It looked a lot cleaner to specifically call out delegates rather than use Func<> everywhere.

src/Microsoft.Azure.WebJobs.Host/IConverterManager.cs
```diff
@@ -2,16 +2,14 @@
...
 
 namespace Microsoft.Azure.WebJobs
 {
     /// <summary>
     /// General service for converting between types for parameter bindings.  
     /// Parameter bindings call this to convert from user parameter types to underlying binding types. 
+    /// Converters can be added via extension configuration rules on <see cref="FluentConverterRules"/>.
     /// </summary>
-    [Obsolete("Not ready for public consumption.")]
     public interface IConverterManager
```
MikeStall: >public interface IConverterManager [](start = 4, length = 34)

This public surface shrinks significantly. The "add" methods all now come through the Fluent binding rules.

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
```diff
@@ -249,7 +249,7 @@ public IConverterManager ConverterManager
         /// <summary>
         /// Gets a helper object for constructing common binding rules for extensions.
         /// </summary>
-        public BindingFactory BindingFactory
+        internal BindingFactory BindingFactory
```
MikeStall: >BindingFactory [](start = 17, length = 14)

Remove from public surface since extensions should be using the binding rules (off ExtensionConfigContext) rather than these interfaces.

MikeStall:
>     public class BindingFactory

The goal is to delete this class entirely and move everybody onto the rule-based system.

---
Refers to: src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs:18 in 88099c7. [](commit_id = 88099c71c6d5f528c9abebca39ffcf93de624b81, deletion_comment = False)

src/Microsoft.Azure.WebJobs.Host/Bindings/ConverterManager.cs
```diff
@@ -2,46 +2,89 @@
...
+using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host.Bindings;
+using Microsoft.Azure.WebJobs.Host.Config;
 using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Azure.WebJobs
-{   
+{
     // Concrete implementation of IConverterManager
     internal class ConverterManager : IConverterManager
```
MikeStall: >ConverterManager [](start = 19, length = 16)

This file is a near complete rewrite. Basically, changes in this file are:
- use the same means for OpenTypes and Exact Types (since exact type is just a subset of open type) 
- remove special casing. 
- the OpenType helpers got moved to that type.

src/Microsoft.Azure.WebJobs.Host/Bindings/OpenType.cs
```diff
@@ -6,28 +6,328 @@
...
+
+        /// <summary>
         /// Return true if and only if given type matches. 
+        /// This is constrained by the current matching context. 
+        /// If the OpenType represents an exact type, then this is just Type.Equals.
         /// </summary>
         /// <param name="type">Type to check</param>
         /// <returns></returns>
-        public abstract bool IsMatch(Type type);
+        public abstract bool IsMatch(Type type, OpenTypeMatchContext context);
```
MikeStall: >OpenTypeMatchContext [](start = 48, length = 20)

Requiring this param is a breaking change. But very few extensions actually derive from OT directly.

src/Microsoft.Azure.WebJobs.Host/Bindings/AsyncCollector/TypedAsyncCollectorAdapter.cs
```diff
@@ -33,10 +33,11 @@ internal class TypedAsyncCollectorAdapter<TSrc, TDest, TAttribute> : IAsyncColle
             _context = context;
         }
 
-        public Task AddAsync(TSrc item, CancellationToken cancellationToken = default(CancellationToken))
+        public async Task AddAsync(TSrc item, CancellationToken cancellationToken = default(CancellationToken))
         {
-            TDest x = _convert(item, _attrResolved, _context);
-            return _inner.AddAsync(x, cancellationToken);
+            TDest result = await _convert(item, _attrResolved, _context);
```
MikeStall: >await [](start = 27, length = 5)

These extra awaits are a big deal. I plumbed async through all the converters.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/AsyncCollectorBindingProvider.cs
```diff
@@ -206,17 +206,34 @@ public IEnumerable<BindingRule> GetRules()
...
-                var types = cm.GetPossibleSourceTypesFromDestination(attribute.GetType(), typeof(TType));
+            {
+                // TType is OpenType 
+                // TType is CloudMessage,   String --> CloudMessage
+
+                var ot = OpenType.FromType<TType>();
 
-                // search in precedence 
-                foreach (var target in new Type[] { typeof(JObject), typeof(byte[]), typeof(string) })
+                // Check for a direct match.
```
MikeStall: >// Check for a direct match.  [](start = 16, length = 29)

This section is the fix for [https://github.com/Azure/azure-webjobs-sdk/issues/1415](https://github.com/Azure/azure-webjobs-sdk/issues/1415)

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/BindToInputBindingProvider.cs
```diff
@@ -46,9 +46,7 @@ public Task<IBinding> TryCreateAsync(BindingProviderContext context)
                 return Task.FromResult<IBinding>(null);
             }
 
-            var type = typeof(ExactBinding<>).MakeGenericType(typeof(TAttribute), typeof(TType), typeUser);
-            var method = type.GetMethod("TryBuild", BindingFlags.Static | BindingFlags.NonPublic | BindingFlags.Public);
-            var binding = BindingFactoryHelpers.MethodInvoke<IBinding>(method, this, context);
```
MikeStall: Before, ConverterMAnager only had <T> overloads, so you had to do this MakeGenericType to call it with a runtime type. Now, It has loose signarures that just take a Type, so it's easier to call without generics.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/BindToInputBindingProvider.cs
```diff
@@ -115,51 +113,54 @@ public Type GetDefaultType(Attribute attribute, FileAccess access, Type requeste
             return null;
         }
 
-        private class ExactBinding<TUserType> : BindingBase<TAttribute>
+        private class ExactBinding : BindingBase<TAttribute>
         {
-            private readonly Func<object, object> _buildFromAttribute;
```
MikeStall: >Func<object, object>  [](start = 29, length = 21)

Good to be getting rid of these adhoc signatures. Everything that does a converter is a FuncAsyncConverter.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/BindToStreamBindingProvider.cs
```diff
@@ -0,0 +1,518 @@
...
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Protocols;
+using Microsoft.Azure.WebJobs.Host.Config;
+
+namespace Microsoft.Azure.WebJobs.Host.Bindings
+{
+    // BindToStream.
+    // Read: Stream, TextReader, string,  byte[]
+    // Write: Stream, TextWriter, out string,  out byte[]
+    internal class BindToStreamBindingProvider<TAttribute> :
```
MikeStall: >BindToStreamBindingProvider [](start = 19, length = 27)

This is the new Bind To Stream rule.  [Blob] is also updated to leverage this.

src/Microsoft.Azure.WebJobs.Host/Blobs/Bindings/BlobAttributeBindingProvider.cs
```diff
@@ -6,181 +6,319 @@
...
 using Microsoft.Azure.WebJobs.Host.Indexers;
+using Microsoft.Azure.WebJobs.Host.Protocols;
 using Microsoft.Azure.WebJobs.Host.Storage;
 using Microsoft.Azure.WebJobs.Host.Storage.Blob;
 using Microsoft.WindowsAzure.Storage.Blob;
 
 namespace Microsoft.Azure.WebJobs.Host.Blobs.Bindings
 {
-    internal class BlobAttributeBindingProvider : IBindingProvider, IBindingRuleProvider
+    internal class BlobExtension : IExtensionConfigProvider,
```
MikeStall: >BlobExtension [](start = 19, length = 13)

This is making Blob a true extension, using IExtensionConfigProvider.  It calls the new BindToStream rule. 
This is similar to what we did for Table and Queue.

src/Microsoft.Azure.WebJobs.Host/Blobs/Triggers/BlobTriggerAttributeBindingProvider.cs
```diff
@@ -18,14 +18,65 @@
...
+        IAsyncConverter<IStorageBlob, Stream>
+    {
+        private IStorageAccountProvider _accountProvider;
+
+        public void Initialize(ExtensionConfigContext context)
+        {
+            _accountProvider = context.Config.GetService<IStorageAccountProvider>();
+
+            var rule = context.AddBindingRule<BlobTriggerAttribute>();
+            rule.BindToTrigger<IStorageBlob>();
```
MikeStall: >BindToTrigger [](start = 17, length = 13)

This is new. This new rule allows triggers to leverage our converter pipeline (same as the regular BindToInput rules) and avoid having to implement their own value-providers.  @brettsam - CosmoDb may be able to benefit from this too.

src/Microsoft.Azure.WebJobs.Host/Blobs/Triggers/BlobTriggerBinding.cs
```diff
@@ -25,7 +25,6 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs.Triggers
     internal class BlobTriggerBinding : ITriggerBinding
     {
         private readonly ParameterInfo _parameter;
-        private readonly IArgumentBinding<IStorageBlob> _argumentBinding;
```
MikeStall: >IArgumentBinding [](start = 25, length = 16)

We no longer need the ArgumentBinding because the new BindToTrigger rule will bind the argument for us.

src/Microsoft.Azure.WebJobs.Host/Config/FluentBindingRule.cs
```diff
@@ -106,6 +108,43 @@ public FluentBindingRule<TAttribute> WhenIsNotNull(string propertyName)
             return this;
         }
 
+        /// <summary>
+        /// The subsequent Bind* operations only apply when the Attribute's property is not null. 
+        /// </summary>
+        /// <param name="propertyName"></param>
+        /// <returns></returns>
+        public FluentBindingRule<TAttribute> When<TValue>(string propertyName, TValue expected)
```
MikeStall: >FluentBindingRule [](start = 15, length = 17)

Blob uses this  (only supports CloudBlobStream on Access=Write). Graph binding (or anything else with a "mode") will benefit from it too.

src/Microsoft.Azure.WebJobs.Host/Config/FluentBindingRule.cs
```diff
@@ -122,9 +161,8 @@ internal FluentBindingRule<TAttribute> SetPostResolveHook(Func<TAttribute, Param
         /// <returns></returns>
         public void BindToInput<TType>(IConverter<TAttribute, TType> builderInstance)
         {
-            var bf = _parent.BindingFactory;
-            var rule = bf.BindToInput<TAttribute, TType>(builderInstance);
-            Bind(rule);
+            var pm = PatternMatcher.New(builderInstance);
```
MikeStall: >PatternMatcher.New [](start = 21, length = 18)

I used PatternMatcher helpers to sanitize the various overloads.
Also, C# can't overload Func<TSrc, TDest> vs. Func<TSrc, Task<TDest>> 
But if we toss in a cancelation token, that changes the signature and then you can get the task overloads.

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
```diff
@@ -515,6 +522,41 @@ private static Assembly GetHostAssembly(IEnumerable<MethodInfo> methods)
             }
         }
 
+        #region Hackery for ExtensionLocator
+
+        // create IConverterManager adapters to any legacy ICloudBlobStreamBinder<T>. 
+        static void AddStreamConverters(IExtensionTypeLocator extensionTypeLocator, ConverterManager cm)
```
MikeStall: >AddStreamConverters [](start = 20, length = 19)

This shims the old ICloudBlobStreamBinder  onto the new converter-manager interfaces.  
Essentially demonstrating we have a fix for [https://github.com/Azure/azure-webjobs-sdk/issues/995](https://github.com/Azure/azure-webjobs-sdk/issues/995), [https://github.com/Azure/azure-webjobs-sdk/issues/995](https://github.com/Azure/azure-webjobs-sdk/issues/995) 
We can remove ICloudBlobStreamBinder  totally separately - that was just more churn and breaks and this change was already pretty large.

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
```diff
@@ -206,6 +206,13 @@ private bool HasJobAttribute(CustomAttributeData attributeData)
             if (triggerBinding != null)
             {
                 bindingDataContract = triggerBinding.BindingDataContract;
+                
+                // See if a regular binding can handle it.
```
MikeStall: >// See if a regular binding can handle it.  [](start = 16, length = 43)

This is part of the support to let Triggers leverage Converters with a BindToTrigger<> rule.

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/QueueBindingProvider.cs
```diff
@@ -59,7 +59,8 @@ public void Initialize(ExtensionConfigContext context)
                 var binding = context.AddBindingRule<QueueAttribute>();
                 binding
                     .AddConverter<byte[], IStorageQueueMessage>(ConvertByteArrayToCloudQueueMessage)
-                    .AddConverter<string, IStorageQueueMessage>(ConvertStringToCloudQueueMessage);
+                    .AddConverter<string, IStorageQueueMessage>(ConvertStringToCloudQueueMessage)
+                    .AddOpenConverter<OpenType.Poco, IStorageQueueMessage>(ConvertPocoToCloudQueueMessage);
```
MikeStall: > .AddOpenConverter<OpenType.Poco, IStorageQueueMessage>(ConvertPocoToCloudQueueMessage); [](start = 19, length = 88)

This is formalizing our "Poco --> Queue" rule. Before, it was kind of a hacky representation. This also means other extensions can implement the same rule using the same definition of "Poco" that Queue uses.

MikeStall:
>         public IBindingProvider BindToGenericValueProvider<TAttribute>(Func<TAttribute, Type, Task<IValueBinder>> builder)

The difference between this and BindToInput is that this lets you write back after the function is complete. It's the In/Out 2-way behavior. Again, we want to move away from requiring IValueBinder /IValueProvider. I'd like to replace this with a BindToInput().WriteBack() pattern. Separate change. 

---
Refers to: src/Microsoft.Azure.WebJobs.Host/Bindings/BindingFactory.cs:95 in 88099c7. [](commit_id = 88099c71c6d5f528c9abebca39ffcf93de624b81, deletion_comment = False)

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/AsyncCollectorBindingProvider.cs
```diff
@@ -206,17 +206,34 @@ public IEnumerable<BindingRule> GetRules()
...
         {
             if (access == FileAccess.Write)
-            {              
-                var cm = (ConverterManager)this._converterManager;
-                var types = cm.GetPossibleSourceTypesFromDestination(attribute.GetType(), typeof(TType));
+            {
+                // TType is OpenType 
+                // TType is CloudMessage,   String --> CloudMessage
+
+                var ot = OpenType.FromType<TType>();
```
mathewc: nit: openType

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/AsyncCollectorBindingProvider.cs
```diff
@@ -206,17 +206,34 @@ public IEnumerable<BindingRule> GetRules()
...
                 {
-                    if (types.Contains(target))
+                    if (ot.IsMatch(target))
+                    {
+                        return typeof(IAsyncCollector<>).MakeGenericType(target);
+                    }
+                }
+
+                // Check if there's a converter 
+                var cm = this._converterManager;
```
mathewc: why do you need this extra local? Can't you just use _converterManager below?

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/BindToStreamBindingProvider.cs
```diff
@@ -0,0 +1,518 @@
...
+                    }
+                }
+                if (access == FileAccess.Write)
+                {
+                    if (CanWrite(this._access))
+                    {
+                        return typeof(Stream);
+                    }    
+                }
+                // Read write stream area not supported.
```
mathewc: nit: are

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/BindToStreamBindingProvider.cs
```diff
@@ -0,0 +1,518 @@
...
+                else
+                {
+                    if (!CanWrite(declaredAccess.Value))
+                    {
+                        errorMsg = "Write";
+                    }
+                }
+                if (errorMsg != null)
+                {
+                    throw new InvalidOperationException($"The parameter type is a '{errorMsg}' binding, but the Attribute's access type is '{declaredAccess}'");
```
mathewc: I wonder if in error messages, rather than say things like "the Attribute's access" we shouldn't avoid mentioning attributes. This will yield better user messages for AF (often they're not dealing with attributes, e.g. Node).

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/ItemBindingProvider.cs
```diff
@@ -16,13 +16,16 @@ internal class ItemBindingProvider<TAttribute> : IBindingProvider
     {
         private readonly INameResolver _nameResolver;
         private readonly Func<TAttribute, Type, Task<IValueBinder>> _builder;
+        private readonly OpenType _filter;
 
         public ItemBindingProvider(
                     INameResolver nameResolver,
-                    Func<TAttribute, Type, Task<IValueBinder>> builder)
+                    Func<TAttribute, Type, Task<IValueBinder>> builder,
+                    OpenType filter)
```
mathewc: It is strange to see an instance of an OpenType be called a "filter". Rename the param to openType?

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/TriggerHelperBindingProvider.cs
```diff
@@ -0,0 +1,183 @@
...
+using System.Reflection;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Config;
+using Microsoft.Azure.WebJobs.Host.Protocols;
+
+namespace Microsoft.Azure.WebJobs.Host.Bindings
+{
+    // Regular BindToInput has to do a TAttribute --> Value creation. 
+    // But triggers already have a Listener that provided the initial object; and we're just
+    // converting it ot the user's target parameter.
```
mathewc: nit: sp "ot"

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/TriggerHelperBindingProvider.cs
```diff
@@ -0,0 +1,183 @@
...
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Config;
+using Microsoft.Azure.WebJobs.Host.Protocols;
+
+namespace Microsoft.Azure.WebJobs.Host.Bindings
+{
+    // Regular BindToInput has to do a TAttribute --> Value creation. 
+    // But triggers already have a Listener that provided the initial object; and we're just
+    // converting it ot the user's target parameter. 
+    internal class TriggerHelperBindingProvider<TAttribute, TTriggerValue> :
```
mathewc: wonder if this shouldn't be renamed to TriggerAdapterBindingProvider?

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/TriggerHelperBindingProvider.cs
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        public Type GetDefaultType(Attribute attribute, FileAccess access, Type requestedType)
+        {
+            return typeof(object);
+        }
+
+        public IEnumerable<BindingRule> GetRules()
+        {
+            yield break;
```
mathewc: probably more efficient to return Enumerable.Empty<BindingRule>()? Not sure if the compiler does an optimization on this.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/TriggerHelperBindingProvider.cs
```diff
@@ -0,0 +1,183 @@
...
+                    // TTriggerValue --> Stream --> TUserType
+                    var c1 = cm.GetConverter<TTriggerValue, Stream, TAttribute>();
+                    if (c1 != null)
+                    {
+                        var c2 = cm.GetConverter<Stream, TUserType, TAttribute>();
+                        if (c2 != null)
+                        {
+                            converter = async (TTriggerValue src, Attribute attr, ValueBindingContext ctx) =>
+                            {
+                                Stream o1 = await c1(src, attr, ctx);
```
mathewc: please use better names for all these locals to make the code clearer. E.g. c1 => streamConverter, c2 => userTypeConverter, etc.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/TriggerHelperBindingProvider.cs
```diff
@@ -0,0 +1,183 @@
...
+                // Never called, since a trigger alreayd has an object. 
+                throw new NotImplementedException();
+            }
+
+            public ParameterDescriptor ToParameterDescriptor()
+            {
+                throw new NotImplementedException();
+            }
+
+            // Caller already filterered on TAttribute
```
mathewc: what are these floating detached comments?

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/BindToStreamBindingProvider.cs
MikeStall: Can you suggest a wording?  "attribute --> "declared access". 
"...  but the binding's declared access is ..."

---
In reply to: [150924524](https://github.com/Azure/azure-webjobs-sdk/pull/1422#discussion_r150924524) [](ancestors = 150924524)

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/ItemBindingProvider.cs
MikeStall: OpenTypes are filters (hence the IsMatch method). In this case, we're filtering the rule to only match on the given OpenType. I can add a comment to clarify this. 

---
In reply to: [150925434](https://github.com/Azure/azure-webjobs-sdk/pull/1422#discussion_r150925434) [](ancestors = 150925434)

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/TriggerHelperBindingProvider.cs
MikeStall: will do. 

---
In reply to: [150925974](https://github.com/Azure/azure-webjobs-sdk/pull/1422#discussion_r150925974) [](ancestors = 150925974)

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/TriggerHelperBindingProvider.cs
MikeStall: copy& paste error. will fix. 

---
In reply to: [150927366](https://github.com/Azure/azure-webjobs-sdk/pull/1422#discussion_r150927366) [](ancestors = 150927366)

src/Microsoft.Azure.WebJobs.Host/Bindings/ConverterManager.cs
```diff
@@ -331,273 +297,144 @@ public Type[] GetPossibleSourceTypesFromDestination(Type typeAttribute, Type typ
...
+            // but now allow some well-defined intermediate conversions. 
             // If this is "wrong" for your type, then it should provide an exact match to override.
-
             // Byte[] --[builtin]--> String --> TDest
             if (typeSource == typeof(byte[]))
             {
-                var bytes2string = TryGetConverter<byte[], TAttribute, string>();
-
-                return (src, attr, context) =>
+                var func2 = GetMiddle<TAttribute, string>(typeSource, typeDest);
```
mathewc: GetIntermediate?

mathewc: Also use a more descriptive name for func2 - what is it a converter?

src/Microsoft.Azure.WebJobs.Host/Bindings/ConverterManager.cs
```diff
@@ -331,273 +297,144 @@ public Type[] GetPossibleSourceTypesFromDestination(Type typeAttribute, Type typ
...
-                };
             }
+            return null;
         }
 
-        // Match a generic type with 1 generic arg. 
-        // like IEnumerable<T>,  IQueryable<T>, etc. 
-        private class SingleGenericArgOpenType : OpenType
+
+        // List of all converters. This may refer to an pen type or an exact match.
```
mathewc: This comment refers to a list, but is on a class? Also, can you use a better name for the class? What does it represent?

src/Microsoft.Azure.WebJobs.Host/Bindings/OpenType.cs
```diff
@@ -6,28 +6,328 @@
 
 namespace Microsoft.Azure.WebJobs.Host.Bindings
 {
+    // Ensures that open Types match consistently 
+    public class OpenTypeMatchContext
+    {
+        internal Type _match;
+
+        internal bool CheckArg(Type type)
```
mathewc: This is a public type, with no public members?

src/Microsoft.Azure.WebJobs.Host/Bindings/OpenType.cs
```diff
@@ -6,28 +6,328 @@
...
+        {
+            if (_match == null)
```
mathewc: I don't understand what this code is doing.

src/Microsoft.Azure.WebJobs.Host/Bindings/OpenType.cs
mathewc: Should this be a protected abstract, since it is just the implementation for IsMatch(Type)? Do they both need to be public?

src/Microsoft.Azure.WebJobs.Host/Bindings/OpenType.cs
```diff
@@ -6,28 +6,328 @@
...
+            internal override string GetDisplayName()
+            {
+                return _inner.GetDisplayName() + "[]";
+            }
+        }
+                 
+        // Formalize definition of a "poco"
+        // This can be used for general "T --> JObject" bindings. 
+        // The exact definition here comes from the WebJobs v1.0 Queue binding.
+        public class Poco : OpenType
```
mathewc: PocoOpenType, or perhaps CustomOpenType?

src/Microsoft.Azure.WebJobs.Host/Bindings/OpenType.cs
MikeStall: This is making sure that when using OpenTypes for both src and dest, they match consistently.  For example, suppose we register a converter:
     IEnumerable<T> --> Widget<T> 
We need to make sure that T is the same in both src and dest. So we have a "context" that is shared across both OpenTypes matches and it verifies consistency. 
The methods themselves are non-public since derived OpenTypes don't need to worry about the exact contract, just that the instance of the contract is the same. 

---
In reply to: [150930248](https://github.com/Azure/azure-webjobs-sdk/pull/1422#discussion_r150930248) [](ancestors = 150930248)

src/Microsoft.Azure.WebJobs.Host/Bindings/PatternMatcher.cs
```diff
@@ -20,11 +22,60 @@ public static PatternMatcher New(Type typeBuilder, params object[] constructorAr
...
+        // T1,T2 may be open types.
+        public static PatternMatcher New<T1, T2>(IAsyncConverter<T1, T2> instance)
         {
             return new CreateViaInstance(instance);
         }
-                        
+
+        public static PatternMatcher New<T1, T2>(IConverter<T1, T2> instance)
+        {
+            return new CreateViaInstance(instance);
```
mathewc: might consider renaming these "vias" to "froms", e.g. CreateFromInstance

src/Microsoft.Azure.WebJobs.Host/Bindings/OpenType.cs
MikeStall: This is a nested type right now (more discoverable), so it's "OpenType.Poco".   Happy to rename (or make unnested). 


---
In reply to: [150931977](https://github.com/Azure/azure-webjobs-sdk/pull/1422#discussion_r150931977) [](ancestors = 150931977)

src/Microsoft.Azure.WebJobs.Host/Bindings/PatternMatcher.cs
```diff
@@ -268,21 +329,22 @@ private static bool CheckArg(Type openType, Type specificType, Dictionary<string
         // Helper for getting a converter func that invokes the converter interface on an object.
         private abstract class InvokerBase
         {
-            public abstract Func<object, object> Work(object instance);
+            // This should get changed to a FuncAsyncConverter
```
mathewc: Comment no longer applies?

src/Microsoft.Azure.WebJobs.Host/Bindings/PatternMatcher.cs
```diff
@@ -268,21 +329,22 @@ private static bool CheckArg(Type openType, Type specificType, Dictionary<string
...
-            public abstract Func<object, object> Work(object instance);
+            // This should get changed to a FuncAsyncConverter 
+            public abstract FuncAsyncConverter Work(object instance);
         }
 
         // Get a converter function that invokes IConverter on an object. 
         private class Invoker<TSrc, TDest> : InvokerBase
         {
-            public override Func<object, object> Work(object instance)
+            public override FuncAsyncConverter Work(object instance)
```
mathewc: The "Work" naming is strange. I'd expect InvokerBase to have an Invoke, or Execute method.

src/Microsoft.Azure.WebJobs.Host/Bindings/PatternMatcher.cs
```diff
@@ -379,5 +441,35 @@ protected override object GetInstance(Dictionary<string, Type> genericArgs)
                 return _instance;
             }
         }
+
+        // Wrapper for matching against an instance. 
+        private class CreateViaDelegate : PatternMatcher
```
mathewc: This class name/declaration is confusing.

src/Microsoft.Azure.WebJobs.Host/Blobs/BindableBlobPath.cs
```diff
@@ -1,45 +0,0 @@
...
-
-            if (containerNameTemplate.HasParameters || blobNameTemplate.HasParameters)
-            {
-                return new ParameterizedBlobPath(containerNameTemplate, blobNameTemplate);
-            }
-
-            BlobClient.ValidateContainerName(parsedPattern.ContainerName);
-            if (!string.IsNullOrEmpty(parsedPattern.BlobName))
-            {
-                BlobClient.ValidateBlobName(parsedPattern.BlobName);
```
mathewc: Where does the validation happen now?

src/Microsoft.Azure.WebJobs.Host/Blobs/BindableBlobPath.cs
```diff
@@ -1,45 +0,0 @@
...
-                !BlobPath.TryParse(pattern, isContainerBinding, out parsedPattern))
-            {
-                // if failed => not blob/container pattern
-                // we try pattern url
-                BindingTemplate urlTemplate = BindingTemplate.FromString(pattern);
-                if (urlTemplate.ParameterNames.Count() == 1)
-                {
-                    return new ParameterizedBlobUrl(urlTemplate);
-                }
-                throw new FormatException($"Invalid blob path '{pattern}'. Paths must be in the format 'container/blob' or 'blob Url'.");
```
mathewc: Where does this happen now?

src/Microsoft.Azure.WebJobs.Host/Bindings/OpenType.cs
MikeStall: IsMatch(Type) is just convenience. 
IsMatch(Type, context) is the one that needs to be overriden. Derived classes will either ignore the context, or pass the context through (if the OT is composed of other OTs, like Foo<T> ) 

---
In reply to: [150931321](https://github.com/Azure/azure-webjobs-sdk/pull/1422#discussion_r150931321) [](ancestors = 150931321)

src/Microsoft.Azure.WebJobs.Host/Blobs/Bindings/BlobAttributeBindingProvider.cs
mathewc: All the other extensions that implement IExtensionConfigProvider have names follwing the pattern "FooExtensionConfig", or "FooConfig". Also, update the file name to match class name

src/Microsoft.Azure.WebJobs.Host/Blobs/Bindings/BlobAttributeBindingProvider.cs
```diff
@@ -6,181 +6,319 @@
...
+        // Converter to produce an IEnumerable<T> for binding to multiple blobs. 
+        // T must have been matched by MultiBlobType        
+        class MultiBlobConverer<T> : IAsyncConverter<MultiBlobContext, IEnumerable<T>>
         {
-            ParameterInfo parameter = context.Parameter;
-            var blobAttribute = TypeUtility.GetResolvedAttribute<BlobAttribute>(parameter);
+            private readonly FuncAsyncConverter<IStorageBlob, T> _converter;
 
-            if (blobAttribute == null)
+            public MultiBlobConverer(BlobExtension parent)
```
mathewc: Type spelled wrong

src/Microsoft.Azure.WebJobs.Host/Blobs/Bindings/BlobAttributeBindingProvider.cs
```diff
@@ -6,181 +6,319 @@
...
+        {
+            var blob = await GetBlobAsync(blobAttribute, cancellationToken, typeof(T));
+            return (T)(blob.SdkObject);
+        }
+        
+        #endregion
+        
+        #region Support for binding to Multiple blobs 
+        // Open type matching types that can bind to an IEnumerable<T> blob collection. 
+        class MultiBlobType : OpenType
```
mathewc: perhaps rename to BlobEnumerableType or BlobCollectionType?

src/Microsoft.Azure.WebJobs.Host/Blobs/BindableBlobPath.cs
MikeStall: I'll create a BlobName validation data annotation attribute.  This goes through attribute cloner, so the parameterized and non-parameterized cases get unified and it will catch the case below too.

---
In reply to: [150936782](https://github.com/Azure/azure-webjobs-sdk/pull/1422#discussion_r150936782) [](ancestors = 150936782)

src/Microsoft.Azure.WebJobs.Host/Blobs/Triggers/BlobTriggerAttributeBindingProvider.cs
```diff
@@ -18,14 +18,65 @@
 using Microsoft.Azure.WebJobs.Host.Triggers;
 using Microsoft.Extensions.Logging;
 using Microsoft.WindowsAzure.Storage.Blob;
+using Microsoft.Azure.WebJobs.Host.Config;
+using System.Threading;
 
 namespace Microsoft.Azure.WebJobs.Host.Blobs.Triggers
 {
+    internal class BlobTriggerExtension : IExtensionConfigProvider,
```
mathewc: nit: same comment as before on naming consistency. E.g. should this be BlobTriggerExtensionConfig? I can't see all the other extensions and the pattern they follow, but I don't think we're consistent across the extrensions.

src/Microsoft.Azure.WebJobs.Host/Blobs/Triggers/BlobTriggerAttributeBindingProvider.cs
```diff
@@ -18,14 +18,65 @@
...
+        }
+
+        async Task<Stream> IAsyncConverter<IStorageBlob, Stream>.ConvertAsync(IStorageBlob input, CancellationToken cancellationToken)
+        {
+            WatchableReadStream watchableStream = await ReadBlobArgumentBinding.TryBindStreamAsync(input, cancellationToken);
+            return watchableStream;
+        }
+
+        // For descirbing InvokeStrings.
+        private async Task<IStorageBlob> ConvertX(DirectInvokeString input, Attribute attr, ValueBindingContext context)
```
mathewc: can you please rename this?

src/Microsoft.Azure.WebJobs.Host/Blobs/Triggers/BlobTriggerAttributeBindingProvider.cs
```diff
@@ -18,14 +18,65 @@
...
+            context.AddConverter(new StorageBlobToCloudAppendBlobConverter());
+        }
+
+        async Task<Stream> IAsyncConverter<IStorageBlob, Stream>.ConvertAsync(IStorageBlob input, CancellationToken cancellationToken)
+        {
+            WatchableReadStream watchableStream = await ReadBlobArgumentBinding.TryBindStreamAsync(input, cancellationToken);
+            return watchableStream;
+        }
+
+        // For descirbing InvokeStrings.
```
mathewc: nit: sp

src/Microsoft.Azure.WebJobs.Host/Blobs/Triggers/BlobTriggerAttributeBindingProvider.cs
```diff
@@ -18,14 +18,65 @@
...
+            // Blob type is a property of an existing blob. 
+            context.AddConverter(new StorageBlobToCloudBlobConverter());
+            context.AddConverter(new StorageBlobToCloudBlockBlobConverter());
+            context.AddConverter(new StorageBlobToCloudPageBlobConverter());
+            context.AddConverter(new StorageBlobToCloudAppendBlobConverter());
+        }
+
+        async Task<Stream> IAsyncConverter<IStorageBlob, Stream>.ConvertAsync(IStorageBlob input, CancellationToken cancellationToken)
+        {
+            WatchableReadStream watchableStream = await ReadBlobArgumentBinding.TryBindStreamAsync(input, cancellationToken);
```
mathewc: So is all the parameter watcher stuff still wired up for these bindings? The stuff that shows in progress blob download/upload percentage progress in the Dashboard?

src/Microsoft.Azure.WebJobs.Host/Blobs/Triggers/BlobTriggerBinding.cs
```diff
@@ -252,10 +235,10 @@ private static IAsyncObjectToTypeConverter<IStorageBlob> CreateConverter(IStorag
                 throw new InvalidOperationException("Unable to convert trigger to IStorageBlob.");
             }
 
-            Host.Bindings.IValueProvider valueProvider = await _argumentBinding.BindAsync(conversionResult.Result, context);
+            // Leave ValueProvider null - other binding rules will take care of it.
```
mathewc: this comment seems like it should just be removed, since it applies to the code a few lines below. Perhaps you should add a new ctor to TriggerData that doesn't take a value provider, and put comments there describing that scenario?

src/Microsoft.Azure.WebJobs.Host/Config/Apply.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Host.Config
+{
+    // Used to register a converter that applies a New value to an existing value. 
+    // This is common for Streams. The new value is written to an existing stream instantation.
```
mathewc: nit: spelling

src/Microsoft.Azure.WebJobs.Host/Config/Apply.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Host.Config
+{
+    // Used to register a converter that applies a New value to an existing value. 
+    // This is common for Streams. The new value is written to an existing stream instantation. 
+    class Apply<TNew, TExisting>
```
mathewc: Can you fix this type up? Apply isn't a good name for a top level type in this namespace. It's not clear what it represents, or in what context it should be used. Also, please add an explicit access modifier.

src/Microsoft.Azure.WebJobs.Host/Config/DirectInvokeString.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Host.Config
+{
+    // A short string serialization used to rehydrate the trigger. Sometimes this is the contents, and sometimes it's a moniker.  For example:
+    // For Blob, the DirectInvokeString is the blob path; whereas the String is the blob contents. (because blobs can be MB large) 
+    // For queue, both the DirectInvokeString and String are the queue contents. (because queue messages are generally small)
+    class DirectInvokeString
```
mathewc: internal class

src/Microsoft.Azure.WebJobs.Host/Config/DirectInvokeString.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        public string Value { get; set; }
+
+        public static DirectInvokeString New(string value)
```
mathewc: What is the benefit of this static New function, over a ctor? E.g. `new DirectInvokeString(...)` is more common/natural than `DirectInvokeString.New(...)` right?

src/Microsoft.Azure.WebJobs.Host/Config/ExtensionConfigContext.cs
```diff
@@ -16,8 +16,8 @@ public class ExtensionConfigContext : FluentConverterRules<Attribute, ExtensionC
         // List of actions to flush from the fluent configuration. 
         private List<Action> _updates = new List<Action>();
 
-        // track which TAttribute rules have been added. 
-        private HashSet<object> _existingRules = new HashSet<object>();
+        // Map of tyepof(TAttribute) --> FluentBindingRule<TAttribute>
```
mathewc: Can the dictionary be strongly typed then?

src/Microsoft.Azure.WebJobs.Host/Config/DirectInvokeString.cs
mathewc: Also, this is just a wrapper around a string with no added behavior - why is this needed?

src/Microsoft.Azure.WebJobs.Host/Config/ExtensionConfigContext.cs
```diff
@@ -31,11 +31,11 @@ public class ExtensionConfigContext : FluentConverterRules<Attribute, ExtensionC
         /// </summary>
         public TraceWriter Trace { get; set; }
 
-        internal override IConverterManager Converters
+        internal override ConverterManager Converters
```
mathewc: I think this property should be renamed to ConverterManager?

src/Microsoft.Azure.WebJobs.Host/Config/FluentBindingRule.cs
```diff
@@ -24,7 +26,7 @@ public class FluentBindingRule<TAttribute> : FluentConverterRules<TAttribute, Fl
     {
         private readonly JobHostConfiguration _parent;
 
-        private List<IBindingProvider> _binders = new List<IBindingProvider>();
+        private List<FluentBindToInputOptions> _binders = new List<FluentBindToInputOptions>();
```
mathewc: Either the type name is off here, or the member name _binders. FluentBindToInputOptions doesn't seem like a comprehensible type name :)

src/Microsoft.Azure.WebJobs.Host/Config/FluentBindingRule.cs
```diff
@@ -106,6 +108,43 @@ public FluentBindingRule<TAttribute> WhenIsNotNull(string propertyName)
...
+        /// <summary>
+        /// The subsequent Bind* operations only apply when the Attribute's property is not null. 
+        /// </summary>
+        /// <param name="propertyName"></param>
+        /// <returns></returns>
+        public FluentBindingRule<TAttribute> When<TValue>(string propertyName, TValue expected)
+        {
+            var prop = ResolveProperty(propertyName);
+
+            if (!typeof(TValue).IsEnum)
```
mathewc: I don't understand where in the method definition it's clear that TValue must be an enum? Perhaps you need a generic constraint or a different method name or something?

src/Microsoft.Azure.WebJobs.Host/Config/FluentBindingRule.cs
```diff
@@ -106,6 +108,43 @@ public FluentBindingRule<TAttribute> WhenIsNotNull(string propertyName)
...
+            {
+                throw new InvalidOperationException($"Rule filter for '{propertyName}' can only be used with enums.");
+            }
+
+            var propType = prop.PropertyType;            
+            if (!propType.IsAssignableFrom(typeof(TValue))) // Handles nullable
+            {
+                throw new InvalidOperationException(
+                    $"Type mismatch. Property '{propertyName}' is type '{TypeUtility.GetFriendlyName(propType)}'. " +
```
mathewc: nit: is of Type

src/Microsoft.Azure.WebJobs.Host/Config/FluentBindingRule.cs
```diff
@@ -106,6 +108,43 @@ public FluentBindingRule<TAttribute> WhenIsNotNull(string propertyName)
...
+            if (!propType.IsAssignableFrom(typeof(TValue))) // Handles nullable
+            {
+                throw new InvalidOperationException(
+                    $"Type mismatch. Property '{propertyName}' is type '{TypeUtility.GetFriendlyName(propType)}'. " + 
+                    $"Can't compare to a value of type '{TypeUtility.GetFriendlyName(typeof(TValue))}");
+            }
+
+            Func<TAttribute, bool> func = (attribute) =>
+            {
+                object x = prop.GetValue(attribute);
```
mathewc: nit rename x to value

src/Microsoft.Azure.WebJobs.Host/Config/FluentBindingRule.cs
```diff
@@ -120,11 +159,10 @@ internal FluentBindingRule<TAttribute> SetPostResolveHook(Func<TAttribute, Param
         /// <typeparam name="TType"></typeparam>
         /// <param name="builderInstance"></param>
         /// <returns></returns>
-        public void BindToInput<TType>(IConverter<TAttribute, TType> builderInstance)
+        public FluentBindToInputOptions BindToInput<TType>(IConverter<TAttribute, TType> builderInstance)
```
mathewc: Having a BindToInput method return an "options" class doesn't seem right

src/Microsoft.Azure.WebJobs.Host/Config/FluentBindingRule.cs
```diff
@@ -167,20 +201,133 @@ public void BindToInput<TType>(IAsyncConverter<TAttribute, TType> builderInstanc
...
+        #region BindToStream
+
+        /// <summary>
+        /// Bind an attribute to a stream. This ensures the stream is flushed after the user function returns. 
+        /// It uses the attribute's Access property to determine direction (Read/Write). 
+        /// It includes rules for additional types of TextReader,string, byte[], and TextWriter,out string, out byte[].
+        /// </summary>
+        /// <param name="func"></param>
+        /// <param name="fileAccess"></param>
+        public void BindToStream(Func<TAttribute, ValueBindingContext, Task<Stream>> func, FileAccess fileAccess)
```
mathewc: can you rename func? I think you use "builder" in other places? Make consistent throughout

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
```diff
@@ -515,6 +522,41 @@ private static Assembly GetHostAssembly(IEnumerable<MethodInfo> methods)
...
+        {
+            if (extensionTypeLocator == null)
+            {
+                return;
+            }
+
+            foreach (var type in extensionTypeLocator.GetCloudBlobStreamBinderTypes())
+            {
+                var inst = Activator.CreateInstance(type);
```
mathewc: can you use better local names? :)

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
```diff
@@ -515,6 +522,41 @@ private static Assembly GetHostAssembly(IEnumerable<MethodInfo> methods)
             }
         }
 
+        #region Hackery for ExtensionLocator
```
mathewc: perhaps we should link to an issue if we have one? Hopefully the hackery is temporary?

src/Microsoft.Azure.WebJobs.Host/Indexers/DefaultBindingProvider.cs
```diff
@@ -33,7 +33,7 @@ internal static class DefaultBindingProvider
         {
             List<IBindingProvider> innerProviders = new List<IBindingProvider>();
                      
-            innerProviders.Add(new BlobAttributeBindingProvider(nameResolver, storageAccountProvider, extensionTypeLocator, blobWrittenWatcherGetter));
+            // innerProviders.Add(new BlobAttributeBindingProvider(nameResolver, storageAccountProvider, extensionTypeLocator, blobWrittenWatcherGetter));
```
mathewc: delete commented out code

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
```diff
@@ -175,14 +175,6 @@ public AsyncChainEndToEndTests(TestFixture fixture)
                 Assert.True(queueProcessorFactory.CustomQueueProcessors.All(p => p.Context.Queue.Name.StartsWith("asynce2eq")));
                 Assert.True(queueProcessorFactory.CustomQueueProcessors.Sum(p => p.BeginProcessingCount) >= 2);
                 Assert.True(queueProcessorFactory.CustomQueueProcessors.Sum(p => p.CompleteProcessingCount) >= 2);
-
-                Assert.Equal(19, storageClientFactory.TotalBlobClientCount);
```
mathewc: How did these change? A bit quirky, but I had these in place I believe to verify custom factories were being called?

src/Microsoft.Azure.WebJobs.Host/Config/ExtensionConfigContext.cs
MikeStall: No. Because the objects have a generic signature. 

---
In reply to: [151523014](https://github.com/Azure/azure-webjobs-sdk/pull/1422#discussion_r151523014) [](ancestors = 151523014)

src/Microsoft.Azure.WebJobs.Host/Config/FluentBindingRule.cs
MikeStall: This is fluent. Nobody will see the typenames, just the "rule.BindToInput().AddAvalidator()" etc.  I'll rename the type to whatever you want. 

---
In reply to: [151525218](https://github.com/Azure/azure-webjobs-sdk/pull/1422#discussion_r151525218) [](ancestors = 151525218)

src/Microsoft.Azure.WebJobs/BlobAttribute.cs
```diff
@@ -77,6 +76,7 @@ public BlobAttribute(string blobPath, FileAccess access)
 
         /// <summary>Gets the path of the blob to which to bind.</summary>
         [AutoResolve]
+        [BlobNameValidation]
```
MikeStall: >[BlobNameValidation] [](start = 8, length = 20)

This is a huge step forward. The error messages now come from the DataAnnotation system. This lets us remove some explicit error checking; and it also unifies the various %%, {} cases.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
MikeStall: These were very implementation-specific and the implementation is changed.  They weren't really testing something meaningful to an end-user. 

---
In reply to: [151528273](https://github.com/Azure/azure-webjobs-sdk/pull/1422#discussion_r151528273) [](ancestors = 151528273)

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/BlobBindingEndToEndTests.cs
```diff
@@ -94,7 +94,7 @@ public BlobBindingEndToEndTests(TestFixture fixture)
             var ex = await Assert.ThrowsAsync<FunctionInvocationException>(() =>
            _fixture.Host.CallAsync(typeof(BlobBindingEndToEndTests).GetMethod("CloudBlockBlobBinding_WithUrlBinding"), arguments));
             // CloudBlockBlobBinding_WithUrlBinding is suppose to bind to a blob
-            Assert.Equal($"Invalid absolute blob url: {poco.A}", ex.InnerException.InnerException.Message);
```
mathewc: Can we keep the bit verifying that the error message is as expected?

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/BlobBindingEndToEndTests.cs
```diff
@@ -111,7 +111,8 @@ public BlobBindingEndToEndTests(TestFixture fixture)
             var ex = await Assert.ThrowsAsync<FunctionInvocationException>(() =>
             _fixture.Host.CallAsync(typeof(BlobBindingEndToEndTests).GetMethod("CloudBlobContainerBinding_WithModelBinding"), arguments));
             // CloudBlobContainerBinding_WithModelBinding is suppose to bind to a container
-            Assert.Equal($"Invalid container name: {poco.A}", ex.InnerException.InnerException.Message);
+            // Assert.Equal($"Invalid container name: {poco.A}", ex.InnerException.InnerException.Message);
```
mathewc: Same here

src/Microsoft.Azure.WebJobs.Host/Config/DirectInvokeString.cs
MikeStall: This is used for semantics with Converter Manager where we want to specify how an invoke string goes to a real object. This was needed in cases where String-->Obj is different than InvokeString-->Obj.

---
In reply to: [151523420](https://github.com/Azure/azure-webjobs-sdk/pull/1422#discussion_r151523420) [](ancestors = 151523420)

src/Microsoft.Azure.WebJobs.Host/Config/FluentBindingRule.cs
MikeStall: C# doesn't allow generic constraints on Enums. :(   I can make the method more explicit. 

---
In reply to: [151524466](https://github.com/Azure/azure-webjobs-sdk/pull/1422#discussion_r151524466) [](ancestors = 151524466)

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/BlobBindingEndToEndTests.cs
MikeStall: We don't know the user is trying to give us a URL, so it's ambiguus what error message we specifically get. 

---
In reply to: [151538959](https://github.com/Azure/azure-webjobs-sdk/pull/1422#discussion_r151538959) [](ancestors = 151538959)

PR #1425, opened 2017-11-10
Allow SDK to run purely off just a SAS connection string to a container

MikeStall:
Allow when we have no full Connection strings involved, just a SAS URL to storage container. Added a new property on JobHostConfiguration to describe the sas. 

Config looks like:
    config.HostId = "hostid1234";
    config.StorageConnectionString = null;
    config.DashboardConnectionString = null;
    **config.SaSContainerConnectionString= "xxxx my container SAS xxxx"**

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
@@ -128,9 +128,19 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                 IDistributedLockManager lockManager = services.GetService<IDistributedLockManager>();
                 if (lockManager == null)
                 {
-                    lockManager = new BlobLeaseDistributedLockManager(
-                        storageAccountProvider,
-                        logger);
+                    var sas = config.InternalStorageConfiguration;
+                    if (sas != null && sas.InternalContainer != null)
+                    {
+                        lockManager = new BlobLeaseDistributedLockManager.SasContainer(
+                              sas.InternalContainer,
+                              logger);
+                    }
+                    else
+                    {
+                        lockManager = new BlobLeaseDistributedLockManager.DedicatedStorage(
+                            storageAccountProvider,
+                            logger);
+                    }
                     services.AddService<IDistributedLockManager>(lockManager);
                 }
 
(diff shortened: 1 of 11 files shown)


src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
```diff
@@ -127,10 +127,20 @@ public static ServiceProviderWrapper CreateStaticServices(this JobHostConfigurat
                 IDistributedLockManager lockManager = services.GetService<IDistributedLockManager>();
                 if (lockManager == null)
                 {
-                    lockManager = new BlobLeaseDistributedLockManager(
-                        storageAccountProvider,
-                        trace,
-                        logger);
+                    if (config.SaSContainerConnectionString != null)
+                    {
+                        lockManager = new BlobLeaseDistributedLockManager.SasContainer(
```
mathewc: So we're saying that we're enforcing SAS wins for system services?

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
```diff
@@ -197,6 +197,12 @@ public string StorageConnectionString
             set { _storageAccountProvider.StorageConnectionString = value; }
         }
 
+        /// <summary>
+        /// Gets or sets a SAS connection string for a container. If this is set, <see cref="StorageConnectionString"/> and <see cref="DashboardConnectionString"/>
+        /// can be set to null and the runtime will use the container. 
+        /// </summary>
+        public string SaSContainerConnectionString { get; set; }
```
mathewc: Perhaps rename to BlobContainerSAS? If we ever support additional SAS resources (Tables, Queues) it might be good to have the names scoped to services.

Actually, if we think there is a possibility that we might move in that direction. I suggest we add a new JobHostSASConfiguration object and corresponding SASConnections property. That gives us room to expand SAS support in the future w/o creating a bunch of top level properties.

src/Microsoft.Azure.WebJobs.Host/Singleton/BlobLeaseDistributedLockManager.cs
```diff
@@ -404,5 +399,53 @@ private static string FormatErrorCode(StorageException exception)
...
+        {
+            private readonly StorageBlobContainer _container;
+
+            public SasContainer(
+                string sasContainer,
+                TraceWriter trace,
+                ILogger logger) : base(trace, logger)
+            {
+                var uri = new Uri(sasContainer);
+                var container2 = new CloudBlobContainer(uri);
```
mathewc: why is this called container2?

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
mathewc: Also, it looks like there is a hard coded precedence now - SAS before dedicated for certain system services. Should probably document that?

mathewc:
How does this work relate to supporting SAS connections more generally? I think we have issues tracking this. E.g. we have an existing StorageAccountAttribute that allows you to specify a storage account for various storage based triggers/bindings. There should be a SAS counterpart to that, either the same attribute extended somehow, or a new attribute.

MikeStall:
This is orthogonal to SAS for bindings (ie, StorageAccountAttribute ). This is focused on the SDK's internal usage of Storage (in support of Singleton and Timer) and not explicit user binding to Storage.

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
MikeStall: I'll add JobHostSASConfiguration. I expect this config part to get superseded once we finish the DI work.

PR #1426, opened 2017-11-13
Changing configuration setup order so environment variables take precendence over app settings

### LINKED ISSUE #1408: Appsettings.json override Environment variable values
Looking at https://github.com/Azure/azure-webjobs-sdk/blob/v3.0.0-beta2/src/Microsoft.Azure.WebJobs.Host/ConfigurationUtility.cs I see that the BuildConfiguration function adds the Environment variables first and then the appsettings.json file.

#### Repro steps

See the code mentioned in the link above

#### Expected behavior

I would expect that the Environment variables would override the appsettings.json values (but please correct me if I'm wrong). At least this seems to be the correct practice according to the documentation available at: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration?tabs=basicconfiguration. See excerpt:

> It's typical to have different configuration settings for different environments, for example, development, test, and production. The CreateDefaultBuilder extension method in an ASP.NET Core 2.x app (or using AddJsonFile and AddEnvironmentVariables directly in an ASP.NET Core 1.x app) adds configuration providers for reading JSON files and system configuration sources:2
> 
> - appsettings.json
> - appsettings.<EnvironmentName>.json
> - environment variables
> 

So instead of this:
```
        private static IConfigurationRoot BuildConfiguration()
        {
            var configurationBuilder = new ConfigurationBuilder()
                .AddEnvironmentVariables()
                .AddJsonFile("appsettings.json", optional: true);

            return configurationBuilder.Build();
        }
```

I would expect this:
```
        private static IConfigurationRoot BuildConfiguration()
        {
            var configurationBuilder = new ConfigurationBuilder()
                .AddJsonFile("appsettings.json", optional: true)
                .AddEnvironmentVariables();

            return configurationBuilder.Build();
        }
```

#### Actual behavior

Appsettings.json values override Environment variables.

#### Known workarounds

No known workarounds.

#### Related information 

* See documentation on how to set up configuration settings in ASP.NET Core/.NET Core: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration?tabs=basicconfiguration

* Package version: v3.0.0-beta2
* Links to source:
https://github.com/Azure/azure-webjobs-sdk/blob/v3.0.0-beta2/src/Microsoft.Azure.WebJobs.Host/ConfigurationUtility.cs

fabiocav:
Resolves #1408 

This is a temporary fix as the `ConfigurationUtility` type will be removed as part of the DI work, but the default behavior introduced with this change will be maintained.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/ConfigurationUtility.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/ConfigurationUtility.cs
@@ -53,8 +53,8 @@ namespace Microsoft.Azure.WebJobs.Host
         private static IConfigurationRoot BuildConfiguration()
         {
             var configurationBuilder = new ConfigurationBuilder()
-                .AddEnvironmentVariables()
-                .AddJsonFile("appsettings.json", optional: true);
+                .AddJsonFile("appsettings.json", optional: true)
+                .AddEnvironmentVariables();
 
             return configurationBuilder.Build();
         }
--- a/test/Microsoft.Azure.WebJobs.Host.UnitTests/ConfigurationUtilityTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.UnitTests/ConfigurationUtilityTests.cs
@@ -49,14 +49,14 @@ namespace Microsoft.Azure.WebJobs.Host.UnitTests
         }
 
         [Fact]
-        public void GetSettingFromConfigOrEnvironment_ConfigAndEnvironment_ConfigWins()
+        public void GetSettingFromConfigOrEnvironment_ConfigAndEnvironment_EnvironmentWins()
         {
             ConfigurationUtility.Reset();
 
             Environment.SetEnvironmentVariable("DisableSetting0", "1");
 
             string value = ConfigurationUtility.GetSetting("DisableSetting0");
-            Assert.Equal("0", value);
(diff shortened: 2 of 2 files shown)


src/Microsoft.Azure.WebJobs.Host/ConfigurationUtility.cs
```diff
@@ -53,8 +53,8 @@ public static bool IsSettingEnabled(string settingName)
         private static IConfigurationRoot BuildConfiguration()
         {
             var configurationBuilder = new ConfigurationBuilder()
-                .AddEnvironmentVariables()
-                .AddJsonFile("appsettings.json", optional: true);
+                .AddJsonFile("appsettings.json", optional: true)
+                .AddEnvironmentVariables();
```
mathewc: So now we're consistent with the v2.x behavior that was here: https://github.com/Azure/azure-webjobs-sdk/blob/v2.x/src/Microsoft.Azure.WebJobs.Host/ConfigurationUtility.cs#L21

fabiocav: It's actually the opposite of the previous behavior. I mentioned that during triage, but with this changes, the default will differ than what we had in V2

PR #1437, opened 2017-11-20
Adding protected host state transition methods

mathewc:
Adding protected virtuals to allow host subclasses to implement custom logic at certain stages of host startup. The problem is that currently initialization, indexing and listener startup are all baked into RunAndBlock. In some scenarios (e.g. Azure Functions - see PR https://github.com/Azure/azure-webjobs-sdk-script/pull/2147) subclasses need to implement custom state transition logic at the right points. These changes enable this.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/JobHost.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/JobHost.cs
@@ -123,9 +123,11 @@ namespace Microsoft.Azure.WebJobs
 
         private async Task StartAsyncCore(CancellationToken cancellationToken)
         {
-            await EnsureHostStartedAsync(cancellationToken);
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.WebJobs.Host/JobHost.cs
```diff
@@ -123,10 +123,12 @@ public Task StartAsync(CancellationToken cancellationToken = default(Cancellatio
 
         private async Task StartAsyncCore(CancellationToken cancellationToken)
         {
-            await EnsureHostStartedAsync(cancellationToken);
```
mathewc: As you can see by the fact that below we're starting the listeners, this EnsureHostStarted method was named incorrectly

PR #1438, opened 2017-11-20
removing TraceWriter from internal logging; everything now using ILogger

brettsam:
Removing the usage of TraceWriter internally.

Some major changes to call out:
- ~~Removal of support for `TextWriter` as a logger from within a function.~~
- Removal of `TraceLevelAttribute`. Filters can now handle this.
- Removal of various other `TraceWriter`-specific classes.
- Log categories have shifted. `Host.Executor` is now `Function.{FunctionName}` and logs coming from the user's function have the category `Function.{FunctionName}.User`. This allows the `FileLogger` in Script to route logs to the right file. Also allows the `SystemLogger` to ignore user logs.

Some future todos:
- Implement Dashboard and FastLogger as ILoggers (#1447)


sample/SampleHost/Program.cs
```diff
@@ -36,7 +36,7 @@ private static void CheckAndEnableAppInsights(JobHostConfiguration config)
             {
                 var filter = new LogCategoryFilter();
                 filter.DefaultLevel = LogLevel.Debug;
-                filter.CategoryLevels[LogCategories.Function] = LogLevel.Debug;
+                filter.CategoryLevels["Function"] = LogLevel.Debug;
```
MikeStall: >["Function" [](start = 37, length = 11)

Switching from a symbol to a magical string literal seems like a step backwards... why the change?

src/Microsoft.Azure.WebJobs/TraceLevelAttribute.cs
```diff
@@ -1,38 +0,0 @@
...
-namespace Microsoft.Azure.WebJobs
-{
-    /// <summary>
-    /// Attribute used to specify <see cref="TraceLevel"/> for job functions.
-    /// This setting will affect both Console logging as well as Dashboard logging.
-    /// </summary>
-    /// <remarks>
-    /// For very high throughput functions where performance is key you can set the
-    /// level to <see cref="TraceLevel.Off"/> if you don't want any logs (Console or
-    /// Dashboard) to be produced. Dashboard logging is somewhat expensive, so this
```
MikeStall: Is this scenario preserved?

src/Microsoft.Azure.WebJobs.Host/Bindings/TraceWriter/TraceWriterBindingProvider.cs
```diff
@@ -31,8 +28,7 @@ public Task<IBinding> TryCreateAsync(BindingProviderContext context)
             }
 
             ParameterInfo parameter = context.Parameter;
-            if (parameter.ParameterType != typeof(TraceWriter) &&
-                parameter.ParameterType != typeof(TextWriter))
```
MikeStall: >TextWriter [](start = 50, length = 10)

This is a significant breaking change! ... (I use it frequently in my own projects) . What's the recommended pattern that people switch to?

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/BlobListenerFactory.cs
```diff
@@ -53,92 +51,21 @@ internal class BlobListenerFactory : IListenerFactory
...
-
-            _hostIdProvider = hostIdProvider;
-            _queueConfiguration = queueConfiguration;
-            _blobsConfiguration = blobsConfiguration;
-            _exceptionHandler = exceptionHandler;
-            _blobWrittenWatcherSetter = blobWrittenWatcherSetter;
-            _messageEnqueuedWatcherSetter = messageEnqueuedWatcherSetter;
-            _sharedContextProvider = sharedContextProvider;
-            _trace = trace;
+            _hostIdProvider = hostIdProvider ?? throw new ArgumentNullException("hostIdProvider");
```
MikeStall: >hostIdProvider [](start = 81, length = 14)

nit: as long as we're using fancy new features here ... we could use nameof(hostIdProvider) rather than the string literal.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/SharedBlobQueueListenerFactory.cs
```diff
@@ -33,53 +32,16 @@ internal class SharedBlobQueueListenerFactory : IFactory<SharedBlobQueueListener
...
-                throw new ArgumentNullException("blobWrittenWatcher");
-            }
-
-            _hostAccount = hostAccount;
-            _sharedQueueWatcher = sharedQueueWatcher;
-            _hostBlobTriggerQueue = hostBlobTriggerQueue;
-            _queueConfiguration = queueConfiguration;
-            _exceptionHandler = exceptionHandler;
-            _trace = trace;
+            _hostAccount = hostAccount ?? throw new ArgumentNullException("hostAccount");
```
MikeStall: >"hostAccount" [](start = 74, length = 13)

nameof

src/Microsoft.Azure.WebJobs.Host/Bindings/TraceWriter/TraceWriterBindingProvider.cs
brettsam: I've removed `TextWriter` -- but `TraceWriter` is still usable. Do we need to keep supporting `TextWriter` going forward? This is a chance to remove old stuff if we want to. We can keep it, but we've had some bugs in the past with this and I wasn't sure anyone used it anymore.

sample/SampleHost/Program.cs
brettsam: It's actually a bad sample; I'll fix it. It used to be that all user function logs came with "Function" and all executor logs came with "Host.Executor", but to make filtering nicer (replacing `TraceLevelAttribute`), you now can use the category "Function.Function1" to control a single function's logs. Or "Function.Function1.User" to control the user logs from a specific function. There are helpers for creating these categories: `LogCategories.CreateFunctionLog(categoryName)`. I'll use those here as a better example.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/BlobListenerFactory.cs
brettsam: Yeah, I noticed that as well right when I sent this PR. These were all done with VS auto-correct so I didn't notice that it didn't auto-correct to `nameof()`. But I agree with you --  I'll update all of these with my next set of changes.

src/Microsoft.Azure.WebJobs.Host/Bindings/TraceWriter/TraceWriterBindingProvider.cs
MikeStall: I use it all the time as a replacement for Console.Out.  ;)    One scenario is to have "Admin" commands that are [ManualInvoke] that bind to TextWriter to write diagnostic information to the dashboard. It's like a poor-man's ACIS. 
Is this easy to shim on TraceWriter? 

---
In reply to: [154748132](https://github.com/Azure/azure-webjobs-sdk/pull/1438#discussion_r154748132) [](ancestors = 154748132)

brettsam: If you use TraceWriter or ILogger from within the function, it will still log to the dashboard. Or does TextWriter have some kind of behavior that's unique?

For the sake of this PR, what about if I add it back (it should be fairly straightforward) and we can follow up later to see if it's worth pulling out?

MikeStall: That'd be fine. Thanks.

src/Microsoft.Azure.WebJobs.Host/Bindings/TraceWriter/TraceWriterBindingProvider.cs
```diff
@@ -10,9 +10,7 @@
 namespace Microsoft.Azure.WebJobs.Host.Bindings
 {
     /// <summary>
-    /// Binding provider handling bindings to both <see cref="TraceWriter"/> and <see cref="TextWriter"/>.
-    /// <remarks>
-    /// </remarks>
+    /// Binding provider handling bindings to <see cref="TraceWriter"/>.
```
mathewc: It still supports TextWriter though

src/Microsoft.Azure.WebJobs.Host/ConsoleTraceWriter.cs
```diff
@@ -1,53 +0,0 @@
...
-using System.Diagnostics;
-using System.IO;
-
-namespace Microsoft.Azure.WebJobs.Host
-{
-    /// <summary>
-    /// This <see cref="TraceWriter"/> delegates to a collection of inner user <see cref="TraceWriter"/>s (from <see cref="JobHostConfiguration.Tracing"/>,
-    /// as well as a Console <see cref="TextWriter"/>.
-    /// </summary>
-    internal class ConsoleTraceWriter : CompositeTraceWriter
```
mathewc: Where is console logging and coloration handled now?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInstanceTraceWriter.cs
```diff
@@ -1,39 +0,0 @@
...
-        }
-
-        public override void Trace(TraceEvent traceEvent)
-        {
-            if (traceEvent == null)
-            {
-                throw new ArgumentNullException("traceEvent");
-            }
-
-            traceEvent.AddFunctionInstanceDetails(_hostInstanceId, _instance.FunctionDescriptor, _instance.Id);
```
mathewc: Where does this happen in your new ILogger pipeline? Also, if we're still supporting a TraceWriter binding, don't we still need this to happen for such traces?

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
```diff
@@ -87,11 +86,6 @@ public static ServiceProviderWrapper CreateStaticServices(this JobHostConfigurat
             services.AddService(blobWrittenWatcherAccessor);
             ISharedContextProvider sharedContextProvider = new SharedContextProvider();
 
-            // Create a wrapper TraceWriter that delegates to both the user 
-            // TraceWriters specified via config (if present), as well as to Console
-            TraceWriter trace = new ConsoleTraceWriter(config.Tracing, consoleProvider.Out);
```
mathewc: Verify that user traces are still written to console

src/Microsoft.Azure.WebJobs/TraceLevelAttribute.cs
mathewc: Doesn't look like it. Why not? It was useful - should be easy to carry forward for ILogger

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
```diff
@@ -372,9 +363,6 @@ private bool HasJobAttribute(CustomAttributeData attributeData)
         {
             var disabled = HostListenerFactory.IsDisabled(method, nameResolver, jobActivator);
 
-            // Determine the TraceLevel for this function (affecting both Console as well as Dashboard logging)
-            TraceLevelAttribute traceAttribute = TypeUtility.GetHierarchicalAttributeOrNull<TraceLevelAttribute>(method);
```
mathewc: Why are we not supporting this anymore?

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
```diff
@@ -337,7 +325,6 @@ public StorageClientFactory StorageClientFactory
         /// </summary>
         public void UseDevelopmentSettings()
         {
-            Tracing.ConsoleLevel = TraceLevel.Verbose;
```
mathewc: There isn't an ILogger analog to this? Configure the logger for maximal debug output?

src/Microsoft.Azure.WebJobs.Host/Listeners/HostListenerFactory.cs
```diff
@@ -49,7 +47,6 @@ internal class HostListenerFactory : IListenerFactory
                 if (functionDefinition.Descriptor.IsDisabled)
                 {
                     string msg = string.Format("Function '{0}' is disabled", functionDefinition.Descriptor.ShortName);
-                    _trace.Info(msg, TraceSource.Host);
```
mathewc: Can't continue to flow category info? Or are you saying this isn't important in this case?

src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonListener.cs
```diff
@@ -52,7 +50,6 @@ public SingletonListener(FunctionDescriptor method, SingletonAttribute attribute
             if (_lockHandle == null)
             {
                 string msg = string.Format(CultureInfo.InvariantCulture, "Unable to acquire Singleton lock ({0}).", _lockId);
-                _trace.Verbose(msg, source: TraceSource.Execution);
                 _logger?.LogDebug(msg);
```
mathewc: Same question here and throughout about preservation of category/source info

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
```diff
@@ -510,95 +417,17 @@ public void FunctionFailures_LogsExpectedTraceEvent()
...
-            TraceEvent[] traceErrors = trace.Traces.Where(p => p.Level == TraceLevel.Error).ToArray();
-            Assert.Equal(0, traceErrors.Length);
-
             // Validate Logger
             LogMessage[] logErrors = _loggerProvider.GetAllLogMessages().Where(l => l.Level == Microsoft.Extensions.Logging.LogLevel.Error).ToArray();
             Assert.Equal(0, logErrors.Length);
         }
 
-        [Fact]
-        public async Task FunctionTraceLevelOverride_ProducesExpectedOutput()
```
mathewc: Add tests back if we preserve this feature

src/Microsoft.Azure.WebJobs/TraceLevelAttribute.cs
brettsam: You can use the Filters now, rather than Attributes. You can select which levels you want for each function.

src/Microsoft.Azure.WebJobs.Host/Bindings/TraceWriter/TraceWriterBindingProvider.cs
brettsam: Oh, right. I re-added that and didn't fix the comment.

src/Microsoft.Azure.WebJobs.Host/ConsoleTraceWriter.cs
brettsam: You now get the default ConsoleLogger that ships with .NET. It doesn't look quite as minimal as what we currently have, though. We should be able to tweak it and use something custom.

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
brettsam: Yep, they are -- they go through the normal ConsoleLogger.

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
brettsam: I mentioned it above -- it's handled via the Filters you can build. You can control each category -- so you can tweak individual function levels.

src/Microsoft.Azure.WebJobs.Host/Listeners/HostListenerFactory.cs
brettsam: Category info is already tracked in ILoggers -- when you create one, it has a category attached for every log it writes. So it's per-logger rather than per-message.

src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonListener.cs
brettsam: Same answer :-)

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
brettsam: There may be something we can do -- like change the default logger level up to Debug.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInstanceTraceWriter.cs
brettsam: These were useful for people plugging in their own TraceWriters to our pipeline. Today, they'd plug in their own ILogger. We wrap all the calls to functions in a logger.BeginScope(<function details>) call, which makes the function details available to any loggers that are registered. This is how the App Insights logger gets the details about a specific message.

The instance details we're providing are slightly different -- but I did notice that I'm missing the hostInstanceId, for example. I'll add that to the scope, and we can continue adding more details as we get requests to do so.

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
brettsam: This is a little tougher to do now since the Filter is created by the user and we don't have direct access to it -- I'll need to refactor. I've filed https://github.com/Azure/azure-webjobs-sdk/issues/1471.

Issue #1427, opened 2017-11-13
Increase total amount of messages we enqueue for blob trigger


watashiSHUN:
currently when we scan blob or scan logs
we will add the blob discovered one by one to the worker queue
https://github.com/Azure/azure-webjobs-sdk/blob/def317119bc15b4c653fcff1f538df4e6d8420f4/src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs#L134-L138

https://github.com/Azure/azure-webjobs-sdk/blob/def317119bc15b4c653fcff1f538df4e6d8420f4/src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/PollLogsStrategy.cs#L108-L111

should make them concurrent to take benefit of 50 default connection limit:  https://github.com/Azure/azure-webjobs-sdk-script/commit/acdc441c9b935cd59465cfa1f59a782940658e4f

this should increase the number of blob trigger execution per unit time

PR #1440, opened 2017-11-21
make blob enqueue concurrent

watashiSHUN:
tested with very simple blob trigger
here is the result of processing 100 new blobs (running on my local machine)
1. sequential:
image
average time spent on writing to sharedBlobQueue is (51150-171)/100=509 ms/blob
2. parallel:
image
average time spent on writing to sharedBlobQueue is (1488-122)/100=13 ms/blob
since my socket upper limit is 50(same as in Azure) , there is basically two group send, each cost (1488-122)/2=683 ms (a little longer than single message send, but the quantity make up for it)

Here is a simulation using the old method (when function execution is fairly short)
```
blobListener thread                      queueListener thread
finished enqueuing blobA
started enqueuing blobB                   dequeued blobA
finished enqueuing blobB                  finished blobA
started enqueuing blobC                   dequeued blobB
```
when enqueue speed is similar to the dequeue and execution speed, we will see blob jobs 
being executed sequentially, no matter what the batch size user specified, we are always dequeuing 1 message at a time, which cost bandwidth and also stop us from scaling. https://github.com/Azure/azure-webjobs-sdk/issues/1427

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/PollLogsStrategy.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/PollLogsStrategy.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
@@ -13,6 +14,7 @@ using Microsoft.Azure.WebJobs.Host.Storage;
 using Microsoft.Azure.WebJobs.Host.Storage.Blob;
 using Microsoft.Azure.WebJobs.Host.Timers;
 using Microsoft.WindowsAzure.Storage;
+using ThreadState = System.Threading.ThreadState;
 
 namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
 {
@@ -26,10 +28,12 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
         private readonly Thread _initialScanThread;
         private readonly ConcurrentQueue<IStorageBlob> _blobsFoundFromScanOrNotification;
         private readonly CancellationTokenSource _cancellationTokenSource;
+        private readonly TraceWriter _trace;
+        private readonly Stopwatch _stopwatch;
         private bool _performInitialScan;
         private bool _disposed;
 
-        public PollLogsStrategy(bool performInitialScan = true)
+        public PollLogsStrategy(bool performInitialScan = true, TraceWriter trace = null)
         {
             _registrations = new Dictionary<IStorageBlobContainer, ICollection<ITriggerExecutor<IStorageBlob>>>(
                 new StorageBlobContainerComparer());
@@ -38,6 +42,8 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
             _blobsFoundFromScanOrNotification = new ConcurrentQueue<IStorageBlob>();
             _cancellationTokenSource = new CancellationTokenSource();
             _performInitialScan = performInitialScan;
+            _trace = trace;
+            _stopwatch = new Stopwatch();
         }
 
         public async Task RegisterAsync(IStorageBlobContainer container, ITriggerExecutor<IStorageBlob> triggerExecutor,
@@ -86,6 +92,9 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
         {
             ThrowIfDisposed();
 
+            ConcurrentBag<IStorageBlob> failedNotifications = new ConcurrentBag<IStorageBlob>(); // NotifyRegistrationAsync is concurrent
+            List<Task> notifications = new List<Task>();
+
             // Drain the background queue (for initial container scans and blob written notifications).
             while (true)
             {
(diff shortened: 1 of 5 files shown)


src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/PollLogsStrategy.cs
```diff
@@ -97,18 +115,76 @@ public void Notify(IStorageBlob blobWritten)
                     break;
                 }
 
-                await NotifyRegistrationsAsync(blob, cancellationToken);
+                notifications.Add(NotifyRegistrationsAsync(blob, failedNotifications, cancellationToken));
```
mathewc: What are the bounds on the number of things that can be in the background queue? While we want to parallelize things, we also need to constrain things so that we're not attempting to start thousands of tasks simultaneously.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/PollLogsStrategy.cs
```diff
@@ -97,18 +115,76 @@ public void Notify(IStorageBlob blobWritten)
...
+                    var blobs = await logListener.GetRecentBlobWritesAsync(cancellationToken);
+                    readTimes[index] = _stopWatch.ElapsedMilliseconds - before;
+
+                    List<Task> enqueues = new List<Task>();
+                    foreach (IStorageBlob blob in blobs)
+                    {
+                        cancellationToken.ThrowIfCancellationRequested();
+                        enqueues.Add(NotifyRegistrationsAsync(blob, failedNotifications, cancellationToken));
+                    }
+                    await Task.WhenAll(enqueues);
```
mathewc: Same question here on bounds - what if there were 10,000 new blobs here? Probably don't want to start 10k tasks in parallel. Instead we can do batching.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/PollLogsStrategy.cs
```diff
@@ -97,18 +115,76 @@ public void Notify(IStorageBlob blobWritten)
...
+
+                }));
+            }
+
+            await Task.WhenAll(scanLogs);
```
mathewc: This set of tasks is probably constrained enough - it is the number of distinct blob containers we're monitoring? How many "log listeners" are there?

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/PollLogsStrategy.cs
```diff
@@ -97,18 +115,76 @@ public void Notify(IStorageBlob blobWritten)
...
+                {
+                    _traceString.AppendLine($"log inventory ==> Enqueue: {beforeLogTryEnqueue} messages, {beforeLogSuccess} successes, {beforeLogSkipped} skipped. Cost: {beforeLogTime}ms");
+                }
+                if (logTryEnqueue > 0)
+                {
+                    _traceString.AppendLine($"log new ==> Enqueue: {logTryEnqueue} messages, {logSuccess} successes, {logSkipped} skipped. Average log cost: {averageReadTime}ms, total cost: {logEnqueueTime}ms");
+                }
+                if (_traceString.Length > 0)
+                {
+                    _trace.Verbose(_traceString.ToString());
```
mathewc: I'm concerned about the noise these logs will introduce. I realize you've made them Verbose, but they're really beyond that from the customer's point of view - they're Debug/Diagnostic level. We have customers who have set their logs to Verbose and I don't think they'd want to see these logs.

Recommend separating out your tracing proposal from the parallelization work since they are separate, so we can decide on what to do there separately.

src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/PollLogsStrategy.cs
```diff
@@ -154,17 +230,29 @@ public void Dispose()
                 return;
             }
 
+            bool succeeded = true;
+            // assume only one registration
```
mathewc: What does this comment mean?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionResult.cs
```diff
@@ -66,5 +66,10 @@ public FunctionResult(bool succeeded, Exception exception)
         /// Gets or sets the exception that occurred during function execution.
         /// </summary>
         public Exception Exception { get; private set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether the function was exited early.
+        /// </summary>
+        public bool ReturnedEarly { get; set; }
```
mathewc: I don't think we should add this to this public type - the new property doesn't have much meaning in many of the existing usages of the type.

I think you're adding this to drive your logging stuff, so if you don't do the logging in this PR, this wouldn't be needed? Another way to do this would be for you to create your own subclass for use by the blob executor.

paulbatum:
@watashiSHUN whats the status on this pr?

watashiSHUN:
it is deprioritized as I was working on some deployment stuff, I will have it updated this week

paulbatum:
@watashiSHUN ping!

fabiocav:
@watashiSHUN closing this as there has been no activity. Please rebase/reopen when this work is ready.

PR #1445, opened 2017-11-28
Namespace updates, remove [Obsolete]s

MikeStall:
Anything used to decorate a binding attribute goes into the Microsoft.Azure.WebJobs.Description namespace.

Notably [AppSetting] and [AutoResolve]
Remove more [Obsolete] attributes.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Config/ExtensionConfigContextExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Config/ExtensionConfigContextExtensions.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
-using Microsoft.Azure.WebJobs.Host.Bindings;
 using Newtonsoft.Json;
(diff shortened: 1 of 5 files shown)


src/Microsoft.Azure.WebJobs.Host/Config/ExtensionConfigContextExtensions.cs
```diff
@@ -14,64 +13,6 @@ namespace Microsoft.Azure.WebJobs.Host.Config
     [Obsolete("Not ready for public consumption.")]
```
fabiocav: Did you mean to remove this?

MikeStall: Not yet - I think we can remove this entire class once we do the DI work. The remaining methods are just about config things.

fabiocav: Sounds good.

Issue #1457, opened 2017-12-11
SingletonAttribute LockAcquisitionTimeOut property not settable


mathewc:
If you try to specify a value for this nullable attribute property, e.g.:

```csharp
[Singleton("%test%", SingletonScope.Function, LockAcquisitionTimeout = 1)]
```

You'll get a compilation error "'LockAcquisitionTimeout' is not a valid named attribute argument because it is not a valid attribute parameter type". We need to make this non-nullable. Surprised we don't have tests verifying this property override works.

Let's be sure to get a fix for this out ASAP.

mathewc:
Fixed

PR #1458, opened 2017-12-11
SingletonAttribute LockAcquisitionTimeout property fix

mathewc:
Fix for https://github.com/Azure/azure-webjobs-sdk/issues/1457

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonManager.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonManager.cs
@@ -89,13 +89,11 @@ namespace Microsoft.Azure.WebJobs.Host
 
         public async virtual Task<RenewableLockHandle> LockAsync(string lockId, string functionInstanceId, SingletonAttribute attribute, CancellationToken cancellationToken)
         {
-            RenewableLockHandle lockHandle = await TryLockAsync(lockId, functionInstanceId, attribute, cancellationToken);
+            var lockHandle = await TryLockAsync(lockId, functionInstanceId, attribute, cancellationToken);
 
             if (lockHandle == null)
             {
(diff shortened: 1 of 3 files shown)


test/Microsoft.Azure.WebJobs.Host.UnitTests/Singleton/SingletonManagerTests.cs
```diff
@@ -554,10 +554,32 @@ public void GetLockPeriod_ReturnsExpectedValue()
...
+            config.LockAcquisitionTimeout = TimeSpan.FromSeconds(3);
+            result = SingletonManager.GetLockAcquisitionTimeout(attribute, config);
+            Assert.Equal(config.LockAcquisitionTimeout, result);
+        }
+
         private static void TestJob()
         {
         }
 
+        [Singleton("Function", Mode = SingletonMode.Function, LockAcquisitionTimeout = 5)]
```
mathewc: This is the repro - wouldn't compile until after the fix.

src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonManager.cs
```diff
@@ -286,6 +281,15 @@ public static SingletonAttribute GetListenerSingletonOrNull(Type listenerType, F
             return singletonAttribute;
         }
 
+        internal static TimeSpan GetLockAcquisitionTimeout(SingletonAttribute attribute, SingletonConfiguration config)
+        {
+            TimeSpan acquisitionTimeout = attribute.LockAcquisitionTimeout != -1
+                    ? TimeSpan.FromSeconds(attribute.LockAcquisitionTimeout) :
```
fabiocav: nit: I know this code just moved and this is how it was before, but can you move the `:` to the line below? :)

PR #1464, opened 2017-12-13
Allow specifying a timeout at the JobHostConfiguration level.

MikeStall:
Fix https://github.com/Azure/azure-webjobs-sdk-script/issues/2105
Still need an update in Script to set this.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
@@ -317,6 +317,12 @@ namespace Microsoft.Azure.WebJobs
         }
 
         /// <summary>
+        /// Default timeout configuration to apply to all functions. 
+        /// If <see cref="TimeoutAttribute"/> is explicitly applied to a function, the values specified by the attribute will take precedence
+        /// </summary>
+        public JobHostFunctionTimeoutConfiguration FunctionTimeout { get; set; }
+
+        /// <summary>
         /// Configures various configuration settings on this <see cref="JobHostConfiguration"/> to
         /// optimize for local development.
         /// </summary>
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
@@ -264,6 +264,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
 
                 if (functionIndexProvider == null)
                 {
+                    var defaultTimeout = config.FunctionTimeout?.ToAttribute();
                     functionIndexProvider = new FunctionIndexProvider(
                         services.GetService<ITypeLocator>(),
                         triggerBindingProvider,
@@ -274,7 +275,8 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                         singletonManager,
                         trace,
                         loggerFactory,
-                        hostSharedQueue);
+                        hostSharedQueue,
+                        defaultTimeout);
 
                     // Important to set this so that the func we passed to DynamicHostIdProvider can pick it up. 
                     services.AddService<IFunctionIndexProvider>(functionIndexProvider);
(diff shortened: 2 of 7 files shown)


src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
```diff
@@ -316,6 +316,11 @@ public StorageClientFactory StorageClientFactory
             }
         }
 
+        /// <summary>
+        /// IF set, apply this timeout attribute to all functions. 
+        /// </summary>
+        public TimeoutAttribute DefaultTimeout { get; set; }
```
mathewc: This should be a TimeSpan value, not an attribute. Also, I think this should be named "FunctionTimeout" to make it clear that it's a timeout for function invocations.

MikeStall: The reason I used the attribute was because there are several knobs involved here (as you noted in the comments).  So we'd end up with a "TimeoutConfig" which is a clone of the attribute, so I just used the attribute.

mathewc: My vote is for a TimeoutConfiguration class to hold the knobs we want to expose host level. Gives us more flexibility to control what we expose, and also addresses the weirdness of having an attribute specified on host config.

MikeStall: Ok - but for now, will that config mirror the attribute exactly?

mathewc: Sure it could

mathewc: You could even define an interface if you wanted for the settings, so attribute and config both implement. The JobHostConfiguration property should probably be called FunctionTimeout

MikeStall: Currently, the attribute *is* the config (with 4 knobs so far), and all the knobs are non-nullable (ie, as soon as you add an attribute, you're opting into setting all of them).  

We can categorize our options by what type of property we're adding to JobHostConfiguration:
1. a new interface, ITimeoutConfiguration.  For correctness, this must match the existing TimeoutAttribute.  But interfaces are a bad choice for config bags since it's a breaking change to add new knobs. 
2. a new class, TimeoutConfiguration.  Again, this will literally be a clone of  TimeoutAttribute, and we copy the properties over. This raises questions about which of these is the ultimately source of truth (ie, does FunctionDescriptor still hold onto a TimeoutAttribute, or do we change it to hold onto a TimeoutConfiguration). 
3. the existing TimeoutAttribute class.   This keeps the simplest invariants and avoids redundant types, but it may be weird to have the JobHostConfiguration hold onto an attribute.

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
```diff
@@ -316,6 +316,11 @@ public StorageClientFactory StorageClientFactory
             }
         }
 
+        /// <summary>
+        /// If set, apply this timeout attribute to all functions.
```
mathewc: "Default timeout configuration to apply to all functions. If TimeoutAttribute is explicitly applied to a function, the values specified by the attribute will take precedence"

src/Microsoft.Azure.WebJobs.Host/JobHostTimeoutConfiguration.cs
```diff
@@ -0,0 +1,48 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Globalization;
+using System.Threading;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// Control timeout configuration for function execution.
```
mathewc: nit: "Configuration options for controlling function execution timeout behavior"?

src/Microsoft.Azure.WebJobs.Host/JobHostTimeoutConfiguration.cs
```diff
@@ -0,0 +1,48 @@
...
+using System;
+using System.Globalization;
+using System.Threading;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// Control timeout configuration for function execution. 
+    /// </summary>
+    public class JobHostTimeoutConfiguration
```
mathewc: I wonder if this should be named JobHostFunctionTimeoutConfiguration?

PR #1496, opened 2018-01-09
Improved binding error messaging. Fixes #1396.

### LINKED ISSUE #1396: Give better error message on binding error from a dll mismatch
We're seeing several binding errors that appear to be DLL mismatches. Can we give a more helpful warning?  The user refers to Storage SDK Ver X, our runtime binds against Ver Y, and so we don't recognize the 'CloudQueue' parameter type.  

- https://stackoverflow.com/questions/46812713/azure-functions-cant-bind-queue-to-type-cloudqueue/46819927#46819927
- https://stackoverflow.com/questions/46815207/error-binding-string-output-queue-in-azure-function#comment80618935_46815207

Can we detect this and give a better error? Maybe a warning if the type-string matches? 

We had a similar problem with EventHub.   They type 'EventData' is defined twice:
"Microsoft.Azure.EventHubs.EventData"  from Microsoft.Azure.EventHubs.1.0.3 
"Microsoft.ServiceBus.Messaging.EventData" from WindowsAzure.ServiceBus.3.4.5

kashimiz:
* If a binding provider fails to bind to a parameter, it will log
information in the context. If no provider successfully binds to a
parameter, the logged errors will be returned as part of the exception
message. This should improve user ability to detect assembly mismatches.


src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/GenericCompositeBindingProvider.cs
```diff
@@ -67,7 +68,19 @@ public GenericCompositeBindingProvider(IEnumerable<IBindingProvider> providers)
...
                 resourceName = resourceName.Substring(0, resourceName.Length - Suffix.Length);
             }
-            throw new InvalidOperationException("Can't bind " + resourceName + " to type '" + context.Parameter.ParameterType + "'.");
+
+            StringBuilder exceptionMessage = new StringBuilder($"Can't bind {resourceName} to type '{context.Parameter.ParameterType}'.");
+
+            if (context.BindingErrors.Count > 0)
+            {
+                exceptionMessage.AppendLine().AppendLine("Possible causes:");
+                foreach (var error in context.BindingErrors.Select((value, index) => new { Value = value, Index = index + 1 }))
```
MikeStall: >Select((value, index) => new { Value = value, Index = index + 1 }) [](start = 60, length = 66)

nit: can we avoid this extra select clause and just use a local? 
Ie, something like: 
    var x = index+1
    $"{x}) {error.value}"

src/Microsoft.Azure.WebJobs.Host/Queues/Bindings/QueueBindingProvider.cs
```diff
@@ -79,9 +79,6 @@ public void Initialize(ExtensionConfigContext context)
 
                 binding.SetPostResolveHook(ToReadWriteParameterDescriptorForCollector)
                         .BindToInput<CloudQueue>(builder);
```
MikeStall: Why are we deleting this?

kashimiz: It's a duplicate line of code.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/GenericCompositeBindingProvider.cs
kashimiz: Sure; I thought the numbering improved readability, but I can also remove it entirely if preferred.

PR #1497, opened 2018-01-09
 Enable service bus binding provider tests. Fixes #2265

### LINKED ISSUE #2265: [AppInsights] Expose EndpointAddress via settings
In order to use a government cloud you need to adjust the ServerTelemetryChannel's EndpointAddress. You should be able to do this via DI today with something like:

`builder.Services.AddSingleton<ITelemetryChannel>(_ => new ServerTelemetryChannel { EndpointAddress = "https://dc.applicationinsights.us/v2/track" });`

... but it'd be nice to have it exposed as an easy setting that we auto-apply for you, especially in Functions.

alrod:


alrod:
@MikeStall can you please take a look? I did the same thing as for BlobTrigger/Blob Output.

src/Microsoft.Azure.WebJobs.ServiceBus/ServiceBusAttribute.cs
```diff
@@ -38,6 +38,18 @@ public sealed class ServiceBusAttribute : Attribute, IConnectionProvider
...
             QueueOrTopicName = queueOrTopicName;
+            EntityType = EntityType.Queue;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ServiceBusAttribute"/> class.
+        /// </summary>
+        /// <param name="queueOrTopicName">The name of the queue or topic to bind to.</param>
+        /// <param name="queueOrTopicName">The type of the entity to bind to.</param>
+        public ServiceBusAttribute(string queueOrTopicName, EntityType entityType)
```
MikeStall: curious: why do we need to move this to a ctor parameter?

alrod: because we added: `metadata["entityType"] = "Queue";` in this change. In V1 EntityType property is set up in script repo.

Issue #1501, opened 2018-01-11
Fix Analyzer Project failures


christopheranderson:
Getting failures from the Analyzer projects. Need to fix that.

```
:\projects\azure-webjobs-sdk-881mw\src\Analyzer\WebJobs.VSAnalyzer.Core\WebJobs.VSAnalyzer.Core.csproj(210,5): error : This project references NuGet package(s) that are missing on this computer. Use NuGet Package Restore to download them.  For more information, see http://go.microsoft.com/fwlink/?LinkID=322105. The missing file is ..\..\..\packages\NETStandard.Library.2.0.1\build\netstandard2.0\NETStandard.Library.targets.
C:\projects\azure-webjobs-sdk-881mw\src\Analyzer\WebJobs.VSAnalyzer.Vsix\Analyzer2.Vsix.csproj(150,3): error MSB4019: The imported project "C:\Program Files\dotnet\sdk\2.1.2\Microsoft\VisualStudio\v15.0\VSSDK\Microsoft.VsSDK.targets" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk.
```

MikeStall:
That looks like a problem with the MS build system and nuget?

paulbatum:
The analyzer projects are relatively new, chat with @MikeStall, he should be able to take care of this.

MikeStall:
This looks like a bug in Nuget/msbuild. Do we have a contact on that team?

christopheranderson:
I have a PR with a fix out. #1503. The csproj for the vsix had a reference to a targets file that didn't exist and the csproj for core was failing during a post build step that I just commented out because it wasn't being used anywhere.

christopheranderson:
This is still happening. We probably need to just remove this from the build or we need to update the CI to install the SDK for this.

fabiocav:
If there's an additional dependency introduced by the new projects (which seems to be the case with the the required SDK), that would need to be added to the build process. That's another good reason to keep that in a separate solution and keep the build project separate.

@MikeStall do we need the VSIX project in this solution?

PR #1503, opened 2018-01-11
Fixed project issues with the Analyzer Projects

christopheranderson:
Fixes https://github.com/Azure/azure-webjobs-sdk/issues/1501

### DIFF
--- a/src/Analyzer/WebJobs.VSAnalyzer.Vsix/Analyzer2.Vsix.csproj
+++ b/src/Analyzer/WebJobs.VSAnalyzer.Vsix/Analyzer2.Vsix.csproj
@@ -147,5 +147,5 @@
 
 
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
-  <Import Project="$(VSToolsPath)\VSSDK\Microsoft.VsSDK.targets" Condition="'$(VSToolsPath)' != ''" />
+  <!--<Import Project="$(VSToolsPath)\VSSDK\Microsoft.VsSDK.targets" Condition="'$(VSToolsPath)' != ''" />-->
 </Project>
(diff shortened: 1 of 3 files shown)


src/Analyzer/WebJobs.VSAnalyzer.Vsix/Analyzer2.Vsix.csproj
```diff
@@ -147,5 +147,5 @@
 
 
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
-  <Import Project="$(VSToolsPath)\VSSDK\Microsoft.VsSDK.targets" Condition="'$(VSToolsPath)' != ''" />
+  <!--<Import Project="$(VSToolsPath)\VSSDK\Microsoft.VsSDK.targets" Condition="'$(VSToolsPath)' != ''" />-->
```
fabiocav: Is this removal required if you build using the VS Developer command prompt?

christopheranderson: I didn't try the command prompt, but it was failing in VS. The targets file doesn't exist. Do you expect that to change from the VS command prompt?

fabiocav: Not if the SDK is not installed. This essentially introduces the SDK dependency, which you'd need to install, and at that point, building from the dev command prompt should work. I just wonder if removing this will have some impact on the analyzer projects.

christopheranderson: I installed the SDK and it still failed. It's fair to say I didn't do any explicit testing the analyzer project, just that it built. Is there a flow to test it I can follow?

christopheranderson:
Opened https://github.com/Azure/azure-webjobs-sdk/issues/1515 to address Fabio's comments about the Analyzer project needing investigation as to whether it is still functional.

PR #1506, opened 2018-01-11
Improved error message when running against stopped storage emulator. (v2)

kashimiz:
* Fixes Azure/azure-webjobs-sdk-script#2237 (v2)
* Corresponding PR for v1 is #1507

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
@@ -97,7 +97,6 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
```diff
@@ -55,6 +58,11 @@ internal class DefaultStorageCredentialsValidator : IStorageCredentialsValidator
                     account.Type = StorageAccountType.Premium;
                     return;
                 }
+                else if (isDevStoreAccount)
+                {
+                    // If using the storage emulator, it might not be running
+                    throw new InvalidOperationException(Constants.CheckAzureStorageEmulatorMessage);
```
paulbatum: Please propagate the inner exception. i.e. pass the caught exception into the InvalidOperationException constructor

kashimiz: Should we be doing this for the invalid credentials case as well?

PR #1507, opened 2018-01-11
Improved error message when running against stopped storage emulator. (v1)

kashimiz:
* Fixes Azure/azure-webjobs-sdk-script#2237 (v1)
* Corresponding PR for v2 is #1506

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
@@ -61,7 +61,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                 else if (isDevStoreAccount)
                 {
                     // If using the storage emulator, it might not be running
-                    throw new InvalidOperationException(Constants.CheckAzureStorageEmulatorMessage);
+                    throw new InvalidOperationException(Constants.CheckAzureStorageEmulatorMessage, e);
                 }
                 else
                 {
@@ -69,7 +69,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                     string message = String.Format(CultureInfo.CurrentCulture,
                         "Invalid storage account '{0}'. Please make sure your credentials are correct.",
                         account.Credentials.AccountName);
-                    throw new InvalidOperationException(message);
+                    throw new InvalidOperationException(message, e);
                 }
             }
 
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Net;
+using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host.Storage;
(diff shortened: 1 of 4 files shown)


src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
```diff
@@ -88,5 +96,23 @@ internal class DefaultStorageCredentialsValidator : IStorageCredentialsValidator
                 throw;
             }
         }
+
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA2204:Literals should be spelled correctly", MessageId = "IsDevStoreAccount")]
```
brettsam: nit: can you move this to the GlobalSuppressions? That's where we keep all our supressions.

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
```diff
@@ -88,5 +96,23 @@ internal class DefaultStorageCredentialsValidator : IStorageCredentialsValidator
...
+        internal static bool GetIsDevStoreAccountFromCloudStorageAccount(CloudStorageAccount account)
+        {
```
brettsam: Instead of private reflection, can we check for the emulator account URL somehow, instead?

test/Microsoft.Azure.WebJobs.Host.UnitTests/Executors/DefaultStorageCredentialsValidatorTests.cs
```diff
@@ -15,6 +16,45 @@ namespace Microsoft.Azure.WebJobs.Host.UnitTests.Executors
...
+                .Returns(new StorageCredentials("name", string.Empty));
+
+            storageMock.Setup(s => s.CreateBlobClient(null))
+                .Returns(blobClientMock.Object)
+                .Verifiable();
+
+            storageMock.Setup(s => s.SdkObject)
+                .Returns(() => {
+                    var account = new CloudStorageAccount(new StorageCredentials("name", string.Empty), false);
+                    typeof(CloudStorageAccount).GetProperty("IsDevStoreAccount", BindingFlags.NonPublic | BindingFlags.Instance).SetValue(account, true);
```
brettsam: Instead of updating this property, can you use actual emulator storage credentials? That'd make sure that it works e2e.

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
kashimiz: I had thought of this, but hesitated since we don't have control of that URL and it could change in the future. If it's preferable to private reflection, though, I'll go ahead with it.

brettsam: As I'm checking the [emulator overview docs](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-emulator), it also looks like there's a way to use the `devstoreaccount1` account name, and even a way to specify a development proxy. So I guess it's not as simple as just checking a URL as I thought it would be.

I think it's okay to do private reflection as long as our test hits this code path (via a 'real' e2e manner like I commented elsewhere). That will alert us pretty quickly if the private reflection has changed.

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
brettsam: I see this in GlobalSuppressions now, but did you forget to remove it from here?

Issue #1499, opened 2018-01-10
Using blob only storage account in webjobs 3.x


NatMarchand:
When using another storage account which is blob only for binding, the operation fail with a Service Unavailable exception.

#### Repro steps
Have two storage account :
- account1 which is a general purpose account used as AzureWebJobsStorage account
- account2 which is a blob only account

```        [FunctionName("MyFunc")]
        public static async Task MyFuncAsync(
            [ServiceBusTrigger("myqueue", Connection = "servicebus")]Message queueMessage,
            Binder binder,
            TraceWriter traceWriter)
        {
            try
            {
                var stream = await binder.BindAsync<Stream>(new Attribute[]
                {
                    new StorageAccountAttribute("account2"),
                    new BlobAttribute($"test/{queueMessage.MessageId}", FileAccess.Write)
                });
/* ... */
            }
            catch(Exception e)
            { traceWriter.Error(e.ToString(), e); }
        }
```

#### Expected behavior
I expect to have a valid stream to the blob

#### Actual behavior

I end up with an exception in BindAsync :

> [10/01/2018 15:12:35] A ScriptHost error has occurred
[10/01/2018 15:12:35] Exception while executing function: SendMailFunction. Microsoft.WindowsAzure.Storage: Service Unavailable.
[10/01/2018 15:12:35] Exception while executing function: SendMailFunction
[10/01/2018 15:12:35] Exception while executing function: SendMailFunction. Microsoft.WindowsAzure.Storage: Service Unavailable.
[10/01/2018 15:12:35] Function completed (Failure, Id=1081697d-7b44-4675-92b7-c7e7281e0d07, Duration=26473ms)
[10/01/2018 15:12:35]
[10/01/2018 15:12:35] Executed 'SendMailFunction' (Failed, Id=1081697d-7b44-4675-92b7-c7e7281e0d07)
[10/01/2018 15:12:35] System.Private.CoreLib: Exception while executing function: SendMailFunction. Microsoft.WindowsAzure.Storage: Service Unavailable.
[10/01/2018 15:12:35]   Function had errors. See Azure WebJobs SDK dashboard for details. Instance ID is '1081697d-7b44-4675-92b7-c7e7281e0d07'
[10/01/2018 15:12:35] System.Private.CoreLib: Exception while executing function: SendMailFunction. Microsoft.WindowsAzure.Storage: Service Unavailable.

The exception come from this line :
[...]

PR #1510, opened 2018-01-12
Fixing ValidateCredentialsAsync when running under netcore or running…

NatMarchand:
… behind a proxy

See issue #1499 for details

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
 using System.Globalization;
 using System.Net;
 using System.Reflection;
@@ -12,7 +11,6 @@ using Microsoft.Azure.WebJobs.Host.Storage;
 using Microsoft.Azure.WebJobs.Host.Storage.Blob;
 using Microsoft.Azure.WebJobs.Host.Storage.Queue;
 using Microsoft.WindowsAzure.Storage;
-using Microsoft.WindowsAzure.Storage.Auth;
 
 namespace Microsoft.Azure.WebJobs.Host.Executors
 {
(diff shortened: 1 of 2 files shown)


NatMarchand:
Is there anything I can do to make the build pass ?
I don't think it's my code which broke the build.

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
```diff
@@ -71,21 +69,30 @@ internal class DefaultStorageCredentialsValidator : IStorageCredentialsValidator
...
-                throw;
+                // Inner exception when using WindowsAzure.Storage for net45
+                case WebException webException:
+                    return webException.Status == WebExceptionStatus.NameResolutionFailure;
+                // Inner exception when using WindowsAzure.Storage for netstandard1.3 as WebException does not exist
+                case System.Net.Http.HttpRequestException httpRequestException:
+                    return (httpRequestException.HResult & 0xFFFF) == errorWinhttpNameNotResolved;
+                // When running behind a proxy
+                case null:
+                    return storageException.RequestInformation?.HttpStatusCode == serviceUnavailable;
```
paulbatum: Can you explain this case? Inferring a blob-only storage account from something as generic as a 503 doesn't seem right but perhaps I'm missing something.

paulbatum:
@NatMarchand there's an issue with our build of public PRs (and therefore with no secrets) for v2. We'll be able to confirm these changes are good by doing a full test run (with secrets).

@brettsam can you review this PR?

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
NatMarchand: In fact, I've tried this code at home and at work. At work, we're behind a corporate http proxy (squid) which "catch" the resolution error and gives a 503. Not sure if it's specific to squid or all the proxy. Maybe we can go without ?

NatMarchand:
Hi!
Were you able to run the tests for this PR ?
Regards,

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultStorageCredentialsValidator.cs
paulbatum: @NatMarchand This doesn't seem resolved. I still have issues with inferring a blob-only account from a generic 503...

NatMarchand: Ok, will remove the case when rebasing then :)

NatMarchand:
Rebased and rewritten to remove the 503 case 😀

Issue #1498, opened 2018-01-09
Allow explicit  binding a parameter to binding data. 


MikeStall:
If a parameter does not have a binding attribute, WebJobs will implicitly bind it to binding data. However, that's not available to non-C# languages. Allow explicitly binding to a binding data value. 

This can then show up in Function.json and enables several things: 

1. For HTTP, this can allow user functions to skip request parsing and bind directly to structured parameters. Swagger generation can also then use this to generate stronger wrappers. 

2. Allow other languages to get access to binding data.  (ie, https://social.msdn.microsoft.com/Forums/en-US/f2cead5e-bcdd-4222-b957-ce836da1b58a/python-azure-function-get-name-of-the-triggered-blob?forum=AzureFunctions ) 

3. This can provide direct access to system parameters in the system data (https://github.com/Azure/azure-webjobs-sdk/blob/86cb51ff269bfcab802c061ae8131174f0848f5c/src/Microsoft.Azure.WebJobs.Host/Bindings/SystemBindingData.cs )

MikeStall:
We already implement this in our tests: https://github.com/Azure/azure-webjobs-sdk/blob/3696dd28df66996b1adcc0697f0d0f88f865cb64/test/Microsoft.Azure.WebJobs.Host.UnitTests/Common/BindingPathAttribute.cs
This means making that public.

mathewc:
Related to https://github.com/Azure/azure-webjobs-sdk-script/issues/1339

PR #1513, opened 2018-01-16
Add BindingExpression attribute which lets you specify a parameter th…

MikeStall:
…at resolves against binding data.

SDK implicitly would do this for C# parameters without attributes, but:
1. the explicit attribute lets you do this for other languages and specify in Function.json, Script.
2. the SDK default still only lets x bind to {x}.
Fixes https://github.com/Azure/azure-webjobs-sdk/issues/1498
This already existed in test, moving to product.


src/Microsoft.Azure.WebJobs.Host/Bindings/BindingExpressionAttribute.cs
```diff
@@ -0,0 +1,31 @@
...
+using Microsoft.Azure.WebJobs.Description;
+using Microsoft.Azure.WebJobs.Host.Config;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// Attribute to allow binding directly to binding data. 
+    /// </summary>
+    [Binding]
+    public class BindingExpressionAttribute : Attribute
```
mathewc: I think this attribute should be defined in the core project (peer to QueueAttribute, etc.)

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingExpressionAttribute.cs
```diff
@@ -0,0 +1,31 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Azure.WebJobs.Description;
+using Microsoft.Azure.WebJobs.Host.Config;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// Attribute to allow binding directly to binding data.
```
mathewc: Can you put a more user friendly/descriptive help text here? Something that would make people understand how to use this if they see it in intellisense :)

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingExpressionAttribute.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// Constructor
+        /// </summary>
+        /// <param name="source">binding expression, such as '{x}' or '%appsetting%'. </param>
+        public BindingExpressionAttribute(string source)
+        {
+            this.Source = source;
+        }
+
+        /// <summary>
+        /// Binding expression for where to pull data.
```
mathewc: Improve xml doc

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingExpressionExtensionConfigProvider.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// Extension provider for <see cref="BindingExpressionAttribute"/>
+    /// </summary>
+    internal class BindingExpressionExtensionConfigProvider : IExtensionConfigProvider
+    {
+        public void Initialize(ExtensionConfigContext context)
+        {
+            var rule = context.AddBindingRule<BindingExpressionAttribute>();
```
mathewc: I don't think these input bindings will be immediately accessible to script functions without changes to the runtime. Have you tried this end to end yet?

fabiocav:
Are we still planning on working on this? Or should this be closed?

mathewc:
Since the PR still has unaddressed comments and we don't have any feature drivers requiring this currently, I vote we don't add this now - it can always be added later in a non-breaking way.

fabiocav:
Sounds good. Closing. We can re-open if needed.

PR #1516, opened 2018-01-17
delay sharedQueue initialization error

watashiSHUN:
previous behavior:
if user does not provide a storage acc
=> logError for sharedQueue, continue execution with in memory implementation
current behavior:
if user does not provide a storage acc 
=> fall back to in memory implementation, but only **logWarning** when developer tried to **register** sharedQueue callback (actual consumer)

Technically I could initialize the sharedQueue when `registerHandler` is first called, so that initialization error would be logged when encountered, but since extensions are free to call `registerHandler`...I don't want to deal with synchronization where there are simultaneous registrations from different extensions (unless we enforce registration to be done in listener constructor, which are being invoked sequentially)

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Dispatch/SharedQueueHandler.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Dispatch/SharedQueueHandler.cs
@@ -22,6 +22,8 @@ namespace Microsoft.Azure.WebJobs.Host.Dispatch
 {
     internal class SharedQueueHandler
     {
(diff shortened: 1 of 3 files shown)


watashiSHUN:
of cause the downsite of this change is when multiple extensions are using sharedqueue, we will get multiple exceptions

brettsam:
Does this need to go into `v2.x` as well as `dev`?

watashiSHUN:
merged https://github.com/Azure/azure-webjobs-sdk/commit/abb6c1cb46df3eaecdc74303d6d5565e18d0d7c8

PR #1527, opened 2018-01-22
adding warning if ServicePointManager.DefaultConnectionLimit is set to 2

### LINKED ISSUE #909: Warn if WebJobs application using the default ServicePointManager.DefaultConnectionLimit
From #908:

>  the problem is that by default DefaultConnectionLimit is 2. That means you are only allowed to have 2 open connections to any URL -- which means you can only have 2 open connections to anything in your blob store, which includes the Lease we use for Singletons. So if you kick off a handful of image downloads and uploads, those 2 connections fill up pretty quickly and everything else is queued behind it. Upping that number shouldn't be affect you negatively. The old guidance was 12*CPUs but now it looks like Web projects set this to Int32.MaxValue: http://stackoverflow.com/questions/5488235/system-net-servicepointmanager-defaultconnectionlimit-24-bug/5649013#5649013.

We should pick a large-enough number here but make it configurable just in case.

brettsam:
Fixes #909.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
+using System.Net;
 using System.Reflection;
 using System.Text;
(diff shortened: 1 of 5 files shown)


brettsam:
@mathewc / @fabiocav -- would one of you mind taking a look at this? It's really WebJobs-only since we handle the DefaultConnectionLimit in functions, but catches a specific scenario that I've seen people hit.

Issue #1523, opened 2018-01-19
CosmosDBTrigger broken in dev


brettsam:
It looks like a change to the binding infrastructure has broken the CosmosDBTrigger (and possibly others). 

Repro: run any sample from here: https://github.com/Azure/azure-webjobs-sdk-extensions/blob/dev/src/ExtensionsSample/Samples/CosmosDBTriggerSamples.cs. Be sure to add `[Disable]` to all but one of the samples (since they all watch the same collection), then add a document to the collection. The trigger will fire, but will throw a NullRef exception at this line: https://github.com/Azure/azure-webjobs-sdk/blob/08709277db41b8dac7cb1e5835005a800af4f304/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/TriggerAdapterBindingProvider.cs#L159

It looks like `_getInvokeString` is null right here b/c we don't support an `IReadOnlyList<Document> -> DirectInvokeString` converter, which is created here: https://github.com/Azure/azure-webjobs-sdk/blob/08709277db41b8dac7cb1e5835005a800af4f304/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/TriggerAdapterBindingProvider.cs#L119

PR #1528, opened 2018-01-22
Fix TriggerAdapter bug

MikeStall:
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/1523

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Config/DirectInvokeString.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Config/DirectInvokeString.cs
@@ -14,5 +14,7 @@ namespace Microsoft.Azure.WebJobs.Host.Config
         {
             this.Value = value;
         }
+
+        public static DirectInvokeString None = new DirectInvokeString("???");
     }
 }
--- a/test/Microsoft.Azure.WebJobs.Host.UnitTests/Common/TriggerAdapterTest.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.UnitTests/Common/TriggerAdapterTest.cs
@@ -1,0 +1,136 @@
+﻿using Microsoft.Azure.WebJobs.Host.Bindings;
+using Microsoft.Azure.WebJobs.Host.Config;
+using Microsoft.Azure.WebJobs.Host.Listeners;
+using Microsoft.Azure.WebJobs.Host.Protocols;
+using Microsoft.Azure.WebJobs.Host.TestCommon;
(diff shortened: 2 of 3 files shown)


src/Microsoft.Azure.WebJobs.Host/Config/DirectInvokeString.cs
```diff
@@ -14,5 +14,7 @@ public DirectInvokeString(string value)
         {
             this.Value = value;
         }
+
+        public static DirectInvokeString None = new DirectInvokeString("???");
```
brettsam: Would it make more sense for this to be `string.Empty`? Although rare, "???" is still a viable option as an invoke string so it may be a little confusing.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Common/TriggerAdapterTest.cs
```diff
@@ -0,0 +1,93 @@
...
+            _value = data;
+        }
+
+        public static void FuncAsInt([FakeQueueTrigger] int data)
+        {
+            _value = data;
+        }
+
+        [Fact]
+        public async Task Test2()
```
brettsam: Nit: can you add a more descriptive name for this test?

test/Microsoft.Azure.WebJobs.Host.UnitTests/Common/TriggerAdapterTest.cs
```diff
@@ -0,0 +1,93 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Host.UnitTests.Common
+{
+    // Test for Trigger adapter class. 
+    public class TriggerAdapterTest
```
brettsam: Can you add a second test to validate the "no string converter" scenario?

brettsam:
@MikeStall -- are you able to get this in soon? We're blocked on a couple of CosmosDB changes until this is merged.

src/Microsoft.Azure.WebJobs.Host/Config/DirectInvokeString.cs
MikeStall: The invoke string is displayed in the dashboard UI, so seeing ??? would be more self-explanatory. This is a tactical concession - handling string.empty may require dashboard changes.

PR #1529, opened 2018-01-23
Add error details on Azure Storage batched operation failures (v2)

### LINKED ISSUE #1300: Storage collector error details
We need to ensure that failures from batched operations are clearly reported to consumers.

If an item in a batched operation fails, the message today does not contain information that sufficiently identifies the issue:

Example:
```
 ---> System.InvalidOperationException : Error while handling parameter outputTable after function returned: 
    ---> Microsoft.WindowsAzure.Storage.StorageException : Element 0 in the batch returned an unexpected response code.
```

From https://stackoverflow.com/questions/45716862/azure-function-fails-with-storageexception

mhoeger:
Adding extension method for getting and showing extra data on StorageException's from failed batch operations. This is a workaround to patch the unspecific error message 'Element {i} in the batch returned an unexpected response code'.

Resolves #1300

(v1 version of the PR: https://github.com/Azure/azure-webjobs-sdk/pull/1480)

Tested manually locally, tested v1 change in prod too
image

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Tables/TableEntityWriter.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Tables/TableEntityWriter.cs
@@ -172,7 +172,7 @@ namespace Microsoft.Azure.WebJobs.Host.Tables
                 {
                     if (!e.IsNotFoundTableNotFound())
                     {
-                        throw new StorageException(e.GetDetailedErrorMessage());
+                        throw new StorageException(e.GetDetailedErrorMessage(), e);
                     }
 
                     exception = e;
@@ -180,24 +180,19 @@ namespace Microsoft.Azure.WebJobs.Host.Tables
 
                 if (exception != null)
                 {
-                    await CreateTableAndExecuteBatch(batch, cancellationToken);
+                    // Make sure the table exists
+                    await _table.CreateIfNotExistsAsync(cancellationToken);
+
+                    // Commit the batch
+                    try
+                    {
+                        await _table.ExecuteBatchAsync(batch, cancellationToken);
+                    }
+                    catch (StorageException e)
+                    {
+                        throw new StorageException(e.GetDetailedErrorMessage(), e);
+                    }
                 }
-            }
-        }
-
-        private async Task CreateTableAndExecuteBatch(IStorageTableBatchOperation batch, CancellationToken cancellationToken)
-        {
-            // Make sure the table exists
-            await _table.CreateIfNotExistsAsync(cancellationToken);
-
-            // Commit the batch
-            try
-            {
-                await _table.ExecuteBatchAsync(batch, cancellationToken);
-            }
-            catch (StorageException e)
-            {
-                throw new StorageException(e.GetDetailedErrorMessage());
             }
         }
 
--- a/src/Microsoft.Azure.WebJobs.Host/Tables/TableEntityWriter.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Tables/TableEntityWriter.cs
@@ -172,7 +172,7 @@ namespace Microsoft.Azure.WebJobs.Host.Tables
                 {
                     if (!e.IsNotFoundTableNotFound())
                     {
-                        throw;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.WebJobs.Host/Tables/TableEntityWriter.cs
```diff
@@ -172,23 +172,35 @@ private static ITableEntity Copy(ITableEntity item)
...
-                    await _table.CreateIfNotExistsAsync(cancellationToken);
-
-                    // Commit the batch
-                    await _table.ExecuteBatchAsync(batch, cancellationToken);
+                    await CreateTableAndExecuteBatch(batch, cancellationToken);
                 }
             }
         }
 
+        private async Task CreateTableAndExecuteBatch(IStorageTableBatchOperation batch, CancellationToken cancellationToken)
```
brettsam: Since you're not doing anything special with the `CreateIfNotExistsAsync` call here (and it looks like it's only called from one place) -- can you remove this method and instead wrap the call to `ExecuteBatchAsync` in-line?

src/Microsoft.Azure.WebJobs.Storage/StorageExceptionExtensions.cs
```diff
@@ -815,5 +815,31 @@ public static string GetErrorCode(this StorageException exception)
...
+            if (exception == null)
+            {
+                throw new ArgumentNullException("exception");
+            }
+
+            string message = exception.Message;
+
+            if (exception.RequestInformation != null)
+            {
+                message += String.Format(" (HTTP status code {0}: {1}. {2})",
```
brettsam: Rather than using `string.Format`, we've started using string interpolation whenever we add new strings (if we remember). It makes it a little easier to read. So can you change this to something like `$" (HTTP status code {exception.RequestInformation.HttpStatusCode}: {exception.RequestInformation.ExtendedErrorInformation?.ErrorCode}. {exception.RequestInformation.ExtendedErrorInformation?.ErrorMessage})"`?

src/Microsoft.Azure.WebJobs.Host/Tables/TableEntityWriter.cs
```diff
@@ -172,23 +172,35 @@ private static ITableEntity Copy(ITableEntity item)
...
+            await _table.CreateIfNotExistsAsync(cancellationToken);
+
+            // Commit the batch
+            try
+            {
+                await _table.ExecuteBatchAsync(batch, cancellationToken);
+            }
+            catch (StorageException e)
+            {
+                throw new StorageException(e.GetDetailedErrorMessage());
```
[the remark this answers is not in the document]
mhoeger: @brettsam I wasn't sure if it would be more confusing to add the original exception as the inner exception, since I'm just adding more detail to the original (unhelpful) StorageException message.

`---> System.InvalidOperationException : Error while handling parameter outputTable after function returned: 
    ---> Microsoft.WindowsAzure.Storage.StorageException : Element 0 in the batch returned an unexpected response code. (HTTP status code 500: Table not found. Detailed message here)`

becomes

`---> System.InvalidOperationException : Error while handling parameter outputTable after function returned: 
    ---> Microsoft.WindowsAzure.Storage.StorageException : Element 0 in the batch returned an unexpected response code. (HTTP status code 500: Table not found. Detailed message here) ---> Microsoft.WindowsAzure.Storage.StorageException : Element 0 in the batch returned an unexpected response code. `

Is that ok?

src/Microsoft.Azure.WebJobs.Storage/StorageExceptionExtensions.cs
mhoeger: Thanks!!

fabiocav:
LGTM, please squash and feel free to merge. 

Also, see my question above about having this work ported to v2.x. If needed, you can open an issue to track that work.

mhoeger:
Should be in v2.x already! Thanks!!
image

fabiocav:
Great! Thank you!

PR #1537, opened 2018-01-25
Applying Obsolete to non-baked APIs

mathewc:
Applying Obsolete to some more new APIs that we've added since 2.0, that we likely want more bake-time for. This is in preparation for the forthcoming 2.1 release.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Config/FluentBindingRule.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Config/FluentBindingRule.cs
@@ -6,8 +6,6 @@ using System.Collections.Generic;
 using System.IO;
 using System.Reflection;
 using System.Text;
-using System.Threading.Tasks;
-using Microsoft.Azure.WebJobs.Host;
 using Microsoft.Azure.WebJobs.Host.Bindings;
 using Microsoft.Azure.WebJobs.Host.Protocols;
 using Microsoft.Azure.WebJobs.Host.Triggers;
@@ -19,6 +17,7 @@ namespace Microsoft.Azure.WebJobs.Host.Config
     /// Helpers for adding binding rules to a given attribute.
     /// </summary>
     /// <typeparam name="TAttribute"></typeparam>
+    [Obsolete("Not ready for public consumption.")]
     public class FluentBindingRule<TAttribute> : FluentConverterRules<TAttribute, FluentBindingRule<TAttribute>>
         where TAttribute : Attribute
     {
--- a/src/Microsoft.Azure.WebJobs.Host/HostQueueNames.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/HostQueueNames.cs
@@ -25,18 +25,22 @@ namespace Microsoft.Azure.WebJobs.Host
             return HostBlobTriggerQueuePrefix + hostId;
         }
 
-        /// <summary>Gets the shared host queue name.</summary>
+        /// <summary>
+        /// Gets the shared host queue name.
+        /// </summary>
         /// <param name="hostId">The host ID.</param>
         /// <returns>The shared host queue name.</returns>
-        public static string GetHostSharedQueueName(string hostId)
+        internal static string GetHostSharedQueueName(string hostId)
         {
             return HostSharedQueuePrefix + hostId;
         }
 
-        /// <summary>Gets the shared host poison queue name.</summary>
+        /// <summary>
+        /// Gets the shared host poison queue name.
+        /// </summary>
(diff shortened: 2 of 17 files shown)


src/Microsoft.Azure.WebJobs.Host/HostQueueNames.cs
```diff
@@ -25,18 +25,22 @@ public static string GetHostBlobTriggerQueueName(string hostId)
...
         }
 
-        /// <summary>Gets the shared host queue name.</summary>
+        /// <summary>
+        /// Gets the shared host queue name.
+        /// </summary>
         /// <param name="hostId">The host ID.</param>
         /// <returns>The shared host queue name.</returns>
-        public static string GetHostSharedQueueName(string hostId)
+        internal static string GetHostSharedQueueName(string hostId)
```
mathewc: I think we only have internal usages of these currently?

src/Microsoft.Azure.WebJobs.Host/Listeners/ListenerFactoryContext.cs
```diff
@@ -18,7 +18,35 @@ public class ListenerFactoryContext
...
         private IDispatchQueueHandler _dispatchQueue;
 
+        /// <summary>
+        /// Creates a new instance
+        /// </summary>
+        /// <param name="descriptor">The <see cref="FunctionDescriptor"/> to create a listener for.</param>
+        /// <param name="executor">The <see cref="ITriggeredFunctionExecutor"/> that should be used to invoke the
+        /// target job function when the trigger fires.</param>
+        /// <param name="cancellationToken">The <see cref="CancellationToken"/> to use.</param>
+        public ListenerFactoryContext(FunctionDescriptor descriptor, ITriggeredFunctionExecutor executor, CancellationToken cancellationToken)
```
mathewc: Adding this public ctor back - it was removed when this work was done, which was a breaking change. The shared queue capability is preserved via an internal constructor.

src/Microsoft.Azure.WebJobs.Host/Listeners/ListenerFactoryContext.cs
```diff
@@ -27,8 +55,8 @@ internal ListenerFactoryContext(FunctionDescriptor descriptor, ITriggeredFunctio
 
             Descriptor = descriptor;
```
watashiSHUN: only set sharedqueue

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
```diff
@@ -400,6 +402,7 @@ public void AddService<TService>(TService serviceInstance)
         /// Add an extension to register new binding attributes and converters.
         /// </summary>
         /// <param name="extension"></param>
+        [Obsolete("Not ready for public consumption.")]
         public void AddExtension(IExtensionConfigProvider extension)
```
MikeStall: Probably keep this one. It's stable. Script will need it

src/Microsoft.Azure.WebJobs.Host/Config/FluentBindingRule.cs
```diff
@@ -19,6 +17,7 @@ namespace Microsoft.Azure.WebJobs.Host.Config
     /// Helpers for adding binding rules to a given attribute.
     /// </summary>
     /// <typeparam name="TAttribute"></typeparam>
+    [Obsolete("Not ready for public consumption.")]
     public class FluentBindingRule<TAttribute> : FluentConverterRules<TAttribute, FluentBindingRule<TAttribute>>
```
MikeStall: These Fluent ones are likely stable.

MikeStall (approved):

src/Microsoft.Azure.WebJobs.Host/Listeners/ListenerFactoryContext.cs
mathewc: done

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
mathewc: ok

src/Microsoft.Azure.WebJobs.Host/Config/FluentBindingRule.cs
mathewc: It's base class FluentConverterRules is also public and is Obsolete. Given that, I think it makes sense to be consistent here. We can always remove these in the next release once we're finished with the APIs.

PR #1546, opened 2018-02-02
Fix race in logging.

### LINKED ISSUE #1545: Race condition can lead to 'Never finished' functions in the dashboard
There's likely a race condition in this chunk of code https://github.com/Azure/azure-webjobs-sdk/blob/v2.x/src/Microsoft.Azure.WebJobs.Logging/Internal/LogWriter.cs#L279-L297

While we’re writing out the batches, if another completed function comes in, it will get added to the _completedFunctions list, which will mean it can get cleared from the list of activeFuncs without being written.

MikeStall:
Fixes #1545

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging/Internal/LogWriter.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging/Internal/LogWriter.cs
@@ -249,51 +249,46 @@ namespace Microsoft.Azure.WebJobs.Logging
             }
         }
 
-
-        private FunctionInstanceLogItem[] Update()
-        {
-            FunctionInstanceLogItem[] items;
-            lock (_lock)
-            {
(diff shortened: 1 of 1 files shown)


src/Microsoft.Azure.WebJobs.Logging/Internal/LogWriter.cs
```diff
@@ -249,51 +249,46 @@ public Task AddAsync(FunctionInstanceLogItem item, CancellationToken cancellatio
             }
         }
 
-
-        private FunctionInstanceLogItem[] Update()
+        // Could flush on a timer. 
+        private async Task FlushIntancesAsync()
```
brettsam: nit: spelling of this method (missing an 's')

brettsam:
Can you port this to `v2.x` as well?

Issue #1500, opened 2018-01-11
 Build errors need to stop CI


christopheranderson:
CI builds are currently failing at `dotnet build -v q`, but we swallow the error. Need to fail when this fails. It's hiding issues.

fabiocav:
Resolved by #1547

PR #1547, opened 2018-02-02
Skipping packaging and exiting with proper code on build failure

fabiocav:
Quick fix for #1500


fabiocav:
Build is actually succeeding now. And failures will be reported.

PR #1551, opened 2018-02-04
EventHubTrigger not resolved in .net standard

DamiaoCastro:
Related to:
https://github.com/Azure/Azure-Functions/issues/652
https://github.com/Azure/Azure-Functions/issues/620

The problem resides between the difference of what is generated inside the "function.json" and what is expected in the EventHubTriggerAttribute constructor.

"path" vs "eventhubname".

If this is the best place to do the change, I don't know


fabiocav:
@DamiaoCastro thank you for the contribution. The right fix for V2 is to remove the special casing we currently have in place for that property. I've opened an issue [here ](https://github.com/Azure/azure-functions-vs-build-sdk/issues/180) to track.

PR #1553, opened 2018-02-06
Removing analyzers from WebJobs solution and introducing a solution with them

fabiocav:
Removing analyzers from the default solution. 

The analyzer projects were in the old project format, introducing some build issues and additional dependencies. #1554 tracks some work that needs to be done to fix the current analyzer project structure.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ obj
 csx
 .vs
 .DS_Store
+.vscode
 
 *.user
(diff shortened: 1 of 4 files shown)


christopheranderson:
LGTM - might want to update build.sh so travis points to the right sln

fabiocav:
Thanks! Yeah.... missed that. Updated now

Issue #1324, opened 2017-08-31
Allow Triggers to pass Invocation Ids to Executor


brettsam:
Right now there's no way to correlate the work that a trigger does with the actual function invocation. We need a way to allow the trigger to either access or provide the GUID identifier so that correlation can happen with logging.

paulbatum:
It would be really really nice to have this for functions v1 as well as for v2

oising:
Ping. 

I followed up thinking #1563 was a solution, but it was closed in favour of "other work" -- I can't track this any further as the scent has gone cold... how does one use an external correlation ID for a function (v2) ?

PR #1563, opened 2018-02-09
Adds ability to pass Id along on TriggeredFunctionData

christopheranderson:
fixes https://github.com/Azure/azure-webjobs-sdk/issues/1324

@cgillum / @brettsam, you both had some use cases in mind for this. I didn't end up consuming this surface area, but totally possible to pass an id along now when you call TryExecuteAsync via the TriggeredFunctionData object. LMK if this works for your scenarios.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionData.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionData.cs
@@ -12,6 +12,10 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
     public class TriggeredFunctionData
     {
         /// <summary>
+        /// The ID for the triggered function invocation
+        /// </summary>
+        public Guid? Id { get; set; }
+        /// <summary>
         /// The parent ID for the triggered function invocation.
         /// </summary>
         public Guid? ParentId { get; set; }
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionExecutor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionExecutor.cs
@@ -35,7 +35,8 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
             var context = new FunctionInstanceFactoryContext<TTriggerValue>()
             {
                 TriggerValue = (TTriggerValue)input.TriggerValue,
-                ParentId = input.ParentId
+                ParentId = input.ParentId,
+                Id = input.Id.GetValueOrDefault(Guid.NewGuid())
             };
 
             if (input.InvokeHandler != null)
--- a/test/Microsoft.Azure.WebJobs.Host.UnitTests/Executors/TriggeredFunctionExecutorTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.UnitTests/Executors/TriggeredFunctionExecutorTests.cs
@@ -14,13 +14,17 @@ namespace Microsoft.Azure.WebJobs.Host.UnitTests.Executors
 {
     public class TriggeredFunctionExecutorTests
     {
-        [Fact]
-        public async Task TryExecuteAsync_WithInvokeHandler_InvokesHandler()
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public async Task TryExecuteAsync_WithInvokeHandler_InvokesHandler(bool passesInvocationId)
         {
+            Guid? actualInvocationId = null;
             var mockExecutor = new Mock<IFunctionExecutor>();
             mockExecutor.Setup(m => m.TryExecuteAsync(It.IsAny<IFunctionInstance>(), It.IsAny<CancellationToken>())).
                 Returns<IFunctionInstance, CancellationToken>((x, y) =>
                 {
+                    actualInvocationId = x.Id;
                     x.Invoker.InvokeAsync(null, null).Wait();
                     return Task.FromResult<IDelayedException>(null);
                 });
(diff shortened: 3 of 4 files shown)


src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionExecutor.cs
```diff
@@ -35,7 +35,8 @@ public FunctionDescriptor Function
             var context = new FunctionInstanceFactoryContext<TTriggerValue>()
             {
                 TriggerValue = (TTriggerValue)input.TriggerValue,
-                ParentId = input.ParentId
+                ParentId = input.ParentId,
+                Id = input.Id.GetValueOrDefault(Guid.NewGuid())
```
fabiocav: Being on the invocation path, please use `input.Id ?? Guid.NewGuid()` for efficiency.

brettsam:
@MikeStall had some valid concerns about letting triggers generate their own Ids. Let me loop you guys in on that conversation and maybe we can come up with a solution.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Executors/TriggeredFunctionExecutorTests.cs
```diff
@@ -53,11 +57,27 @@ public class TriggeredFunctionExecutorTests
...
+            }
+
             var result = await triggerExecutor.TryExecuteAsync(triggerData, CancellationToken.None);
 
             Assert.True(result.Succeeded);
             Assert.True(customInvokerInvoked);
             Assert.True(innerInvokerInvoked);
+
+            if (passesInvocationId)
+            {
```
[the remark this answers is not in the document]
christopheranderson: `passesInvocationId` is flag, passed via `InlineData`. I'm not sure why I'd assert it?

src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionExecutor.cs
christopheranderson: Oh cool, I didn't know that ?? was faster.

mathewc:
Because a trigger binding is the originator of a function invocation (it is responsible for calling ITriggeredFunctionExecutor.TryExecuteAsync), if the listener needs to correlate logs with the invocation, it needs to know the ID before TryExecuteAsync is called. This is a natural way of allowing for that.

src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionData.cs
```diff
@@ -11,6 +11,10 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
     /// </summary>
     public class TriggeredFunctionData
     {
+        /// <summary>
+        /// The ID for the triggered function invocation
+        /// </summary>
+        public Guid? Id { get; set; }
```
mathewc: We call this FunctionInstanceId everywhere else

test/Microsoft.Azure.WebJobs.Host.UnitTests/Executors/TriggeredFunctionExecutorTests.cs
```diff
@@ -53,11 +57,27 @@ public class TriggeredFunctionExecutorTests
...
+                expectedInvocationId = Guid.NewGuid();
+                triggerData.Id = expectedInvocationId;
+            }
+
             var result = await triggerExecutor.TryExecuteAsync(triggerData, CancellationToken.None);
 
             Assert.True(result.Succeeded);
             Assert.True(customInvokerInvoked);
             Assert.True(innerInvokerInvoked);
```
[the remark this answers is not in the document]
mathewc: I think you should add a new targeted test for your scenario, rather than complicate this test. Your test could be simpler than this existing test I think.

brettsam:
@mathewc -- @MikeStall had some concerns with this approach when he started trying to tackle this a few weeks ago. I'm doing my due diligence and making sure someone chats with him before we merge this.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Executors/TriggeredFunctionExecutorTests.cs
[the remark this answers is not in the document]
fabiocav: The assert I'm referring to is over the `actualInvocationId.HasValue`, not `passesInvocation`. That would cleanup some unnecessary code .

src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionExecutor.cs
fabiocav: It's not just ??. With the current implementation, you're always creating a NewGuid, instead of allowing this to short circuit in the scenarios where the id is already set, which the recommended change would do.

paulbatum:
I wonder, are the concerns related to billing and/or the dashboard? I think there are a few downstream systems that assume each function invocation has a unique ID. For example, do we track execution counts by counting unique invocation IDs? If so, a custom trigger that uses the same invocation ID every time might be able to circumvent that.. Similarly will non unique invocation IDs blow up the dashboard? Sure the dashboard is less important now for main functions scenarios (because we prefer App Insights) but its not going away completely (e.g. Azure Stack still needs it).

paulbatum:
On the topic of flowing the request ID through... I suspect it should not flow through as the invocation ID for the same reasons discussed above. Yes we want it to be easy to correlate from Front End tables to functions logs tables, but we could do that by having an activity ID that is NOT the invocation ID.

fabiocav:
@christopheranderson this is no longer applicable, right? Will you be closing this in favor of other work being done?

PR #1572, opened 2018-02-13
Don't query non-existent queues (#770)

mathewc:
Fix for https://github.com/Azure/azure-webjobs-sdk/issues/770

We currently generate a LOT of exceptions polling non-existent queues. While they are handled internally, it makes local debugging hard if you have "break on all exceptions" set. More importantly (as in the issue) these show up in App Insights. 

To see the issue for yourself, simply set the SampleHost project as the startup project and run with "break on all exceptions" set. You'll see lots :) With these changes, you'll see none.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Queues/Listeners/QueueListener.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Queues/Listeners/QueueListener.cs
@@ -32,6 +32,7 @@ namespace Microsoft.Azure.WebJobs.Host.Queues.Listeners
         private readonly QueueProcessor _queueProcessor;
         private readonly TimeSpan _visibilityTimeout;
 
+        private bool? _queueExists;
         private bool _foundMessageSinceLastDelay;
         private bool _disposed;
         private TaskCompletionSource<object> _stopWaitingTaskSource;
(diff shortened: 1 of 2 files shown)


mathewc:
My expectation is that AppInsights is smart enough to treat actual errors/exceptions differently than non-failed operations (e.g. `CloudQueue.ExistsAsync()`). I can see why us generating a bunch of exceptions would result in error logs. If checks like `CloudQueue.ExistsAsync()` also generate noise/errors in AppInsights, that I think that is a problem THEY need to fix - that doesn't seem right. Are you saying that with these changes AI still shows errors?

src/Microsoft.Azure.WebJobs.Host/Queues/Listeners/QueueListener.cs
```diff
@@ -145,17 +146,36 @@ public void Dispose()
...
-            IEnumerable<IStorageQueueMessage> batch;
+            IEnumerable<IStorageQueueMessage> batch  = null;
             try
             {
-                batch = await _queue.GetMessagesAsync(_queueProcessor.BatchSize,
-                    _visibilityTimeout,
-                    options: null,
-                    operationContext: null,
-                    cancellationToken: cancellationToken);
+                if (!_queueExists.HasValue || !_queueExists.Value)
```
fabiocav: nit: consider using `if (!_queueExists.GetValueOrDefault(false))`

src/Microsoft.Azure.WebJobs.Host/Queues/Listeners/QueueListener.cs
```diff
@@ -145,17 +146,36 @@ public void Dispose()
...
+                        operationContext: null,
+                        cancellationToken: cancellationToken);
+                }
             }
             catch (StorageException exception)
             {
+                // if we get ANY errors querying the queue reset our existence check
+                // doing this on all errors rather than trying to special case not
+                // found, because correctness is the most important thing here
+                _queueExists = null;
```
fabiocav: A bit concerned with possible races with this...

brettsam:
With automatic Dependency Tracking enabled, App Insights tracks the underlying HTTP requests using .NET Framework EventListeners. They don't know what API you used, they just see that an HTTP request was made somewhere in the .NET Framework and the result was a 404. They can't tell if 404 is expected or not, so they call it an error all the time. The link I had above is one way to prevent that. You look at their generated dependencies and then decide in the context of your application whether it's a failure or not (maybe you should never see a 404, so you always want to be alerted).

I could add a utility to help with this (or maybe one exists already; I'll have to investigate). It has just seemed odd to do the work here since we don't use the dependency tracking ourselves and it's not directly our issue.

I ran some tests this morning with ExistsAsync to confirm:
image

src/Microsoft.Azure.WebJobs.Host/Queues/Listeners/QueueListener.cs
mathewc: Can you elaborate? This entire method is called in a single threaded manner on a background timer.

src/Microsoft.Azure.WebJobs.Host/Queues/Listeners/QueueListener.cs
mathewc: I'll make that change

Issue #922, opened 2016-11-23
Investigate moving away of use of StorageException Extended Error Info in Exception helpers


mathewc:
From very early on, the SDK has used a set of [StorageException extension methods](https://github.com/Azure/azure-webjobs-sdk/blob/a2215b59f9138cfa98f2f9c6b95434bd73500a42/src/Microsoft.Azure.WebJobs.Storage/StorageExceptionExtensions.cs) to inspect and classify Azure Storage exceptions and to make decisions based on that info.

However, we've seen some instances that suggest that this information might not be reliable. We should investigate our use of this info for any important decisions and move away from it.

For example, we've received the following stack trace from a customer:

```text
OnExceptionReceived got unhandled exception: Microsoft.WindowsAzure.Storage.StorageException: The remote server returned an error: (404) Not Found. ---> 
System.Net.WebException: The remote server returned an error: (404) Not Found.
at Microsoft.WindowsAzure.Storage.Shared.Protocol.HttpResponseParsers.ProcessExpectedStatusCodeNoException[T](HttpStatusCode expectedStatusCode, HttpStatusCode actualStatusCode, T retVal, StorageCommandBase`1 cmd, Exception ex) in c:\Program Files (x86)\Jenkins\workspace\release_dotnet_master\Lib\Common\Shared\Protocol\HttpResponseParsers.Common.cs:line 50
at Microsoft.WindowsAzure.Storage.Blob.CloudBlob.<DeleteBlobImpl>b__29(RESTCommand`1 cmd, HttpWebResponse resp, Exception ex, OperationContext ctx) in c:\Program Files (x86)\Jenkins\workspace\release_dotnet_master\Lib\ClassLibraryCommon\Blob\CloudBlob.cs:line 3118
at Microsoft.WindowsAzure.Storage.Core.Executor.Executor.EndGetResponse[T](IAsyncResult getResponseResult) in c:\Program Files (x86)\Jenkins\workspace\release_dotnet_master\Lib\ClassLibraryCommon\Core\Executor\Executor.cs:line 299 --- End of inner exception stack trace --- Server stack trace:
at Microsoft.WindowsAzure.Storage.Core.Executor.Executor.EndExecuteAsync[T](IAsyncResult result) in c:\Program Files (x86)\Jenkins\workspace\release_dotnet_master\Lib\ClassLibraryCommon\Core\Executor\Executor.cs:line 50
[...]

PR #1575, opened 2018-02-14
changing AzureStorageDeploymentValidator to work with Storage SDK 9.0

### LINKED ISSUE #1574: Storage v9 runtime failures
From @baywet [here](https://github.com/Azure/azure-webjobs-sdk/issues/1505#issuecomment-364135691):

> Running into the same issue with .net 4.6.2, webjob 2.0.0 (or 2.1.0, doesn't matter) and storage v9. It looks like the namespace Microsoft.WindowsAzure.Storage.Table.DataServices has been removed in v9

This ends up being a run-time (rather than a build-time error), which stinks.

We need to see if this is something we can work around. If not, we may need to throw a nuget maximum version on our nuspec.

brettsam:
Fixes #1574.

Storage SDK version 9.0.0 removed `TableServiceContext`, which means this check fails at runtime. I changed the check to be reflection- and version-based to make sure that we continue to enforce the check on versions where it is required (those under 8,0.0). 

I ran all the tests on both version 7.2.1 and 9.0.0 and they all passed. I manually verified that removing one of the assemblies (Microsoft.Data.Services.Client, Microsoft.Data.OData, Microsoft.Data.Edm) continued to fail this check in 7.2.1.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/AzureStorageDeploymentValidator.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/AzureStorageDeploymentValidator.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Linq;
 using Microsoft.WindowsAzure.Storage.Table;
 using Microsoft.WindowsAzure.Storage.Table.DataServices;
 
@@ -22,7 +23,11 @@ namespace Microsoft.Azure.WebJobs.Host
         {
             try
             {
-                VerifyTableServiceAssemblyLoad();
+                // Later versions of the Storage SDK removed the dependencies on these assemblies, so this check is skipped in those cases.
+                if (typeof(CloudTableClient).Assembly.GetReferencedAssemblies().Any(a => a.Name == "Microsoft.Data.Services.Client"))
+                {
+                    VerifyTableServiceAssemblyLoad();
+                }
             }
             catch (Exception ex)
             {
--- a/test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
@@ -136,8 +136,8 @@ namespace Microsoft.Azure.WebJobs.Host.EndToEndTests
                     "User TextWriter log (TestParam)"
                 }.OrderBy(p => p).ToArray();
 
-                bool hasError = consoleOutputLines.Any(p => p.Contains("Function had errors"));
-                Assert.False(hasError);
+                var errors = consoleOutputLines.Where(p => p.Contains("Function had errors"));
+                Assert.False(errors.Any(), string.Join(Environment.NewLine, errors));
 
                 // Validate console output
                 for (int i = 0; i < expectedOutputLines.Length; i++)


brettsam:
@fabiocav / @mathewc: thoughts on the approach here? This type was Obsoleted previously and has now been removed in v9 of the Storage SDK.

fabiocav:
Curious, why are we using that type specifically? What assemblies do we care to load at this point? Can we use a type that exists in both versions and move away from the reflection based code?

brettsam:
It came from here: https://github.com/Azure/azure-webjobs-sdk/issues/922#issuecomment-263374023.

I don't fully understand all of the scenarios we're trying to catch with this code, so I didn't want to alter the end result too much.

brettsam:
But we run on 7.2.1 so it'd still be possible to hit this... but you have more context on the severity of the original issue. If we can get rid of the check, then I'm all for that.

mathewc:
Sorry - I missed this was for v2.x

fabiocav:
So based on the issue, it does look like we could use another type to force the assembly load to perform that check, which would allow us to do this without the reflection code.

brettsam:
v9 removed dependencies on those nugets completely, so we can't rely on directly loading them ourselves.

The issue is going to be finding a type that loads `Microsoft.Data.Services.Client` in v7 and still exists in v9. Since those types were all obsoleted, I'm guessing they're mostly gone. 

I've found a type (`TableStorageModel`) that loads two of the assemblies (Edm/OData) in v7, but it looks like it's been removed in v9.

I'm going to try the original approach here: https://github.com/Azure/azure-webjobs-sdk/issues/922#issuecomment-263370807 and see if that still works for both versions. It looks pretty gnarly, though.

davidmatson:
I think this scenario is when one of the following is true:
1. One of the table-only DLLs is missing.
2. One of the table-only DLLs is present but of a different version than the Web Jobs SDK was compiled against, and a required binding redirect is missing.

Going down any code path that would trigger loading one of these DLLs should be sufficient. I can dig more if needed (both into the exact list of DLLs and what types cause loading them).

brettsam:
Thanks @davidmatson. The issue I'm running into is finding an type that loads the three DLLs and also exists in both v7 and v9 of the Storage client. It looks like a lot of these were removed along with the dependencies on those three assemblies.

This check (https://github.com/Azure/azure-webjobs-sdk/issues/922#issuecomment-263370807) seems to only valdate Microsoft.Data.Services.Client; if Edm or OData are missing, the check succeeds.

davidmatson:
Yep, it looks like they're basically self-contained now (which is quite nice).

Maybe the best option is to make the existing check conditional on whether Microsoft.WidnowsAzure.Storage.dll has an assembly reference to Microsoft.Data.Services.Client.dll? (If we're using a self-contained version, we know we're not broken?) I'm assuming there's a way via reflection to tell fairly easily?

Checking whether error deserialization is functionally broken is also a legitimate option, but that code is much uglier due to the dependencies involved. Thoughts?

davidmatson:
Yeah, this check seems to work (says 8.7.0 is not self-contained but 9.0.0 is):
```c#
bool isSelfContained = !typeof(CloudStorageAccount).Assembly.GetReferencedAssemblies().Any(a => a.Name == "Microsoft.Data.Services.Client");
```

brettsam:
Yes, I like that better. Thanks!

@mathewc / @fabiocav -- does this latest change sit better with you guys? 

I've done some initial testing but want to check that it works with an actual nuget install.

PR #1576, opened 2018-02-14
adding utility to make binding resolution easier for custom IResolutionPolicy authors

brettsam:


src/Microsoft.Azure.WebJobs.Host/Bindings/Path/BindingTemplate.cs
```diff
@@ -54,22 +53,36 @@ internal IEnumerable<BindingTemplateToken> Tokens
...
         /// </summary>
         public bool HasParameters => ParameterNames.Any();
 
+        /// <summary>
+        /// A utility for resolving individual tokens against binding data.
+        /// </summary>
+        /// <param name="token">A token string to resolve from the binding data. For example, "a.b.c".</param>
+        /// <param name="bindingData">The binding data to resolve the parameter against.</param>
+        /// <returns>The resolved parameter.</returns>
+        public static string ResolveToken(string token, IReadOnlyDictionary<string, object> bindingData)
```
mathewc: Is this equivalent to the below? Wondering who is going to use this new method and if the below isn't sufficient.

```csharp
var template = BindingTemplate.FromString(token);
template.Bind(bindingData);
```

mathewc:
@brettsam Can this be closed?

brettsam:
Yeah, the issue is captured in #1726. I put this together first before we realized it required much larger changes.

PR #1578, opened 2018-02-15
Add error log from eventhublistener

### LINKED ISSUE #1357: Log Event Hubs related exceptions captured by ProcessErrorAsync
This code appears to be silently throwing away errors that come from event processor host.
https://github.com/Azure/azure-webjobs-sdk/blob/901764c1aa610685ce850748103c8554aef3d250/src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubListener.cs#L117-L121

We need to wire this code path up to both system logs and application logs.

pragnagopa:
Fixes #1357 in V2

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
@@ -104,14 +104,6 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                 config.AddExtension(new BlobTriggerExtensionConfig());
             }
 
-            ILoggerFactory loggerFactory = services.GetService<ILoggerFactory>();
-
-            if (loggerFactory == null)
-            {
-                loggerFactory = new LoggerFactory();
-                services.AddService<ILoggerFactory>(loggerFactory);
-            }
-
             ExtensionConfigContext context = new ExtensionConfigContext
             {
                 Config = config,
@@ -123,7 +115,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
 
             if (singletonManager == null)
             {
-                var logger = loggerFactory?.CreateLogger(LogCategories.Singleton);
+                var logger = config.LoggerFactory?.CreateLogger(LogCategories.Singleton);
 
                 IDistributedLockManager lockManager = services.GetService<IDistributedLockManager>();
                 if (lockManager == null)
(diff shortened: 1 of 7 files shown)


src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubListener.cs
```diff
@@ -116,7 +122,8 @@ public Task OpenAsync(PartitionContext context)
 
             public Task ProcessErrorAsync(PartitionContext context, Exception error)
```
brettsam: Do you have any way to verify this is getting called and logged? I wonder if there's an e2e test we can write that exercises this handling somehow.

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostConfigurationExtensions.cs
```diff
@@ -110,6 +110,7 @@ public static ServiceProviderWrapper CreateStaticServices(this JobHostConfigurat
             {
                 loggerFactory = new LoggerFactory();
                 services.AddService<ILoggerFactory>(loggerFactory);
+                config.LoggerFactory = loggerFactory;
```
pragnagopa: @brettsam - Added loggerfactory to the config object. Is this ok? Other option is to add NullHostInstanceLogger as part of services in [Jobhostconfiguration](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs#L110)

brettsam: Let's remove this entire `if` block, and add the `NullLoggerFactory.Instance` directly to JobHostConfiguration. It makes it easier to reason through how this works.

brettsam: Actually, I take that back. Instead of the `NullLoggerFactory`, use a `new LoggerFactory()`. That way people can immediately start adding to it without needing to new-up another one.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubListener.cs
pragnagopa: I could not find figure out a way to test this.  Let me know if there is anyone we could contact to get a working sample that triggers an error.

pragnagopa: I have opened issue #1604 for investigating end to end testing. Will go ahead and merge this PR

PR #1585, opened 2018-02-21
moving flaky test away from WaitHandles

brettsam:
AppVeyor runs on fewer cores than our dev machines, so I suspect a lot of these tests that mix background Tasks with WaitHandles may be getting deadlocked more often there. When I find a failing tests, I'll swap it out to use TaskCompletionSource instead.

### DIFF
--- a/test/Microsoft.Azure.WebJobs.Host.TestCommon/TestHelpers.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.TestCommon/TestHelpers.cs
@@ -1,6 +1,12 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using System;
+using System.Diagnostics;
+using System.Linq;
+using System.Reflection;
+using System.Threading;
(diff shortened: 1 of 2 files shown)


test/Microsoft.Azure.WebJobs.Host.TestCommon/TestHelpers.cs
```diff
@@ -9,27 +15,28 @@
...
-using System.Threading;
-using System.Threading.Tasks;
 using Xunit;
 
 namespace Microsoft.Azure.WebJobs.Host.TestCommon
 {
     public static class TestHelpers
     {
-        public static async Task Await(Func<bool> condition, int timeout = 60 * 1000, int pollingInterval = 2 * 1000)
+        public static async Task Await(Func<bool> condition, int timeout = 60 * 1000, int pollingInterval = 2 * 1000, bool throwWhenDebugging = false, string userMessage = null)
```
brettsam: Moving these useful changes over from Functions.

brettsam:
@fabiocav and I have a new theory. Killing this PR.

PR #1591, opened 2018-02-22
removing a disposal from TaskSeriesTimer

brettsam:
In investigating the flaky test [Dispose_TriggersCommandCancellationToken](https://github.com/Azure/azure-webjobs-sdk/blob/v2.x/test/Microsoft.Azure.WebJobs.Host.UnitTests/Timers/TaskSeriesTimerTests.cs#L760), @fabiocav noticed that the comments in `TaskSeriesTimer.Dispose` indicate that the `cancellationTokenSource` should *not* be disposed, yet it was being disposed. That would explain the flakiness of this test: there was a race between when the token is canceled and the token is disposed. The test expects the token's `WaitHandle` to be usable, so occasionally you'll end up with an `ObjectDisposedException` coming from a background Task (which was ignored -- an issue with the test). 

That led to some more targeted logging so we could see that this is exactly what is happening:
```
[19652 ticks][TID 17] starting to wait for executeStarted
[34653 ticks][TID 16] executeStarted is set
[34712 ticks][TID 17] executeStarted completed
[34807 ticks][TID 17] product disposed
[34824 ticks][TID 17] starting to wait for executeFinished
[55280 ticks][TID 16] System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.Azure.WebJobs.Host.UnitTests.Timers.TaskSeriesTimerTests.<>c__DisplayClass27_3.<Dispose_TriggersCommandCancellationToken>b__1(CancellationToken cancellationToken) in C:\projects\azure-webjobs-sdk\test\Microsoft.Azure.WebJobs.Host.UnitTests\Timers\TaskSeriesTimerTests.cs:line 782
```
The call to `Dispose()` was added several years ago when enabling FxCop: https://github.com/Azure/azure-webjobs-sdk/commit/0ca360d4d449efc5cf1974ca7c4e99d2de95bbf7#diff-806e353b065475b5fe0ed75414c79056. My assumption, given the original comment, is that this was inadvertent and we should revert this call.


brettsam:
@mathewc -- Even though the original change was made 2+ years ago, any chance you remember whether this was intentional or not? Given the comment, @fabiocav and I figured it wasn't intended.

brettsam:
Before I check this in, I want to understand why we see a `NullReferenceException` in the log, yet if I manually try to force this to happen I get an `ObjectDisposedException` as I'd expect.

fabiocav:
A null ref on that particular line is indeed a strange...

brettsam:
I now understand what's happening (see below) so I'm going to get this in. I'm also going to change the logging I added since I now see there's a chance of an exception while using a StringBuilder from multiple threads like this.

I hopped onto AppVeyor and set the test to run 100 times in a loop. I saw a mix of `ObjectDisposedException` and `NullReferenceException` (and sometimes `ArgumentOutOfRangeException` on `StringBuilder`) on the line that called `cancellationToken.WaitHandle.WaitOne(1000);` When debugging in VS I got better details on the exception and the message said "System.Threading.CancellationToken.WaitHandle.get returned null." But when inspecting it in the debugger, I see an `ObjectDisposedException`. Interesting.

So I looked in reflector and I see that there's also an internal race inside of `CancellationToken.WaitHandle.get`, which directly returns the parent `CancellationTokenSource.WaitHandle`. That getter looks like:

```
internal WaitHandle WaitHandle
{
    get
    {
        this.ThrowIfDisposed();
        if (this.m_kernelEvent == null)
        {
            // do some init
        }
        return this.m_kernelEvent;
    }
}
```

And `CancellationTokenSource.Dispose` looks like:
```
protected virtual void Dispose(bool disposing)
{
    if (disposing && !this.m_disposed)
    {
        // snip
        if (this.m_kernelEvent != null)
        {
            this.m_kernelEvent.Close();
            this.m_kernelEvent = null;
        }
        this.m_disposed = true;
    }
}
```

So if you call `Dispose()` from one thread and simultaneously try to access the `WaitHandle` from another (which this test is doing), there's a chance that the check to `ThrowIfDisposed()` succeeds just before `m_disposed` is set to `true`. If that happens, `m_kernelEvent` will be null, which means the `WaitHandle` is null, hence the `NullReferenceException`.

In summary: more races.

PR #1592, opened 2018-02-22
removing App Insights HTTP result handling; updating to App Insights 2.5 RTM

brettsam:
Removing this as the HTTP pipeline is different in v2 and this logging wasn't working as intended. Filed issue https://github.com/Azure/azure-functions-host/issues/2447 with details.


brettsam:
@fabiocav this is the (hopefully temporary) removal of how we hand HTTP in the SDK. Some corresponding changes to Functions tests will come when I get this moved over.

PR #1599, opened 2018-02-28
build improvements

brettsam:
Improving some build processes:
- all projects now use `CommonAssemblyInfo.cs`
- inject the build number into `AssemblyFileVersion`
- inject the build number and commit SHA into `AssemblyInformationalVersion`
- remove need for `IncludeBuildNumberInVersion` environment variable. Now we'll append the build number *unless* the commit we're building has a tag. Then we assume it's official and leave it off.

### DIFF
--- a/WebJobs.proj
+++ b/WebJobs.proj
@@ -124,8 +124,36 @@
       </Binplace>
     </ItemGroup>
   </Target>
+    
+  <Target Name="UpdateAssemblyInfo">   
+    <PropertyGroup>
+      <CommonAssemblyInfoFile>$(MSBuildProjectDirectory)\src\Common\CommonAssemblyInfo.cs</CommonAssemblyInfoFile>
+      <BuildMessage Condition="'$(BuildNumber)' != ''">build $(BuildNumber), </BuildMessage>
+      <CommitMessage Condition="'$(CommitHash)' != ''">$(BuildMessage)commit $(CommitHash)</CommitMessage>
+    </PropertyGroup>
+    
+    <Message 
+      Condition="'$(BuildNumber)' != ''"
+      Importance="High"
+      Text="Updating the AssemblyFileVersion to include the build number: $(BuildNumber)" />     
+    <ReplaceFileText 
+      Condition="'$(BuildNumber)' != ''"
+      FileName="$(CommonAssemblyInfoFile)"
+      MatchExpression="(?&lt;=\[assembly: AssemblyFileVersion\(&quot;\d+.\d+.)\d+(?=.\d+&quot;\)\])"      
+      ReplacementText="$(BuildNumber)" />
 
-  <Target Name="Build" DependsOnTargets="RestorePackages;GetBinplace">
+    <Message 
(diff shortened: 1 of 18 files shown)


WebJobs.proj
```diff
@@ -124,8 +124,36 @@
       </Binplace>
     </ItemGroup>
   </Target>
+    
+  <Target Name="UpdateAssemblyInfo">
```
mathewc: I assume you're going to apply the same changes to the other 2 repos so we're consistent (extensions, script)?

WebJobs.proj
```diff
@@ -294,4 +322,23 @@
...
+  <ParameterGroup>
+    <FileName ParameterType="System.String" Required="true" />    
+    <MatchExpression ParameterType="System.String" Required="true" />
+    <ReplacementText ParameterType="System.String" Required="true" />
+  </ParameterGroup>
+  <Task>
+    <Reference Include="System.Core" />
+    <Using Namespace="System" />
+    <Using Namespace="System.IO" />
+    <Using Namespace="System.Text.RegularExpressions" />
```
mathewc: The way the TeamCity jobs worked was that all the repos requiring this common behavior pulled WebJobs.proj from a common repo http://vstfrd:8080/Azure/One/_git/AzureJobsInternal. If you look there, all the logic you're adding in this PR is in that repo.

It'd be good I think if we had one version of this to maintain, rather than copies in each repo? That'd mean a shared public repo with the common build stuff.

src/Dashboard/Properties/AssemblyInfo.cs
```diff
@@ -1,10 +1,6 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
-using System.Reflection;
 using System.Runtime.CompilerServices;
 
-[assembly: AssemblyVersion("2.1.0.0")]
```
mathewc: Nice to remove all this duplication!

WebJobs.proj
brettsam: Yep. I got most of this from Script -- except it doesn't do the commit stamping. I'll move that over. And yeah, I'll do this to extensions as well.

WebJobs.proj
brettsam: That's a good idea. I've filed https://github.com/Azure/azure-webjobs-sdk/issues/1610 so we (or I :-)) can tackle that in the future.

src/Dashboard/Properties/AssemblyInfo.cs
brettsam: Yeah, I was thinking about that as well -- I'll see if there's a straightforward way to get this working there when I tackle that repo next.

mathewc: Actually, I think these should all be 2.2.0.0 - looks like I missed doing this when I revved the minor version last

brettsam: Doh! You're right. Good catch, I didn't even notice that.

Issue #1605, opened 2018-03-02
Failed ServiceBus messages can lead to reprocessing


mathewc:
Repro function below. For the repro, set up another thread pumping messages once per second into the queue.

On the first delivery we cause a failure and we expect the message to be reprocessed. However in steady state, we don't expect the delivery count to go above 2. I mention steady state because if you ungracefully kill the repro, any in progress messages will be abandoned and get retried, so you might see a few re-deliveries.

What actually ends up happening is messages get reprocessed multiple times with delivery counts going over 3.

```csharp
public static void Run([ServiceBusTrigger("myqueue")] string message, string messageId, int deliveryCount, DateTime enqueuedTimeUtc, TraceWriter log)
{
    if (deliveryCount == 1)
    {
        throw new ApplicationException($"Exception for message {messageId}! Retry please.");
    }
    else if (deliveryCount > 2)
    {
        log.Info($"Unexpected delivery count Id={messageId}, DeliveryCount={deliveryCount}, EnqueuedTime={enqueuedTimeUtc}");
    }
    else
    {
        log.Info($"Message Id={messageId}, DeliveryCount={deliveryCount} processed successfully.");
    }
}
```

krislw:
I have tried this now that the SDK has been released to all servers. I am **still** getting this issue where messages are being replayed, despite going over the specified dequeue count, and more importantly, successful messages are also being replayed.

Is there a specific client version of these APIs that we should be building our functions/webjobs with?
I am also struggling to find any best practice information regarding exception/error handling on message processing.

PR #1606, opened 2018-03-02
ServiceBus MessageProcessor fix (#1605)

mathewc:
Fix for https://github.com/Azure/azure-webjobs-sdk/issues/1605

As you can see from the ServiceBus [message pump source code](https://github.com/Azure/azure-service-bus-dotnet/blob/90ce19dd89cad09065103ed7bb7994e73c6cc8bb/src/Microsoft.Azure.ServiceBus/MessageReceivePump.cs#L143) the contract when using the MessageReceiver.OnMessageAsync callback pattern is that SB handles the message state for you based on whether your callback throws an exception or not.

Here's what was happening:

- user function throws an exception
- our MessageProcessor calls `BrokeredMessage.AbandonAsync()` but **does not propagate exception**
- SB interprets lack of exception as success and internally calls `BrokeredMessage.CompleteAsync()`, completing an already abandoned message
- from my testing this causes all sorts of issues and message reprocessing 

Before/After

- with the fix in this PR, the repro app can run for 5 minutes with NO reprocessed messages
- without the changes, in 5 minutes my repro app resulted in TONS of reprocessed messages (like 170 or so)

### DIFF
--- a/src/Microsoft.Azure.WebJobs.ServiceBus/MessageProcessor.cs
+++ b/src/Microsoft.Azure.WebJobs.ServiceBus/MessageProcessor.cs
@@ -70,8 +70,11 @@ namespace Microsoft.Azure.WebJobs.ServiceBus
             }
             else
             {
+                // if the invocation failed, we must propagate the
+                // exception back to SB so it can handle message state
+                // correctly
                 cancellationToken.ThrowIfCancellationRequested();
-                await message.AbandonAsync();
+                throw result.Exception;
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.WebJobs.ServiceBus/MessageProcessor.cs
```diff
@@ -70,8 +70,11 @@ public MessageProcessor(OnMessageOptions messageOptions)
             }
             else
             {
+                // if the invocation failed, we must to propagate the
```
ahmelsayed: nit: extra `to` :)

mathewc:
Yes, I looked way back in history and we've always called AbandonAsync. See code from 2 years ago [here](https://github.com/Azure/azure-webjobs-sdk/blob/445e080d8d9b15c0a5203ab285ac182654ac89b9/src/Microsoft.Azure.WebJobs.ServiceBus/Listeners/ServiceBusListener.cs#L105).

PR #1608, opened 2018-03-02
Restricting CI build branches. Ignoring .vscode folder

fabiocav:


fabiocav:
dev is already setup this way

PR #1647, opened 2018-03-28
Host Startup Partial Mode (#2497)

mathewc:
Core changes required to address Functions issue https://github.com/Azure/azure-functions-host/issues/2497. In Functions, we'll set this new mode to true. The default behavior for Functions will be the same as we have now in that we allow the host to start partially - however we'll no longer ignore listener startup failures. Listeners will now be retried in the background.

While this feature is being added primarily for Azure Functions usage, it can be generally useful. E.g. in a continuous WebJob with many different functions, the back end service for one listener might be down for a bit preventing the listener from starting. However all the other functions can run fine. With this feature, rather than the host not being able to start at all until ALL listeners can start (i.e. the standard continuous WebJobs restart loop), most functions can start running while we wait for the one to recover.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
@@ -37,6 +37,7 @@ namespace Microsoft.Azure.WebJobs.Host.Indexers
         private readonly ILogger _logger;
         private readonly SharedQueueHandler _sharedQueue;
         private readonly TimeoutAttribute _defaultTimeout;
+        private readonly bool _allowPartialHostStartup;
 
         public FunctionIndexer(
             ITriggerBindingProvider triggerBindingProvider,
@@ -49,7 +50,8 @@ namespace Microsoft.Azure.WebJobs.Host.Indexers
             ILoggerFactory loggerFactory,
             INameResolver nameResolver = null,
             SharedQueueHandler sharedQueue = null,
-            TimeoutAttribute defaultTimeout = null)
+            TimeoutAttribute defaultTimeout = null,
+            bool allowPartialHostStartup = false)
         {
             if (triggerBindingProvider == null)
             {
@@ -97,6 +99,7 @@ namespace Microsoft.Azure.WebJobs.Host.Indexers
             _logger = loggerFactory?.CreateLogger(LogCategories.Startup);
             _sharedQueue = sharedQueue;
             _defaultTimeout = defaultTimeout;
+            _allowPartialHostStartup = allowPartialHostStartup;
         }
 
         public async Task IndexTypeAsync(Type type, IFunctionIndexCollector index, CancellationToken cancellationToken)
@@ -109,9 +112,18 @@ namespace Microsoft.Azure.WebJobs.Host.Indexers
                 }
                 catch (FunctionIndexingException fex)
                 {
+                    if (_allowPartialHostStartup)
+                    {
+                        fex.Handled = true;
+                    }
+
                     fex.TryRecover(_trace, _logger);
+
                     // If recoverable, continue to the rest of the methods.
                     // The method in error simply won't be running in the JobHost.
+                    string msg = $"Function '{method.GetShortName()}' failed indexing and will be disabled.";
+                    _trace.Warning(msg);
+                    _logger?.LogWarning(msg);
                     continue;
                 }
             }
--- a/src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
@@ -84,6 +84,21 @@ namespace Microsoft.Azure.WebJobs
         }
 
         /// <summary>
+        /// Gets or sets a value indicating whether the host should be able to start partially
+        /// when some functions are in error. The default is false.
+        /// </summary>
+        /// <remarks>
+        /// Normally when a function encounters an indexing error or its listener fails to start
+        /// the error will propagate and the host will not start. However, with this option set,
+        /// the host will be allowed to start in "partial" mode:
+        ///   - Functions without errors will run normally
+        ///   - Functions with indexing errors will not be running
+        ///   - Functions listener startup failures will be retried in the background
+        ///     until they start.
+        /// </remarks>
+        public bool AllowPartialHostStartup { get; set; }
+
+        /// <summary>
         /// Gets a value indicating whether the <see cref="JobHost"/> is running in a Development environment.
         /// You can use this property in conjunction with <see cref="UseDevelopmentSettings"/> to default
         /// configuration settings to values optimized for local development.
--- a/src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
@@ -4,8 +4,8 @@
 using System;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.Azure.WebJobs.Host.Loggers;
 using Microsoft.Azure.WebJobs.Host.Protocols;
+using Microsoft.Azure.WebJobs.Host.Timers;
 using Microsoft.Azure.WebJobs.Logging;
 using Microsoft.Extensions.Logging;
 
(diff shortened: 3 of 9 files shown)


src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
```diff
@@ -109,7 +112,13 @@ internal class FunctionIndexer
                 }
                 catch (FunctionIndexingException fex)
                 {
+                    if (_allowPartialHostStartup)
+                    {
+                        fex.Handled = true;
```
mathewc: When this mode is set, we'll now be defaulting these to handled. In Functions currently we're setting this manually (code [here](https://github.com/Azure/azure-functions-host/blob/749564ab175f7c0d2e4ef811cb33f354e460803f/src/WebJobs.Script/Host/ScriptHost.cs#L1800)). We'll be able to remove that in favor of setting the new host config property.

src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
```diff
@@ -54,7 +57,54 @@ public void Dispose()
...
+                }
+
+                flx.TryRecover(_trace, _logger);
+
+                if (_allowPartialHostStartup)
+                {
+                    // if to here, the listener exception was handled and
+                    // we're in partial startup mode.
+                    // we spin up a background task retrying to start the listener
+                    Task taskIgnore = Task.Run(() => RetryStartWithBackoffAsync(cancellationToken), cancellationToken);
```
paulbatum: Just curious, what does the additional Task.Run(..) achieve here? How is it different from just 
`Task taskIgnore = RetryStartWithBackoffAsync(cancellationToken);

src/Microsoft.Azure.WebJobs.Host/Utility.cs
```diff
@@ -0,0 +1,68 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Host
```
mathewc: These utility methods and tests for exponential backoff are copied from an Azure Functions internal helper I also wrote.

src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
```diff
@@ -54,7 +57,54 @@ public void Dispose()
...
+                    unit: TimeSpan.FromSeconds(1),
+                    min: TimeSpan.FromSeconds(2),
+                    max: TimeSpan.FromMinutes(2));
+
+                try
+                {
+                    string message = $"Retrying to start listener for function '{_descriptor.ShortName}' (Attempt {attempt})";
+                    _trace.Info(message);
+                    _logger?.LogInformation(message);
+
```
paulbatum: I wonder, is this going to be really noisy in the case of misconfigured functions where we never successfully start the listener? If so, can we use some sort of category here to make it easy to filter the noise out?

test/Microsoft.Azure.WebJobs.Host.UnitTests/Listeners/FunctionListenerTests.cs
```diff
@@ -35,6 +35,56 @@ public FunctionListenerTests()
...
+            {
+                var lastTrace = trace.Traces.Last();
+                return lastTrace.Message == "Listener successfully started for function 'testfunc' after 3 retries.";
+            });
+
+            badListener.Verify(p => p.StartAsync(ct), Times.Exactly(4));
+            
+            var validators = new Action<string>[]
+            {
+                p => Assert.Equal(p, "The listener for function 'testfunc' was unable to start."),
```
mathewc: These are the logs users will see while this is going on in the background

src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
```diff
@@ -54,7 +57,54 @@ public void Dispose()
...
+                    message = $"Listener successfully started for function '{_descriptor.ShortName}' after {attempt} retries.";
+                    _trace.Info(message);
+                    _logger?.LogInformation(message);
+                }
+                catch (Exception ex) when (!ex.IsFatal())
+                {
+                    // swallow and log exceptions since we're in a retry loop
+                    // at this point doing this in the background
+                    _trace.Error(ex.Message, ex);
+                    _logger?.LogError(0, ex, ex.Message);
```
paulbatum: Same as above, wondering about whether this will generate lots of log noise..

src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs
```diff
@@ -83,6 +83,21 @@ public JobHostConfiguration(string dashboardAndStorageConnectionString)
...
+        /// <remarks>
+        /// Normally when a function encounters an indexing error or its listener fails to start
+        /// the error will propagate and the host will not start. However, with this option set,
+        /// the host will be allowed to start in "partial" mode:
+        ///   - Functions without errors will run normally
+        ///   - Functions with indexing errors will not be running
+        ///   - Functions listener startup failures will be retried in the background
+        ///     until they start.
+        /// </remarks>
+        public bool AllowPartialHostStartup { get; set; }
```
mathewc: Still considering the right name for this

paulbatum:
Its a bit hard for me to see from the diff, is there any public surface area change here? It looks like the flag is on internal types but maybe I'm misunderstanding..

src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
mathewc: My understanding is that invoking directly (fire and forget) will capture the current synchronization context and resume on that context. Task.Run will just start and complete on a thread pool thread without capturing/restoring a context.

src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
mathewc: What do you mean by "misconfigured"? Many issues will be indexing time failures (e.g. app setting for Connection doesn't exist, and in many cases even an invalid connection string for bindings like queue that do up front validation). If the function is truly broken, the backoff will max out to once every 2 minutes, which won't be very noisy. But yes, there will be an initial flurry of retries.

I'm open to putting them in a category

src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
mathewc: Noise for us you mean? Or for customers - I don't think it is noise for customers. They should be seeing these.

mathewc:
The new public surface area is the new JobHostConfiguration property which is public and is a new feature that people can use.

paulbatum:
Ahh yes that makes sense, thanks.

src/Microsoft.Azure.WebJobs.Host/Utility.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// </summary>
+        /// <param name="exponent">The backoff exponent. E.g. for backing off in a retry loop,
+        /// this would be the retry attempt count.</param>
+        /// <param name="cancellationToken">The <see cref="CancellationToken"/> to use.</param>
+        /// <param name="unit">The time unit for the backoff delay. Default is 1 second, producing a backoff sequence
+        /// of 2, 4, 8, 16, etc. seconds.</param>
+        /// <param name="min">The minimum delay.</param>
+        /// <param name="max">The maximum delay.</param>
+        /// <returns>A <see cref="Task"/> representing the computed backoff interval.</returns>
+        public static async Task DelayWithBackoffAsync(int exponent, CancellationToken cancellationToken, TimeSpan? unit = null, TimeSpan? min = null, TimeSpan? max = null)
```
brettsam: We already have a `RandomizedExponentialBackoffStrategy` -- can you use that existing logic here?

mathewc: Yes, I'll move to that and remove the new code I added for this. I had considered that previously but thought the core backoff stuff was tied too deeply with the TaskSeriesTimer stuff (I was wrong).

PR #1650, opened 2018-03-30
Improving ServiceBus listener state management

mathewc:


src/Microsoft.Azure.WebJobs.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -36,13 +37,14 @@ public Task StartAsync(CancellationToken cancellationToken)
         {
             ThrowIfDisposed();
 
-            if (_receiver != null)
```
mathewc: Using _receiver to determine that the listener was successfully started is incorrect

PR #1656, opened 2018-04-06
Extra logging that will help For Type Version mismatch

vsadams:
In the case of a ReflectionTypeLoadException, write out the loader exception to assist in debugging.

This can happen if the runtime has a version of a library loaded that is not the same version that a customer library is using.


vsadams:
@fabiocav  I have made your suggested updates!

fabiocav:
Thank you! Merged!

PR #1676, opened 2018-04-25
fixes #1675

### LINKED ISSUE #1675: Cannot set `GracePeriod` on `TimeoutAttribute`
It is not possible to specify a `GracePeriod` on a `TimeoutAttribute` because `TimeSpan` is not a valid attribute parameter type

#### Repro steps

1. Attempt to set the `GracePeriod` parameter when using the `TimeoutAttribute`

```
public class Functions
    {

        [Timeout("00:10:00", GracePeriod = TimeSpan.FromSeconds(15), ThrowOnTimeout = true, TimeoutWhileDebugging = true)]
        public async Task MyFunction([QueueTrigger("myqueue")] CloudQueueMessage message, TextWriter log, Microsoft.Azure.WebJobs.ExecutionContext context, CancellationToken cancellationToken)
        {
    ...
```

#### Expected behavior

Expect to have some means of specifying the `GracePeriod` on a `TimeoutAttribute` - perhaps via an overload to the constructor with a string to parse

#### Actual behavior

Build error:

> CS0655	'GracePeriod' is not a valid named attribute argument because it is not a valid attribute parameter type

#### Known workarounds

I believe it is possible to control the timeout behaviourd via the `JobHostConfiguration` rather than an attribute on the function.

stewart-r:
fixes #1675


mathewc:
Rebased and merged in https://github.com/Azure/azure-webjobs-sdk/commit/7629644f5c6f8759321d6b978b83badd1b4bbcb7.

Thanks for the contribution!

Issue #1691, opened 2018-05-09
Race condition when retrying failed listeners


brettsam:
Customer reported an issue and when I looked into it, I could see that we were running a disposed host for several hours. When walking back through the logs, I noticed that it appears we are retrying a failed listener after the Job host has been stopped. This causes the stopped, disposed host to continue running functions indefinitely, which causes constant failures until the host is restarted.

Change that implemented this was https://github.com/Azure/azure-webjobs-sdk/pull/1647

Relevant log lines:

````
2018-05-08 11:16:44.9554933    The listener for function 'Functions.functionname' was unable to start....
2018-05-08 11:18:55.9265356    Retrying to start listener for function 'Functions.functionname' (Attempt 1)
2018-05-08 11:19:11.4812976    Job host stopped
2018-05-08 11:19:14.9707997    Listener successfully started for function 'Functions.functionname' after 1 retries.
````

**Comment by brettsam** [comment]
I believe this may be a bug in the Event Hub trigger. It looks like it's not set up to handle `StopAsync()` being called while `StartAsync()` is executing. If `Start()` is in the middle of registering a handler while `Stop()` is called, it looks like it'll be a no-op and the listener will continue running:

https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubListener.cs#L48-L62

**Comment by brettsam** [comment]
Rather than create a new issue -- I'll reactivate this one. It looks like we still have an issue here. I can see an EventHubListener getting started during a retry after the host is stopped. It tries to process events on a disposed host for a while:
```
FunctionsLogs
| where PreciseTimeStamp > datetime(2019-02-13 07:20) and PreciseTimeStamp < datetime(2019-02-13 10:00)
| where AppName == "rdc-func-app"
| where HostInstanceId == "95ea2cfb-05cd-4952-8cb2-36177aed7912"
| project PreciseTimeStamp, Details, Summary, HostInstanceId, Level, HostVersion
```

PR #1692, opened 2018-05-09
Fixing listener retry race condition (#1691)

mathewc:
Fix for https://github.com/Azure/azure-webjobs-sdk/issues/1691.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
@@ -19,6 +19,7 @@ namespace Microsoft.Azure.WebJobs.Host.Listeners
         private readonly FunctionDescriptor _descriptor;
         private readonly TraceWriter _trace;
         private readonly ILogger _logger;
+        private readonly SemaphoreSlim _semaphoreSlim = new SemaphoreSlim(1, 1);
         private bool _started = false;
         private bool _allowPartialHostStartup;
         private CancellationTokenSource _retryCancellationTokenSource;
@@ -55,6 +56,7 @@ namespace Microsoft.Azure.WebJobs.Host.Listeners
         {
             _listener.Dispose();
             _retryCancellationTokenSource?.Cancel();
+            _semaphoreSlim.Dispose();
         }
 
         public async Task StartAsync(CancellationToken cancellationToken)
@@ -114,6 +116,17 @@ namespace Microsoft.Azure.WebJobs.Host.Listeners
                         message = $"Listener successfully started for function '{_descriptor.ShortName}' after {attempt} retries.";
                         _trace.Info(message);
                         _logger?.LogInformation(message);
+
+                        if (cancellationToken.IsCancellationRequested)
+                        {
+                            // stop has been called while we were in the process of starting
+                            // so we need to stop this listener
+                            await StopAsync(cancellationToken);
+
+                            message = $"Listener for function '{_descriptor.ShortName}' stopped. A stop was initiated while starting.";
+                            _trace.Info(message);
+                            _logger?.LogInformation(message);
+                        }
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


test/Microsoft.Azure.WebJobs.Host.UnitTests/Listeners/FunctionListenerTests.cs
```diff
@@ -150,6 +150,62 @@ public FunctionListenerTests()
...
+            bool stopCalled = false;
+            badListener.Setup(bl => bl.StopAsync(It.IsAny<CancellationToken>()))
+                .Callback(() => stopCalled = true)
+                .Returns(Task.CompletedTask);
+
+            await listener.StartAsync(ct);
+
+            await TestHelpers.Await(() =>
+            {
+                return stopCalled;
```
mathewc: Before the fix, this test would hang. The listener was never stopped

src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
```diff
@@ -114,6 +114,17 @@ public void Dispose()
...
+
+                        if (cancellationToken.IsCancellationRequested)
+                        {
+                            // stop has been called while we were in the process of starting
+                            // so we need to stop this listener
+                            await StopAsync(cancellationToken);
+
+                            message = $"Listener for function '{_descriptor.ShortName}' stopped.";
+                            _trace.Info(message);
+                            _logger?.LogInformation(message);
```
brettsam: Can you move this log into `StopAsync()` itself?

src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
```diff
@@ -114,6 +114,17 @@ public void Dispose()
                         message = $"Listener successfully started for function '{_descriptor.ShortName}' after {attempt} retries.";
                         _trace.Info(message);
                         _logger?.LogInformation(message);
+
+                        if (cancellationToken.IsCancellationRequested)
+                        {
+                            // stop has been called while we were in the process of starting
+                            // so we need to stop this listener
+                            await StopAsync(cancellationToken);
```
brettsam: Now there's a chance that Stop can be called multiple times on a single listener. Do we know how different listeners will react to that?

src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
mathewc: We're not doing any start/stop/cancel etc. logging for listener operations currently. If we moved this, we'd want to add logging to all the other methods as well. And I think that would result in a lot of noise in the logs. That's why I kept this on the retry path only. What do you think?

src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
mathewc: You're talking about the case where the retry succeeds to start the listener, sets _started to true, context switches to an outer StopAsync call which will cancel the token and call Stop, context switch back and the retry loop sees the cancellation and calls stop again? Yes, there is a race there I'll need to synchronize.

mathewc: Ok, I've added synchronization on StopAsync

src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
brettsam: We can leave it here, that's okay. It may be nitpicky, but I'd suggest we add something to this message that indicates this is only b/c of a retry and that it's not a general 'listener stopping' message. 

My concern was that it could be confusing see "Listener for function 'abc' stopped" under this one specific condition, yet we don't log it generally. Someone analyzing failures in the future may assume that no other listeners were stopped b/c they don't see that log message.

src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
brettsam: Looks good!

src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
mathewc: Updated message to "Listener for function 'xxx' stopped. A stop was initiated while starting."

Issue #1689, opened 2018-05-07
Environment Variable Caching issue


mathewc:
The core SDK is using IConfiguration, which caches environment variables at the Process level [here](https://github.com/Azure/azure-webjobs-sdk/blob/24447418dcfd5657e320937c7c7ce5d305c0f28c/src/Microsoft.Azure.WebJobs.Host/ConfigurationUtility.cs#L55).

We fixed this in Azure Functions here: https://github.com/Azure/azure-functions-host/pull/2728. We should apply the same fix.

PR #1694, opened 2018-05-10
Fixing IConfiguration environment variable caching

mathewc:
Addresses https://github.com/Azure/azure-webjobs-sdk/issues/1689

Applying same fix that we did for Azure Functions in https://github.com/Azure/azure-functions-host/commit/29397abfc80b2850164cf2c3da4ec380f4ede607. There are Azure Functions tests/scenarios currently disabled due to this bug (e.g. [this test](https://github.com/Azure/azure-functions-host/blob/ef311c8c4a46e6494e0b6b4af359586bb1295667/test/WebJobs.Script.Tests/Binding/CoreExtensionsScriptBindingProviderTests.cs#L17)).

Eventually just like in Functions the static config that ConfigurationUtility is doing will go away - this is a stepping stone to get things working consistently up and down the stack.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/ConfigurationUtility.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/ConfigurationUtility.cs
@@ -2,8 +2,9 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
-using System.Globalization;
+using System.Linq;
 using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Configuration.EnvironmentVariables;
 
 namespace Microsoft.Azure.WebJobs.Host
 {
@@ -32,7 +33,6 @@ namespace Microsoft.Azure.WebJobs.Host
         public static void SetConfigurationFactory(Func<IConfiguration> configurationRootFactory)
         {
             _configurationFactory = configurationRootFactory;
-            Reset();
         }
 
         public static bool IsSettingEnabled(string settingName)
@@ -53,15 +53,10 @@ namespace Microsoft.Azure.WebJobs.Host
         private static IConfigurationRoot BuildConfiguration()
         {
             var configurationBuilder = new ConfigurationBuilder()
-                .AddJsonFile("appsettings.json", optional: true)
-                .AddEnvironmentVariables();
+                .AddJsonFile("appsettings.json", optional: true);
+            configurationBuilder.Sources.Add(new WebJobsEnvironmentVariablesConfigurationSource());
 
             return configurationBuilder.Build();
-        }
-
-        internal static void Reset()
-        {
-            _configuration = new Lazy<IConfiguration>(_configurationFactory);
         }
     }
 }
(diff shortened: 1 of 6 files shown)


src/Microsoft.Azure.WebJobs.Host/ConfigurationUtility.cs
```diff
@@ -56,12 +56,16 @@ private static IConfigurationRoot BuildConfiguration()
                 .AddJsonFile("appsettings.json", optional: true)
                 .AddEnvironmentVariables();
 
-            return configurationBuilder.Build();
-        }
+            // replace the default environment source with our own
+            IConfigurationSource envVarsSource = configurationBuilder.Sources.OfType<EnvironmentVariablesConfigurationSource>().FirstOrDefault();
+            if (envVarsSource != null)
+            {
+                configurationBuilder.Sources.Remove(envVarsSource);
```
brettsam: Rather than remove this, why not just skip adding it altogether?

src/Microsoft.Azure.WebJobs.Host/WebJobsEnvironmentVariablesConfigurationSource.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        private class WebHostEnvironmentVariablesProvider : EnvironmentVariablesConfigurationProvider
+        {
+            public WebHostEnvironmentVariablesProvider() : base()
+            {
+            }
+
+            public override bool TryGet(string key, out string value)
+            {
+                value = Environment.GetEnvironmentVariable(key);
```
brettsam: It looks like the EnvConfigProvider does some normalization that we'd be bypassing here: https://github.com/aspnet/Configuration/blob/dev/src/Config.EnvironmentVariables/EnvironmentVariablesConfigurationProvider.cs.

mathewc: I say we skip that funky logic - for environment variables we just return the truth. I'm not sure what that logic is for, but we've never had it for Functions and we've been fine :)

brettsam: The "__" -> ":" for hierarchical settings was what I was most worried about. From [the docs](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-2.1&tabs=basicconfiguration#configuration-by-environment):

`For hierarchical config values specified in environment variables, a colon (:) may not work on all platforms. Double underscore (__) is supported by all platforms.`

src/Microsoft.Azure.WebJobs.Host/ConfigurationUtility.cs
mathewc: How?

src/Microsoft.Azure.WebJobs.Host/WebJobsEnvironmentVariablesConfigurationSource.cs
mathewc: Updated the code to support this normalization, and added test verifying that and case insensitivity as well

src/Microsoft.Azure.WebJobs.Host/ConfigurationUtility.cs
brettsam: Right above this -- remove `AddEnvironmentVariables()` from the `ConfigurationBuilder`.

mathewc: doh! done

PR #1704, opened 2018-05-17
Exception logging for MessageReceiver background exceptions

mathewc:
Addresses https://github.com/Azure/azure-webjobs-sdk/issues/1703

I'll be making similar changes in v2

PR #1709, opened 2018-05-18
Exception logging for MessageReceiver background exceptions

mathewc:
Merging https://github.com/Azure/azure-webjobs-sdk/pull/1704 to dev. Code is a bit different in dev branch due to API differences.

### DIFF
--- a/sample/SampleHost/Functions.cs
+++ b/sample/SampleHost/Functions.cs
@@ -2,7 +2,9 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs;
+using Microsoft.Extensions.Logging;
 using Newtonsoft.Json.Linq;
 using SampleHost.Filters;
 using SampleHost.Models;
@@ -32,6 +34,19 @@ namespace SampleHost
             [QueueTrigger("test")] WorkItem workItem)
         {
             Console.WriteLine($"Processed work item {workItem.ID}");
+        }
+
+        public static async Task ProcessWorkItem_ServiceBus(
+            [ServiceBusTrigger("test-items")] WorkItem item,
+            string messageId,
+            int deliveryCount,
+            ILogger log)
+        {
+            log.LogInformation($"Processing ServiceBus message (Id={messageId}, DeliveryCount={deliveryCount})");
+
+            await Task.Delay(1000);
+
+            log.LogInformation($"Message complete (Id={messageId})");
         }
     }
 }
--- a/src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Constants/LogCategories.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Constants/LogCategories.cs
@@ -25,6 +25,11 @@ namespace Microsoft.Azure.WebJobs.Logging
         public const string Singleton = "Host.Singleton";
 
         /// <summary>
+        /// The category for all logs written by the function executor.
+        /// </summary>
+        public const string Executor = "Host.Executor";
+
+        /// <summary>
         /// The category for logs written by the function aggregator.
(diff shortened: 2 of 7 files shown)


sample/SampleHost/Functions.cs
```diff
@@ -33,5 +35,18 @@ public static class Functions
...
+
+        public static async Task ProcessWorkItem_ServiceBus(
+            [ServiceBusTrigger("test-items")] WorkItem item,
+            string messageId,
+            int deliveryCount,
+            ILogger log)
+        {
+            log.LogInformation($"Processing ServiceBus message (Id={messageId}, DeliveryCount={deliveryCount})");
+
+            await Task.Delay(TimeSpan.FromMinutes(7));
```
mathewc: Will revert this back to 1 second after checkin. Still testing lock loss scenarios + logging

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Constants/LogCategories.cs
```diff
@@ -24,6 +24,11 @@ public static class LogCategories
         /// </summary>
         public const string Singleton = "Host.Singleton";
 
+        /// <summary>
+        /// The category for all logs written by the function executor.
+        /// </summary>
+        public const string Executor = "Host.Executor";
```
mathewc: This exists in the v2 branch, but not here for some reason

src/Microsoft.Azure.WebJobs.ServiceBus/Config/ServiceBusExtensionConfig.cs
```diff
@@ -54,6 +57,20 @@ public void Initialize(ExtensionConfigContext context)
                 throw new ArgumentNullException("context");
             }
 
+            // Register an exception handler for background exceptions
+            // coming from MessageReceivers.
+            Config.ExceptionHandler = (e) =>
```
mathewc: Our setting this here will override the logic used by our default MessageReceiverOptions that we set up. If the user sets ServiceBusConfiguration.MessageOptions to their own instance with their own handler, they completely override, and our handler doesn't get used (by design). For WebJobs scenarios that's what we want - user has full control. For Functions, they can't override our MessageReceiverOptions instance, so our logging will always be in effect.

src/Microsoft.Azure.WebJobs.ServiceBus/Config/ServiceBusConfiguration.cs
```diff
@@ -54,7 +53,7 @@ public string ConnectionString
         }
 
         /// <summary>
-        /// Gets or sets the default <see cref="MessageHandlerOptions"/> that will be used by
+        /// Gets the default <see cref="MessageHandlerOptions"/> that will be used by
```
fabiocav: nit: This still has a setter.

src/Microsoft.Azure.WebJobs.ServiceBus/Config/ServiceBusExtensionConfig.cs
```diff
@@ -54,6 +57,20 @@ public void Initialize(ExtensionConfigContext context)
...
+            {
+                var sbex = e.Exception as ServiceBusException;
```
fabiocav: nit: you should be able to inline this (e.g. `(if e.Exception is ServiceBusException sbx && !sbex.IsTransient)`)

mathewc: But then I have to cast the exception twice

src/Microsoft.Azure.WebJobs.ServiceBus/Config/ServiceBusConfiguration.cs
mathewc: Fixed

Issue #1705, opened 2018-05-17
Log background errors coming from EventProcessorHost


mathewc:
When setting up the EventProcessorHost, we don't currently add an event handler to [EventProcessorOptions.ExceptionReceived](https://docs.microsoft.com/en-us/dotnet/api/microsoft.servicebus.messaging.eventprocessoroptions.exceptionreceived?view=azure-dotnet). This means that any background exceptions will be silently ignored.

See related issue https://github.com/Azure/azure-webjobs-sdk/issues/1703

pragnagopa:
Related info: Fix for Issue #1357 added logging in [EventHubListener.cs#L131](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubListener.cs#L131). 

Issue tracking to add E2E test #1604.

PR #1714, opened 2018-05-22
(Dev) Exception logging for EventProcessorHost background exceptions …

mathewc:
…(#1705)

This PR is the dev branch counterpart to https://github.com/Azure/azure-webjobs-sdk/pull/1712

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubConfiguration.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubConfiguration.cs
@@ -11,7 +11,9 @@ using Microsoft.Azure.EventHubs;
 using Microsoft.Azure.EventHubs.Processor;
 using Microsoft.Azure.WebJobs.Host.Bindings;
 using Microsoft.Azure.WebJobs.Host.Config;
+using Microsoft.Azure.WebJobs.Logging;
 using Newtonsoft.Json;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.WebJobs.ServiceBus
 {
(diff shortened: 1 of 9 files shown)


src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubConfiguration.cs
```diff
@@ -43,27 +43,18 @@ public class EventHubConfiguration : IExtensionConfigProvider
...
-        public EventHubConfiguration()
-            : this(null)
-        {
-        }
-
         /// <summary>
         /// Constructs a new instance.
         /// </summary>
-        /// <param name="options">The optional <see cref="EventProcessorOptions"/> to use when receiving events.</param>
-        public EventHubConfiguration(EventProcessorOptions options)
```
mathewc: Making the options a get/set property, symmetric with ServiceBusConfiguration.MessageOptions. All our other config classes have default constructors with settable properties.

src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubConfiguration.cs
```diff
@@ -437,6 +439,13 @@ private static Task<object> ConvertPocoToEventData(object arg, Attribute attrRes
             return Task.FromResult<object>(ConvertString2EventData(JsonConvert.SerializeObject(arg)));
         }
 
+        internal Action<ExceptionReceivedEventArgs> ExceptionHandler { get; set; }
+
+        private void ExceptionReceivedHandler(ExceptionReceivedEventArgs args)
```
mathewc: This somewhat strange level of indirection is necessary because the options class is constructed (and handler wired up) before initialize runs and can configure the actual handler. The handler is a ctor param of EventProcessorOptions so must be set up early.

Note that in the v1 version of the EH library the handler is an **event** so we don't have to do this - we just add an event handler.

test/Microsoft.Azure.WebJobs.EventHubs.UnitTests/EventHubConfigurationTests.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            // ensure the EventHubTriggerAttributeBindingProvider was registered
+            triggerBindingProviders = extensions.GetExtensions<ITriggerBindingProvider>().ToArray();
+            EventHubTriggerAttributeBindingProvider triggerBindingProvider = (EventHubTriggerAttributeBindingProvider)triggerBindingProviders.Single();
+            Assert.NotNull(triggerBindingProvider);
+
+            // ensure the EventProcessorOptions ExceptionReceived event is wired up
+            var eventProcessorOptions = eventHubConfiguration.EventProcessorOptions;
+            var ex = new EventHubsException(false, "Kaboom!");
+            var ctor = typeof(ExceptionReceivedEventArgs).GetConstructors(BindingFlags.NonPublic | BindingFlags.Instance).Single();
```
mathewc: They've decided to make their event args ctor internal so you can't mock it at all :(

PR #1718, opened 2018-05-24
Fixing UseEventHub namespace issue

mathewc:
For proper discoverability, all extension registration methods are supposed to be in the root Microsoft.Azure.WebJobs so they are discoverable in a vanilla console host immediately after adding the extension package. All the other extensions follow this pattern, but we missed this on EH.

### DIFF
--- a/sample/SampleHost/Program.cs
+++ b/sample/SampleHost/Program.cs
@@ -4,7 +4,6 @@
 using System;
 using Microsoft.Azure.WebJobs;
 using Microsoft.Azure.WebJobs.Logging;
-using Microsoft.Azure.WebJobs.ServiceBus;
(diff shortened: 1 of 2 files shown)


brettsam (approved):

sample/SampleHost/Program.cs
```diff
@@ -4,7 +4,6 @@
 using System;
 using Microsoft.Azure.WebJobs;
 using Microsoft.Azure.WebJobs.Logging;
-using Microsoft.Azure.WebJobs.ServiceBus;
```
mathewc: The new EH sample function demonstrates the problem nicely. Today I added the EH sample function to this app, but after referencing the extension assembly, doing "config." for intellisense didn't show "UseEventHub".

Issue #1719, opened 2018-05-24
EventHubs extension in ServiceBus namespace


mathewc:
Now that we've moved this extension out into its own assembly, we need to fix the namespace - it's still under **Microsoft.Azure.WebJobs.ServiceBus** when it should be **Microsoft.Azure.WebJobs.EventHubs**.

We'll need to coordinate this with Azure Functions template updates of course.

PR #1720, opened 2018-05-24
Fixing EventHubs extension namespacing

mathewc:
Fix for https://github.com/Azure/azure-webjobs-sdk/issues/1719

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubAttribute.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubAttribute.cs
@@ -4,7 +4,7 @@
 using System;
 using Microsoft.Azure.WebJobs.Description;
 
-namespace Microsoft.Azure.WebJobs.ServiceBus
+namespace Microsoft.Azure.WebJobs
 {
     /// <summary>
     /// Setup an 'output' binding to an EventHub. This can be any output type compatible with an IAsyncCollector.
--- a/src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubAsyncCollector.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubAsyncCollector.cs
@@ -8,14 +8,14 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.EventHubs;
 
-namespace Microsoft.Azure.WebJobs.ServiceBus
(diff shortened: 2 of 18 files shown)


src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubAttribute.cs
```diff
@@ -4,7 +4,7 @@
 using System;
 using Microsoft.Azure.WebJobs.Description;
 
-namespace Microsoft.Azure.WebJobs.ServiceBus
```
mathewc: Just like the "UseEventHubs" attribute should be in the root namespace, so should the core attributes

paulbatum:
high level question for you - why doesn't this namespace have "extensions" in it? 

Examples from other extensions:
`namespace Microsoft.Azure.WebJobs.Extensions.EventGrid`
`namespace Microsoft.Azure.WebJobs.Extensions.CosmosDB`

mathewc:
@paulbatum Good question. It's historical. ServiceBus binding existed before extensibility, and when I pushed it out into its own assembly and made it an extension, it retained its existing namespace/assembly name. It was the very first extension, and that naming pattern hadn't been established yet. Later, I created the Extrensions repo and the first set of extensions, and that naming convention. Never went back to rename Service bus namespace/assembly/package to match the convention. EventHub extension started out as part of ServiceBus, so it inherited all this.

We could rename it all to follow the pattern if we want. In addition to the changes in this PR, we'd update the ServiceBus namespace as well. We'd also want to update the assembly names as well for both of them. Not sure it's worth it.

paulbatum:
@mathewc So if we're going to change the event hubs namespace anyway and endure a breaking change, we might as well make it consistent with the rest of the stuff and call it an extension. For service bus and storage, I'm not sure if its worth the break just to achieve consistency. Do you agree?

paulbatum:
I'll add that I suspect its not worth changing any assembly names here.

mathewc:
If we're going to add "Extensions" to the namespace, I think we should do it for ServiceBus as well, and probably for the new Storage extension that is being created. Since none of these are baked into the core anymore, they're all extensions.

mathewc:
Paul and I discussed in person and agreed that we'll update the namespace AND assembly name for the EventHub extension to start with **Microsoft.Azure.WebJobs.Extensions.EventHubs**. We'll leave the ServiceBus extension inconsistent for now, but may also fix that soon as well while we're doing the Storage extension refactorings.

PR #1724, opened 2018-05-24
Changes to ExtensionTypeProvider 

fabiocav:
Changes to ExtensionTypeProvider to avoid the use of the DefaultTypeProvider. Removal of the code path that loads ICloudBlobStreamBinder<T> implementations.

**THIS IS A TACTICAL CHANGE**. Proper fix should come after the hosting/DI changes land, and would be the removal of all deprecated interfaces (e.g. `IExtensionTypeProvider`, `ICloudBlobStreamBinder<T>`)

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Indexers/ExtensionTypeLocator.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Indexers/ExtensionTypeLocator.cs
@@ -10,6 +10,7 @@ namespace Microsoft.Azure.WebJobs.Host.Indexers
     {
         private readonly ITypeLocator _typeLocator;
         private IReadOnlyList<Type> _cloudBlobStreamBinderTypes;
+        private readonly static IReadOnlyList<Type> _emptyList = new List<Type>().AsReadOnly();
 
         public ExtensionTypeLocator(ITypeLocator typeLocator)
         {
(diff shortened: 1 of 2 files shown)


fabiocav:
@mathewc @MikeStall this is a tactical change to eliminate the use of the default type locator in the functions initialization path. I’m trying to get this in to initiate the release process for V2 and address some issues impacting the new assembly binding logic.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AzureStorageEndToEndTests.cs
```diff
@@ -191,13 +191,13 @@ public AzureStorageEndToEndTests(TestFixture fixture)
             await EndToEndTest(uploadBlobBeforeHostStart: false);
         }
 
-        [Fact]
+        [Fact(Skip = "This test relies on ICloudBlobStreamBinder<T> support, whith has been removed")]
```
mathewc: Fix spelling error, and reference the bug https://github.com/Azure/azure-webjobs-sdk/issues/995

src/Microsoft.Azure.WebJobs.Host/Indexers/ExtensionTypeLocator.cs
```diff
@@ -10,6 +10,7 @@ internal class ExtensionTypeLocator : IExtensionTypeLocator
     {
         private readonly ITypeLocator _typeLocator;
         private IReadOnlyList<Type> _cloudBlobStreamBinderTypes;
+        private readonly static IReadOnlyList<Type> _emptyList = new List<Type>().AsReadOnly();
```
mathewc: Can't this entire ExtensionTypeLocator be removed? I think that'd be best, particularly since you've removed the link to the ICloudBlobStreamBinder type in this source, so we may forget to clean this up.

MikeStall: This will already happen with the DI, and we wanted to minimize the merge conflicts.

fabiocav: Indeed. As mentioned in the description, the proper fix (removal of the mentioned interfaces, etc.) is a bit invasive and will certainly introduce some painful conflicts. We'll address those once the DI changes are in place.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AzureStorageEndToEndTests.cs
fabiocav: done.

PR #1733, opened 2018-06-01
Updating version to 3.0.0-beta7

fabiocav:


salmangadit:
@brettsam @fabiocav - has this been published as a nuget package yet? If not, what is the ETA? I need the latest update to see if Webjobs work with .NET Core 2.1 :)

PR #1739, opened 2018-06-07
improving TestLogger against race conditions

brettsam:


brettsam:
There's also general test logging improvements sprinkled throughout to help us better understand flaky failures.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
```diff
@@ -118,7 +119,6 @@ public AsyncChainEndToEndTests(TestFixture fixture)
                     "Microsoft.Azure.WebJobs.Host.EndToEndTests.AsyncChainEndToEndTests.ReadResultBlob",
                     "Microsoft.Azure.WebJobs.Host.EndToEndTests.AsyncChainEndToEndTests.SystemParameterBindingOutput",
                     "Function 'AsyncChainEndToEndTests.DisabledJob' is disabled",
-                    "ServicePointManager.DefaultConnectionLimit is set to the default value of 2.",
```
mathewc: Why this removal?

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
```diff
@@ -783,6 +782,11 @@ public static void DisabledJob([QueueTrigger(Queue1Name)] string message)
             Assert.Equal("async works", _finalBlobContent);
         }
 
+        private void WaitForFunctionComplete(int waitTimeout = 60000)
```
mathewc: Can we set the max to 30s to minimize the duration of hanging/failed tests? We should never have any test that takes this long to run, right?

brettsam: I've made it 30 sec in my 'dev' PR. I'll be porting that here.

I was seeing some real odd behavior a few days ago where blob triggers were taking *minutes* to run the first time due to a really long delay in parsing the Blob logs. I've not been able to repro it now but I want to keep an eye on it. This was my attempt to debug that on AppVeyor.

PR #1744, opened 2018-06-11
removing e2e test race condition

brettsam:
In the `AsyncChain...` e2e tests, there are two tests in the chain that go:

1. Add message to queue and write blob.
2. Trigger off of queue message inserted in 1 and read blob from 1.

Because the queue message was inserted first, there was a chance that the next function started before the blob write had completed. That function could throw a 404 on the missing blob and then re-run successfully. For most of the e2e tests, this was fine as they only cared whether the full chain of functions completed successfully. However, the `AggregatorOnly` test was looking at the aggregation events. It'd see that failed-and-retried function and fail because it was unexpected.

Branch `v2.x` has this same problem. Will be following up there as well.

### DIFF
--- a/test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
@@ -431,15 +431,14 @@ namespace Microsoft.Azure.WebJobs.Host.EndToEndTests
         }
 
         [NoAutomaticTrigger]
-        public static async Task WriteStartDataMessageToQueue(
-            [Queue(Queue1Name)] ICollector<string> queueMessages,
-            [Blob(ContainerName + "/" + NonWebJobsBlobName, FileAccess.Write)] Stream nonSdkBlob,
-            CancellationToken token)
+        public static void WriteStartDataMessageToQueue(
+            [Queue(Queue1Name)] out string queueMessage,
+            [Blob(ContainerName + "/" + NonWebJobsBlobName, FileAccess.Write)] Stream nonSdkBlob)
         {
-            queueMessages.Add(" works");
+            queueMessage = " works";
 
             byte[] messageBytes = Encoding.UTF8.GetBytes("async");
-            await nonSdkBlob.WriteAsync(messageBytes, 0, messageBytes.Length);
+            nonSdkBlob.Write(messageBytes, 0, messageBytes.Length);
         }
 
         [NoAutomaticTrigger]
@@ -499,6 +498,7 @@ namespace Microsoft.Azure.WebJobs.Host.EndToEndTests
             CloudBlobClient blobClient = _storageAccount.CreateCloudBlobClient();
             CloudBlobContainer container = blobClient.GetContainerReference(_resolver.ResolveInString(ContainerName));
             CloudBlockBlob blob = container.GetBlockBlobReference(NonWebJobsBlobName);
+
             string blobContent = await blob.DownloadTextAsync();
 
             trace.Info("User TraceWriter log 1");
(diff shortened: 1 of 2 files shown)


test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
```diff
@@ -436,10 +436,10 @@ public void FunctionFailures_LogsExpectedMessage()
             [Blob(ContainerName + "/" + NonWebJobsBlobName, FileAccess.Write)] Stream nonSdkBlob,
             CancellationToken token)
         {
-            queueMessages.Add(" works");
-
             byte[] messageBytes = Encoding.UTF8.GetBytes("async");
             await nonSdkBlob.WriteAsync(messageBytes, 0, messageBytes.Length);
+
+            queueMessages.Add(" works");
         }
```
brettsam: Well... that didn't seem to work. Thinking through this again. Maybe WriteAsync isn't guaranteed to be completed after it returns?

brettsam:
It looks like according to this: https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs#L936 we try to process queues last. However, when using a Collector, we add the messages immediately so the sorting doesn't really matter. We don't call `CommitAsync()` on the blob stream until after the function has exited as part of output parameter processing, which means the next function in the chain may run before that occurs.

mathewc:
Yes, the BindStepOrder stuff for Queue output bindings only applies when binding to a single output queue message. In that case the queue message is written when the function completes, in BindStepOrder. As you say, with a collector the messages are send when added.

brettsam:
@MikeStall / @mathewc -- here's an odd scenario that's been around a while it looks like.

See the conversation above (with myself :-))  for what is happening. It looks like we try to make sure queue params are handled last, yet the collectors don't batch, so it doesn't guarantee anything. One of our tests was actually checking which functions were being invoked -- so it fails every once in a while when the blob hasn't been committed.

Mike, I know you said this was something we should revisit in v2 in a couple of different areas: 
- Making sure we batch/flush every AsyncCollector.
- Removing the BindStepOrder thing completely -- since it's much harder now that the number of extensions have exploded.

This is just for stabilizing the test, but wanted to make sure we open up the right issues to track the improvements.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
```diff
@@ -499,6 +499,13 @@ public static void DisabledJob([QueueTrigger(Queue1Name)] string message)
             CloudBlobClient blobClient = _storageAccount.CreateCloudBlobClient();
             CloudBlobContainer container = blobClient.GetContainerReference(_resolver.ResolveInString(ContainerName));
             CloudBlockBlob blob = container.GetBlockBlobReference(NonWebJobsBlobName);
+
+            // The blob may not have committed before this function started.
+            while (!await blob.ExistsAsync())
```
mathewc: Rather than polling like this, I think the scenario of this test would remain as intended if the writer was changed from a collector binding to using an output binding. It only writes a single message anyways. That way things will happen in the correct order - when the message is received, we know the blob is ready for processing.

brettsam: Yeah good point. Done.

mathewc:
It is a very common pattern to do some work, produce one or more artifacts, then write a queue message indicating that the next stage in the process can begin (i.e. the artifacts for processing are ready).

The funky BindStepOrder stuff handles the case where the flush for output bindings is controlled by the binding pipeline - we ensure the blob stream is flushed, then we write the queue message. However if writes/flushes are controlled by the user internally to the function (using ICollector, Binder, etc.) they control the ordering. Couldn't this test also have been updated to first write and flush the blob, then add the message to the collector? That should have produced the correct ordering as well.

PR #1746, opened 2018-06-11
fixing race in e2e test

brettsam:
`v2.x` version of #1744


brettsam:
@mathewc -- fyi this is the `v2.x` version. As long as it passes CI, I'll merge it.

Issue #1743, opened 2018-06-11
Add ServiceBus QueueClient binding


mathewc:
In the new .NET Core ServiceBus SDK, the BrokeredMessage type was removed and changed to Message. As part of that SB SDK change, all the message operations (Complete, Abandon, etc.) no longer live as methods hanging off the message.

Therefore, we need to add direct QueueClient binding capabilities, similar to the way our Azure Storage Queue binding allows you to bind to a CloudQueue:

```csharp
public static void ProcessWorkItem(
    [QueueTrigger("test")] WorkItem workItem,
    [Queue("test")] CloudQueue queue)
{
    Console.WriteLine($"Processed work item {workItem.ID}");
}
```

For ServiceBus that would then be:

```csharp
public static async Task ProcessWorkItem(
    [ServiceBusTrigger("test-items")] WorkItem item,
    [ServiceBus("test-items")] QueueClient queueClient,
    string messageId
    ILogger log)
{
    log.LogInformation($"Processing ServiceBus message (Id={messageId})");
}
```

In addition to allowing binding to a queue using the attribute to bind to an arbitrary queue, we should also support a direct binding to the queue for the triggering message:

```charp
public static async Task ProcessWorkItem(
    [ServiceBusTrigger("test-items")] Message message,
    QueueClient queueClient,
    string messageId
    ILogger log)
{
    log.LogInformation($"Processing ServiceBus message (Id={messageId})");
}
```

ryancrawcour:
Agree that the change from BrokeredMessage to Message and losing access to methods like Abandon(), Complete(), Deadletter() has caused some problems with some current scenarios. So putting this back in would be awesome!

ryancrawcour:
The docs for [QueueClient ](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.servicebus.queueclient?view=azure-dotnet) say to use [MessageSender ](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.servicebus.core.messagesender?view=azure-dotnet)or [MessageReceiver ](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.servicebus.core.messagesender?view=azure-dotnet) for advanced set of functionality. 

Would it make sense to (also) expose access to MessageReceiver and MessageSender?

ryancrawcour:
QueueClient.CompleteAsync method (and others like AbandonAsync, DeadletterAsync etc.) require a locktoken … 

[...]

PR #1753, opened 2018-06-14
ServiceBus MessageReceiver/MessageSender bindings (#1743)

mathewc:
Addresses https://github.com/Azure/azure-webjobs-sdk/issues/1743

Adding:

- new ServiceBusAttribute binding for MessageSender. This allows a user to bind to a sender if they want to do the sending themselves. Similar to the Azure Storage CloudQueue binding, and other bindings that allow you to bind to a service client
- new binding data direct bind capability for lockToken and triggering MessageReceiver. Using these two values, a function can perform direct operations on the message (e.g. renew lease, abandon, etc.)

The end to end test shows example functions using these new bindings.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.ServiceBus/Bindings/MessageSenderArgumentBindingProvider.cs
+++ b/src/Microsoft.Azure.WebJobs.ServiceBus/Bindings/MessageSenderArgumentBindingProvider.cs
@@ -1,0 +1,80 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.ServiceBus.Core;
+using Microsoft.Azure.WebJobs.Host.Bindings;
+
+namespace Microsoft.Azure.WebJobs.ServiceBus.Bindings
+{
+    internal class MessageSenderArgumentBindingProvider : IQueueArgumentBindingProvider
+    {
+        public IArgumentBinding<ServiceBusEntity> TryCreate(ParameterInfo parameter)
+        {
+            if (parameter.ParameterType != typeof(MessageSender))
+            {
+                return null;
+            }
+
+            return new MessageSenderArgumentBinding();
+        }
+
+        internal class MessageSenderArgumentBinding : IArgumentBinding<ServiceBusEntity>
+        {
+            public Type ValueType
+            {
+                get { return typeof(MessageSender); }
+            }
+
+            public Task<IValueProvider> BindAsync(ServiceBusEntity value, ValueBindingContext context)
+            {
+                if (context == null)
+                {
+                    throw new ArgumentNullException("context");
+                }
+
+                IValueProvider provider = new MessageSenderValueBinder(value.MessageSender);
+
+                return Task.FromResult(provider);
+            }
+
+            private class MessageSenderValueBinder : IValueBinder
+            {
+                private readonly MessageSender _messageSender;
+
+                public MessageSenderValueBinder(MessageSender messageSender)
+                {
+                    _messageSender = messageSender;
+                }
+
+                public BindStepOrder StepOrder
+                {
+                    get { return BindStepOrder.Enqueue; }
+                }
+
+                public Type Type
+                {
+                    get { return typeof(MessageSender); }
+                }
+
+                public Task<object> GetValueAsync()
+                {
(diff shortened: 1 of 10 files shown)


src/Microsoft.Azure.WebJobs.ServiceBus/Bindings/ServiceBusBinding.cs
```diff
@@ -41,7 +42,7 @@ public bool FromAttribute
             context.CancellationToken.ThrowIfCancellationRequested();
 
             string boundQueueName = _path.Bind(context.BindingData);
-            var messageSender = new MessageSender(_account.ConnectionString, boundQueueName);
+            var messageSender = _messagingProvider.CreateMessageSender(boundQueueName, _account.ConnectionString);
```
mathewc: Now routing creation of senders through the factory, as we do for receivers. This was an oversight when the factory was introduced originally.

src/Microsoft.Azure.WebJobs.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -35,6 +35,8 @@ public ServiceBusListener(string entityPath, ServiceBusTriggerExecutor triggerEx
             _messageProcessor = config.MessagingProvider.CreateMessageProcessor(entityPath, _serviceBusAccount.ConnectionString);
         }
 
+        internal MessageReceiver Receiver => _receiver;
```
mathewc: Done to allow the trigger binding to access the receiver, so it can add it to the binding data collection

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ServiceBusEndToEndTests.cs
```diff
@@ -321,12 +326,20 @@ protected static void SBTopicListener2Impl(Message message)
...
+                MessageReceiver messageReceiver,
+                string lockToken,
+                [ServiceBus(SecondQueueName)] MessageSender messageSender)
             {
+                Assert.Equal(FirstQueueName, messageReceiver.Path);
                 Assert.Equal(1, deliveryCount);
-                message = SBQueue2SBQueue_GetOutputMessage(start);
+
+                // verify the message receiver and token are valid
+                await messageReceiver.RenewLockAsync(lockToken);
```
mathewc: This was the goal - allow direct operations on the triggering message.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ServiceBusEndToEndTests.cs
```diff
@@ -321,12 +326,20 @@ protected static void SBTopicListener2Impl(Message message)
         public class ServiceBusTestJobs : ServiceBusTestJobsBase
         {
             // Passes service bus message from a queue to another queue
-            public static void SBQueue2SBQueue(
+            public static async Task SBQueue2SBQueue(
                 [ServiceBusTrigger(FirstQueueName)] string start, int deliveryCount,
-                [ServiceBus(SecondQueueName)] out string message)
+                MessageReceiver messageReceiver,
+                string lockToken,
+                [ServiceBus(SecondQueueName)] MessageSender messageSender)
```
mathewc: Use of the new sender binding.

mathewc:
We have existing tests that verify that the ServiceBus binding works for both Queues and Topics. They are treated the same internally - both end up using a MessageSender with the entity path being the queue or topic.

alrod (approved):

src/Microsoft.Azure.WebJobs.ServiceBus/Bindings/MessageSenderArgumentBindingProvider.cs
```diff
@@ -0,0 +1,80 @@
...
+using System;
+using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.ServiceBus.Core;
+using Microsoft.Azure.WebJobs.Host.Bindings;
+
+namespace Microsoft.Azure.WebJobs.ServiceBus.Bindings
+{
+    internal class MessageSenderArgumentBindingProvider : IQueueArgumentBindingProvider
```
MikeStall: Could this be done with the v2 converter interfaces, rather than the v1 argumentbinder ?

mathewc: The entire binding needs to be ported to new model. That work is orthogonal to this fix.

PR #1754, opened 2018-06-14
Fix LogWriter memory leak and resiliency (webjobs V3.x)

paulbatum:
Resolves https://github.com/Azure/azure-functions-host/issues/2690

There's two main changes here:
1. When the background flusher hits an error, clear out the task so that it is resumed later (otherwise no more background flushes occur).
2. Set a upper limit on how many entries will be buffered so that if there are errors in flush for a long period of time (say a storage key was rolled over), there is an upper bound on how much memory will be consumed.

webjobs v2.x port is here:
https://github.com/Azure/azure-webjobs-sdk/pull/1762

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging/Internal/LogWriter.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging/Internal/LogWriter.cs
@@ -19,8 +19,7 @@ namespace Microsoft.Azure.WebJobs.Logging
     {
         // Logs from AddAsync() are batched up. They can be explicitly flushed via FlushAsync() and 
         // they get autotmatically flushed at Interval. 
-        // Calling AddAsync() will startup the background flusher. Calling FlushAsync() explicitly will disable it. 
-        private static TimeSpan _flushInterval = TimeSpan.FromSeconds(45);
+        // Calling AddAsync() will startup the background flusher. Calling FlushAsync() explicitly will disable it.
         private CancellationTokenSource _cancelBackgroundFlusher = null;
         private Task _backgroundFlusherTask = null;
 
@@ -50,6 +49,9 @@ namespace Microsoft.Azure.WebJobs.Logging
         private CloudTableInstanceCountLogger _instanceLogger;
 
         private Action<Exception> _onException;
+
+        public int MaxBufferedEntryCount { get; set; } = 10000;
+        public TimeSpan FlushInterval { get; set; } = TimeSpan.FromSeconds(45);
 
         public LogWriter(string hostName, string machineName, ILogTableProvider logTableProvider, Action<Exception> onException = null)
         {
@@ -81,15 +83,17 @@ namespace Microsoft.Azure.WebJobs.Logging
             {
                 try
                 {
-                    await Task.Delay(_flushInterval, cancellationToken);
+                    await Task.Delay(FlushInterval, cancellationToken);
+                    await this.FlushCoreAsync();
                 }
-                catch (OperationCanceledException)
+                catch
                 {
-                    // Don't return yet. One last chance to flush 
-                    return;
+                    lock (_lock)
+                    {
+                        _backgroundFlusherTask = null;
+                        return;
+                    }
                 }
-
-                await this.FlushCoreAsync();
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.WebJobs.Logging/Internal/LogWriter.cs
```diff
@@ -152,6 +156,15 @@ public Task AddAsync(FunctionInstanceLogItem item, CancellationToken cancellatio
             // Both Start and Completed log here. Completed will overwrite a Start entry. 
             lock (_lock)
             {
+                // Permanent failures when flushing to storage can result in many log entries being buffered.
+                // This basically becomes a memory leak. Mitigate by enforcing a max.
+                if (_activeFuncs.Count >= MaxBufferedEntryCount || _completedFunctions.Count >= MaxBufferedEntryCount)
+                {
+                    _onException?.Invoke(new Exception($"The limit on the number of bufferable log entries was reached. A total of '{MaxBufferedEntryCount}' log entries were dropped."));
```
brettsam: nit: is 'bufferable' a word? :-) Maybe 'queued'?

src/Microsoft.Azure.WebJobs.Logging/Internal/LogWriter.cs
```diff
@@ -152,6 +156,15 @@ public Task AddAsync(FunctionInstanceLogItem item, CancellationToken cancellatio
             // Both Start and Completed log here. Completed will overwrite a Start entry. 
             lock (_lock)
             {
+                // Permanent failures when flushing to storage can result in many log entries being buffered.
+                // This basically becomes a memory leak. Mitigate by enforcing a max.
+                if (_activeFuncs.Count >= MaxBufferedEntryCount || _completedFunctions.Count >= MaxBufferedEntryCount)
```
brettsam: Should this be comparing `_activeFuncs.Count + _completedFunctions.Count >= MaxBufferedEntryCount`? Otherwise it seems like you could approach 2x that number of queued functions.

src/Microsoft.Azure.WebJobs.Logging/Internal/LogWriter.cs
brettsam: Also -- that total number dropped may not be accurate. Can you use the actual counts for the message?

src/Microsoft.Azure.WebJobs.Logging/Internal/LogWriter.cs
```diff
@@ -237,6 +250,9 @@ public Task AddAsync(FunctionInstanceLogItem item, CancellationToken cancellatio
                     }
                 }
 
+                // Just making a note while I'm in this code fixing something else. 
+                // It looks like this will drop data if there is any type of error when communicating with storage. 
+                // Other code paths drop the data after successfully writing to storage
```
brettsam: Can you file an issue so we can track it?

src/Microsoft.Azure.WebJobs.Logging/Internal/LogWriter.cs
paulbatum: I had it this way originally and changed it because a given log entry is often written to both of these, which made things confusing because you would hit the buffer limit twice as fast as expected. I think its more intuitive this way.

src/Microsoft.Azure.WebJobs.Logging/Internal/LogWriter.cs
paulbatum: Fixed the non-word :)

See my comment below about how a log entry is often written to both collections. If I show in the error message the active+completed count it will probably be misleading. I'll keep this part as is.

src/Microsoft.Azure.WebJobs.Logging/Internal/LogWriter.cs
paulbatum: Filed https://github.com/Azure/azure-webjobs-sdk/issues/1761

paulbatum:
I've done an end to end test and confirmed that:

1. the logs appear in kusto:

image

2. The truncation effectively limits the amount of leakable memory. Here's a comparison between two dumps after running for 30 minutes in a state where log entries can't be written (triggered by a storage key rollover):

image

PR #1764, opened 2018-06-21
re-enabling ScanBlobScanLogHybridPollingStrategyTests

brettsam:


test/FakeStorage/Blob.cs
```diff
@@ -1,145 +0,0 @@
...
-using Microsoft.WindowsAzure.Storage.Shared.Protocol;
-using Microsoft.WindowsAzure.Storage.Table;
-using System;
-using System.Collections.Generic;
-using System.Threading;
-using System.Threading.Tasks;
-
-namespace FakeStorage
-{
-    public class FakeBlobClient : CloudBlobClient
```
MikeStall: Where'd this go?

brettsam: It's not used by anything -- we have FakeStorageBlobClient now.

PR #1765, opened 2018-06-21
Fix test layering

MikeStall:
Dll layering: Host.UnitTests and TestCommon should not depend on Storage/ServiceBus.
Move storage tests from Unit into Functional.

No product change.


test/Microsoft.Azure.WebJobs.Host.EndToEndTests/TestExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.WebJobs.Host.EndToEndTests
+{
+    static class TestExtensions
+    {
+        public static XStorageAccount GetStorageAccount(this IHost host)
+        {
+            var provider = host.Services.GetRequiredService<StorageAccountProvider>(); // $$$ ok?
```
brettsam: You can probably remove the $$$ here -- this should be okay.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncCancellationEndToEndTests.cs
```diff
@@ -37,6 +37,7 @@ public AsyncCancellationEndToEndTests()
 
             _host = new HostBuilder()
                 .ConfigureDefaultTestHost<AsyncCancellationEndToEndTests>()
+                .AddAzureStorage()
```
mathewc: ConfigureDefaultTestHost already has a call to AddAzureStorage inside of it. Why is this needed everywhere?

MikeStall:
The most interesting change here is the project file. ;)   The code changes are either moving whole tests or splitting test classes.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/TestExtensions.cs
mathewc: ?

MikeStall: Just moved from somewhere else.  We still need to chase down the tail of $$$ todo markers.

test/Microsoft.Azure.WebJobs.Host.TestCommon/TestHelpers.cs
```diff
@@ -121,8 +121,7 @@ public static IHostBuilder ConfigureDefaultTestHost(this IHostBuilder builder, p
                    // Register this to fail a test if a background exception is thrown
                    services.AddSingleton<IWebJobsExceptionHandlerFactory, TestExceptionHandlerFactory>();
                })
-               .ConfigureTestLogger()
-               .AddAzureStorage();
+               .ConfigureTestLogger();
```
brettsam: Nice, I was going to suggest this as well -- it's easy enough to call `AddAzureStorage()`  when you need it, but hard for tests to remove it.

test/Microsoft.Azure.Webjobs.Extensions.Storage.UnitTests/Extensions.cs
```diff
@@ -9,12 +9,35 @@
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Microsoft.WindowsAzure.Storage.Blob;
+using Microsoft.WindowsAzure.Storage.Queue;
 using Microsoft.WindowsAzure.Storage.Table;
 
 namespace Microsoft.Azure.WebJobs
 {
     static class MoreStorageExtensions
```
brettsam: I assume there's a `StorageExtensions` somewhere? Can these just go there? :-)

Issue #1731, opened 2018-06-01
Distributed tracing in Azure Functions


lmolkova:
I'll go through the planned changes (split by several PRs) and describe high-level design and expected outcome. 

Here is a messy prototype for all changes: https://github.com/lmolkova/azure-webjobs-sdk/commit/22a7affce96f0c34363b3746d0af5797a1c1c593

**I'm open to any feedback and suggestions and want to make sure we are on the same page before I start the implementation**

# Motivation 
We want to enable distributed tracing for Azure Functions using ApplicationInsights users are able to track calls to and from the Azure Functions and correlate it to the rest of telemetry reported by other applications.

As a result, we want to enable experiences like [End-to-end transaction viewer](https://docs.microsoft.com/en-us/azure/application-insights/app-insights-transaction-diagnostics#transaction-diagnostics-experience) and [Application Map](https://docs.microsoft.com/en-us/azure/application-insights/app-insights-app-map) for all users automagically.

## Goal 1: supported auto-tracking (triggers only are tracked now)

ApplicationInsights can track HTTP, ServiceBus, EventHub, SQL calls automagically if configured so. It's a component responsibility to participate in distributed tracing, i.e. when new components (e.g. ServiceGrid) implement specific instrumentation, they just light up - requiring no changes from Azure Functions.

Enabling current tracking capabilities gives us following level of support:

| Type | 1.x | 2.x | Trigger | Input | Output | Calls Tracking |
|------|-----|-----|---------|-------|--------|----------|
| Blob Storage | ✔ | ✔ | ✔ | ✔ | ✔| ✔ (Http) |
| Cosmos DB | ✔ | ✔ | ✔ | ✔ | ✔| ✔ (Http - Azure Functions use deafult configuration, i.e. all communication is Http based)|
| Event Grid | ✔ | ✔ | ✔ |  |  | N/A (trigger only) |
| Event Hubs | ✔ | ✔ | ✔ | | ✔ | ✔| ✔|
| External File |✔ | | | ✔| ✔ | :x: |
| External Table |✔ | |  |✔ | ✔ | :x: |
| Http | ✔ | ✔ | ✔ |  | ✔| ✔ |
| Microsoft Graph Excel tables | | ✔ | | ✔ | ✔ | ✔ (Http) | 
| Microsoft Graph OneDrive files | | ✔ | | ✔ | ✔ | ✔ (Http) | 
[...]

PR #1767, opened 2018-06-21
Distributed tracing: Correlation and Http dependency collection

lmolkova:
This change introduces:

1. Http dependency collection for bindings and user code (.NET only).
When http call is done by the host process, it is tracked by ApplicaitonInsights.
If it happens to be in the scope of bindings or user code, it is tracked as a dependency.

2. Correlation improvements: all telemetry is correlated: requests, binding calls, user code (.NET).
Correlation is achieved using .NET tracing context `System.Diagnostics.Activity` that flows with async calls. ApplicationInsights uses it internally to correlate all telemetry.
In this change, all control is given to AI to set telemetry OperationId, ParentId and Id properly.

In the next PRs I'm going to start receiving context with the external calls.

3. Some DI improvements:
`TelemetryConfiguration.Active` static singleton is the best (not ideal though) way Functions users can use AI SDK explicitly while sharing the same configuration with host. This change configures Active.

See #1731. This change corresponds to PR2 (limited to HTTP tracking only for now).

@brettsam @fabiocav @paulbatum @SergeyKanzhelev


lmolkova:
ping @brettsam @fabiocav @paulbatum :)

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsHostBuilderExtensions.cs
```diff
@@ -57,52 +60,57 @@ public static IHostBuilder AddApplicationInsights(this IHostBuilder builder, str
...
+                        channel,
+                        provider.GetServices<ITelemetryInitializer>(),
+                        provider.GetServices<ITelemetryModule>());
 
-                    if (samplingSettings != null)
+                    // Function users have no access to TelemetryConfiguration from host DI container,
+                    // so we'll expect user to work with TelemteryConfiguration.Active
+                    // Also, some ApplicaitonInsights internal operations (heartbeats) depend on
+                    // the TelemetryConfiguration.Active being set so, we'll set up Active once per process lifetime.
+                    TelemetryConfiguration activeConfig = TelemetryConfiguration.Active;
```
brettsam: Customers could change this config from within their code, right? Would that affect anything critical?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsHostBuilderExtensions.cs
```diff
@@ -57,52 +60,57 @@ public static IHostBuilder AddApplicationInsights(this IHostBuilder builder, str
...
+                    if (string.IsNullOrEmpty(activeConfig.InstrumentationKey))
                     {
-                        config.TelemetryProcessorChainBuilder.Use((next) =>
-                            new AdaptiveSamplingTelemetryProcessor(samplingSettings, null, next));
+                        SetupTelemetryConfiguration(
+                            activeConfig,
+                            instrumentationKey,
+                            filter,
+                            samplingSettings,
+                            new ServerTelemetryChannel(),
```
brettsam: Is this meant to get a new instance of the channel? Or can it use the same as the other config?

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/ApplicationInsightsEndToEndTests.cs
```diff
@@ -501,11 +568,17 @@ private static SeverityLevel GetSeverityLevel(LogLevel logLevel)
             }
         }
 
-        private static void ValidateException(ExceptionTelemetry telemetry, string expectedCategory, string expectedOperationName)
```
brettsam: Nice test!

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/ApplicationInsightsEndToEndTests.cs
```diff
@@ -103,26 +108,20 @@ private IHost ConfigureHost(LogCategoryFilter filter = null)
                 string expectedFunctionUserCategory = LogCategories.CreateFunctionUserCategory(testName);
```
brettsam: Won't AddApplicationInsights initialize this for us now?

brettsam: Won't `AddApplicationInsights()` initialize this for us now?

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/HttpDependencyCollectionTests.cs
```diff
@@ -0,0 +1,417 @@
...
+        private CloudQueue _triggerQueue;
+
+        [Fact]
+        public async Task BindingsAreNotReportedWhenDisabled()
+        {
+            using (var host = ConfigureHost(LogLevel.Warning))
+            {
+                await host.StartAsync();
+                await host.GetJobHost()
+                    .CallAsync(typeof(HttpDependencyCollectionTests).GetMethod("BlobInputAndOutputBindings"));
```
brettsam: nit: Can you use `nameof()` for these method names? It makes navigating easier when debugging in the future (we're slowly trying to change them all as we see them).

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/HttpDependencyCollectionTests.cs
```diff
@@ -0,0 +1,417 @@
...
+        private string _outputQueueName;
+        private string _triggerQueueName;
+
+        private CloudBlobClient _blobClient;
+        private RandomNameResolver _resolver;
+        private CloudQueueClient _queueClient;
+        private CloudQueue _triggerQueue;
+
+        [Fact]
+        public async Task BindingsAreNotReportedWhenDisabled()
```
brettsam: nit: maybe change this test name to `...WhenFiltered()`?

test/Microsoft.Azure.WebJobs.Host.UnitTests/Loggers/ApplicationInsightsConfigurationTests.cs
```diff
@@ -78,5 +82,24 @@ public void DepednencyInjectionConfiguration_ConfiguresSampling()
...
+                var config = host.Services.GetService<TelemetryConfiguration>();
+                Assert.Equal(4, config.TelemetryProcessors.Count);
+                Assert.IsType<QuickPulseTelemetryProcessor>(config.TelemetryProcessors[0]);
+                Assert.IsType<FilteringTelemetryProcessor>(config.TelemetryProcessors[1]);
+                Assert.IsType<AdaptiveSamplingTelemetryProcessor>(config.TelemetryProcessors[2]);
+
+                Assert.Equal(samplingSettings.MaxTelemetryItemsPerSecond, ((AdaptiveSamplingTelemetryProcessor)config.TelemetryProcessors[2]).MaxTelemetryItemsPerSecond);
+
+            }
+        }
```
brettsam: Can you add a test (or even a validation step in one of these tests) that shows `TelemetryConfiguration.Active` is set up as expected?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/FilteringTelemetryProcessor.cs
```diff
@@ -41,6 +42,11 @@ private bool IsEnabled(ITelemetry item)
                 {
                     // If no category is specified, it will be filtered by the default filter
                     categoryName = string.Empty;
+
+                    if (telemetry is OperationTelemetry)
+                    {
+                        enabled = false;
+                    }
```
brettsam: This is where the "magic" happens, right? We'll only log OperationTelemetry if it has a category attached -- and that will only happen when wrapped in our scope.

It may be worth putting a comment here for others to briefly explain how this works.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsHostBuilderExtensions.cs
lmolkova: 1. WebJobs SDK customers:
We'll need to update documentation and encourage them to modify telemetry configuration through the dependency injection. They can do all kinds of harmful things  - remove initializers, processors and break things.
The  good thing - we can refer to existing ApplicationInsights documentation on how to do it correctly without breaking things.

2.  Azure Function customers:
 They do not have access to DI and only have access to TelemetryConfiguration.Active
They may change Active config, but it would not affect the instance host uses for logging. I.e. changes would only reflect telemetry that is reported directly from the user code.
Active is needed to just support the scenario when telemetry is reported from the user code rather than modifications.
I think the only appropriate way to modify host configuration is still through the settings file and is quite limited for functions users.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsHostBuilderExtensions.cs
lmolkova: This meant to have a new ServerTelemetryChannel. As host could be disposed, it will cause TelemtryConfiguration disposal and the underlying channel will also be disposed. 
If channel is reused TelemetryConfiguration.Active becomes broken after host dies, so it has to have another instance which lifetime is process lifetime.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/FilteringTelemetryProcessor.cs
lmolkova: Exactly, this is the place that filters unwanted calls. I'll add a comment, good point.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/ApplicationInsightsEndToEndTests.cs
lmolkova: Perhaps I don't understand the question: we're trying to validate if the category is set on the telemetry. ApplicationInsights has no control over which categories are assigned. So, could you please elaborate?

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/ApplicationInsightsEndToEndTests.cs
lmolkova: thanks!

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/HttpDependencyCollectionTests.cs
lmolkova: 👍

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/HttpDependencyCollectionTests.cs
lmolkova: 👍

test/Microsoft.Azure.WebJobs.Host.UnitTests/Loggers/ApplicationInsightsConfigurationTests.cs
lmolkova: sure

lmolkova:
@brettsam Thanks for the review!

 I wonder why the build fails  with 

`MSBUILD : error MSB1011: Specify which project or solution file to use because this folder contains more than one project or solution file.`

It does not look like something related to my changes, but apparently, other PRs fails for other reasons.

fabiocav:
@lmolkova the PR was failing because the wrong build job was being triggered. We have different build jobs for members of the team (with end-to-end tests) and the simple job seems to be broken. I've updated the PR with a label that will trigger a full build moving forward

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsHostBuilderExtensions.cs
brettsam: Great, this all makes sense. It's still a step in the right direction to allow TelemetryConfiguration.Active have the same setup as our client uses. 

Another option I'd considered was having a TelemetryClient binding where we'd pass you the one we use in DI directly.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/ApplicationInsightsEndToEndTests.cs
brettsam: Hmm, it's sticking my comment on the wrong line somehow (I used VS tools to make my comments -- something must have been screwed up). I'll re-make the comment on the right line...

brettsam: My comment was meant to be around `SetActiveTelemetryConfiguration(host.Services.GetService<TelemetryConfiguration>());`, but it looks like you cleaned that up, which was exactly what I was asking for.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsHostBuilderExtensions.cs
lmolkova: > Another option I'd considered was having a TelemetryClient binding where we'd pass you the one we use in DI directly.

This would be awesome!

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/ApplicationInsightsEndToEndTests.cs
lmolkova: yeah, Active lifetime is more than one test, so we have to clean it up after each run. I corrected it and moved to Dispose method

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsHostBuilderExtensions.cs
```diff
@@ -28,10 +30,11 @@ public static class ApplicationInsightsHostBuilderExtensions
         /// </summary>
         /// <param name="builder">The host builder.</param>
         /// <param name="instrumentationKey">The Application Insights instrumentation key.</param>
+        /// <param name="samplingSettings">The <see cref="SamplingPercentageEstimatorSettings"/> to use for configuring adaptive sampling. If null, sampling is disabled.</param>
         /// <returns>A <see cref="IHostBuilder"/> for chaining additional operations.</returns>
         public static IHostBuilder AddApplicationInsights(this IHostBuilder builder, string instrumentationKey, SamplingPercentageEstimatorSettings samplingSettings)
```
fabiocav: This is a nit, and I realize this pattern was already here prior to this PR,  but can we have an overload without the `samplingSettings` argument (and ideally, validate that it isn't `null` in the overload that takes the parameter) since that is optional, and not just rely on `null` being passed? That will make it an explicit choice and a clearer API and avoid issues if a `null` reference is unintentionally passed. The

lmolkova: Sure, I will do this. But I won't have time today, so please wait till tomorrow.

Thanks for the review!

fabiocav: No need to block on this, as it could come later if there's any urgency on getting this in.

Thanks again for sending this!

brettsam:
@lmolkova -- we've added you to azure-functions-team so your PRs should get the full build treatment from now on, without the need for the special tag.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsHostBuilderExtensions.cs
brettsam: @fabiocav -- these settings are exposed so that Script can pass in settings via host.json. Should we try to move this to using `IOptions<>` so we can bind directly? Having (only) these settings exposed seems odd now.

Doesn't have to be done here, but we can file an issue to clean up in a later PR...

lmolkova: Hey, I just created an issue for this: https://github.com/Azure/azure-webjobs-sdk/issues/1780.

So it seems `AddApplicationInsights` would eventually have IConfiguration/IOptions overload and some set of default parameters. 
For now, I believe having extra overloads that set default parameters is somewhat useless, so I'd rather remove it completely.

what do you think?

JasonBSteele:
Hi, I'm struggling to work out the current state of automatic dependency tracking and correlation id propagation within V2 Functions.

Does it just work now? Or do we still need to do it manually through explicit AI methods? 
What has been implemented - it sounds like this PR is about Trigger bindings, but what about using HttpClient within an Function to call another Function?  Does that magically propagate the Correlation Id and log the dependency?

Thanks

brettsam:
Making HTTP calls from within a function is now tracked as a Dependency in App Insights (i.e. you'll get response code, duration, etc, logged for each outbound request you make from within a function). This includes using output bindings for HTTP-based services like Storage and Service Bus.

Correlation across calls is not yet implemented -- other than some specific cases involving Service Bus, which has implemented the parent-id tracking for these calls. We're working with App Insights to get all of this working across Functions calls, and also talking about how to make this work nicely with Durable Functions.

If you have a specific scenario, let me know -- I want to beef up our docs and wiki with specific scenarios so you can have a better view of what works today.

JasonBSteele:
@brettsam, many thanks for the update. We are currently not seeing CorrelationId of ServiceBus messages getting set automatically, but it sounds like that should be working?

There is no specific example, but a table telling us which calls from resource to resource will be captured as dependencies and which propagate the correlation id would be really useful. Especially if it is kept up to date and shows what's coming next. Thanks

brettsam:
Can you share how you're creating and sending your SB messages?

JasonBSteele:
Hopefully this shows enough. We find that we have to explicitly set the CorrelationId:

```
 public class QueueClientEnqueuer : IQueueClientEnqueuer
    {
        private readonly IJsonOperations _jsonOperations;
        private readonly Lazy<IQueueClient> _processorQueueClient;

        public QueueClientEnqueuer(IJsonOperations jsonOperations, Lazy<IQueueClient> processorQueueClient)
        {
            _jsonOperations = jsonOperations;
            _processorQueueClient = processorQueueClient;
        }

        public Task EnqueueAsync(ExportContext argument, string correlationId)
        {
            if (string.IsNullOrWhiteSpace(correlationId))
            {
                throw new ArgumentException(nameof(correlationId));
            }

            var brokeredMessage = CreateMessage(argument, correlationId);
            return _processorQueueClient.Value.SendAsync(brokeredMessage);
        }

        private Message CreateMessage(ExportContext argument, string correlationId)
        {
            var stream = _jsonOperations.SerializeToMemoryStream(argument);
            return new Message(stream.ToArray())
            {
                ContentType = "application/json",
                MessageId = Guid.NewGuid().ToString(),
                CorrelationId = correlationId
            };
        }
    }
```

lmolkova:
@JasonBSteele CorrelationId in queues serves a different purpose - it's not for logging/tracing. What ServiceBus uses is called 'Diagnostic-Id' in user properties. When producer is instrumented with AppInsights (or another tracing system), the unique id is sent in this property. 

When the consumer (function) received the message and finds Diagnostics-Id it uses it as a parent and populates telemetry accordingly.

Here is more info on how this works: https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-end-to-end-tracing

PR #1768, opened 2018-06-21
Fix sampling settings AddApplicationInsights overload

lmolkova:
Fixes typo: sampling settings were not passed

@zhencui, thanks for noticing it!


lmolkova:
fixed in #1767

PR #1771, opened 2018-06-25
enabling DispatchQueue tests

brettsam:
Some notable changes here:
- We no longer fall back to an in-memory dispatch queue implementation if the given implementation fails to initialize (removed `InMemoryDispatchQueueHandler`) -- it was introducing too many layers. Now, with DI, if the registered `ILoadBalancerQueue` implementation fails during initialization, we'll delay-throw that when someone tries to use it. The default is still in-memory... calling `AddAzureStorage()` adds the storage implementation.
- I used DataFlow for the `InMemoryLoadBalancerQueue` and made sure we had a test for it.
- Changed capitalization to `ILoadBalancerQueue`
- Cleaned up some warnings.


fabiocav:
Closing this stale PR. Please rebase/reopen if still applicable.

PR #1775, opened 2018-06-26
adding Logging.FunctionalTests to build

brettsam:
After a bunch of debugging, I enabled fusion logs on the Appveyor VM and saw that we weren't skipping strong name validation on this assembly. Tracked it back to the SkipStrongNames.xml file. Now all should be building and running.


brettsam:
@fabiocav -- here's that Logging.Functional fix I mentioned (along with QuickPulse fixes)

PR #1781, opened 2018-06-28
Enable Application Insights heartbeats

lmolkova:
This change adds ApplicaitonInsights heartbeats.

Heartbeat is a simple custom event, sent to customer ApplicationInsights resource every 15 minutes with infrastructure information such as OS,  runtime version, and app-service details: service name, stamp, owner and host.

this information 
1. allows navigating from Application Map directly to Function/App Service resource 
2. measure application insights usage

/cc @fabiocav  @brettsam


lmolkova:
@brettsam @fabiocav could you guys please have a look when you have time? Thanks!

Issue #1752, opened 2018-06-14
Don't checkpoint when Event Hubs Listener is closing


jeffhollan:
Here we do a checkpoint if an EventHubListener is closing [because of a server shutdown](https://docs.microsoft.com/en-us/dotnet/api/microsoft.servicebus.messaging.closereason?view=azure-dotnet)
https://github.com/Azure/azure-webjobs-sdk/blob/700947786067834c775bce588483c4560c1eb0ca/src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubListener.cs#L118-L121

I'm not sure why we would want to checkpoint in this case.  Since this appears to be invokable in parallel to an execution that may be in progress (and not yet completed), seems it could result in an invalid checkpoint for in flight data.

Seems we shouldn't have these lines.

paulbatum:
I agree this looks like an issue. Here's an example sequence:

1. Functions are in the process of executing.
2. For one of several reasons, we decide we need to start a graceful shutdown of the host. This calls `StopAsync` on the listener.
3. Stopping the listener triggers a call to CloseAsync on each partition listener, which cancels the cancellation token passed into user code.
4. User code aborts the execution (assuming they are handling the token cancellation). Their functions did not finish running
5. The line that Jeff highlighted checkpoints anyway, and so the system "loses" the data from the aborted executions.

@mathewc Can you see any mistakes in the above reasoning?

mathewc:
Agreed - this seems like an issue. Recommend we create a simple repro that demonstrates the problem that we can use to verify the fix.

mathewc:
Possibly related: ICM 73233348

jeffhollan:
Thanks for looking into this. Yes that ICM is what first made me inspect the code and find a potential gap. I don't think necessarily this is what they were hitting but is a possibility as it is clear during the logs of their incident during data loss was a transition of partition listeners between instances (one instance took a partition lock away from another) - so possible is related

PR #1783, opened 2018-06-29
Fixes for EventHubTrigger checkpointing (#1752)

mathewc:
Fix for https://github.com/Azure/azure-webjobs-sdk/issues/1752.

I'll be making the same changes to the dev branch.


src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubListener.cs
```diff
@@ -12,180 +12,204 @@
 
 namespace Microsoft.Azure.WebJobs.ServiceBus
 {
-    // Created from the EventHubTrigger attribute to listen on the EventHub.
```
mathewc: I reworked this entire file quite a bit to make it easier to understand and also to enable me to improve the test coverage.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubListener.cs
```diff
@@ -12,180 +12,204 @@
...
-        
-        // This will get called per-partition. 
+         
         IEventProcessor IEventProcessorFactory.CreateEventProcessor(PartitionContext context)
         {
-            return new Listener(this);
+            return new EventProcessor(_config, new Checkpointer(), _executor, _singleDispatch);
         }
 
-        internal static Func<Func<Task>, Task> CreateCheckpointStrategy(int batchCheckpointFrequency)
```
mathewc: This javascript-like pattern and the way it was taking a closure was hard to understand. Simplifying it and making things more testable

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubListener.cs
```diff
@@ -12,180 +12,204 @@
...
+            public Task CloseAsync(PartitionContext context, CloseReason reason)
             {
-                this._cts.Cancel(); // Signal interuption to ProcessEventsAsync()
+                // signal cancellation for any in progress executions 
+                _cts.Cancel();
 
-                // Finish listener
-                if (reason == CloseReason.Shutdown)
-                {
-                    await context.CheckpointAsync();
```
mathewc: This was the bug fix - we should never have been checkpointing here

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubListener.cs
```diff
@@ -12,180 +12,204 @@
...
                     {
-                        await Task.WhenAll(dispatches);
+                        await Task.WhenAll(invocationTasks);
                     }
+
+                    // Only consider this success if ALL events in the batch were processed
+                    // successfully. If we were cancelled in the middle of the batch and
+                    // haven't processed all the events, that is also failure and we shouldn't
+                    // checkpoint.
+                    int successCount = invocationTasks.Count(p => p.Result.Succeeded);
```
mathewc: This was the second big issue I noticed - we were checkpointing the batch without ever checking the FunctionResult! I.e. even if an invocation failed we'd checkpoint. We should only checkpoint if the invocation(s) were successful.

paulbatum (approved):
Only had time for a quick read but this looks good. Would feel better if you can get one more sign off (maybe Brett).

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubListener.cs
```diff
@@ -12,180 +12,198 @@
...
                 {
-                    await _checkpoint(context);
+                    await CheckpointAsync(context);
                 }
             }
-        } // end class Listener 
+
+            private async Task CheckpointAsync(PartitionContext context)
+            {
+                if (_config.BatchCheckpointFrequency == 1)
```
fabiocav: We're losing some validation we had with the previous implementation to ensure `BatchCheckpointFrequency` wasn't set to 0.

fabiocav: Not a huge deal, but this is also slightly less efficient.

fabiocav (approved):
Small comment. Looks good.

src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubListener.cs
mathewc: I removed the validation because if you look at the setter for EventHubConfiguration it already has the validation - it can't be set to an invalid value.

Regarding inefficiency - you're talking about the extra property access? I'll fix that.

PR #1787, opened 2018-07-02
Fix FunctionExecutor null ref

mathewc:
Fix for https://github.com/Azure/azure-functions-host/issues/2785

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
@@ -111,7 +111,6 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                     };
 
                     exceptionInfo = ExceptionDispatchInfo.Capture(exception);
-
                     exceptionInfo = await InvokeExceptionFiltersAsync(parameterHelper.JobInstance, exceptionInfo, functionInstance, parameterHelper.FilterContextProperties, cancellationToken);
                 }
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -133,8 +132,11 @@ public HostOutputMessage HostOutputMessage
                     logCompletedCancellationToken = cancellationToken;
                 }
 
-                await NotifyCompleteAsync(instanceLogEntry, functionCompletedMessage.Arguments, exceptionInfo);
-                _resultsLogger?.LogFunctionResult(instanceLogEntry);
+                if (instanceLogEntry != null)
```
mathewc: Avoid operating on instanceLogEntry if it is null

brettsam: Why is this ever null? That would mean we're missing some important details about the function executing.

mathewc: See above - if NotifyPreBind fails it can be null.

brettsam: What exception does the user see now?

mathewc: "Unable to resolve function name 'xxx'." See the repro test case I added in https://github.com/Azure/azure-functions-host/pull/3094

brettsam: Ok, cool. I just wanted to make sure we weren't running real functions that were having these calls to `Notify***` skipped.

brettsam (approved):

PR #1797, opened 2018-07-06
 Enable ServiceBus dependency and request collection

lmolkova:
This change enables ServicBus calls tracking
ApplicaitonInsights listens to ServiceBus SDK events and tracks requests automatically, enriching them with ServiceBus specific properties such as endpoint, queue name, function called (Send/Process)

Such events are translated to telemetry so that 'Send' is tracked as dependency and 'Process' is a request. 
ServiceBus SDK also ensures that is there is a trace context in the message, it will be used.

There is a small caveat with request tracking: AppInsights SDK tracks request from SB automatically and therefore Request tracked by webjobs SDK, in this case, is not needed.
Based on the presence of Activity.Current, we can say whether someone tracks request already and avoid double-tracking.

We've had an offline conversation with @brettsam about HTTP requests, and we've decided that 'outer', 'automagical' request should be tracked: it contains specific info (such as SB message Id or queue name, or in case of HTTP request URL and response code).
'Outer' request will have a bit bigger duration than a pure function call; for now, actual function duration is added into the telemetry properties.

Request operation name still is set to function name as portal UI depends on it.

ServiceBus package was updated to the latest stable version 3.0.2 because of NullReferenceException bug in diagnostics code https://github.com/Azure/azure-service-bus-dotnet/issues/459

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
@@ -3,6 +3,8 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
+using System.Globalization;
 using System.Linq;
 using Microsoft.ApplicationInsights;
 using Microsoft.ApplicationInsights.DataContracts;
@@ -164,13 +166,6 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
 
             var customScopeProperties = scopeProperties.Where(p => !SystemScopeKeys.Contains(p.Key, StringComparer.Ordinal));
             ApplyProperties(telemetry, customScopeProperties, LogConstants.CustomPropertyPrefix);
-
-            // Set the invocation id, if there is one. Set it under the friendly key name.
-            string functionInvocationId = scopeProperties.GetValueOrDefault<string>(ScopeKeys.FunctionInvocationId);
-            if (!string.IsNullOrEmpty(functionInvocationId))
-            {
-                ApplyProperty(telemetry, LogConstants.InvocationIdKey, functionInvocationId);
-            }
         }
 
         private void LogException(LogLevel logLevel, IEnumerable<KeyValuePair<string, object>> values, Exception exception, string formattedMessage)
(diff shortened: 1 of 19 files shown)


lmolkova:
@brettsam @fabiocav please review when you have time

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/WebJobsTelemetryInitializer.cs
```diff
@@ -26,36 +28,75 @@ public void Initialize(ITelemetry telemetry)
...
-
-            telemetry.Context.Operation.Name = scopeProps.GetValueOrDefault<string>(ScopeKeys.FunctionName);
+            IDictionary<string, object> scopeProps =
+                DictionaryLoggerScope.GetMergedStateDictionary() ?? new Dictionary<string, object>();
 
-            // Apply Category and LogLevel to all telemetry
-            if (telemetry is ISupportProperties telemetryProps)
+            // this could be telemetry tracked in scope of function call - then we should apply the logger scope
+            // or RequestTelemetry tracked by the WebJobs SDK or AppInsight SDK - then we should apply Activity.Tags
+            if (request == null && scopeProps.Any())
```
brettsam: How are you determining whether the `RequestTelemetry` is part of a function call here? Shouldn't `RequestTelemetry` not be null in that case?

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/ServiceBusRequestAndDependencyCollectionTests.cs
```diff
@@ -0,0 +1,151 @@
...
+            using (var host = ConfigureHost())
+            {
+                await host.StartAsync();
+                await host.GetJobHost()
+                    .CallAsync(typeof(ServiceBusRequestAndDependencyCollectionTests).GetMethod(nameof(ServiceBusOut)), new {input = "message" + DateTime.UtcNow.ToString("o")});
+
+                _functionWaitHandle.WaitOne();
+
+                //TODO : clean me
+                await Task.Delay(5000);
```
brettsam: Note -- one thing I've been doing lately to remove these delays is to instead use `TestHelpers.Await()` and wait for the telemetries to appear. It'll auto-timeout if it takes too long, and that way the test is more deterministic.

brettsam:
FYI -- I'll do another pass through when it's all ready to merge for a final review

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsHostBuilderExtensions.cs
```diff
@@ -93,12 +93,14 @@ public static IHostBuilder AddApplicationInsights(this IHostBuilder builder, str
...
+                    var includedActivities = dependencyCollector.IncludeDiagnosticSourceActivities;
+                    includedActivities.Add("Microsoft.Azure.ServiceBus");
+
                     return dependencyCollector;
                 });
                 services.AddSingleton<ITelemetryModule, AppServicesHeartbeatTelemetryModule>();
 
-                ServerTelemetryChannel serverChannel = new ServerTelemetryChannel();
-                services.AddSingleton<ITelemetryChannel>(serverChannel);
+                services.AddSingleton<ITelemetryChannel, ServerTelemetryChannel>();
```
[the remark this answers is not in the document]
lmolkova: exactly

lmolkova:
> FYI -- I'll do another pass through when it's all ready to merge for a final review
thanks, I'm debugging the new test that always passes locally, I'll fix it and go through your comments.

I'll let you know when I'm done, please ignore all debugging commits for now...

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/WebJobsTelemetryInitializer.cs
lmolkova: RequestTelemetry is never null (initializer runs on tracked telemetry item). The question is: who reported RequestTelemetry: functions in LogFunctionResult or AppInsights SDK automagically

lmolkova:
@brettsam 
I've done with my changes, but I found the issue in ApplicaitonInsights SDK that seems to affect tests only: https://github.com/Microsoft/ApplicationInsights-dotnet-server/pull/948.

The test fails because other service bus tests might run before it and it affects magical auto-collection :(

We are fixing the issue in the next ApplicaitonInsights release (stable in a 1-2 weeks) and I'll update the version and enable the test. 

Please advise whether it's ok to merge like that.

lmolkova:
@brettsam could you please have a look? There was an issue with ServiceBus dependency collection that we've fixed in AppInsights 2.7, so if you're ok with this, it is ready to merge

lmolkova:
Still blocked because of https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/970

lmolkova:
Finally ready. All issues fixed and tests passed

lmolkova:
quick pulse tests still seem to be unstable. But this change is ready for merge.
@brettsam

lmolkova:
@brettsam could you please merge it if you don't have other concerns

brettsam:
Taking another look now... I've kicked off another test run to make sure that it's just flakiness with that QuickPulse test. I made some improvements to it in v2.x but I need to port those over to dev.

src/Microsoft.Azure.WebJobs.ServiceBus/WebJobs.ServiceBus.csproj
```diff
@@ -9,8 +9,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.ServiceBus" Version="2.0.0" />
-    <PackageReference Include="StyleCop.Analyzers" Version="1.1.0-beta004">
+    <PackageReference Include="Microsoft.Azure.ServiceBus" Version="3.0.2" />
```
brettsam: @mathewc / @fabiocav -- this change is bumping our ServiceBus dependency. Do we have any concerns with that? I know we wanted to do it anyway, but wanted to make sure it was clear that it was happening here.

brettsam:
It looks like we're seeing an additional trace with those QuickPulse tests. Do you see that locally?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -301,59 +302,88 @@ private void LogFunctionResultAggregate(IEnumerable<KeyValuePair<string, object>
...
+                }
+                if (scope.TryGetValue(LogConstants.CategoryNameKey, out object category))
+                {
+                    currentActivity.AddTag(LogConstants.CategoryNameKey, category.ToString());
+                }
+
+                if (scope.TryGetValue(LogConstants.LogLevelKey, out object logLevel))
+                {
+                    currentActivity.AddTag(LogConstants.LogLevelKey, logLevel.ToString());
                 }
```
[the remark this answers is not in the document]
brettsam: Sure, thanks! Merging now.

lmolkova:
> It looks like we're seeing an additional trace with those QuickPulse tests. Do you see that locally?

I'm not able to repo it locally. On the build server, if the new test introduced in this PR is enabled, it causes some strange side effects on quick pulse test. 

I'm investigating this and it seems to take a while...

brettsam:
Okay -- don't worry too much about that if it doesn't happen locally. I'll port over my improvements from the other branch and maybe that'll make things better. If everything else is ready and passing but that test is still failing sometimes, we can still get this in. I'll take care of improving it.

PR #1801, opened 2018-07-09
Add ApplicationIdProvider to support multi-instrumentation key scenarios

lmolkova:
ApplicationInsights can show AppMap and end-to-end trace viewer in case when different sub-components report to different ApplicationInsights resources.
This feature is available now HTTP only and we are working on extending support for other protocols.

This change enables multiple instrumentation key feature. 

/cc  @brettsam @fabiocav

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
@@ -111,8 +111,8 @@ namespace Microsoft.Extensions.DependencyInjection
                         new ServerTelemetryChannel(),
                         provider.GetServices<ITelemetryInitializer>(),
                         provider.GetServices<ITelemetryModule>(),
-                        appIdProvider,
-                        new ApplicationInsightsApplicationIdProvider());
+                        new ApplicationInsightsApplicationIdProvider(),
+                        filterOptions);
                 }
                 return config;
             });
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
@@ -10,6 +10,7 @@ using Microsoft.ApplicationInsights.Channel;
 using Microsoft.ApplicationInsights.DependencyCollector;
 using Microsoft.ApplicationInsights.Extensibility;
 using Microsoft.ApplicationInsights.Extensibility.Implementation;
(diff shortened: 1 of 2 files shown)


lmolkova:
@brettsam could you please have a look?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
```diff
@@ -103,6 +111,7 @@ public static IServiceCollection AddApplicationInsights(this IServiceCollection
                         new ServerTelemetryChannel(),
                         provider.GetServices<ITelemetryInitializer>(),
                         provider.GetServices<ITelemetryModule>(),
+                        new ApplicationInsightsApplicationIdProvider(),
```
fabiocav: So we don't want to use the same registered instance here?

lmolkova: Yes, we have to initialize configuration for this host AND Active (static singleton) config. Active is initialized just once and lives till process dies. Function users access Active instance (as they don't have access to DI) if they want to manually track any telemetry from the function code.
ApplicaitonIdProvider is disposable and dies together with WebHost, but we don't want to same for the Active one.

So, I'd love to avoid using Active at all and never even initialize it, and it would be possible if there was a binding for TelemtryClient for functions. I think @brettsam mentioned that it is something he was considering. IF it's done we can get rid of Active config for good.

lmolkova: Awesome! BTW, it would be a breaking change to stop initializing Active config, if we'll ever release current version. Active initialization is a new thing and has never been released yet.

If the binding is planned after v2 GA, I should probably remove  Active initialization so we don't break it later. What do you think?

brettsam: @lmolkova -- That may be a good idea to remove this initialization. Accessing `Active` makes me a little nervous b/c Functions can restart the host and dispose services at any time. Having some Static service that lives across that barrier seems troublesome.

I think it'd be much nicer if we prioritized the direct access to the registered service. And like you said, since we've never released this, it technically wouldn't be cutting anything.

lmolkova: great, I'll send another PR to remove Active initialization

lmolkova:
BTW, I no longer have write permissions and cannot merge PRs, could you guys please merge it?

brettsam:
I pulled this, rebased it, and squashed it. Committed now with https://github.com/Azure/azure-webjobs-sdk/commit/0ee4f6f7ef02593ec835a655cdbec165649c6fec.

PR #1810, opened 2018-07-19
Check queue message dequeuecount before executing the functions

### LINKED ISSUE #1809: Messages never move to poison queue incase of timeout exceptions
_From @pragnagopa on July 18, 2018 20:14_

#### Investigative information

Please provide the following:

- Timestamp:
- Function App version (1.0 or 2.0-beta): 1.0.11913.0 and V2 Beta
- Function App name: pgopafunctestprod
- Function name(s) (as appropriate):
- Invocation ID:
- Region:

#### Repro steps

Update functionTimeout in host.json to 30secs - to repro issue faster than the default timeout of 5 minutes

```
using System;

public static void Run(string myQueueItem, TraceWriter log)
{
    log.Info($"C# Queue trigger function processed: {myQueueItem}");
   // Intentionally sleep for 40 secs to trigger timeout exception
    System.Threading.Thread.Sleep( 40000);
   log.Info($"C# Queue trigger done function processed: {myQueueItem}");
    
}

```

#### Expected behavior

Queue message moves to poison-queue after 5 retries

#### Actual behavior

ScriptHost keeps restarting and continues to retry

#### Known workarounds

none

#### Related information 

Kusto logs:

FunctionsLogs
| where PreciseTimeStamp >= datetime(2018-07-17 15:15:00)
| where AppName == "pgopafunctestprod" 
| project PreciseTimeStamp ,  AppName , FunctionName , Details , Summary , HostVersion, HostInstanceId  , Source, FunctionInvocationId 

_Copied from original issue: Azure/azure-functions-host#3157_

pragnagopa:
Fixes #1809

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Queues/Listeners/QueueListener.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Queues/Listeners/QueueListener.cs
@@ -271,6 +271,7 @@ namespace Microsoft.Azure.WebJobs.Host.Queues.Listeners
             {
                 if (!await _queueProcessor.BeginProcessingMessageAsync(message.SdkObject, cancellationToken))
                 {
+                    await _queueProcessor.MoveMessageToPoisonQueueAsync(message.SdkObject, cancellationToken);
                     return;
                 }
 
--- a/src/Microsoft.Azure.WebJobs.Host/Queues/Listeners/QueueListener.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Queues/Listeners/QueueListener.cs
@@ -271,7 +271,6 @@ namespace Microsoft.Azure.WebJobs.Host.Queues.Listeners
             {
                 if (!await _queueProcessor.BeginProcessingMessageAsync(message.SdkObject, cancellationToken))
                 {
-                    await _queueProcessor.MoveMessageToPoisonQueueAsync(message.SdkObject, cancellationToken);
                     return;
                 }
 
--- a/src/Microsoft.Azure.WebJobs.Host/Queues/QueueProcessor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Queues/QueueProcessor.cs
@@ -92,6 +92,10 @@ namespace Microsoft.Azure.WebJobs.Host.Queues
         /// <returns>True if the message processing should continue, false otherwise.</returns>
         public virtual async Task<bool> BeginProcessingMessageAsync(CloudQueueMessage message, CancellationToken cancellationToken)
         {
+            if (message.DequeueCount > MaxDequeueCount)
+            {
+                return await Task.FromResult<bool>(false);
+            }
             return await Task.FromResult<bool>(true);
         }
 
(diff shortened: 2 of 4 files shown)


src/Microsoft.Azure.WebJobs.Host/Queues/Listeners/QueueListener.cs
```diff
@@ -271,6 +271,7 @@ private TaskSeriesCommandResult CreateSucceededResult()
             {
                 if (!await _queueProcessor.BeginProcessingMessageAsync(message.SdkObject, cancellationToken))
                 {
+                    await _queueProcessor.MoveMessageToPoisonQueueAsync(message.SdkObject, cancellationToken);
```
mathewc: Customers can override BeginProcessingMessage and return false for their own reasons. I don't think it's safe to assume that if they return false the message should be moved to poison queue.

Instead, move this into BeginProcessingMessage.

src/Microsoft.Azure.WebJobs.Host/Queues/QueueProcessor.cs
```diff
@@ -142,6 +136,30 @@ public QueueProcessor(QueueProcessorFactoryContext context)
             }
         }
 
+        /// <summary>
+        /// Copies properties from the specified message and moves it to the poison queue.
+        /// </summary>
+        /// <param name="message">The poison message</param>
+        /// <param name="cancellationToken">The <see cref="CancellationToken"/> to use</param>
+        /// <returns></returns>
+        internal async Task MoveMessageToPoisonQueueAsync(CloudQueueMessage message, CancellationToken cancellationToken)
```
mathewc: nit: maybe rename this to HandlePoisonMessageAsync, to avoid the similar name to CopyToPoisonQueueAsync. Handle makes sense, since we're doing multiple operations - copying the message to poison queue then deleting it.

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Queues/QueueListenerTests.cs
```diff
@@ -20,77 +20,74 @@ namespace Microsoft.Azure.WebJobs.Host.FunctionalTests.Queues
...
+        private StorageQueue _storagePoisonQueue;
+        private Mock<ITriggerExecutor<IStorageQueueMessage>> _mockTriggerExecutor = new Mock<ITriggerExecutor<IStorageQueueMessage>>(MockBehavior.Strict);
+
         public QueueListenerTests(TestFixture fixture)
         {
-            Fixture = fixture;
+            _trace = new TestTraceWriter(TraceLevel.Verbose);
+            _queue = fixture.CreateNewQueue();
+            _poisonQueue = fixture.CreateNewQueue();
+            _queuesConfig = new JobHostQueuesConfiguration { MaxDequeueCount = 2 };
```
mathewc: Have you taken a look at QueueProcesorTests? Those are isolated tests for QueueProcessor - I think it makes more sense for your new test to be there. E.g. it should be easy then for you to have a test that calls directly into BeginProcessingMessageAsync with a message whose dequeue count is over limit.

PR #1812, opened 2018-07-20
Check queue message dequeue count before executing function.

pragnagopa:
Port fix for #1809  to V3


test/Microsoft.Azure.Webjobs.Extensions.Storage.UnitTests/Queue/QueueProcessorTests.cs
```diff
@@ -27,6 +27,7 @@ public QueueProcessorTests(TestFixture fixture)
         {
             _queue = fixture.Queue;
             _poisonQueue = fixture.PoisonQueue;
+
```
mathewc: nit: remove

Issue #1820, opened 2018-08-01
Finish DI + Storage refactor work


mathewc:
During the recent refactorings, a bunch of code/tests were commented out and a bunch of temporary hacks were put in place. Searching the code for $$$ and TODO markers turns up hundreds of instances. E.g. entire test suites are still commented out. We need to clean all this up.

paulbatum:
The tests aspect of this is the should-have part.

fabiocav:
@mathewc @brettsam this issue is hasn't been touched in a while. Do we need to decompose this into more specific issues for things we still need to address?

PR #1821, opened 2018-08-01
Don't query non-existent queues (#770)

mathewc:
Merging fix for https://github.com/Azure/azure-webjobs-sdk/pull/1572 to dev - this was missed long ago.

As part of this enabling the QueueListenerTests suite that was $$$ disabled as part of the recent DI changes. I logged https://github.com/Azure/azure-webjobs-sdk/issues/1820 to address the rest of the remaining work there.


src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Listeners/QueueListener.cs
```diff
@@ -137,17 +138,36 @@ public void Dispose()
...
-            IEnumerable<CloudQueueMessage> batch;
+            IEnumerable<CloudQueueMessage> batch = null;
             try
             {
-                batch = await _queue.GetMessagesAsync(_queueProcessor.BatchSize,
-                    _visibilityTimeout,
-                    options: null,
-                    operationContext: null,
-                    cancellationToken: cancellationToken);
+                if (!_queueExists.HasValue || !_queueExists.Value)
```
mathewc: This is all just a direct port of code + tests from the v2 commit

test/Microsoft.Azure.Webjobs.Extensions.Storage.UnitTests/Queue/QueueListenerTests.cs
```diff
@@ -1,182 +0,0 @@
...
-using Microsoft.Azure.WebJobs.Host.Queues.Listeners;
-using Microsoft.Azure.WebJobs.Host.TestCommon;
-using Microsoft.Azure.WebJobs.Host.Timers;
-using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Hosting;
-using Microsoft.WindowsAzure.Storage.Queue;
-using Moq;
-using Xunit;
-
-#if false // $$$ enable
```
mathewc: I reenabled all these tests. The rest of the rename/move changes here are consolidating tests (we had two QueueListenerTests.cs files under two different directories).

We also had Queue unit tests in two different directories (Queue and Queues). Consolidated these and fixed the naming of these directories to match FX folders Queues, Blobs, Tables

PR #1824, opened 2018-08-06
[AppInsights] moving to newer .NET logging dependencies

### LINKED ISSUE #1780: Configure ApplicationInsights through  IConfiguration or options
ApplicationInsights is now configured via explicit parameters in `AddApplicationInsights` signature

https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsHostBuilderExtensions.cs#L50

```
        public static IHostBuilder AddApplicationInsights(
            this IHostBuilder builder, 
            string instrumentationKey,
            Func<string, LogLevel, bool> filter,
            SamplingPercentageEstimatorSettings samplingSettings)
```
The list of parameters may grow and it becomes hard to manage default values and overloads. But now Script host passes parameters explicitly. We should revisit this and configure the AI via IConfiguration and/or configuration options.

See  also https://github.com/Azure/azure-webjobs-sdk/pull/1763#discussion_r198633087 and https://github.com/Azure/azure-webjobs-sdk/pull/1767#discussion_r198616628

brettsam:
Fixes #1780.

Moves our Application Insights logging to use `IConfiguration`, `ILoggingBuilder`, and `LogFilterOptions` like other providers.

One tricky piece -- the App Insights pipeline doesn't filter logs until *after* they've gone to the QuickPulse processor. This is so you can get an accurate view of your site, even with logging turned off. There's not a real mechanism for this in the logging types. So I register a filter that allows all logs through for our logger, then I 'capture' the logger rules and apply them later in the pipeline.

All tests are passing, but I want to add some new ones for this new behavior before committing. Sending now for early comments.


brettsam:
Had to take a different approach to our filtering scenarios. Just pushed that. When testing I realized that any filters a customer would add to "AppInsights" wouldn't get properly applied. 

I'm going to have to rebase this on #1825 once that goes in, then it should be ready for a final signoff.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsHostBuilderExtensions.cs
```diff
@@ -28,6 +29,13 @@ public static IHostBuilder ConfigureWebJobsHost(this IHostBuilder builder, Actio
                 config.AddEnvironmentVariables();
             });
 
+
+            builder.ConfigureLogging((context, logging) =>
+            {
+                // Loggers can be configured under "Logging" in config
+                logging.AddConfiguration(context.Configuration.GetSection("Logging"));
+            });
```
brettsam: @fabiocav -- I decided to auto-register this. It seems like a common pattern and is overrideable if needed (I have a test that verifies this).

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsLoggingBuilderExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+                    return r.ProviderName == typeof(ApplicationInsightsLoggerProvider).FullName
+                        || r.ProviderName == ApplicationInsightsLoggerProvider.Alias;
+                }).ToArray();
+
+                foreach (var rule in matchingRules)
+                {
+                    o.Rules.Remove(rule);
+                }
+
+                o.Rules.Add(new ApplicationInsightsLoggerFilterRule(matchingRules.ToList()));
```
brettsam: Here's the new behavior -- we capture all rules applying to our logger and store them inside a custom rule for re-enabling later.

brettsam:
@fabiocav -- ping

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsHostBuilderExtensions.cs
fabiocav: Sounds good.  Do we need to think about whether this  would be on the root, though? In the case of host.json, this would be scoped under "AzureFunctionsJobHost", do we need to special case it?

fabiocav: I guess we could also have some logic in Script where `AddConfiguration` is called with the appropriate section.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsLoggingBuilderExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+
+            builder.Services.PostConfigure<LoggerFilterOptions>(o =>
+            {
+                // We want all logs to flow through the logger so they show up in QuickPulse.
+                // To do that, we'll hide all registered rules inside of this one. They will be re-populated
+                // and used by the FilteringTelemetryProcessor futher down the pipeline.
+
+                var matchingRules = o.Rules.Where(r =>
+                {
+                    return r.ProviderName == typeof(ApplicationInsightsLoggerProvider).FullName
```
fabiocav: nit: move `typeof(...)` outside of lambda and use resolved name

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsLoggingBuilderExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+                        || r.ProviderName == ApplicationInsightsLoggerProvider.Alias;
+                }).ToArray();
```
fabiocav: Can we just call `ToList` here and avoid the call below?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,225 @@
...
+        }
+
+        public static IServiceCollection AddApplicationInsights(this IServiceCollection services)
+        {
+            // Bind to the configuration section registered with 
+            services.AddOptions<ApplicationInsightsLoggerOptions>()
+                .Configure<ILoggerProviderConfiguration<ApplicationInsightsLoggerProvider>>((options, config) =>
+                {
+                    config.Configuration?.Bind(options);
+                });
```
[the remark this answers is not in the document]
fabiocav: 👍

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,227 @@
...
+
+                return dependencyCollector;
+            });
+            services.AddSingleton<ITelemetryModule, AppServicesHeartbeatTelemetryModule>();
+
+            ServerTelemetryChannel serverChannel = new ServerTelemetryChannel();
+            services.AddSingleton<ITelemetryChannel>(serverChannel);
+            services.AddSingleton<TelemetryConfiguration>(provider =>
+            {
+                ApplicationInsightsLoggerOptions options = provider.GetService<IOptions<ApplicationInsightsLoggerOptions>>().Value;
```
fabiocav: Would it make sense to wrap this logic up in a factory?

brettsam: Yes, I was going to bring that up with @lmolkova as well. It's getting pretty big and may be nice to split out. But not as part of this change.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsLoggingBuilderExtensions.cs
brettsam: Yup, will do.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsLoggingBuilderExtensions.cs
brettsam: Yup, good call.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsHostBuilderExtensions.cs
brettsam: Yeah, you bring up a good point. I'll actually pull this out and we can chat about it. The logging docs show that you should add this yourself -- but I was trying to eliminate some concepts for people getting started. But maybe it won't work quite as smoothly as I'd hoped. It's easy to add back later.

lmolkova:
@davidfowl @brettsam 

we also heavily use logger scopes here to track, scope and enrich AppInsights telemetry. It will be problematic with the 'Microsoft.Extensions.Logging.ApplicationInsights' that does not know about webjob/functions scopes.

If we decide to move the Microsoft.Extensions.Logging.ApplicationInsights here, we'll need to find another level of abstraction to track requests/metrics and set custom properties.

PR #1825, opened 2018-08-07
Refactoring builder APIs, introducing IWebJobsBuilder, moving extension methods

fabiocav:


fabiocav:
Initial work for the configuration updates planned.

This initial set of changes introduce a different builder pattern, with WebJobs services added to the `IWebJobsBuilder` instance returned (when working with `IServiceCollection`) or provided when configuring the host (from `IHostBuilder`).

brettsam:
I just noticed that this was WIP -- but if you have a decent stopping place (which it looks like this may be), I'd suggest just getting it in.

mathewc:
I'll spend an hour or so getting the tests running, then I'll update the PR. Hope is to get this in by afternoon.

fabiocav:
Sounds good. That being the case, I'll merge this and publish packages to MyGet so you have these changes and we don't have to wait for the configuration updates.

PR #1826, opened 2018-08-08
Extension Name Metadata + Misc. Cleanup

mathewc:
In this PR:

- addressing https://github.com/Azure/azure-functions-host/issues/3186, Idea is to have tooling consume the new ExtensionAttribute to get the friendly name for an extension
- removing a `ICloudBlobStreamBinder<T>`, addressing https://github.com/Azure/azure-webjobs-sdk/issues/995
- lots of other cleanup


src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Triggers/BlobTriggerAttributeBindingProvider.cs
```diff
@@ -2,78 +2,21 @@
...
 using Microsoft.Azure.WebJobs.Host.Timers;
 using Microsoft.Azure.WebJobs.Host.Triggers;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
-using Microsoft.WindowsAzure.Storage.Blob;
 
-namespace Microsoft.Azure.WebJobs.Host.Blobs.Triggers
+namespace Microsoft.Azure.WebJobs.Extensions.Storage.Blobs.Triggers
 {
-    internal class BlobTriggerExtensionConfig : IExtensionConfigProvider
```
mathewc: This file was messed up. It's file name said "binding provider" but it was an extension config provider, with a private binding provider impl. I hoisted the binding provider out into this file, and moved the extension config provider to its own file.

src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Config/BlobsExtensionConfigProvider.cs
```diff
@@ -16,27 +18,93 @@
 
 namespace Microsoft.Azure.WebJobs.Host.Blobs.Bindings
 {
-    internal class BlobExtensionConfig : IExtensionConfigProvider,
+    [Extension("AzureBlobs")]
+    internal class BlobsExtensionConfigProvider : IExtensionConfigProvider,
```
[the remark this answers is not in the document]
fabiocav: 👍

src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Triggers/BlobTriggerAttributeBindingProvider.cs
```diff
@@ -114,14 +57,14 @@ internal class BlobTriggerAttributeBindingProvider : ITriggerBindingProvider
...
-        public async Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext context)
+        public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext context)
         {
             ParameterInfo parameter = context.Parameter;
             var blobTriggerAttribute = TypeUtility.GetResolvedAttribute<BlobTriggerAttribute>(context.Parameter);
 
             if (blobTriggerAttribute == null)
             {
-                return null;
+                return Task.FromResult<ITriggerBinding>(null);
```
fabiocav: Nice change! Those are a pain!

src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Triggers/BlobTriggerAttributeBindingProvider.cs
```diff
@@ -114,14 +57,14 @@ internal class BlobTriggerAttributeBindingProvider : ITriggerBindingProvider
             _loggerFactory = loggerFactory;
         }
 
-        public async Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext context)
+        public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext context)
```
mathewc: I found several methods like this that were async but not awaiting anything, so I fixed them

src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Config/QueuesExtensionConfigProvider.cs
```diff
@@ -6,24 +6,24 @@
...
 using Newtonsoft.Json.Linq;
 
-namespace Microsoft.Azure.WebJobs.Host.Queues.Bindings
+namespace Microsoft.Azure.WebJobs.Host.Queues.Config
 {
-    // Write up bindinging rules for [Queue] attribute. 
-    // This is fundemantentally an IAsyncCollector<IStorageQueueMessage>
-    internal class QueueExtension : IExtensionConfigProvider
+    [Extension("AzureQueues")]
+    internal class QueuesExtensionConfigProvider : IExtensionConfigProvider
```
mathewc: TODO : This is the other one that can't be made public due to non-public ctor params

src/Microsoft.Azure.WebJobs.Extensions.Storage/StorageHostBuilderExtensions.cs
```diff
@@ -55,10 +55,9 @@ public static IWebJobsBuilder AddAzureStorage(this IWebJobsBuilder builder)
 
             builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IBindingProvider, CloudStorageAccountBindingProvider>());
 
-            return builder.AddExtension<TableExtension>()
-                .AddExtension<QueueExtension>()
-                .AddExtension<BlobExtensionConfig>()
```
mathewc: Used to be 2 blob extension config providers, now one

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsBuilderExtensions.cs
```diff
@@ -32,6 +32,13 @@ public static IWebJobsBuilder AddExtension(this IWebJobsBuilder builder, IExtens
             return builder;
         }
 
+        public static IWebJobsBuilder AddExtension(this IWebJobsBuilder builder, Type extensionConfigProviderType)
```
mathewc: Adding a non-generic Type based overload (used in my tests)

test/Microsoft.Azure.Webjobs.Extensions.Storage.UnitTests/HostCallTests.cs
```diff
@@ -1726,5 +1655,73 @@ private struct StructTableEntity
         {
             public string Value { get; set; }
         }
+
+        internal class CustomBlobConverterExtensionConfigProvider : IExtensionConfigProvider
+        {
+            public void Initialize(ExtensionConfigContext context)
```
mathewc: Registering custom Blob conversions via an extension.

src/Microsoft.Azure.WebJobs/ExtensionAttribute.cs
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs.Description
+{
+    /// <summary>
+    /// Attribute applied to the <see cref="IExtensionConfigProvider"/> implementation for an extension.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class)]
```
fabiocav: Should this be marked as `Inherited ` false? It would make sense tho prevent that from happening.

src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Triggers/BlobTriggerAttributeBindingProvider.cs
fabiocav: The spots where `null` was returned when a Task is expected are also trouble! So thanks for fixing!

src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Config/BlobsExtensionConfigProvider.cs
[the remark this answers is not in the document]
mathewc: Don't really want or need these config providers to be public - we just need a public way for tooling to statically discover extensions. An alternative we discussed was to specify the name as part of the WebJobsStartup annotations. However WebJobsStartup isn't extensions specific at all and we can't assume each maps to an extension. It's just a way to run startup code.

src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubExtensionConfigProvider.cs
```diff
@@ -6,16 +6,17 @@
...
+using Microsoft.Azure.WebJobs.Description;
 using Microsoft.Azure.WebJobs.Host.Bindings;
 using Microsoft.Azure.WebJobs.Host.Config;
 using Microsoft.Azure.WebJobs.Logging;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
 using Newtonsoft.Json;
 
-namespace Microsoft.Azure.WebJobs.ServiceBus
+namespace Microsoft.Azure.WebJobs.EventHubs
```
brettsam: Should this be `...Extensions.EventHubs`?

src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubsWebJobsStartup.cs
```diff
@@ -1,13 +1,13 @@
...
 
-using Microsoft.Azure.WebJobs.Extensions.EventHubs;
+using Microsoft.Azure.WebJobs.EventHubs;
 using Microsoft.Azure.WebJobs.Hosting;
 using Microsoft.Extensions.Hosting;
 
 [assembly: WebJobsStartup(typeof(EventHubsWebJobsStartup))]
 
-namespace Microsoft.Azure.WebJobs.Extensions.EventHubs
+namespace Microsoft.Azure.WebJobs.EventHubs
```
brettsam: Just mentioned this above -- but it looks like our extensions in the Extensions repo all have the `Extensions` name in there.

test/Microsoft.Azure.Webjobs.Extensions.Storage.UnitTests/HostCallTests.cs
brettsam: Yes, I agree. Was going to ask if you'd moved things to use converters when you removed the CloudBlobStreamBinder (as it was hard to tell due to the reshuffling above). But I think this is a good step to get there.

It also seems like we could support POCO with a converter that we auto-register with the Blob extension.

src/Microsoft.Azure.WebJobs/ExtensionAttribute.cs
```diff
@@ -0,0 +1,24 @@
...
+    public sealed class ExtensionAttribute : Attribute
```
brettsam: Are we going to make this required for ExtensionConfigProviders? (i.e. will tooling require it?)

If it's only meant to be used when the name is different from the class, we could rename this to `ExtensionAlias` to be clear. Logging has a similar `ProviderAlias` attribute.

src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubExtensionConfigProvider.cs
```diff
@@ -6,16 +6,17 @@
...
 {
+    [Extension("EventHubs")]
```
brettsam: Is this required?

src/Microsoft.Azure.WebJobs.ServiceBus/Config/ServiceBusExtensionConfig.cs
```diff
@@ -17,6 +18,7 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Config
     /// <summary>
     /// Extension configuration provider used to register ServiceBus triggers and binders
     /// </summary>
+    [Extension("ServiceBus")]
     public class ServiceBusExtensionConfig : IExtensionConfigProvider
```
brettsam: Rename to `ServiceBusExtensionConfigProvider` to be consistent?

src/Microsoft.Azure.WebJobs/ExtensionAttribute.cs
fabiocav: It isn’t required. We’ll fallback to deriving the name from the type based on convention then to use the full type name, I agree that `ExtensionAlias` would be a better option.

src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubExtensionConfigProvider.cs
mathewc: The namespace for this assembly is currently Microsoft.Azure.WebJobs.EventHub which is in agreement with its assembly name. That's a consistency we want to maintain. Ideally the assembly would be Microsoft.Azure.WebJobs.Extensions.EventHubs then the namespace would change to match. If we want to do that assembly rename, we'd do all of this together as a separate change. @paulbatum and I were discussing that yesterday.

src/Microsoft.Azure.WebJobs/ExtensionAttribute.cs
mathewc: I named this ExtensionAttribute rather than ExtensionName/ExtensionAlias to allow for the attribute to contain other extension metadata in the future.

src/Microsoft.Azure.WebJobs.ServiceBus/Config/ServiceBusExtensionConfig.cs
mathewc: Fixed

src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubExtensionConfigProvider.cs
mathewc: Nothing requires/enforces it currently but we could do that. It's a pattern at this point.

PR #1827, opened 2018-08-08
Adding extension configuration/options binding support. Cleaning up configuration

fabiocav:


src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubExtensionConfigProvider.cs
```diff
@@ -48,14 +48,11 @@ public void Initialize(ExtensionConfigContext context)
...
-            // TODO: Can we bind these during service setup?
-            // apply at eventProcessorOptions level (maxBatchSize, prefetchCount)
-
             EventProcessorOptions options = _options.GetOptions();
             options.SetExceptionHandler(ExceptionReceivedHandler);
-            _configuration.Bind("eventHub", options);
 
             // apply at config level (batchCheckpointFrequency)
+            // TODO: Revisit this... All configurable options should move to the Options type.
             _configuration.Bind("eventHub", _options);
```
fabiocav: We need to cleanup the `EventHubConfiguration` type and remove the direct dependency the config provider has on `IConfiguration`. This class shouldn't be binding here.

src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubWebJobsBuilderExtensions.cs
```diff
@@ -12,7 +13,9 @@ public static class EventHubWebJobsBuilderExtensions
     {
         public static IWebJobsBuilder AddEventHubs(this IWebJobsBuilder builder)
         {
-            builder.AddExtension<EventHubExtensionConfigProvider>();
+            builder.AddExtension<EventHubExtensionConfigProvider>()
+                .BindOptions<EventProcessorOptions>();
```
fabiocav: This is an example of using `BindOptions` to automatically bind to the extension section.

src/Microsoft.Azure.WebJobs.Host/Utility.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            if (extension == null)
+            {
+                throw new ArgumentNullException(nameof(extension));
+            }
+
+            return GetExtensionName(extension.GetType());
+        }
+
+        private static string GetExtensionName(Type extensionType)
```
fabiocav: This will change once @mathewc  merges his changes. We'll update this to also look at the attribute applied to the provider.

src/Microsoft.Azure.WebJobs.ServiceBus/ServiceBusHostBuilderExtensions.cs
```diff
@@ -14,13 +15,12 @@ public static class ServiceBusHostBuilderExtensions
...
-            builder.Services.AddOptions<ServiceBusOptions>()
-                            .Configure<IConnectionStringProvider>((o, p) =>
-                            {
-                                o.ConnectionString = p.GetConnectionString(ConnectionStringNames.ServiceBus);
-                            });
+            builder.AddExtension<ServiceBusExtensionConfig>()
+                .ConfigureOptions<ServiceBusOptions>((config, options) =>
+                {
+                    config.Bind(options);
+                    options.ConnectionString = config.GetConnectionString(ConnectionStringNames.ServiceBus);
```
fabiocav: This is an example of using `ConfigureOptions` (instead of `BindOptions`), which will give you the entire configuration section so you can assign/manager configuration as needed.

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Configuration/WebJobsExtensionOptionsConfigurationTests.cs
```diff
@@ -0,0 +1,90 @@
...
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Options;
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Host.FunctionalTests.Configuration
+{
+    public class WebJobsExtensionOptionsConfigurationTests
+    {
```
fabiocav: These are tests. The test  things.

fabiocav:
Also opened this to track removing `IConnectionStringProvider`: https://github.com/Azure/azure-webjobs-sdk/issues/1828

src/Microsoft.Azure.WebJobs.Host/Properties/AssemblyInfo.cs
```diff
@@ -8,6 +8,7 @@
 [assembly: InternalsVisibleTo("Dashboard.UnitTests")]
 [assembly: InternalsVisibleTo("DynamicProxyGenAssembly2")]
 [assembly: InternalsVisibleTo("Microsoft.Azure.WebJobs.Host.TestCommon")]
+[assembly: InternalsVisibleTo("Microsoft.Azure.WebJobs.Host.EndToEndTests")]
```
brettsam: We've tried to avoid this since the E2E tests are supposed to be how customers would be able to interact with our stuff. Is there something that we need to expose so we can remove this?

src/Microsoft.Azure.WebJobs.Host/Utility.cs
```diff
@@ -0,0 +1,75 @@
...
+            where T : IExtensionConfigProvider
+        {
+            return GetExtensionConfigurationSection(configuration, GetExtensionName<T>());
+        }
+
+        public static IConfigurationSection GetExtensionConfigurationSection(IConfiguration configuration, string extensionName)
+        {
+            string configPath = "extensions";
+
+            if (extensionName != null)
```
brettsam: We'll allow a null extension name? Does that mean it tries to read from directly under "extensions"?

src/Microsoft.Azure.WebJobs.Host/Utility.cs
```diff
@@ -0,0 +1,75 @@
...
+            {
+                throw new ArgumentNullException(nameof(extension));
+            }
+
+            return GetExtensionName(extension.GetType());
+        }
+
+        private static string GetExtensionName(Type extensionType)
+        {
+            string[] defaultSuffixes = new[] { "ExtensionConfigProvider", "ExtensionConfig", "Extension" };
```
brettsam: nit: move this out to a static field?

test/Microsoft.Azure.WebJobs.EventHubs.UnitTests/EventHubTests.cs
```diff
@@ -228,9 +228,9 @@ public void InitializeFromHostMetadata()
                 {
                     c.AddInMemoryCollection(new Dictionary<string, string>
                     {
-                        { "EventHub:MaxBatchSize", "100" },
-                        { "EventHub:PrefetchCount", "200" },
-                        { "EventHub:BatchCheckpointFrequency", "5" },
+                        { "AzureWebJobs:extensions:EventHub:MaxBatchSize", "100" },
+                        { "AzureWebJobs:extensions:EventHub:PrefetchCount", "200" },
+                        { "AzureWebJobs:extensions:EventHub:BatchCheckpointFrequency", "5" },
```
brettsam: Just noticed this -- should this be `EventHubs` (plural)?

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Configuration/WebJobsExtensionOptionsConfigurationTests.cs
brettsam: We'll be the judge of that. 👨‍⚖️

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Configuration/WebJobsExtensionOptionsConfigurationTests.cs
```diff
@@ -0,0 +1,90 @@
...
+        public void ConfigureOptions_BindsToConfiguration()
+        {
+            var host = new HostBuilder()
+                .ConfigureAppConfiguration(b =>
+                {
+                    b.AddInMemoryCollection(new Dictionary<string, string>
+                    {
+                        { "AzureWebJobs:extensions:test:config1", "test1" },
+                        { "AzureWebJobs:extensions:test:config2", "test2" }
+                    });
```
brettsam: I know it's not really testing our stuff -- but maybe add a "config3" here to show that it overrides in the configure call below? It'll make it clearer to newer people how this works.

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Configuration/WebJobsExtensionOptionsConfigurationTests.cs
```diff
@@ -0,0 +1,90 @@
...
+                        o.Config3 = "called";
+                    });
+                }).Build();
+
+            var options = host.Services.GetService<IOptions<TestOptions>>();
+
+            Assert.Equal("test1", options.Value.Config1);
+            Assert.Equal("test2", options.Value.Config2);
+            Assert.Equal("called", options.Value.Config3);
+        }
```
brettsam: Can we have a test that calls `BindOptions<>` and then `ConfigureOptions<>` in one chain?

test/Microsoft.Azure.WebJobs.Host.UnitTests/Configuration/WebJobsExtensionOptionsConfigurationTests.cs
```diff
@@ -0,0 +1,70 @@
...
+            };
+
+            TestOptions options = ConfigureOptions<TestOptions>(config, "testext");
+
+            // Assert
+            Assert.Equal("test1", options.TestValue1);
+            Assert.Equal("test2", options.TestValue2);
+        }
+
+        private TOptions ConfigureOptions<TOptions>(IDictionary<string, string> configValues, string extensionName = null, string rootSectionName = "AzureWebJobs")
```
brettsam: using `rootSectionName` anywhere?

test/Microsoft.Azure.WebJobs.Host.UnitTests/Configuration/WebJobsExtensionOptionsConfigurationTests.cs
```diff
@@ -0,0 +1,70 @@
...
+            Assert.Equal("test1", options.TestValue1);
+            Assert.Equal("test2", options.TestValue2);
+        }
+
+        [Fact]
+        public void Configuration_WithConfiguredRootSectionName_IsBound()
+        {
+            var config = new Dictionary<string, string>
+            {
+                { "AzureWebJobsConfigurationSection", "testroot" },
```
brettsam: This is what Functions will do to swap this out, right?

src/Microsoft.Azure.WebJobs.Host.Storage/RuntimeStorageHostBuilderExtensions.cs
```diff
@@ -42,7 +42,7 @@ public static IWebJobsBuilder AddAzureStorageCoreServices(this IWebJobsBuilder b
             builder.Services.TryAddEnumerable(
                 ServiceDescriptor.Transient<IConfigureOptions<LegacyConfig>, LegacyConfigSetup>());
 
-            builder.Services.TryAddEnumerable(ServiceDescriptor.Transient<IConfigureOptions<JobHostInternalStorageOptions>, JobHostInternalStorageOptionsSetup>());
+            builder.Services.TryAddEnumerable(ServiceDescriptor.Transient<IConfigureOptions<JobHostInternalStorageOptions>, CoreWebJobsOptionsSetup<JobHostInternalStorageOptions>>());
```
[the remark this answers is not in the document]
fabiocav: I kept the existing lifetime, just updating to the shared type (instead of the custom one), but it doesn't have to remain transient.

src/Microsoft.Azure.WebJobs.Host/Configuration/CoreWebJobsOptionsSetup.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Host.Configuration
+{
+    internal class CoreWebJobsOptionsSetup<TOptions> : IConfigureOptions<TOptions> where TOptions : class
```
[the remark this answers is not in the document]
fabiocav: I debated just pushing the pattern with the delegate, but this goal of this type is actually to simplify that and do exactly what you're describing; avoid additional types to configure options. 

This is now a single type that handles binding against the root configuration, abstracting away the logic to grab the correct root scope and binding. This will avoid a lot of repetitive code as we properly configure all of the different core options types.

src/Microsoft.Azure.WebJobs.Host/Configuration/WebJobsExtensionOptionsConfiguration.cs
```diff
@@ -0,0 +1,44 @@
...
+        public WebJobsExtensionOptionsConfiguration(IConfiguration configuration, string extensionName)
+            : this(configuration, extensionName, BindOption)
+        {
+            _configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
+            _extensionName = extensionName;
+        }
+
+        private static void BindOption(IConfigurationSection section, TOptions options)
+        {
+            if (section.Exists())
```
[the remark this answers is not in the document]
fabiocav: I'm just doing a sanity check here. This does avoid additional work, but calling Bind on a section that doesn't exist shouldn't cause any issues.

The `Exists` check is a established pattern in other repos, so I wanted to follow that.

src/Microsoft.Azure.WebJobs.Host/Properties/AssemblyInfo.cs
fabiocav: Yeah, didn't like that either. This was to have access to the newly introduced `Utility` class. I don't really want to make that public.

We could remove the `InternalsVisibleTo` and link to that file. That would prevent us from inadvertently use internal types in tests.

src/Microsoft.Azure.WebJobs.Host/Utility.cs
fabiocav: Kept it like this because this is all temporary. This will be replaced with a single constant as the convention will only be `<Name>ExtensionConfigProvider` and we'll be using the attribute @mathewc is adding.

test/Microsoft.Azure.WebJobs.EventHubs.UnitTests/EventHubTests.cs
fabiocav: It always gets me too, but I verified as I was doing this as every type is `EventHub...` so I kept it consistent here.

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Configuration/WebJobsExtensionOptionsConfigurationTests.cs
fabiocav: Good idea. I'll add that.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Configuration/WebJobsExtensionOptionsConfigurationTests.cs
fabiocav: Good catch. Leftover after refactoring.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Configuration/WebJobsExtensionOptionsConfigurationTests.cs
fabiocav: Yes. This makes the root configurable.

src/Microsoft.Azure.WebJobs.Host/Configuration/IWebJobsExtensionConfiguration.cs
```diff
@@ -0,0 +1,16 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.Azure.WebJobs.Host.Config;
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.Azure.WebJobs.Host.Configuration
+{
+    public interface IWebJobsExtensionConfiguration<T> where T : IExtensionConfigProvider
```
mathewc: The configuration.Configuration naming pattern this introduces does look a little strange.

fabiocav: Didn't like that, but the type name options aren't great considering some of the other types in the SDK. I've renamed the property to `ConfigurationSection` . We can move forward with a type rename later if needed.

PR #1829, opened 2018-08-08
ExtensionAttribute pattern refinements (#3186)

mathewc:
We didn't really want IExtensionConfigProviders to have to be public just for tooling discovery. This actually was problematic in a couple cases (Blobs/Queues) where these couldn't be made public due to the fact that the ctors for the config providers referenced non-public types injected via DI. That will be a general problem for people wishing to leverage DI for their config providers.

So instead I'd like to move to the pattern where it's the WebJobsStartup class for an extension that is annotated. The runtime already discovers these and runs them to register extensions, so these really do define the extensions that are installed at runtime. So for the tooling issue https://github.com/Azure/azure-functions-host/issues/3186 I'll update the tooling to instead search for WebJobsStartup types and those will be added to extensions.json. The runtime already treats these as startup types.

The startup class maps to the set of extensions that will be registered by a startup class. E.g., the AzureStorage extension includes Blobs/Tables/Queues each which as an implementation detail has its own IExtensionConfigProvider. However the unit of registration is AddAzureStorage corresponding to the startup class. Similarly for the "core" extensions in the extensions repo (Files/Timers). [ExtensionsWebJobsStartup](https://github.com/Azure/azure-webjobs-sdk-extensions/blob/dev/src/WebJobs.Extensions/Extensions/ExtensionsWebJobsStartup.cs) registers both Timers and Files. I'll likely give this extension an name like "Core" or something, or perhaps we split it out into a couple startup classes, TBD.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubExtensionConfigProvider.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubExtensionConfigProvider.cs
@@ -17,7 +17,7 @@ using Newtonsoft.Json;
 namespace Microsoft.Azure.WebJobs.EventHubs
 {
     [Extension("EventHubs")]
-    public class EventHubExtensionConfigProvider : IExtensionConfigProvider
+    internal class EventHubExtensionConfigProvider : IExtensionConfigProvider
     {
         private readonly EventHubConfiguration _options;
         private readonly ILoggerFactory _loggerFactory;
--- a/src/Microsoft.Azure.WebJobs.Host/Hosting/IWebJobsStartup.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Hosting/IWebJobsStartup.cs
@@ -1,14 +1,20 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
-using System;
-using System.Collections.Generic;
-using Microsoft.Extensions.Hosting;
-
 namespace Microsoft.Azure.WebJobs.Hosting
 {
+    /// <summary>
+    /// Interface defining a startup configuration action that should be performed
+    /// as part of host startup.
+    /// </summary>
     public interface IWebJobsStartup
     {
+        /// <summary>
+        /// Performs the startup configuration action. The host will call this
+        /// method at the right time during host initialization.
+        /// </summary>
+        /// <param name="builder">The <see cref="IWebJobsBuilder"/> that can be used to
+        /// configure the host.</param>
         void Configure(IWebJobsBuilder builder);
     }
 }
--- a/src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsStartupAttribute.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsStartupAttribute.cs
@@ -5,10 +5,20 @@ using System;
 
 namespace Microsoft.Azure.WebJobs.Hosting
 {
+    /// <summary>
+    /// Attribute used to declare <see cref="IWebJobsStartup"/> Types that should be registered and invoked
+    /// as part of host startup.
+    /// </summary>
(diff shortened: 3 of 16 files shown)


src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubsWebJobsStartup.cs
```diff
@@ -9,6 +10,7 @@
 
 namespace Microsoft.Azure.WebJobs.EventHubs
 {
+    [Extension("EventHubs")]
```
mathewc: Actually, I think it might make sense at this point to generalize this attribute to WebJobsStartupAliasAttribute

fabiocav: If this becomes a requirement, we might just want to add the name to the existing `WebJobsStartupAttribute`. The issue with the dependency on internal types isn't that pressing as we could just rely on abstractions. My concern is with some of the downsides we've discussed.

src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubExtensionConfigProvider.cs
```diff
@@ -17,7 +17,7 @@
 namespace Microsoft.Azure.WebJobs.EventHubs
 {
     [Extension("EventHubs")]
```
mathewc: The current config name for EventHubs extension in v1 is "eventHub" singular. I think we want it to be plural.

fabiocav: Why are all the type names singular?

src/Microsoft.Azure.WebJobs.Host/Hosting/IWebJobsStartup.cs
```diff
@@ -1,14 +1,20 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
-using System;
```
fabiocav: Keep `System`?

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsStartupAttribute.cs
```diff
@@ -5,10 +5,20 @@
...
     public class WebJobsStartupAttribute : Attribute
     {
-        public WebJobsStartupAttribute(Type startupType)
+        /// <summary>
+        /// Constructs a new instance.
+        /// </summary>
+        /// <param name="startupType">The Type of the <see cref="IWebJobsStartup"/> class to register.</param>
+        /// <param name="alias">The friendly name alias for the startup action. If null, the name will be
+        /// defaulted based on naming convention.</param>
+        public WebJobsStartupAttribute(Type startupType, string alias = null)
```
fabiocav: Are we keeping the alias? Did we want this to be a description?

src/Microsoft.Azure.WebJobs/ExtensionAttribute.cs
```diff
@@ -6,19 +6,25 @@
...
 
         /// <summary>
         /// Gets the friendly name of the extension.
         /// </summary>
         public string Name { get;  }
+
+        /// <summary>
+        /// Gets the configuration section name for this extension.
+        /// </summary>
+        public string ConfigName { get; }
```
fabiocav: Should this be more specific? `ConfigurationSection`?

mathewc: Fixed

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsStartupAttribute.cs
mathewc: Clarified comments that this is a human readable alias, used for logging, etc.

PR #1831, opened 2018-08-09
Updating extension configuration section name resolution logic

fabiocav:


src/Microsoft.Azure.WebJobs/ExtensionAttribute.cs
```diff
@@ -13,8 +13,8 @@ public sealed class ExtensionAttribute : Attribute
     {
         public ExtensionAttribute(string name, string configurationSection = null)
         {
-            Name = name;
-            ConfigurationSection = configurationSection ?? name;
+            Name = name ?? throw new ArgumentNullException(nameof(name));
+            ConfigurationSection = configurationSection;
```
fabiocav: Modified this to allow configuration section to be `null`, as using the name would be problematic as it is a human readable name.

fabiocav:
@brettsam / @mathewc merging this so I can pull the changes into the host. Please review and I'll address any feedback you may have.

src/Microsoft.Azure.WebJobs/ExtensionAttribute.cs
fabiocav: FYI @mathewc

Issue #1251, opened 2017-07-21
Cleanup tests failing on macOS/Linux


fabiocav:
We have a few tests currently failing on macOS/Linux. Most of the failures are due to improper line endings. We need to fix those and enable full test runs on Travis.

paulbatum:
We should just switch this over to appveyor linux. Once we do so, we need to add the badges back on the main readme page.

PR #1837, opened 2018-08-10
Update badges in readme

paulbatum:


paulbatum:
We can update this again once https://github.com/Azure/azure-webjobs-sdk/issues/1251  is resolved

PR #1838, opened 2018-08-10
Do not setup Active TelemetryConfiguration

lmolkova:
As discussed in https://github.com/Azure/azure-webjobs-sdk/pull/1801, TelemetryConfiguration.Active is static singleton that is used by customers to report telemetry manually from the custom code.

Static singletons do not work well with the disposable hosts and DI. We should not initialize Active and instead, to provide the way for users to report telemetry in a consistent matter, we should provide access to TelemtryConfiguration registered in DI container.

This change removes Active initialization. it will still work to the same extent as in V1: it will report telemetry, but no quick pulse. AI config applied by the host (telemetry initializers) will not affect Active. 
Correlation should still work automagically.


lmolkova:
@brettsam @fabiocav please review

brettsam:
@lmolkova -- when this is all rebased and ready, can you squash it down to a single commit? Then I can more easily use the Github UX to merge it into dev for us.

lmolkova:
@brettsam sure, will do and let you know as soon as all tests pass

lmolkova:
@brettsam, squashed and ready

PR #1843, opened 2018-08-13
Removing IConnectionStringProvider

mathewc:
Removing IConnectionStringProvider and associated types/implementations in favor of IConfiguration


src/Microsoft.Azure.WebJobs.Host.Storage/LegacyConfig.cs
```diff
@@ -1,35 +1,32 @@
...
 using Microsoft.WindowsAzure.Storage;
-using System;
-using System.Collections.Generic;
-using System.Text;
 
 namespace Microsoft.Azure.WebJobs
 {
 
     // $$$ Validate these?  And what are their capabilities? 
     public class LegacyConfigSetup : IConfigureOptions<LegacyConfig>
```
mathewc: I'm not sure why this LegacyConfigSetup/LegacyConfig stuff was added recently as new public surface area. We shouldn't have this

src/Microsoft.Azure.WebJobs.Host/Extensions/IConfigurationExtensions.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.WebJobs.Host;
+
+namespace Microsoft.Extensions.Configuration
+{
+    public static class IConfigurationExtensions
+    {
+        public static string GetWebJobsConnectionString(this IConfiguration configuration, string connectionStringName)
```
mathewc: Chose to migrate our current prefixing logic into an extension method.

test/Microsoft.Azure.WebJobs.EventHubs.UnitTests/PublicSurfaceTests.cs
```diff
@@ -9,7 +9,7 @@ namespace Microsoft.Azure.WebJobs.Host.UnitTests
     public class PublicSurfaceTests
     {
         [Fact]
-        public void EventHubsPublicSurface_LimitedToSpecificTypes()
+        public void WebJobs_Extensions_EventHubs_VerifyPublicSurfaceArea()
```
mathewc: As part of adding missing surface area tests and cleaning them up, I established a naming pattern for these to make viewing them in the test list easier

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/PublicSurfaceTests.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    /// <summary>
+    /// These tests help maintain our public surface area + dependencies. They will
+    /// fail any time new dependencies or public surface area are added, ensuring
+    /// we review such additions carefully.
+    /// </summary>
+    public class PublicSurfaceTests
+    {
+        [Fact]
+        public void WebJobs_Host_Storage_VerifyPublicSurfaceArea()
```
mathewc: We missed tests for this new assembly

test/Microsoft.Azure.WebJobs.Host.UnitTests/PublicSurfaceTests.cs
```diff
@@ -61,13 +62,44 @@ public void LoggingPublicSurface_LimitedToSpecificTypes()
...
         [Fact]
-        public void WebJobsHostPublicSurface_LimitedToSpecificTypes()
+        public void WebJobs_VerifyPublicSurfaceArea()
+        {
+            // The DLL containing the binding attributes should be truly minimal and have no extra dependencies. 
+            var assembly = typeof(AutoResolveAttribute).Assembly;
+
+            var expected = new[]
+            {
+                "AppSettingAttribute",
```
mathewc: During the recent refactorings, it appears that we lost the tests for the WebJobs core assembly

src/Microsoft.Azure.WebJobs.Host/Extensions/IConfigurationExtensions.cs
brettsam: Love it.

fabiocav: This does work a little different than what we had discussed, right?

This will expect connection strings to be defined in the root, under the connection strings section, as opposed to being in the extension scope/configuration section.

With this approach, are we proposing that all connection strings go under the root? We need to settle on this to avoid variation (which we have a bit of).

src/Microsoft.Azure.WebJobs.ServiceBus/Config/ServiceBusExtensionConfigProvider.cs
```diff
@@ -22,25 +22,25 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Config
...
         /// Creates a new <see cref="ServiceBusExtensionConfigProvider"/> instance.
         /// </summary>
-        /// <param name="serviceBusConfig">The <see cref="ServiceBusOptions"></see> to use./></param>
-        public ServiceBusExtensionConfigProvider(IOptions<ServiceBusOptions> serviceBusConfig,
+        /// <param name="options">The <see cref="ServiceBusOptions"></see> to use./></param>
+        public ServiceBusExtensionConfigProvider(IOptions<ServiceBusOptions> options,
             MessagingProvider messagingProvider,
             INameResolver nameResolver,
-            IConnectionStringProvider connectionStringProvider,
+            IConfiguration configuration,
```
fabiocav: Would be nice to avoid having to depend on the whole configuration here. Ideally, the options instance we take in would come fully populated, with connection strings and all.

src/Microsoft.Azure.WebJobs.Host/Extensions/IConfigurationExtensions.cs
mathewc: This method just takes an IConfiguration which can be an extension config section. So the caller determines the scope. This method then checks connection strings then direct config for the value.

I did think about extensions with these changes, and wondered if I needed to add an overload taking an extension section name. However, all tests pass with these changes - it doesn't look like any extensions require anything else, so I didn't add anything else.

I'm in the process of pulling these updates into Extensions now and will verify everything works.

mathewc: Extensions all work with these changes - see PR https://github.com/Azure/azure-webjobs-sdk-extensions/pull/458.

PR #1844, opened 2018-08-13
Update Microsoft.Azure.EventHubs to 2.1.0

### LINKED ISSUE #1710: Update Event Hub extension to use Microsoft.Azure.EventHubs 2.0.0
There is a new major version of `Microsoft.Azure.EventHubs` (2.x) that has breaking changes. Customers that are trying to use the newest version of this library will need a corresponding extension that they can install.

pragnagopa:
Also updated WindowsAzure.Storage to v9.3.1
Fixes #1710


fabiocav:
🎉

PR #1850, opened 2018-08-21
Throw exception without losing stack trace

michaeltg17:
When using a FunctionInvocationFilterAttribute, exceptions are not keeping their stack trace due to this "thow exception".


mathewc:
Merged https://github.com/Azure/azure-webjobs-sdk/commit/da4c069c0e0eddcbaf6ef8a64f5986c6d786cc0a

PR #1863, opened 2018-08-29
Signing dev (v3.x) branch builds

mhoeger:


test/Microsoft.Azure.WebJobs.Host.UnitTests/Executors/DefaultHostIdProviderTests.cs
```diff
@@ -31,7 +31,7 @@ public async Task GetHostIdAsync_ReturnsExpectedResult()
             // computed should remain static. If this test is failing
             // it likely means we've changed the ID computation algorithm
             // which would be a BREAKING CHANGE
-            string expected = "43f86080ae3050da38a314521eb5f1c2";
+            string expected = "89810a97b95cb5c71e5b7915bb172e5d";
```
mhoeger: We are changing the expected value because the assembly name has changed with strong name signing.

PR #1885, opened 2018-09-06
Disabling a function by FunctionNameAttribute.

alrod:


src/Microsoft.Azure.WebJobs.Host/Listeners/HostListenerFactory.cs
```diff
@@ -78,41 +78,50 @@ public async Task<IListener> CreateAsync(CancellationToken cancellationToken)
 
         internal static bool IsDisabled(MethodInfo method, INameResolver nameResolver, IJobActivator activator)
         {
-            // First try to resolve disabled state by setting
-            string settingName = string.Format(CultureInfo.InvariantCulture, "AzureWebJobs.{0}.Disabled", method.Name);
+            // First try to resolve disabled state by setting and method name
```
mathewc: We don't want to have two ways to disable the same function right? If there is a FunctionName attribute then that's the name we should use, otherwise we use the MethodInfo name.

So I think you should add a Utility.GetFunctionName(MethodInfo) helper that gets the right name, then your existing logic remains pretty much the same.

src/Microsoft.Azure.WebJobs.Host/Utility.cs
```diff
@@ -55,6 +57,17 @@ internal static string GetExtensionAliasFromTypeName(string typeName)
             return typeName;
         }
 
+        internal static string GetFunctionName(MethodInfo methodInfo)
+        {
+            FunctionNameAttribute functionNameAttribute = TypeUtility.GetHierarchicalAttributeOrNull<FunctionNameAttribute>(methodInfo);
+            string settingName = (functionNameAttribute != null) ? functionNameAttribute.Name : methodInfo.Name;
+            return string.Format(CultureInfo.InvariantCulture, "AzureWebJobs.{0}.Disabled", settingName);
+        }
+        private static string GetDisabledSettingName(string name)
```
mathewc: No longer used. Remove

paulbatum:
Resolved https://github.com/Azure/azure-functions-host/issues/3200

PR #1894, opened 2018-09-11
Updating Service Bus package name to Microsoft.Azure.WebJobs.Extensions.ServiceBus

fabiocav:


src/Microsoft.Azure.WebJobs.ServiceBus/WebJobs.ServiceBus.csproj
```diff
@@ -6,6 +6,8 @@
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
     <AssemblyName>Microsoft.Azure.WebJobs.ServiceBus</AssemblyName>
+    <PackageId>Microsoft.Azure.WebJobs.Extensions.ServiceBus</PackageId>
```
mathewc: So both EventHubs and ServiceBus packages now have the right Extensions naming, while their namespace/assembly name remain Microsoft.Azure.WebJobs.EventHubs and Microsoft.Azure.WebJobs.ServiceBus.

As a nit - could you change the project names to WebJobs.Extensions.EventHubs and WebJobs.Extensions.ServiceBus to match WebJobs.Extensions.Storage?

fabiocav:
Was chatting with @jeffhollan about this today. It is a tough one. We’d have to unlist the other package to minimize issues, but this will likely still lead to some confusion. On the other hand, it’s not great to leave it as is, as search quality is low and it’s harder to find this package when it doesn’t follow the same convention.

mathewc:
The ServiceBus package is the only one that doesn't follow the naming convention at this point. I think it's worth the short term pain to make it consistent with the rest of the packages as part of this release. I vote we just do it and get it over with, otherwise we have to deal with the inconsistency for perpetuity.

fabiocav:
I'm favoring the change as well. Even if it adds a bit of pain and confusion now, it will avoid confusion in the long run.

Issue #1869, opened 2018-08-31
Development Settings Configuration regression


mathewc:
In v2 we had a config method UseDevelopmentSettings (code [here](https://github.com/Azure/azure-webjobs-sdk/blob/447a67bb6af15b80485ea71603a4448515e22c12/src/Microsoft.Azure.WebJobs.Host/JobHostConfiguration.cs#L346)) that could be called to put the host in development mode, optimal for testing (e.g. reduced queue polling interval, etc.). Note that this will also affect blob trigger functions, and timer trigger functions.

Doesn't look like this was carried forward in the v3 migration. E.g. I see a bunch of commented out tests that are supposed to test this configuration: https://github.com/Azure/azure-webjobs-sdk/blob/dev/test/Microsoft.Azure.WebJobs.Host.FunctionalTests/JobHostOptionsTests.cs#L52.

This also affects Azure Functions, since AF configures debug mode on the host dynamically, based on portal connection.

PR #1907, opened 2018-09-13
Fixing Development environment options config (#1869)

mathewc:
Addresses https://github.com/Azure/azure-webjobs-sdk/issues/1869

Without these changes the development mode experience isn't very good. E.g. by default:

- after a short period of time the queue polling interval exponentially backs off to 1 minute
- the queue polling interval affects blob trigger as well
- timers can experience issues because the default listener lock period is 1 minute, meaning if you kill the local host and restart it, it might take up to a minute for it to acquire the lease

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsServiceCollectionExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsServiceCollectionExtensions.cs
@@ -13,10 +13,11 @@ using Microsoft.Azure.WebJobs.Host.Loggers;
 using Microsoft.Azure.WebJobs.Host.Timers;
 using Microsoft.Azure.WebJobs.Host.Triggers;
 using Microsoft.Azure.WebJobs.Logging;
+using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Hosting;
-
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.Azure.WebJobs
 {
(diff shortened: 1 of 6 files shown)


src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Config/QueuesOptionsSetup.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    internal class QueuesOptionsSetup : IConfigureOptions<QueuesOptions>
```
brettsam: If you want to avoid these tiny setup classes, you can use this if you want to configure inline with dependencies: https://twitter.com/davidfowl/status/1004248179628392448. 
But both ways work -- I know extra classes can be annoying sometimes.

brettsam (approved):

src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Config/QueuesOptionsSetup.cs
mathewc: Funny - I spent a bit spelunking and looking online for a way to do that but didn't find it and wound up with what I have.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsServiceCollectionExtensions.cs
```diff
@@ -97,6 +99,20 @@ public static IWebJobsBuilder AddWebJobs(this IServiceCollection services, Actio
...
+                {
+                    if (env.IsDevelopment())
+                    {
+                        options.ListenerLockPeriod = TimeSpan.FromSeconds(15);
+                    }
+                })
+                .Configure<IConfiguration>((options, config) =>
+                {
+                    var section = config.GetWebJobsRootConfiguration().GetSection(SingletonConfigSectionName);
+                    section.Bind(options);
```
mathewc: This bind was missing before - so we weren't populating from config

brettsam (approved):

PR #1910, opened 2018-09-15
Adding support for [ConnectionString] attribute on binding attribute properties

brettsam:
Allows binding attributes to set `[ConnectionString]` (as opposed to `[AppSetting]`, which instructs the attribute cloner to look up the key with `IConfiguration.GetConnectionString()` first, before falling back to the root configuration.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubTriggerAttribute.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubTriggerAttribute.cs
@@ -11,7 +11,7 @@ namespace Microsoft.Azure.WebJobs
     /// </summary>
     [AttributeUsage(AttributeTargets.Parameter)]
     [Binding]
-    public sealed class EventHubTriggerAttribute : Attribute, IConnectionProvider
+    public sealed class EventHubTriggerAttribute : Attribute
     {
         /// <summary>
         /// Create an instance of this attribute.
@@ -19,7 +19,7 @@ namespace Microsoft.Azure.WebJobs
         /// <param name="eventHubName">Event hub to listen on for messages. </param>
         public EventHubTriggerAttribute(string eventHubName)
         {
-            this.EventHubName = eventHubName;
+            EventHubName = eventHubName;
         }
 
         /// <summary>
@@ -35,7 +35,6 @@ namespace Microsoft.Azure.WebJobs
         /// <summary>
         /// Gets or sets the optional app setting name that contains the Event Hub connection string. If missing, tries to use a registered event hub receiver.
         /// </summary>
-        [AppSetting]
         public string Connection { get; set; }
     }
 }
--- a/src/Microsoft.Azure.WebJobs.Extensions.EventHubs/Triggers/EventHubTriggerAttributeBindingProvider.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.EventHubs/Triggers/EventHubTriggerAttributeBindingProvider.cs
@@ -4,14 +4,14 @@
 using System;
 using System.Reflection;
 using System.Threading.Tasks;
+using Microsoft.Azure.EventHubs;
 using Microsoft.Azure.WebJobs.Host;
 using Microsoft.Azure.WebJobs.Host.Bindings;
 using Microsoft.Azure.WebJobs.Host.Listeners;
 using Microsoft.Azure.WebJobs.Host.Triggers;
-using Microsoft.Azure.EventHubs;
-using Microsoft.Extensions.Logging;
 using Microsoft.Azure.WebJobs.Logging;
 using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 
 namespace Microsoft.Azure.WebJobs.EventHubs
@@ -61,9 +61,10 @@ namespace Microsoft.Azure.WebJobs.EventHubs
 
             if (!string.IsNullOrWhiteSpace(attribute.Connection))
             {
-                _options.Value.AddReceiver(resolvedEventHubName, _nameResolver.Resolve(attribute.Connection));
+                var connectionString = _config.GetConnectionStringOrSetting(attribute.Connection);
+                _options.Value.AddReceiver(resolvedEventHubName, connectionString);
             }
-            
+
             var eventHostListener = _options.Value.GetEventProcessorHost(_config, resolvedEventHubName, resolvedConsumerGroup);
 
             Func<ListenerFactoryContext, bool, Task<IListener>> createListener =
@@ -74,7 +75,7 @@ namespace Microsoft.Azure.WebJobs.EventHubs
              };
 
             ITriggerBinding binding = BindingFactory.GetTriggerBinding(new EventHubTriggerBindingStrategy(), parameter, _converterManager, createListener);
-            return Task.FromResult<ITriggerBinding>(binding);         
+            return Task.FromResult<ITriggerBinding>(binding);
         }
     } // end class
 }
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
@@ -6,9 +6,9 @@ using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
 using System.Linq;
 using System.Reflection;
-using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Description;
 using Microsoft.Azure.WebJobs.Host.Bindings.Path;
+using Microsoft.Extensions.Configuration;
 using BindingData = System.Collections.Generic.IReadOnlyDictionary<string, object>;
 using BindingDataContract = System.Collections.Generic.IReadOnlyDictionary<string, System.Type>;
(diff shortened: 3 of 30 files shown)


src/Microsoft.Azure.WebJobs.Extensions.EventHubs/EventHubTriggerAttribute.cs
```diff
@@ -35,7 +35,6 @@ public EventHubTriggerAttribute(string eventHubName)
         /// <summary>
         /// Gets or sets the optional app setting name that contains the Event Hub connection string. If missing, tries to use a registered event hub receiver.
         /// </summary>
-        [AppSetting]
```
brettsam: Triggers do not go through the AttributeCloner as they do not use binding rules. This setting wasn't doing anything, so removing to reduce confusion.

src/Microsoft.Azure.WebJobs.Extensions.EventHubs/Triggers/EventHubTriggerAttributeBindingProvider.cs
```diff
@@ -61,9 +61,10 @@ public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext contex
 
             if (!string.IsNullOrWhiteSpace(attribute.Connection))
             {
-                _options.Value.AddReceiver(resolvedEventHubName, _nameResolver.Resolve(attribute.Connection));
+                var connectionString = _config.FindConnectionString(attribute.Connection);
+                _options.Value.AddReceiver(resolvedEventHubName, connectionString);
```
brettsam: The NameResolver does not check ConnectionStrings first. Since we're moving away from that for environment lookups, I'm going directly against IConfiguration now. As mentioned elsewhere, triggers do not go through the AttributeCloner, so they need to resolve these things themselves.

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -94,39 +96,48 @@ private BindingDataResolver GetResolver(PropertyInfo propInfo, INameResolver nam
...
+                throw new InvalidOperationException($"{nameof(ConnectionStringAttribute)}, {nameof(AutoResolveAttribute)}, or {nameof(AppSettingAttribute)} property '{propInfo.Name}' must be of type string.");
             }
+
             var str = (string)originalValue;
 
-            // first try to resolve with app setting
+            // first try to resolve with connection string
+            if (connStrAttr != null)
+            {
+                return GetConfigurationResolver(str, connStrAttr.Default, propInfo, validator, s => _configuration.FindConnectionString(s));
```
brettsam: Both [AppSetting] and [ConnectionString] now go directly against IConfiguration. A lot of the changes here were just plumbing that type through to the AttributeCloner.

src/Microsoft.Azure.WebJobs.Host/DefaultNameResolver.cs
```diff
@@ -11,14 +12,21 @@ namespace Microsoft.Azure.WebJobs
...
         /// <summary>
         /// Resolves tokens by looking first in App Settings and then in environment variables.
         /// </summary>
         /// <param name="name">The token to resolve.</param>
         /// <returns>The token value from App Settings or environment variables. If the token is not found, null is returned.</returns>
         public virtual string Resolve(string name)
         {
-            return ConfigurationUtility.GetSetting(name);
+            return _configuration[name];
         }
```
brettsam: The DefaultNameResolver now goes directly against IConfiguration. ConfigurationUtility should be removed, but for here I just will avoid using it.

src/Microsoft.Azure.WebJobs.Host/Extensions/IConfigurationExtensions.cs
```diff
@@ -69,7 +69,13 @@ public static string GetPrefixedConnectionStringName(string connectionStringName
...
         }
 
-        private static string FindConnectionString(IConfiguration configuration, string connectionName) =>
+        /// <summary>
+        /// Looks for a connection string by first checking the ConfigurationStrings section, and then the root.
+        /// </summary>
+        /// <param name="configuration">The configuration.</param>
+        /// <param name="connectionName">The connection string key.</param>
+        /// <returns></returns>
+        public static string FindConnectionString(this IConfiguration configuration, string connectionName) =>
```
brettsam: I exposed this helper publicly as I was using this pattern in several places.

test/Microsoft.Azure.WebJobs.Host.TestCommon/TestConfigurationBuilderExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+        /// <summary>
+        /// Allows configuration to come from %userprofile%\.azurefunctions\appsettings.tests.json
+        /// </summary>
+        /// <param name="builder">The configuration builder.</param>
+        /// <returns>The modified configuration builder.</returns>
+        public static IConfigurationBuilder AddTestSettings(this IConfigurationBuilder builder)
+        {
+            string configPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), ".azurefunctions", "appsettings.tests.json");
+            return builder.AddJsonFile(configPath, true);
+        }
```
brettsam: Added this in a bunch of places. Made testing variations manually a ton quicker.

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -94,39 +96,48 @@ private BindingDataResolver GetResolver(PropertyInfo propInfo, INameResolver nam
...
             {
-                throw new InvalidOperationException($"Property '{propInfo.Name}' cannot be annotated with both AppSetting and AutoResolve.");
+                throw new InvalidOperationException($"Property '{propInfo.Name}' can only be annotated with one of the types {nameof(AppSettingAttribute)}, {nameof(AutoResolveAttribute)}, and {nameof(ConnectionStringAttribute)}.");
             }
 
             // attributes only work on string properties. 
             if (propInfo.PropertyType != typeof(string))
             {
-                throw new InvalidOperationException($"AutoResolve or AppSetting property '{propInfo.Name}' must be of type string.");
+                throw new InvalidOperationException($"{nameof(ConnectionStringAttribute)}, {nameof(AutoResolveAttribute)}, or {nameof(AppSettingAttribute)} property '{propInfo.Name}' must be of type string.");
```
fabiocav: Eventually, it would be nice to have an analyzer for this restriction to make sure we're failing at compile time.

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -151,21 +162,25 @@ internal BindingDataResolver GetAutoResolveResolver(string originalValue, AutoRe
...
 
-        // Apply AppSetting attribute. 
-        internal static BindingDataResolver GetAppSettingResolver(string originalValue, AppSettingAttribute attr, INameResolver nameResolver, PropertyInfo propInfo, Validator validator)
+        // Both ConnectionString and AppSetting have the same behavior, but perform the lookup differently.
+        internal static BindingDataResolver GetConfigurationResolver(string propertyValue, string defaultValue, PropertyInfo propInfo, Validator validator, Func<string, string> resolveValue)
         {
-            string appSettingName = originalValue ?? attr.Default;
-            string resolvedValue = string.IsNullOrEmpty(appSettingName) ?
-                originalValue : nameResolver.Resolve(appSettingName);
+            string connectionStringName = propertyValue ?? defaultValue;
```
fabiocav: nit: this should probably be named something like `configurationKey` or `configurationName` for clarity, as it won't always be a connection string.

src/Microsoft.Azure.WebJobs.Host/DefaultNameResolver.cs
fabiocav: Thanks! We should probably mark that type as obsolete with a proper description. We want to start moving away from it as it really provides no value today.

src/Microsoft.Azure.WebJobs.Host/Extensions/IConfigurationExtensions.cs
fabiocav: nit: The name wasn't super clear to me at first, I don't have a great suggestion, but can we name this something more obvious like `GetConnectionStringOrSetting` ?

test/Microsoft.Azure.WebJobs.Host.TestCommon/TestConfigurationBuilderExtensions.cs
fabiocav: Nice to take advantage of this in tests. We need to document the ability to do this so devs are aware as this is super helpful.

src/Microsoft.Azure.WebJobs.Host/Extensions/IConfigurationExtensions.cs
brettsam: Done. Used `GetConnectionStringOrSetting`.

src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
brettsam: Done.

Issue #1702, opened 2018-05-16
Newtonsoft.Json.JsonReaderException exception on all queue messages.


FrikkinLazer:
'Newtonsoft.Json.JsonReaderException' thrown from JsonSerialization.cs on any queue message received.
The message does get processed successfully though, despite the exception.

#### Repro steps
I have attached a zip with the solution as well.

Packages:
```
<PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.0-beta5" />
    <PackageReference Include="System.Configuration.ConfigurationManager" Version="4.4.1" />  
```  

Here is the function:
```
public static void TEST1(
            [QueueTrigger("testqueue")] string queueItem
        ) {
            Console.WriteLine(queueItem);
        }
```

The code inside the function gets called successfully. But for every message received on the queue this exception is also thrown:

An exception of type 'Newtonsoft.Json.JsonReaderException' occurred in Newtonsoft.Json.dll but was not handled in user code: 'Error reading JObject from JsonReader. Current JsonReader item is not an object: Integer. Path '', line 1, position 4.'
   at Newtonsoft.Json.Linq.JObject.Load(JsonReader reader, JsonLoadSettings settings)
   at Microsoft.Azure.WebJobs.Host.Protocols.JsonSerialization.ParseJObject(String json) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Protocols\JsonSerialization.cs:line 102
   at Microsoft.Azure.WebJobs.Host.Queues.QueueCausalityManager.GetOwner(IStorageQueueMessage msg) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Queues\QueueCausalityManager.cs:line 56

#### Expected behavior

The exception should not be thrown.

#### Known workarounds

Ignore the exception.

#### Related information 

Provide any related information 

[coregeneral.zip](https://github.com/Azure/azure-webjobs-sdk/files/2009109/coregeneral.zip)

paulbatum:
Do you have first chance exceptions turned on in VS? This exception is thrown and caught as part of the pipeline that attempts to determine if the payload is JSON and whether it should be deserialized as such.

FrikkinLazer:
In VS Code, I have "All Exceptions" unticked, and "User-Unhandled Exceptions" ticked.

paulbatum:
[...]

PR #1930, opened 2018-09-25
Adding try behavior for loading json reader

maiqbal11:
Adding try behavior when loading JObject to improve user experience with respect to #1702.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/QueueCausalityManager.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/QueueCausalityManager.cs
@@ -59,7 +59,7 @@ namespace Microsoft.Azure.WebJobs.Host.Queues
                 return null;
             }
 
-            if (json == null | !json.ContainsKey(ParentGuidFieldName) || json[ParentGuidFieldName].Type != JTokenType.String)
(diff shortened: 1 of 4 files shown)


paulbatum:
As discussed, I suspect this is going to have the same behavior as the previous implementation in that an exception is still thrown and caught.  I'll wait for an updated PR.

src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/QueueCausalityManager.cs
```diff
@@ -59,6 +59,11 @@ public static void SetOwner(Guid functionOwner, JObject token)
                 return null;
             }
 
+            if (json == null)
```
mathewc: nit: just move this check into the if clause below

src/Microsoft.Azure.WebJobs.Protocols/JsonSerialization.cs
```diff
@@ -85,6 +85,17 @@ internal static JsonTextWriter CreateJsonTextWriter(TextWriter textWriter)
             return jsonWriter;
         }
 
+        public static bool IsValidJObjectInput(string input)
```
mathewc: nit: recommend renaming this to IsJsonObject - we're not really checking whether it's valid json, so "valid" is a little strong

PR #1939, opened 2018-10-01
fixing JobHostFunctionTimeoutOptions usage

brettsam:
We weren't passing the default timeout values into our Executor. This means that any changes to `JobHostFunctionTimeoutOptions` wasn't honored. This is how functions handles precompiled timeouts, so it's a bug that needs fixing.

Note that there's a handful of formatting fixes that VS auto-performed for me (removing `this.` and removing unnecessary casts). I love getting these in when I can, but let me know if I need to back them out.


brettsam:
V1 precompiled works at a basic level, at least (I just double-checked). This bug meant that any precompiled function in v2 wasn't using a timeout.

PR #1943, opened 2018-10-03
removing extra Console logging

brettsam:
Fixes https://github.com/Azure/azure-functions-host/issues/3432.

When dashboard was disabled, we were still writing function logs to the Console. This should all be a no-op so removing the Console logging bits completely and replacing with Null versions.


brettsam:
FYI -- before this change if you ran our sample (which calls `AddConsole()` to add a console logger) with a simple function like:

```
public void ProcessWorkItem([QueueTrigger("test")] string message, ILogger logger)
{
    logger.LogInformation($"Processed work item {message}");
}
```

You'd get a double-log in the console:
```
Processed work item ddd
info: Function.ProcessWorkItem.User[0]
      Processed work item ddd
```

Now you just get one log:
```
info: Function.ProcessWorkItem.User[0]
      Processed work item ddd
```

Issue #1876, opened 2018-09-04
[AppInsights] -- Dashboard calls are appearing as dependency failures


brettsam:
@lmolkova FYI.

Ideally these calls would be filtered out as they're part of our host operations and not technically failures. But they happen in the context of the function execution, so they're appearing today. Most people using AI should disable the dashboard, but filing this to track as multiple people have mentioned it.

See https://github.com/Azure/azure-functions-host/issues/3363#issuecomment-417759496.

lmolkova:
@brettsam is there anything special about this calls that we may filter based on? Or can we add another scope to dashboard calls to distinguish them from 'user' calls?

brettsam:
We probably want to add something more generic to cover non-dashboard calls too (I see another one with Singleton that I need to narrow down) -- like maybe a "HostCall" scope value. If that thing is present, it wins out and we ignore the call for dependency tracking. We can then use that wherever we see something like this happening in the scope of an executing function.

harshithkashyap:
@brettsam I found a couple of more issues with the 409 errors.
1. Only HTTP triggered functions are running. Queue and Timer triggered functions do not start. Although manually triggering the queue/timer functions using `/admin/functions` endpoint works.
2. Functions runtime keeps spawning new instances even when there is no load on the application.

I've emailed you the name of our function app.

brandonh-msft:
Add me to the list of those experiencing 4xx errors from storage associated w/ a V2 app. So far I've seen 404, 409, and 412 (had to look that one up! :) )

harshithkashyap:
@brettsam Any update on this issue? We are still seeing the issue on the GA release. I'm curious to see this issue added to Backlog milestone and wasn't included in the GA release.

ianrathbone:
Can confirm also seeing a bunch of 409 PUT errors from storage on a V2 app (Runtime version: 2.0.12115.0).

Container: azure-webjobs-hosts/locks/

muratboduroglu:
I am also on the same boat. Just switched a sample benchmark app V1 to V2 and currently seeing a huge number of httprequest failures, including 404's, 409's and socket connection errors. In V1, all http connections beyond the pool limit (in consumption plan I assume this was 50) were throttled/queued (default behavior of IIS). It looks like in V2 the connections beyond the maximum allowed limit fail instantly. 

Sample socket error msg:
"Exception message	An attempt was made to access a socket in a way forbidden by its access permissions"

[...]

PR #1946, opened 2018-10-04
[AppInsights] improving host dependency tracking

brettsam:
Improves #1876 (but may not catch everything).

Notably:
- Moves dependency filtering in front of the QuickPulse processor. If anything does not have a Category (which we explicitly put on bindings so they show up), it won't show up in the QuickPulse either.
- Fixes one specific instance of a common 409 in a binding. It's harmless but looks bad. Instead, switching to call `ExistsAsync()`, which will almost always return a 200 which will not be as alarming.


src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Processors/DependencyFilteringTelemetryProcessor.cs
```diff
@@ -0,0 +1,34 @@
...
+using Microsoft.ApplicationInsights.Extensibility;
+using Microsoft.ApplicationInsights.Extensibility.Implementation;
+
+namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
+{
+    internal class DependencyFilteringTelemetryProcessor : ITelemetryProcessor
+    {
+        private readonly ITelemetryProcessor _next;
+
+        public DependencyFilteringTelemetryProcessor(ITelemetryProcessor next)
```
lmolkova: `DependencyFilteringTelemetryProcessor` may be confusing as this processor filters *requests* and dependencies,   how about `OperationTelemteryFilteringProcessor`?

brettsam: Good point. Will update.

brettsam:
Gonna go ahead and merge this since I'm getting lots of comments on the ugly Dependency "failures" showing up in QuickPulse and @lmolkova has signed off. @mathewc, @fabiocav, feel free to comment if you have concerns and I can address them.

PR #1957, opened 2018-10-10
Fix default operation name on ServiceBus requests

lmolkova:
Because of https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/1038 some ServiceBus requests do not have operation name set to function name.

This happens because AppInsights stops the Activity prematurely. It will be fixed in the next ApplicationInsights SDK release.

This change provides workaround for this issue. It could be safely removed with update of AI SDK.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
@@ -95,13 +95,18 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
                                 request.Context.Operation.Name = tag.Value;
                                 request.Name = tag.Value;
                                 break;
-                            case LogConstants.DurationKey:
-                                request.Properties[LogConstants.FunctionExecutionTimeKey] = tag.Value;
-                                break;
                             default:
                                 request.Properties[tag.Key] = tag.Value;
                                 break;
                         }
+                    }
+                }
+                else // workaround for https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/1038
+                {
+                    if (request.Properties.TryGetValue(LogConstants.NameKey, out var functionName))
+                    {
+                        request.Context.Operation.Name = functionName;
+                        request.Name = functionName;
                     }
                 }
             }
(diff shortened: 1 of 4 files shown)


lmolkova:
@brettsam please review

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/ServiceBusRequestAndDependencyCollectionTests.cs
```diff
@@ -78,6 +79,33 @@ await host.GetJobHost()
             ValidateServiceBusRequest(sbTriggerRequest, _endpoint, _queueName, nameof(ServiceBusTrigger), operationId, sbOutDependency.Id);
         }
 
+        [Fact]
+        public async Task ServiceBusRequestWithoutParent()
+        {
+            using (var host = ConfigureHost())
+            {
+                var sender = new MessageSender(_connectionString, _queueName);
+                await sender.SendAsync(new Message { Body = Encoding.UTF8.GetBytes("message"), ContentType = "text/plain" });
```
brettsam: This test fails without your change? If so, looks great!

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
```diff
@@ -95,15 +95,20 @@ public void Initialize(ITelemetry telemetry)
                                 request.Context.Operation.Name = tag.Value;
                                 request.Name = tag.Value;
                                 break;
-                            case LogConstants.DurationKey:
-                                request.Properties[LogConstants.FunctionExecutionTimeKey] = tag.Value;
-                                break;
```
brettsam: Why did this get removed?

brettsam:
FYI -- this will address https://github.com/Azure/azure-functions-host/issues/3585.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/ServiceBusRequestAndDependencyCollectionTests.cs
lmolkova: you are right, this test did not repro the issue. I've fixed it -  I moved message sending to happen before AppInsights initializes in ConfigureHost and before it starts tracking calls and injects parent info into the message.

brettsam: Ok, great. So now we know we've got a good regression test.

lmolkova:
@brettsam thanks for the review! please merge

lmolkova:
> Can you squash this down to a single commit? Then I can merge quickly...

done!

PR #1958, opened 2018-10-10
Fix missing cloud role name on the heartbeats

lmolkova:
ApplicationInsights SDK sends heartbeat events that are used on the AppMap to map node to the particular Azure resource. Cloud role name was missing on heartbeats, so mapping was not possible.

This change ensures proper telemetry initializer that sets role name is added to TeletmryConfiguration.Active.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
@@ -14,7 +14,6 @@ using Microsoft.ApplicationInsights.Extensibility.Implementation.ApplicationId;
 using Microsoft.ApplicationInsights.Extensibility.PerfCounterCollector.QuickPulse;
 using Microsoft.ApplicationInsights.SnapshotCollector;
 using Microsoft.ApplicationInsights.WindowsServer;
-using Microsoft.ApplicationInsights.WindowsServer.Channel.Implementation;
 using Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel;
 using Microsoft.Azure.WebJobs;
 using Microsoft.Azure.WebJobs.Logging.ApplicationInsights;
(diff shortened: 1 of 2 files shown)


lmolkova:
@brettsam please review

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
```diff
@@ -83,6 +82,18 @@ public static IServiceCollection AddApplicationInsights(this IServiceCollection
...
+                // Because of https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/943
+                // we have to touch (and create) Active configuration before initializing telemetry modules
+                // Active configuration is used to report AppInsights heartbeats
+                // role environment telemetry initializer is needed to correlate heartbeats to particular host
+                if (!string.IsNullOrEmpty(options.InstrumentationKey) &&
+                    string.IsNullOrEmpty(TelemetryConfiguration.Active.InstrumentationKey))
+                {
+                    TelemetryConfiguration.Active.InstrumentationKey = options.InstrumentationKey;
+                    TelemetryConfiguration.Active.TelemetryInitializers.Add(
+                        new WebJobsRoleEnvironmentTelemetryInitializer());
```
brettsam: We add this initializer to our TelemetryConfiguration as well, right? Do we still need to do that?

lmolkova: yes, because heartbeats are reported through TelemetryConfiguration.Active: https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/943 👎

lmolkova:
hey @brettsam , could you please approve/merge it if you are ok with the change?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
brettsam: Oh! Now I get it. I thought `WebJobsRoleEnvironmentTelemetryInitializer` was the thing that did the hearbeat. Then I realized that's ours, duh. :-)

PR #1959, opened 2018-10-11
Set ApplicaitonInsights RoleInstance to function name

lmolkova:
ApplicationInsights uses RoleInstance on the AppMap to show a number of instances of a role.
Role is node on the app map and is described by RoleName (function site name). 

Now, the roleInstance is set to function instance id that looks like `55eae20aa29dec69f13da780338872e0b0cbfd220429444bad0527b131af78d3`.

It results in several issues in UX. 

E.g. every time function restarts or runs on a new host it results in increase of number of instances on the node. If there are multiple functions on the same host, they are still shown as one instance.

Performacne analysis blade allows to group telemetry by role instance and aggregate data. Having one roleInstance here is also not useful.

It makes sense for this number to match functions on the host rather than instances of functions. And having real number of function instances does not seem really useful in serverless architecture.

This change sets RoleInstance to the function name on all telemetry that is sent in the scope of the function. Other telemetries such as host start/stop events, heartbeats are still sent with function instance id. This should not affect AppMap/performance blade behavior because this telemetry does not participate in distributed tracing.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Constants/LogConstants.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Constants/LogConstants.cs
@@ -140,6 +140,11 @@ namespace Microsoft.Azure.WebJobs.Logging
         public const int MetricEventId = 1;
 
         /// <summary>
+        /// Gets the name of the key used to store a metric name.
+        /// </summary>
+        public const string MetricNameKey = "MetricName";
+
+        /// <summary>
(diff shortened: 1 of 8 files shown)


brettsam:
I believe this is used today by QuickPulse to show the current instance count, isn't it? Or does that use some other value?

Customers actually use that quite often to see if things are scaling out -- is there another way that they can track their actual scale-out? While it's not strictly necessary, it is still very useful to a lot of people when monitoring their system.

PR #1964, opened 2018-10-15
Suggested Code Cleanups

davidrevoledo:
Basically a very few things changed

- complete some `nameof` when a new exception is being thrown
- Change `String` to `string` at the first time I thought you were using String, Int64 but clearly are some `String` randomly there.

- Complete some missing configure await false for some async methods here
https://github.com/Azure/azure-webjobs-sdk/compare/dev...davidrevoledo:dev-CodeCleanUps#diff-cda30319ad04308eeb3d4afa64843fe0

- Changed return `await Task.FromResult(true)` just to return true as there is an existing async method.


azfuncgh:

PR #1965, opened 2018-10-16
Enable AppInsights http requests collection and W3C distributed tracing support

lmolkova:
This change brings correlation with upstream services for HttpTrigger.

ApplicationInsights is configured to:
- collect Http requests with extra info like URL and status code
- support W3C distributed tracing protocol 

Related to https://github.com/Azure/azure-functions-host/issues/2447

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
@@ -9,6 +9,7 @@ using System.Linq;
 using Microsoft.ApplicationInsights;
 using Microsoft.ApplicationInsights.DataContracts;
 using Microsoft.ApplicationInsights.Extensibility;
+using Microsoft.ApplicationInsights.W3C;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
@@ -18,6 +19,7 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
     internal class ApplicationInsightsLogger : ILogger
     {
         private readonly TelemetryClient _telemetryClient;
+        private readonly ApplicationInsightsLoggerOptions _loggerOptions;
         private readonly string _categoryName;
         private readonly bool _isUserFunction = false;
 
@@ -43,9 +45,10 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
                 OperationContext
             };
 
-        public ApplicationInsightsLogger(TelemetryClient client, string categoryName)
+        public ApplicationInsightsLogger(TelemetryClient client, string categoryName, ApplicationInsightsLoggerOptions loggerOptions)
         {
             _telemetryClient = client;
+            _loggerOptions = loggerOptions;
             _categoryName = categoryName ?? DefaultCategoryName;
             _isUserFunction = LogCategories.IsFunctionUserCategory(categoryName);
         }
@@ -450,6 +453,18 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
 
                     // We'll need to store this operation context so we can stop it when the function completes
                     IOperationHolder<RequestTelemetry> operation = _telemetryClient.StartOperation(request);
+
+#pragma warning disable 612, 618
+                    if (_loggerOptions.EnableW3CDistributedTracing)
+                    {
+                        // currently ApplicationInsights supports 2 parallel correlation schemes:
+                        // legacy and W3C, they both appear in telemetry. UX handles all differences in operation Ids. 
+                        // This will be resolved in next .NET SDK on Activity level.
+                        // This ensures W3C context is set on the Activity.
+                        Activity.Current?.GenerateW3CContext();
+                    }
+#pragma warning restore 612, 618
+
                     stateValues[OperationContext] = operation;
                 }
             }
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
@@ -8,12 +8,37 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
 {
     public class ApplicationInsightsLoggerOptions
     {
+        /// <summary>
+        /// Gets or sets Application Insights instrumentation key.
+        /// </summary>
         public string InstrumentationKey { get; set; }
 
+        /// <summary>
+        /// Gets or sets sampling settings.
+        /// </summary>
         public SamplingPercentageEstimatorSettings SamplingSettings { get; set; }
 
+        /// <summary>
+        /// Gets or sets snapshot collection options.
+        /// </summary>
         public SnapshotCollectorConfiguration SnapshotConfiguration { get; set; }
 
+        /// <summary>
+        /// Gets or sets authentication key for Quick Pulse (Live Metrics).
+        /// </summary>
         public string QuickPulseAuthenticationApiKey { get; set; }
+
+        /// <summary>
+        /// Gets or sets flag that enables support of W3C distributed tracing protocol
+        /// (and turns on legacy correlation schema).  Enabled by default.
+        /// </summary>
+        public bool EnableW3CDistributedTracing { get; set; } = true;
+
+        /// <summary>
+        /// Gets or sets a flag that enables injection of multi-component correlation headers into responses.
+        /// This allows Application Insights to construct an Application Map to  when several
+        /// instrumentation keys are used.  Disabled by default.
+        /// </summary>
+        public bool EnableResponseHeaderInjection { get; set; } = true;
     }
 }
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
@@ -5,7 +5,10 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Reflection.Metadata.Ecma335;
 using Microsoft.ApplicationInsights;
+using Microsoft.ApplicationInsights.AspNetCore;
+using Microsoft.ApplicationInsights.AspNetCore.Extensions;
 using Microsoft.ApplicationInsights.Channel;
 using Microsoft.ApplicationInsights.DependencyCollector;
 using Microsoft.ApplicationInsights.Extensibility;
@@ -13,11 +16,13 @@ using Microsoft.ApplicationInsights.Extensibility.Implementation;
 using Microsoft.ApplicationInsights.Extensibility.Implementation.ApplicationId;
 using Microsoft.ApplicationInsights.Extensibility.PerfCounterCollector.QuickPulse;
 using Microsoft.ApplicationInsights.SnapshotCollector;
+using Microsoft.ApplicationInsights.W3C;
 using Microsoft.ApplicationInsights.WindowsServer;
 using Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel;
 using Microsoft.Azure.WebJobs;
 using Microsoft.Azure.WebJobs.Logging.ApplicationInsights;
 using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Configuration;
 using Microsoft.Extensions.Options;
@@ -38,6 +43,8 @@ namespace Microsoft.Extensions.DependencyInjection
 
         public static IServiceCollection AddApplicationInsights(this IServiceCollection services)
         {
+            services.TryAddSingleton<ISdkVersionProvider, WebJobsSdkVersionProvider>();
+
             // Bind to the configuration section registered with 
             services.AddOptions<ApplicationInsightsLoggerOptions>()
                 .Configure<ILoggerProviderConfiguration<ApplicationInsightsLoggerProvider>>((options, config) =>
@@ -47,14 +54,17 @@ namespace Microsoft.Extensions.DependencyInjection
 
             services.AddSingleton<ITelemetryInitializer, HttpDependenciesParsingTelemetryInitializer>();
             services.AddSingleton<ITelemetryInitializer, WebJobsRoleEnvironmentTelemetryInitializer>();
-            services.AddSingleton<ITelemetryInitializer, WebJobsTelemetryInitializer>();
             services.AddSingleton<ITelemetryInitializer, WebJobsSanitizingInitializer>();
+            services.AddSingleton<ITelemetryInitializer, WebJobsTelemetryInitializer>();
+
             services.AddSingleton<ITelemetryModule, QuickPulseTelemetryModule>();
 
             services.AddSingleton<IApplicationIdProvider, ApplicationInsightsApplicationIdProvider>();
 
             services.AddSingleton<ITelemetryModule, DependencyTrackingTelemetryModule>(provider =>
             {
+                var options = provider.GetService<IOptions<ApplicationInsightsLoggerOptions>>().Value;
+
                 var dependencyCollector = new DependencyTrackingTelemetryModule();
                 var excludedDomains = dependencyCollector.ExcludeComponentCorrelationHttpHeadersOnDomains;
                 excludedDomains.Add("core.windows.net");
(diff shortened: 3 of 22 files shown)


lmolkova:
users may set extra request properties inside the function method with 

```
public static async Task<IActionResult> Run([HttpTrigger(AuthorizationLevel.Anonymous, "get", "post", Route = null)]HttpRequest req, ILogger log)
  {
   System.Diagnostics.Activity.Current?.AddTag("my custom property", req.Header["my-header"]);
   ....
}
```

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/ServiceBusRequestAndDependencyCollectionTests.cs
```diff
@@ -67,16 +66,27 @@ await host.GetJobHost()
...
+            string triggerOperationId = sbTriggerRequest.Context.Operation.Id;
 
-            ValidateServiceBusDependency(sbOutDependency, _endpoint, _queueName, "Send", nameof(ServiceBusOut), operationId, manualCallRequest.Id);
-            ValidateServiceBusRequest(sbTriggerRequest, _endpoint, _queueName, nameof(ServiceBusTrigger), operationId, sbOutDependency.Id);
+            // currently ApplicationInsights supports 2 parallel correlation schemes:
+            // legacy and W3C, they both appear in telemetry. UX handles all differences in operation Ids. 
+            // This will be resolved in next .NET SDK on Activity level
+            string dependencyLegacyId = sbOutDependency.Properties.Single(p => p.Key == "ai_legacyRequestId").Value;
+            string triggerCallLegacyRootId = sbTriggerRequest.Properties.Single(p => p.Key == "ai_legacyRootId").Value;
+            string manualCallLegacyRootId = manualCallRequest.Properties.Single(p => p.Key == "ai_legacyRootId").Value;
```
brettsam: These are validating the legacy correllation? Do we validate W3C anywhere?

brettsam:
> users may set extra request properties inside the function method with
> 
> ```cs
> public static async Task<IActionResult> Run([HttpTrigger(AuthorizationLevel.Anonymous, "get", "post", Route = null)]HttpRequest req, ILogger log)
> {
>    System.Diagnostics.Activity.Current?.AddTag("my custom property", req.Header["my-header"]);
>    ....
> }
> ```

Where do these properties end up? On the `RequestTelemetry`?

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/ApplicationInsightsEndToEndTests.cs
```diff
@@ -321,10 +326,57 @@ public async Task ApplicationInsights_OuterRequestIsTrackedOnce()
...
+                {
+                    ["traceparent"] = "00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01"
+                });
+
+                // simulate auto tracked HTTP incoming call
+                var activity = fakeAspNetCore.StartRequest(httpContext);
+
+                MethodInfo methodInfo = GetType().GetMethod(testName, BindingFlags.Public | BindingFlags.Static);
+                await host.GetJobHost().CallAsync(methodInfo, new { input = "input" });
+                fakeAspNetCore.StopRequest(activity, httpContext);
```
brettsam: One thing I'm not following fully -- It seems like our function Logger scope that sets the Category would have fallen out of scope by the time the RequestTelemetry is logged by App Insights. How is it getting that correct Category of Host.Results?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/WebJobs.Logging.ApplicationInsights.csproj
```diff
@@ -24,17 +24,18 @@
...
+  <PackageReference Include="Microsoft.ApplicationInsights.WindowsServer" Version="2.8.0" />
+  <PackageReference Include="Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel" Version="2.8.0" />
   <PackageReference Include="Microsoft.AspNetCore.Mvc.Abstractions" Version="2.1.0" />
   <PackageReference Include="Microsoft.AspNetCore.Mvc.Core" Version="2.1.0" />
   <PackageReference Include="StyleCop.Analyzers" Version="1.1.0-beta004">
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
-    <PackageReference Include="Microsoft.ApplicationInsights" Version="2.7.2" />
-    <PackageReference Include="Microsoft.ApplicationInsights.DependencyCollector" Version="2.7.2" />
+    <PackageReference Include="Microsoft.ApplicationInsights" Version="2.8.0" />
```
MS-TimothyMothra: @lmolkova We just released 2.8.1 today.
Since this hasn't merged yet, please update here :)

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
```diff
@@ -83,21 +90,24 @@ public void Initialize(ITelemetry telemetry)
             // we may track traces/dependencies after function scope ends - we don't want to update those
             else if (request != null)
             {
+                // set SDK version on all requests to mark that they are collected by webjobs/functions
+                request.Context.GetInternalContext().SdkVersion = sdkVersion;
```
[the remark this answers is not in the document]
lmolkova: > and to expose it via a setting that we really don't want anyone but us to touch
The problem here is that we already want functions host to set it. I understand that you somehow override it on the functions side, but I'm not sure how it works and whether it would work for auto-collected requests.

> Is this not set automatically since it's set on the TelemetryClient?
It's set on the TelemetryClient, and used when webjobs explicitly track telemetry through this client. Auto-collected telemetry is tracked using different instances of the client.
To track usage properly we need to stamp azurefunctions/webjobs on the requests. We probably need this configuration in the AI, so that custom internal version could be propagated to internal TelemetryClients. But for now we can only override whatever AI SDK sets automatically.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/ApplicationInsightsEndToEndTests.cs
```diff
@@ -734,5 +786,53 @@ public void Dispose()
         {
             _channel?.Dispose();
         }
+
+        class FakeAspNetCore
```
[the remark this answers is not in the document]
lmolkova: great idea!

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/ServiceBusRequestAndDependencyCollectionTests.cs
lmolkova: we do in http tests. We also test W3C extensively in AI SDK.

Also, this is ServiceBus tests and ServiceBus does not implement W3C just yet, but below in the same test there is tricky correlation validation. 

For the end users UX handles all this tricky things and end users should not notice there are 2 separate correlation schemas involved.

lmolkova:
> Where do these properties end up? On the RequestTelemetry?

Yes

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/ApplicationInsightsEndToEndTests.cs
lmolkova: The timeline is

0. AspNetCore starts a new Activity
1. AI starts tracking request (based on AspNetCore events). 
2. Function execution starts and sets function scope
3. Function execution ends and ApplicationInsightsLogger copies category to the Activity.Tags
4. AspNetCore sends Stop event and ApplicationInsights tracks Request
5. WebJobsTelemetryIntitializer copies Activity.Tags to Request.Properties
6. AspNEtCore stops Activity
7. FilteringProcessor checks category and lets request telemetry through as it has category

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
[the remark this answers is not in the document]
lmolkova: great idea, one caveat here is that TelemetryClient still has to have sdk version set up so that if customer reports any telemetry through it, it will get the version stamped.

Otherwise, for auto-collected dependencies/metrics/etc (but not requests) we should keep the generated SDK version for debugging.

I also want to make initializer easily replaceable, so it does not have any logic.

lmolkova:
@brettsam 
I think I've addressed all feedback, please have a look. Thanks!

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/ApplicationInsightsEndToEndTests.cs
brettsam: Ah ha yes! Step 3 was the one I was missing somehow.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
```diff
@@ -8,12 +8,37 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
...
+        /// <summary>
+        /// Gets or sets flag that enables support of W3C distributed tracing protocol
+        /// (and turns on legacy correlation schema).  Enabled by default.
+        /// </summary>
+        public bool EnableW3CDistributedTracing { get; set; } = true;
+
+        /// <summary>
+        /// Gets or sets flag that enables injection of multi-component correlation headers into response
+        /// which allows to build Application Map where several ApplicaitonInsights
+        /// instrumentation keys are used.  Disabled by default.
```
brettsam: Can we reprhase this slightly? Assuming my version still makes sense, that is :-)

        /// Gets or sets a flag that enables injection of multi-component correlation headers into responses.
        /// This allows Application Insights to construct an Application Map to  when several
        /// instrumentation keys are used.  Disabled by default.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
```diff
@@ -47,14 +55,18 @@ public static IServiceCollection AddApplicationInsights(this IServiceCollection
 
             services.AddSingleton<ITelemetryInitializer, HttpDependenciesParsingTelemetryInitializer>();
             services.AddSingleton<ITelemetryInitializer, WebJobsRoleEnvironmentTelemetryInitializer>();
-            services.AddSingleton<ITelemetryInitializer, WebJobsTelemetryInitializer>();
             services.AddSingleton<ITelemetryInitializer, WebJobsSanitizingInitializer>();
+            services.AddSingleton<ITelemetryInitializer, WebJobsTelemetryInitializer>();
+            services.AddSingleton<ITelemetryInitializer>(new SdkVersionRequestTelemetryInitializer(sdkVersion));
```
brettsam: Can you make this a factory so DI will "own" it if it ever needs disposing? 

            services.AddSingleton<ITelemetryInitializer>(_ => new SdkVersionRequestTelemetryInitializer(sdkVersion));

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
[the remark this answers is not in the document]
brettsam: So one other question here... what TelemetryInitializers do these internal TelemetryClients use? It sounds like they use the same ones as the ones we've registered?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/SdkVersionTelemetryInitializer.cs
```diff
@@ -0,0 +1,29 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.ApplicationInsights.Channel;
+using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.ApplicationInsights.Extensibility;
+using Microsoft.ApplicationInsights.Extensibility.Implementation;
+
+namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
+{
+    internal class SdkVersionRequestTelemetryInitializer : ITelemetryInitializer
```
brettsam: I'm having second thoughts on this now (even though I suggested it). This means that Functions/CLI will need to register an additional initializer to overwrite this one, which feels clunky.

However, I think it's fine for this initial commit, and when I pull it through to Functions, I may come back and revisit. Even having an extra class like an ISdkVersionProvider that we could register once and use in different places may be a better fit. Then each place we'd register our own and trust that it'd be set up correctly everywhere.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -450,6 +451,15 @@ private void StartTelemetryIfFunctionInvocation(IDictionary<string, object> stat
 
                     // We'll need to store this operation context so we can stop it when the function completes
                     IOperationHolder<RequestTelemetry> operation = _telemetryClient.StartOperation(request);
+
+#pragma warning disable 612, 618
+                    // currently ApplicationInsights supports 2 parallel correlation schemes:
+                    // legacy and W3C, they both appear in telemetry. UX handles all differences in operation Ids. 
+                    // This will be resolved in next .NET SDK on Activity level.
+                    // This ensures W3C context is set on the Activity.
+                    Activity.Current?.GenerateW3CContext();
```
brettsam: Elsewhere we check if options.EnableW3CDistributedTracing is enabled before adding initializers. Do we need to check that here or is this enabling something different?

lmolkova: good catch! it's harmless if W3C is disabled, but anyway, let's not do it.

lmolkova:
> So one other question here... what TelemetryInitializers do these internal TelemetryClients use? It sounds like they use the same ones as the ones we've registered?

yes, they use the same TelemteryConfiguration, i.e. the same initializers. Sdk version is different by design as we want to mark telemetry differently. Dependency's sdk version is set accordingly to the mechanism that captured them, request's sdk version explains framework/host they were captured on, telemetry captured manually has another version prefix. We rely on this for debugging and usage tracking.

lmolkova:
@brettsam could you please have a look again?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
brettsam: What were the complaints? Performance?

brettsam:
We really just have the one outstanding question on the default values for one parameter -- but let's get this in while we discuss. We just need to get it right in the next couple of weeks before the next release. Can you rebase/squash this down? Then I can take one final look and merge.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
lmolkova: [bandwidth](https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/613) and limited [memory](https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/739#issuecomment-367399581) on IoT devices.

lmolkova:
> We really just have the one outstanding question on the default values for one property -- but let's get this in while we discuss. We just need to get it right in the next couple of weeks before the next release. Can you rebase/squash this down? Then I can take one final look and merge.

Done!

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
brettsam: I think we should just leave it as "true" if that's what the SDK already has. Like you said, most will go through IoT Hub. As long as we document how to turn it off from Functions (via host.json -- we'll own that documentation), it should be fine. And that's more consistent across the SDKs -- and across our settings (both of these new `Enable...` properties default to true).

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ISdkVersionProvider.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Reflection;
+
+namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
+{
+    public interface ISdkVersionProvider
```
brettsam: Nice!

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
lmolkova: agree, being consistent is good. I just pushed this change

PR #1967, opened 2018-10-16
Functions v1: Update ApplicationInsights to 2.6.4 and enable heartbeats

lmolkova:
This change enables ApplicaitonInsights heartbeats.

It requires ApplicaitonInsights SDK update

AI is updated to 2.6.4 (minimum suitable version)
DiagnosticSource is consolidated across projects to 4.4.0 (this version is enforced by AI)

@brettsam please review

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/packages.config
+++ b/src/Microsoft.Azure.WebJobs.Host/packages.config
@@ -18,7 +18,7 @@
   <package id="System.ComponentModel.EventBasedAsync" version="4.3.0" targetFramework="net45" />
   <package id="System.Console" version="4.3.0" targetFramework="net46" />
   <package id="System.Diagnostics.Debug" version="4.3.0" targetFramework="net46" />
-  <package id="System.Diagnostics.DiagnosticSource" version="4.3.0" targetFramework="net46" />
+  <package id="System.Diagnostics.DiagnosticSource" version="4.4.0" targetFramework="net46" />
   <package id="System.Diagnostics.Tools" version="4.3.0" targetFramework="net46" />
   <package id="System.Diagnostics.Tracing" version="4.3.0" targetFramework="net46" />
   <package id="System.Dynamic.Runtime" version="4.3.0" targetFramework="net45" />
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/app.config
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/app.config
@@ -18,6 +18,14 @@
         <assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30ad4fe6b2a6aeed" culture="neutral" />
         <bindingRedirect oldVersion="0.0.0.0-9.0.0.0" newVersion="9.0.0.0" />
       </dependentAssembly>
+      <dependentAssembly>
+        <assemblyIdentity name="System.Diagnostics.DiagnosticSource" publicKeyToken="cc7b13ffcd2ddd51" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-4.0.2.0" newVersion="4.0.2.0" />
+      </dependentAssembly>
+      <dependentAssembly>
+        <assemblyIdentity name="System.Net.Http" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-4.1.1.0" newVersion="4.1.1.0" />
+      </dependentAssembly>
     </assemblyBinding>
   </runtime>
 </configuration>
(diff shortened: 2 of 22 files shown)


test/Microsoft.Azure.WebJobs.Host.UnitTests/WebJobs.Host.UnitTests.csproj
```diff
@@ -194,6 +194,7 @@
       <Private>True</Private>
     </Reference>
     <Reference Include="System.Web" />
+    <Reference Include="System.Web.Extensions" />
```
lmolkova: referenced by AI Server Telemetry Channel (https://github.com/Microsoft/ApplicationInsights-dotnet/issues/950)

src/Microsoft.Azure.WebJobs.Host/packages.config
```diff
@@ -18,7 +18,7 @@
   <package id="System.ComponentModel.EventBasedAsync" version="4.3.0" targetFramework="net45" />
   <package id="System.Console" version="4.3.0" targetFramework="net46" />
   <package id="System.Diagnostics.Debug" version="4.3.0" targetFramework="net46" />
-  <package id="System.Diagnostics.DiagnosticSource" version="4.3.0" targetFramework="net46" />
+  <package id="System.Diagnostics.DiagnosticSource" version="4.4.0" targetFramework="net46" />
```
brettsam: @fabiocav -- how do you feel about this consolidation? It'll happen anyway in Script since App Insights is referenced there... Just wanted to point it out.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/App.config
```diff
@@ -47,12 +47,16 @@
       </dependentAssembly>
       <dependentAssembly>
         <assemblyIdentity name="System.Diagnostics.DiagnosticSource" publicKeyToken="cc7b13ffcd2ddd51" culture="neutral" />
-        <bindingRedirect oldVersion="0.0.0.0-4.0.1.0" newVersion="4.0.1.0" />
+        <bindingRedirect oldVersion="0.0.0.0-4.0.3.0" newVersion="4.0.3.0" />
```
brettsam: Elsewhere the redirect is to `4.0.2.0`... is this one wrong, or are the others?

brettsam (commented):
Have a couple questions... and wanted @fabiocav to ack as well.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/App.config
lmolkova: this is a mistake, fixed

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/app.config
```diff
@@ -18,6 +18,14 @@
         <assemblyIdentity name="Newtonsoft.Json" publicKeyToken="30ad4fe6b2a6aeed" culture="neutral" />
         <bindingRedirect oldVersion="0.0.0.0-9.0.0.0" newVersion="9.0.0.0" />
       </dependentAssembly>
+      <dependentAssembly>
```
fabiocav: These are usually explicitly removed as the redirects are known to cause issues with NGEN. In script, downgrade in the redirects to avoid issues. We need to be cautious and perform a full NGEN validation if we're touching them.

brettsam: System.Diagnostics.DiagnosticSource should be okay -- we don't downgrade that one: https://github.com/Azure/azure-functions-host/blob/v1.x/src/WebJobs.Script.WebHost/Web.config#L237-L240

System.Net.Http, though, we do downgrade -- https://github.com/Azure/azure-functions-host/blob/v1.x/src/WebJobs.Script.WebHost/Web.config#L229-L232. I'm guessing we can't use this version or our NGEN/cold-start will suffer. We'll need to see if anything breaks when downgrading. We've not run into anything yet...

fabiocav: We don't currently downgrade `System.Diagnostics.DiagnosticSource`, but it was one of the dependencies that caused issues in the past, so we just need to be extra diligent with NGEN validation.

lmolkova:
@brettsam @fabiocav any update on this?

MS-TimothyMothra:
@lmolkova @brettsam @fabiocav What is blocking this PR?

- 2.4.1 was released 7/8/2017
- 2.6.1 was released 5/10/2018

At this point, we're blocking a year old SDK from replacing a two-year-old SDK.

Note that 2.10 is scheduled for release on 5/1/2019

TimothyMothra:
@brettsam @lmolkova Should this PR be abandoned?

My understanding is that Functions v1 is in maintenance mode as cited here: https://docs.microsoft.com/en-us/azure/azure-functions/functions-versions

brettsam:
V1 is in maintenance mode and we'll not be moving dependencies forward unless there's a great reason to.

PR #1969, opened 2018-10-17
Fixing copyright notice in common props

fabiocav:
Invalid copyright notice causing package validation errors.


fabiocav:
@mathewc @brettsam merging this as it's a required metadata change. Let me know if you have any comments.

PR #1788, opened 2018-07-03
Improvements to EventHub/ServiceBus error handling (#1784, #1760)

mathewc:
Fixes for the following issues:

- https://github.com/Azure/azure-webjobs-sdk/issues/1760
- https://github.com/Azure/azure-webjobs-sdk/issues/1784

PR #1971, opened 2018-10-18
Fixing memory leaks

brettsam:
Should fix https://github.com/Azure/azure-functions-host/issues/2907. There's 3 issues being fixed here:

#### Disposing of aggregator
We were never disposing the Aggregator, which uses a timer with a callback to itself. It implemented IDisposable, but we never called it.

#### Removing ConcurrentBag usage
ConcurrentBag stores objects in ThreadLocal storage to highly optimize for same thread read/writes. That’s interesting, but doesn’t explain why the objects are not getting collected.

So now I’m suspecting that we’re hitting this -- https://github.com/dotnet/corefx/issues/23068 (the issue is in the .NET Core repo, but I’d expect that the behavior is the same in full framework). Our object graphs in these bags is huge due to the captured ScriptHost. So I’m pretty certain we’ve got a circular reference in there that’s preventing these objects from getting cleaned up.

I switched the two places we used ConcurrentBag to use ICollections (with simple locking) and the leak went away. At first I implemented IDisposable and made sure we were clearing the dictionaries, which also worked, but finding the right way to dispose of the SharedQueueWatcher was… challenging in v1. I think removing ConcurrentBag altogether will work fine for us. It looks like we can also swap to ConcurrentQueue, which does not use ThreadLocal storage.

#### EventHub ExceptionReceived
We’re not unsubscribing from EventHub’s ExceptionReceived event.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
+++ b/src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
@@ -445,12 +445,12 @@ namespace Microsoft.Azure.WebJobs.ServiceBus
                 .AddConverter<EventData, byte[]>(ConvertEventData2Bytes);
 
             // register the background exception handler
-            MessagingExceptionHandler.Subscribe(_options, context.Trace, context.Config.LoggerFactory);
+            var exceptionHandler = MessagingExceptionHandler.Subscribe(_options, context.Trace, context.Config.LoggerFactory);
 
             // register our trigger binding provider
             INameResolver nameResolver = context.Config.NameResolver;
             IConverterManager cm = context.Config.GetService<IConverterManager>();
-            var triggerBindingProvider = new EventHubTriggerAttributeBindingProvider(nameResolver, cm, this);
+            var triggerBindingProvider = new EventHubTriggerAttributeBindingProvider(nameResolver, cm, this, exceptionHandler);
             context.AddBindingRule<EventHubTriggerAttribute>()
                 .BindToTrigger(triggerBindingProvider);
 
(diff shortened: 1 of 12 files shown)


src/Microsoft.Azure.WebJobs.ServiceBus/EventHubs/EventHubConfiguration.cs
```diff
@@ -445,12 +445,12 @@ void IExtensionConfigProvider.Initialize(ExtensionConfigContext context)
                 .AddConverter<EventData, byte[]>(ConvertEventData2Bytes);
 
             // register the background exception handler
-            MessagingExceptionHandler.Subscribe(_options, context.Trace, context.Config.LoggerFactory);
+            var exceptionHandler = MessagingExceptionHandler.Subscribe(_options, context.Trace, context.Config.LoggerFactory);
```
mathewc: This was a memory leak as well? I had considered this issue when I authored the code originally - see my comments in https://github.com/Azure/azure-webjobs-sdk/pull/1788/files.

brettsam: It certainly looked like it was -- I saw ScriptHosts hanging around and saw this in their GC roots. Let me take another look, though. Thanks for calling it out.

brettsam: I removed that change and ran through my scenario again to double-check -- this time with just one function with an EH Trigger. 

The ScriptHosts and EventHubListeners stay around and the roots in VS look like below. When I unsubscribe, things get cleaned up like I'd expect.

image

PR #1978, opened 2018-10-23
incrementing version to 3.0.2

brettsam:


brettsam:
@mathewc -- I'm going to go ahead and get this in so we don't block this release, but we need to update this test so it doesn't fail with every version increase we do.

PR #1984, opened 2018-10-24
Clean up code in AppInsightsLogger

Zordid:
Mainly converting from types to var, remove cluttering and misleading comments, reduce code duplication


fabiocav:
Thank you for the contribution, we won't be taking this as this does not comply with the standards used in the codebase.

PR #1993, opened 2018-10-29
Log messages for Queue and Service Bus

maiqbal11:


src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionData.cs
```diff
@@ -21,6 +21,11 @@ public class TriggeredFunctionData
         /// </summary>
         public object TriggerValue { get; set; }
 
+        /// <summary>
+        /// The ID for the message being processed. Not applicable for all trigger types.
+        /// </summary>
+        public string MessageId { get; set; }
```
mathewc: Let's not plumb this message ID through all these core types, just to get a log statement. Instead, just add a new log statement at the correct place in the listener for triggers that need it. E.g. for Queues [here](https://github.com/Azure/azure-webjobs-sdk/blob/554b7ba922be3a4e1f380034dc0c62d4efb2aa79/src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Listeners/QueueListener.cs#L263)

brettsam: We need to make sure that we can correlate the invocationId with the message id... I'm not sure we have an invocation id at that stage, do we?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -649,6 +649,7 @@ private FunctionStartedMessage CreateStartedMessageWithoutArguments(IFunctionIns
                 FunctionInstanceId = instance.Id,
                 Function = instance.FunctionDescriptor,
                 ParentId = instance.ParentId,
+                MessageId = instance.MessageId,
                 Reason = instance.Reason,
```
brettsam: I was hoping we could piggy-back on this Reason to add some additional context. Is that not possible? I'm not entirely sure how that thing gets populated.

maiqbal11: I don't think we'll be able to do that since the Reason is bubbled from the trigger as an Enum of type ExecutionReason. MessageId would need to be some kind of Guid or otherwise unique indentifier.

src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionData.cs
mathewc: Right. Not until the listener calls into the executor is the function instance ID known. How about we generalize this then, and add a new "string TriggerMessage" property here, which allows a listener to optionally specify a log message that will be written by the executor as part of the invocation. Then in FunctionExecutor IIF there is a value for this message it gets logged, as a new log line (not part of the executing/executed messages), and it gets logged under the same FunctionInvocationId.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
mathewc: Right - that is a higher level concept, like "manually invoked / HostCall", "automatically triggered", etc.

src/Microsoft.Azure.WebJobs.Host/Loggers/FunctionInstanceLogger.cs
```diff
@@ -22,7 +22,7 @@ public FunctionInstanceLogger(ILoggerFactory loggerFactory)
 
         public Task<string> LogFunctionStartedAsync(FunctionStartedMessage message, CancellationToken cancellationToken)
         {
-            string traceMessage = string.Format(CultureInfo.InvariantCulture, "Executing '{0}' (Reason='{1}', Id={2})", message.Function.ShortName, message.FormatReason(), message.FunctionInstanceId);
+            string traceMessage = string.Format(CultureInfo.InvariantCulture, "Executing '{0}' (Reason='{1}', Id={2}, MessageId={3})", message.Function.ShortName, message.FormatReason(), message.FunctionInstanceId, message.MessageId);
```
mathewc: Based on my suggestions above, if message.TriggerMessage is not null/empty, we'd log a new log line after this, like $"Trigger Details: {message.TriggerDetails}"

src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionData.cs
brettsam: That sounds good to me. How do we want to handle something like batched EH? List the count plus all the ids in the batch?

mathewc: Details need to be a very short summary, so you definitely can't have all the Event IDs in a batch. Here are some suggested details:

- Queues: "MessageId: X, DequeueCount: Y, InsertionTime: Z"
- ServiceBus: "MessageId: X, DeliveryCount: Y, EnqueuedTime: Z"
- EventHub
    - Batched: "BatchCount: X"
    - Single: "SequenceNumber: X, Offset: Y, EnqueuedTime: Z"

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusTriggerExecutor.cs
```diff
@@ -25,9 +25,14 @@ public async Task<FunctionResult> ExecuteAsync(Message value, CancellationToken
             {
                 ParentId = parentId,
                 TriggerValue = value,
-                MessageId = value.MessageId
+                TriggerDetails = TriggerDetails(value)
             };
             return await _innerExecutor.TryExecuteAsync(input, cancellationToken);
         }
+
+        private string TriggerDetails(Message value)
```
mathewc: nit: FormatTriggerDetails

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusTriggerExecutor.cs
```diff
@@ -24,9 +24,15 @@ public async Task<FunctionResult> ExecuteAsync(Message value, CancellationToken
...
-                TriggerValue = value
+                TriggerValue = value,
+                TriggerDetails = TriggerDetails(value)
             };
             return await _innerExecutor.TryExecuteAsync(input, cancellationToken);
         }
+
+        private string TriggerDetails(Message value)
+        {
+            return $"MessageId: {value.MessageId}, TimeToLive: {value.TimeToLive}, ScheduleEnqueueTime: {value.ScheduledEnqueueTimeUtc}";
```
mathewc: What about the details I suggested "MessageId: X, DeliveryCount: Y, EnqueuedTime: Z"? DeliveryCount would be useful. Also, we should use the EnqueuedTime from SystemProperties, not ScheduledEnqueueTime. Not sure if TTL would be useful - recommend replacing that with DeliveryCount (also from SystemProperties)

src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Listeners/QueueTriggerExecutor.cs
```diff
@@ -25,9 +25,15 @@ public async Task<FunctionResult> ExecuteAsync(CloudQueueMessage value, Cancella
...
             {
                 ParentId = parentId,
-                TriggerValue = value
+                TriggerValue = value, 
+                TriggerDetails = TriggerDetails(value)
             };
             return await _innerExecutor.TryExecuteAsync(input, cancellationToken);
         }
+
+        private string TriggerDetails(CloudQueueMessage value)
```
mathewc: nit: FormatTriggerDetails

src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionData.cs
```diff
@@ -21,6 +21,11 @@ public class TriggeredFunctionData
         /// </summary>
         public object TriggerValue { get; set; }
 
+        /// <summary>
+        /// Details of the trigger (e.g. Message ID, insertion time, deque count etc.)
```
mathewc: nit: spelling

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusTriggerExecutor.cs
maiqbal11: I wasn't sure where to pull those out from so I used what looked like the closest equivalents. I'll add them in from the SystemProperties.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusTriggerExecutor.cs
```diff
@@ -21,10 +22,17 @@ public ServiceBusTriggerExecutor(ITriggeredFunctionExecutor innerExecutor)
         public async Task<FunctionResult> ExecuteAsync(Message value, CancellationToken cancellationToken)
         {
             Guid? parentId = ServiceBusCausalityHelper.GetOwner(value);
+            ServiceBusTriggerDetails triggerDetails = new ServiceBusTriggerDetails
```
mathewc: These changes look good overall, though you might consider not adding these new types and running them through json serialization, since that's a performance hit on every invocation that isn't really needed. The string formatting you had in place before was just as good and more performant.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusTriggerExecutor.cs
```diff
@@ -24,9 +24,17 @@ public async Task<FunctionResult> ExecuteAsync(Message value, CancellationToken
...
             };
             return await _innerExecutor.TryExecuteAsync(input, cancellationToken);
         }
+
+        private string FormatTriggerDetails(Message value)
+        {
+            return $"MessageId: {value.MessageId}, " +
+                $"DeliveryCount: {value.SystemProperties.DeliveryCount}, " +
+                $"EnqueuedTime: {value.SystemProperties.EnqueuedTimeUtc}";
+        }
```
brettsam: One other thing that would be super-useful here would be `LockedUntilUtc`. I just had an issue that we've been debugging for a while and I think the issue is ultimately a too-large PrefetchCount. That means messages sit in the local cache and expire before they are sent to the function. If I could have seen that the lock was expiring soon, that would have helped.

src/Microsoft.Azure.WebJobs.Host/Executors/IFunctionInstance.cs
```diff
@@ -9,7 +9,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
     public interface IFunctionInstance
     {
         Guid Id { get; }
-
+        string TriggerDetails { get; }
```
fabiocav: nit: spacing

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInstance.cs
```diff
@@ -31,6 +33,11 @@ public Guid Id
             get { return _id; }
         }
 
+        public string TriggerDetails
```
fabiocav: Having this as a dictionary, as opposed to a string, would give us a bit more flexibility when using (logging) the data as it would allow us to pass the details as distinct arguments, making this easy to query for providers supporting structured logging. It would also ensure the logging is consistent with all the different triggers.

brettsam: Yes this would be a good idea. We should be doing more structured logging wherever we can. Our App Insights logger, for example, will break those structured properties into their own queryable columns so you don't have to parse (or use Contains()) when querying.

maiqbal11: Could elaborate more on how we can take advantage of the structured logging in App Insights? What I'm imagining right now is that I would pass a dictionary of values through the data types, but I'd have to use a ToString() method to convert to string before logging it (a string that has the same format as the one I'm using right now). Where would we plug into the App Insights in this case?

fabiocav: There isn’t anything specific to AI. Will follow up in person

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInstanceFactoryContext.cs
```diff
@@ -11,6 +11,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
     public class FunctionInstanceFactoryContext
     {
         public Guid Id { get; set; }
+        public Dictionary<string, string> TriggerDetails { get; set; }
```
fabiocav: Change to `IDictionary` (similar to `Parameters` below)

src/Microsoft.Azure.WebJobs.Host/Executors/IFunctionInstance.cs
```diff
@@ -10,6 +11,8 @@ public interface IFunctionInstance
     {
         Guid Id { get; }
 
+        Dictionary<string, string> TriggerDetails { get; }
```
fabiocav: Use `IDictionary<string, string>` here instead.

src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionData.cs
```diff
@@ -21,6 +22,11 @@ public class TriggeredFunctionData
         /// </summary>
         public object TriggerValue { get; set; }
 
+        /// <summary>
+        /// Details of the trigger (e.g. Message ID, insertion time, dequeue count etc.)
+        /// </summary>
+        public Dictionary<string, string> TriggerDetails { get; set; }
```
fabiocav: Use `IDictionary<string, string>` here instead.

src/Microsoft.Azure.WebJobs.Host/Loggers/FunctionInstanceLogger.cs
```diff
@@ -24,10 +26,27 @@ public Task<string> LogFunctionStartedAsync(FunctionStartedMessage message, Canc
...
+                LogTemplatizedTriggerDetails(message);
+            }
             return Task.FromResult<string>(null);
         }
 
+        private void LogTemplatizedTriggerDetails(FunctionStartedMessage message)
+        {
+            var templateKeys = message.TriggerDetails.Select(entry => $"{entry.Key}: {{{entry.Key}}}");
+            string messageTemplate = "Trigger Details: " + string.Join(", ", templateKeys);
+            string[] templateValues = message.TriggerDetails.Select(entry => $"{entry.Value}").ToArray();
```
fabiocav: No need for the select here. Use: 
`string[] templateValues = message.TriggerDetails.Values.ToArray();`

src/Microsoft.Azure.WebJobs.Host/Loggers/FunctionInstanceLogger.cs
```diff
@@ -24,10 +26,27 @@ public Task<string> LogFunctionStartedAsync(FunctionStartedMessage message, Canc
...
+            Log(LogLevel.Information, message.Function, messageTemplate, templateValues);
+        }
+
+        private void Log(LogLevel level, FunctionDescriptor descriptor, string message, params object[] args)
+        {
+            ILogger logger = _loggerFactory?.CreateLogger(LogCategories.CreateFunctionCategory(descriptor.LogName));
+            logger?.Log(level, 0, message, args);
```
fabiocav: Is there a reason why we're creating the logger here with every invocation?

maiqbal11: Given the way the code is currently structured, the logger factory takes the function's name as an argument and the logs make it clear which function the logs stem from. It would be a larger change to, perhaps, create a logger at the outset and then inject each function's name into it without using the logger factory. This is because the `ILogger`, from what I can tell, does not allow setting the function name (also known category name). @brettsam to comment on how we could do this differently.

fabiocav: Makes sense. This should reuse the logger internally, so we should be OK.

brettsam:
@fabiocav / @mathewc / @maiqbal11 -- should we make this change to `v2.x` as well? We'll be supporting that for a while and having these details will be really helpful. I'm working with a Functions v1 customer right now, for example, and keep having to ask her send me their App Insights logs because I can't track duplicate SB message ids myself.

mathewc:
Yes, that would be good, I agree

fabiocav:
Agreed. Backporting this makes sense.

maiqbal11:
I'll open a separate PR to get this into `v2.x` as well

PR #1997, opened 2018-11-01
bumping version number

brettsam:


brettsam:
@mathewc -- quick look at this? We should be bumping up right after a release (rather than right before) so our MyGet packages look newer. We just rarely do that :-)

mathewc:
Funny - I just did this for v2 https://github.com/Azure/azure-webjobs-sdk/commit/5759fc792190eb9da07b5300d976d2275d1e7168

mathewc:
What about all the other packages, e.g. Extensions?

brettsam:
Will do extensions as well.

PR #2006, opened 2018-11-07
adding more quickpulse logging

brettsam:
It looks like the dependency tracking can pick up some AppVeyor HTTP calls. I'm going to try to add more logging to figure out the correct fix here. 

We're seeing this URL get called during our runs occassionally: https://github.com/xunit/xunit/blob/9d10262a3694bb099ddd783d735aba2a7aac759d/src/xunit.runner.reporters/AppVeyorClient.cs#L21.


brettsam:
Closing as we determined the issue and fixed it elsewhere.

PR #2013, opened 2018-11-15
fixing issues with auto-tracking of Http and Service Bus triggers

### LINKED ISSUE #2012: [AppInsights] Auto-tracking of ServiceBusTrigger not recording failures
Since adding auto-tracking of ServiceBusTrigger through App Insights, failures are not being properly recorded.

brettsam:
While porting the new Http auto-tracking changes to Functions, I ran into a handful of issues that this commit now fixes:
- Auto-tracked Service Bus requests were getting "200" status code. We now set this to "0".
- The URL for auto-tracked Http requests contained the query string, which we want to prevent.
- Because we overwrite Http request `Name` with the function name, we had lost the HttpMethod. I now pull that out and put it into the `HttpMethod` property, like we did in v1.
- We were not looking at the function result when setting Success. This needs to be pulled from the `Suceeded` property and explicitly set on the telemetry. Fixes #2012.
- During host restarts in Functions, it was possible for us to see auto-tracked `RequestTelemetry` flow through our initializers and processors twice. This is because one TelemetryClient was shutting down while another was starting up, and they were overlapping. Since they both watch for requests with DiagnosticListeners, they were both taking the request through their pipelines. I improved this by
    - Making sure our initializers are idempotent. Even if a telemetry flows through twice, it ends up the same.
    - Setting an explicit property when telemetry has already gone through the processor pipeline. If we see it again, we return without sending it to App Insights.

Tests have been beefed up, and new ones written, to cover all of this.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
@@ -19,11 +19,16 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
         private const string WebSiteInstanceIdKey = "WEBSITE_INSTANCE_ID";
 
         private static readonly string _roleInstanceName = GetRoleInstanceName();
-        private readonly string sdkVersion;
+        private readonly string _sdkVersion;
 
         public WebJobsTelemetryInitializer(ISdkVersionProvider versionProvider)
         {
-            sdkVersion = versionProvider?.GetSdkVersion();
+            if (versionProvider == null)
+            {
+                throw new ArgumentNullException(nameof(versionProvider));
+            }
+
+            _sdkVersion = versionProvider.GetSdkVersion();
         }
 
         public void Initialize(ITelemetry telemetry)
@@ -90,20 +95,18 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
             // we may track traces/dependencies after function scope ends - we don't want to update those
             else if (request != null)
             {
-                request.Context.GetInternalContext().SdkVersion = sdkVersion;
+                UpdateRequestProperties(request);
+
                 Activity currentActivity = Activity.Current;
-                if (currentActivity != null) // should never be null, but we don't want to throw anyway
+                if (currentActivity != null)
                 {
-                    // tags is a list, we'll enumerate it
                     foreach (var tag in currentActivity.Tags)
                     {
-                        if (tag.Key == LogConstants.NameKey)
-                        {
-                            request.Context.Operation.Name = tag.Value;
-                            request.Name = tag.Value;
-                        }
-                        // ignore internal ai tags, but add custom properties
-                        else if (!tag.Key.StartsWith("w3c_") && !tag.Key.StartsWith("ai_"))
+                        // Apply well-known tags and custom properties, 
+                        // but ignore internal ai tags
+                        if (!TryApplyProperty(request, tag) &&
+                            !tag.Key.StartsWith("w3c_") &&
+                            !tag.Key.StartsWith("ai_"))
                         {
                             request.Properties[tag.Key] = tag.Value;
                         }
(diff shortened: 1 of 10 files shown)


src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
```diff
@@ -130,5 +132,69 @@ private static string GetRoleInstanceName()
...
+            }
+
+            // If the Url is not null, it's an actual HttpRequest, as opposed to a
+            // Service Bus or other function invocation that we're tracking as a Request
+            if (request.Url != null)
+            {
+                // We don't want to include the query string in URLs
+                request.Url = new Uri(request.Url.GetLeftPart(UriPartial.Path));
+
+                // App Insights sets this as 'VERB /url/path'. We want to extract the VERB.
```
fabiocav: Is this documented and guaranteed not to change? can we really rely on this being the same?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Processors/OperationFilteringTelemetryProcessor.cs
```diff
@@ -18,6 +21,24 @@ public OperationFilteringTelemetryProcessor(ITelemetryProcessor next)
...
         public void Process(ITelemetry item)
         {
+            // There are scenarios in Functions where we can have two TelemetryClients running for a short period.
+            // We use this property to ensure we only send one of the RequestTelemetry objects (auto-tracked by App 
+            // Insights) to the server.
+            if (item is RequestTelemetry request)
+            {
+                lock (_lock)
+                {
+                    if (request.Properties.ContainsKey(ProcessedKey))
```
fabiocav: nit: can we simplify this by setting if the key is not present instead (eliminate the `else`)

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Processors/OperationFilteringTelemetryProcessor.cs
```diff
@@ -18,6 +21,24 @@ public OperationFilteringTelemetryProcessor(ITelemetryProcessor next)
...
+                    {
+                        return;
+                    }
+                    else
+                    {
+                        request.Properties[ProcessedKey] = "1";
```
fabiocav: What is inspecting this property?

brettsam: This... directly above. if it's there, we stop processing it.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
brettsam: Per-version, yes. I have tests that validate it, so if we move forward and it breaks, we'll know and adjust. @lmolkova -- is there any other way for me to see the method?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Processors/OperationFilteringTelemetryProcessor.cs
brettsam: If it's there, we return -- thereby short-circuiting the telemetry (it won't be uploaded). Note we don't call `_next.Process()` in that case. So if it's there, exit.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Processors/OperationFilteringTelemetryProcessor.cs
fabiocav: I see the property being set if we don't return, but what actually "processes" the event/item?

fabiocav: Nevermind... my issue with the diff

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Processors/OperationFilteringTelemetryProcessor.cs
fabiocav: Sorry... had an issue with my diff (same reason for the question below), the else can still be removed, though :)

brettsam:
I'm going to split this up so that I can fix the SB succeeded/failed issue separately while we work on getting the rest of the HTTP tracking complete.

PR #2014, opened 2018-11-15
Revert "Enable AppInsights http requests collection and W3C distributed tracing support"

brettsam:
This reverts commit 24ff093604e8da25837b27e3a2bf3095314accaa

We found an issue when moving this into Functions and testing end-to-end. Requests coming from the Portal were not traced as expected and there was no quick-and-easy fix for it.
We'll bring this back once we've been able to do more testing.


fabiocav:
Looks like there's a test failure that seems related to the change

PR #2015, opened 2018-11-15
[AppInsights] fixing SB result-reporting

### LINKED ISSUE #2012: [AppInsights] Auto-tracking of ServiceBusTrigger not recording failures

brettsam:
Fixes #2012. 

Also re-adds `ISdkVersionProvider` from the reverted Http tracking commit, as it's a better way for us to update the version in multiple places.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using Microsoft.ApplicationInsights.Channel;
 using Microsoft.ApplicationInsights.DataContracts;
 using Microsoft.ApplicationInsights.Extensibility;
+using Microsoft.ApplicationInsights.Extensibility.Implementation;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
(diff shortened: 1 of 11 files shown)


test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/ServiceBusRequestAndDependencyCollectionTests.cs
```diff
@@ -139,7 +172,10 @@ public async Task ServiceBusRequestWithoutParent()
             Assert.True(double.TryParse(request.Properties[LogConstants.FunctionExecutionTimeKey], out double functionDuration));
             Assert.True(request.Duration.TotalMilliseconds >= functionDuration);
 
-            TelemetryValidationHelpers.ValidateRequest(request, operationName, operationId, parentId, LogCategories.Results);
+            Assert.DoesNotContain(request.Properties, p => p.Key == LogConstants.HttpMethodKey);
```
lmolkova: i think you validate it in the `TelemetryValidationHelpers.ValidateRequest`

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
```diff
@@ -122,5 +129,52 @@ private static string GetRoleInstanceName()
...
+                bool.TryParse(property.Value, out bool success))
+            {
+                // no matter what App Insights says about the response, we always
+                // want to use the function's result for Succeeeded
+                request.Success = success;
+                wasPropertySet = true;
+
+                // Remove the Succeeded property as it's duplicated
+                request.Properties.Remove(LogConstants.SucceededKey);
+            }
```
brettsam: FYI -- this part where we check the SucceededKey is the major change. It will be needed for Http as well.

lmolkova: I wonder if we can prevent it from being set on properties at the first place. It's set in the AppInsightsLogger. but now it's needed because of https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/1038 anyway

PR #2020, opened 2018-11-16
prototype for loggable options

brettsam:
Fixes https://github.com/Azure/azure-functions-host/issues/3507

Here's my thoughts on getting "loggable options". BTW -- I wanted to avoid using "Logger" all over here since that's a specific type and none of these are actually Loggers.

There's an example below of:
- logging a simple type (Queues)
- logging a nested type (ServiceBus)
- wrapping a system type (LoggerFilterOptions)

All API shapes up for discussion, but the mechanism works. When I run this locally, I get this on the console:

```
dbug: Microsoft.Azure.WebJobs.Hosting.LoggableOptionsService[0]
      LoggerFilterOptions
      {
        "MinLevel": 1,
        "Rules": [
          {
            "ProviderName": "Console",
            "CategoryName": "Host.Executor",
            "LogLevel": 6,
            "Filter": null
          }
        ]
      }
dbug: Microsoft.Azure.WebJobs.Hosting.LoggableOptionsService[0]
      QueuesOptions
      {
        "BatchSize": 16,
        "NewBatchThreshold": 8,
        "MaxPollingInterval": "00:00:02",
        "MaxDequeueCount": 5,
        "VisibilityTimeout": "00:00:00"
      }
dbug: Microsoft.Azure.WebJobs.Hosting.LoggableOptionsService[0]
      ServiceBusOptions
      {
        "PrefetchCount": 0,
        "MessageHandlerOptions": {
          "MaxConcurrentCalls": 16,
          "AutoComplete": true,
          "MaxAutoRenewDuration": "00:05:00"
        }
      }
```


brettsam:
FYI -- this is for https://github.com/Azure/azure-functions-host/issues/3507

src/Microsoft.Azure.WebJobs.Host/Hosting/LoggableOptions/ILoggableOptionsWrapper.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Hosting
+{
+    /// <summary>
+    /// Allows framework types to be registered as LoggableTypes.
+    /// </summary>
+    /// <typeparam name="TOptions">The options type.</typeparam>
+    internal interface ILoggableOptionsWrapper<TOptions> : ILoggableOptions where TOptions : class, new()
```
brettsam: I should change this to have it be `GetLoggableOptions(TOptions options)` rather than a two-step set-then-get.

brettsam:
@fabiocav / @mathewc -- any thoughts? I'm going to start writing tests and hardening tomorrow. Any early feedback welcome.

src/Microsoft.Azure.WebJobs.Host/Hosting/LoggableOptions/LoggerFilterOptionsLoggableWrapper.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Hosting
+{
+
+    internal class LoggerFilterOptionsLoggableType : ILoggableOptionsWrapper<LoggerFilterOptions>
```
mathewc: You have to wrap like this because we don't own the LoggerFilterOptions type?

src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Config/QueuesOptions.cs
```diff
@@ -148,5 +149,17 @@ public TimeSpan VisibilityTimeout
                 _visibilityTimeout = value;
             }
         }
+
+        public object GetLoggableOptions()
```
mathewc: I wonder if the contract shouldn't just be a Dictionary<string, string> or Dictionary<string, object> so we can avoid the json serialization?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Config/ServiceBusOptions.cs
```diff
@@ -4,13 +4,14 @@
...
 using Microsoft.Azure.ServiceBus;
+using Microsoft.Azure.WebJobs.Hosting;
 
 namespace Microsoft.Azure.WebJobs.ServiceBus
 {
     /// <summary>
     /// Configuration options for the ServiceBus extension.
     /// </summary>
-    public class ServiceBusOptions
+    public class ServiceBusOptions : ILoggableOptions
```
mathewc: There are other option types that need to be logged, e.g. EventHub, Singleton, etc.

src/Microsoft.Azure.WebJobs.Host/Hosting/LoggableOptions/LoggerFilterOptionsLoggableWrapper.cs
brettsam: We don't own LoggerFilterOptions, correct. There are likely other built-in Options that we want to log as well. Not sure I understand your proposal -- isn't this just a different wrapper type? Are you saying that one, single type would log all the options that we don't own (as opposed to one-wrapper-type-per-options)?

fabiocav: Yeah, I was thinking about the same approach.

brettsam: Oh! Sounds like we're on the same page -- I commented above that I'm going to remove `Set()` and just have it be `GetLoggableOptions(TOptions options)`. You're right that this wouldn't be a wrapper anymore... it'd just log as it's passed in, so I'll adjust the naming.

src/Microsoft.Azure.WebJobs.Host/Hosting/OptionsFormatter/IOptionsFormatter.Generic.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+
+namespace Microsoft.Azure.WebJobs.Hosting
+{
+    /// <summary>
+    /// Allows framework types to be registered as LoggableTypes.
```
mathewc: Is this comment out of date? What is a "LoggableType"?

src/Microsoft.Azure.WebJobs.Host/Hosting/OptionsFormatter/IOptionsFormatter.Generic.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// </summary>
+    /// <typeparam name="TOptions">The options type.</typeparam>
+    public interface IOptionsFormatter<TOptions>
+    {
+        /// <summary>
+        /// Creates an object populated with the keys and values of the current Options.
```
mathewc: Comments out of date

src/Microsoft.Azure.WebJobs.Host/Hosting/OptionsFormatter/IOptionsFormatter.Generic.cs
brettsam: Yeah, out of date. Will update.

src/Microsoft.Azure.WebJobs.Host/Hosting/OptionsFormatter/OptionsLoggingService.cs
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        private async Task ProcessLogs()
+        {
+            ISourceBlock<string> source = _source.LogStream;
+            try
+            {
+                while (await source.OutputAvailableAsync(_cts.Token))
+                {
+                    _logger.LogInformation(await source.ReceiveAsync());
```
brettsam: Should we keep this as Information? Debug? If it's Information, it'll introduce a handful of logs to the CLI, etc, for almost everyone's startup.

mathewc: I think leaving this as Info is good - these logs are useful to customers as well.

fabiocav: I do worry about the noise this will cause in the CLI context. We may need to filter the category out there.

src/Microsoft.Azure.WebJobs.Host/Hosting/OptionsFormatter/OptionsLoggingSource.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Threading.Tasks.Dataflow;
+
+namespace Microsoft.Azure.WebJobs.Hosting
+{
+    internal class OptionsLoggingSource : IOptionsLoggingSource
+    {
+        private BufferBlock<string> _buffer = new BufferBlock<string>();
```
fabiocav: nit: make this `readonly`

src/Microsoft.Azure.WebJobs.Host/Hosting/OptionsFormatter/WebJobsOptionsFactory.cs
```diff
@@ -0,0 +1,61 @@
...
+                formattedOptions = _optionsFormatter.Format(options);
+            }
+            else if (options is IOptionsFormatter optionsFormatter)
+            {
+                formattedOptions = optionsFormatter.Format();
+            }
+
+            if (formattedOptions != null)
+            {
+                string logString = $"{typeof(TOptions).Name}{Environment.NewLine}{formattedOptions}";
```
fabiocav: The one downside of adding the options name at the top is that we'll no longer have valid JSON, which help with queries.

brettsam: Any other alternatives? I'll merge this, but can come back and improve it with a follow-up.

brettsam:
Yes, I'll move this over to functions-host first and get all tests passing (there are likely some logging-related tests that need updating). Then work on using this everywhere else.

src/Microsoft.Azure.WebJobs.Host/Hosting/OptionsFormatter/WebJobsOptionsFactory.cs
fabiocav: I don't know. Either pass as an argument or craft a JSON envelope that would have the contents of what is returned by the formatter.

fabiocav:
About doing this for extensions, more people can assist once this is merged and available in the SDK

PR #2021, opened 2018-11-17
Porting message ID logging to v2.x

maiqbal11:
Message ID logging for web jobs v2.x. For parity with #1993.

### DIFF
--- a/test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
@@ -10,6 +10,7 @@ using System.Net;
 using System.Reflection;
 using System.Runtime.ExceptionServices;
 using System.Text;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host.Executors;
@@ -41,6 +42,8 @@ namespace Microsoft.Azure.WebJobs.Host.EndToEndTests
         private const string Queue1Name = TestArtifactsPrefix + "q1%rnd%";
         private const string Queue2Name = TestArtifactsPrefix + "q2%rnd%";
         private const string TestQueueName = TestArtifactsPrefix + "q3%rnd%";
+
+        private const string TriggerDetailsMessageStart = "Trigger Details:";
 
         private static CloudStorageAccount _storageAccount;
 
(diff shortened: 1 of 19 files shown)


RomainLasso:
Hi @maiqbal11,

It would be great if this feature could be used to pass a custom properties extractor (throught configuration, Dependency Injection or whatever) to retrieve custom metadata contained in a trigger message.

For example, if we want to pass a Correlation ID between multiple micro-services, we could use this feature to retrieve the Correlation ID regardless of the trigger (custom message properties in an Azure Service Bus Topic / Queue, HTTP Header in a HTTP request...) throught the TriggerDetails property (may be with an "ExtendedProperties" / "CustomMetadata" property that contains all the custom properties extracted)

I could make this Pull Request if it makes sense for you guys, once this one is accepted.

Today, I handle these cases with a custom extension, so it might be great if this feature is built into the trigger itself.

Best Regards,

Romain

fabiocav:
@RomainLasso this is a good idea, and something we'd likely want to apply to 3.x, as opposed to 2.x so [this PR](https://github.com/Azure/azure-webjobs-sdk/pull/1993) is a more appropriate reference.

If you could also open an issue to discuss your ideas and what you mean you have in mind when you're talking about a _properties extractor_ that would help iterating on a high level design before you spend any time on implementation.

RomainLasso:
Hi @fabiocav 

Thanks for the answer, I'll open a new issue with more explanations as soon as I have time

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
```diff
@@ -147,13 +152,24 @@ public async Task AsyncChainEndToEnd()
 
                 // Validate Logger output
                 var allLogs = _loggerProvider.CreatedLoggers.SelectMany(l => l.GetLogMessages().SelectMany(m => m.FormattedMessage.Trim().Split(new string[] { Environment.NewLine }, StringSplitOptions.None))).OrderBy(p => p).ToArray();
-                // Logger doesn't log the 'Executing' messages
-                var loggerExpected = expectedOutputLines.Where(l => !l.StartsWith("Executing '")).ToArray();
+                // Logger doesn't log the 'Executing' or 'Trigger Details' messages
```
brettsam: I'd forgotten about this -- it kinda stinks. This means that people using App Insights (which uses ILogger) won't see these logs in functions v1/webjobs v2. I can't remember why this is the case -- can you file an issue so that we can follow up? I'd rather not hold this up as this will still give us valuable details in our backend logs for investigations.

maiqbal11: Sounds good. I'll create a separate issue to track this.

PR #2043, opened 2018-12-03
Make resolution of [ConnectionString] annotated properties consistent…

### LINKED ISSUE #2034: Make resolution of [ConnectionString] annotated properties consistent
There is ordering issue in the `AttributeCloner` code. [Here](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs#L129) we try to get connection string and then we try to [auto resolve](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs#L140) it.

alrod:
…. Fixes #2034

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
@@ -126,7 +126,7 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
             // first try to resolve with connection string
             if (connStrAttr != null)
             {
-                return GetConfigurationResolver(str, connStrAttr.Default, propInfo, validator, s => _configuration.GetConnectionStringOrSetting(s));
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.WebJobs.Host/Bindings/AttributeCloner.cs
```diff
@@ -126,7 +126,7 @@ private BindingDataResolver GetResolver(PropertyInfo propInfo, INameResolver nam
             // first try to resolve with connection string
             if (connStrAttr != null)
             {
-                return GetConfigurationResolver(str, connStrAttr.Default, propInfo, validator, s => _configuration.GetConnectionStringOrSetting(s));
+                return GetConfigurationResolver(str, connStrAttr.Default, propInfo, validator, s => _configuration.GetConnectionStringOrSetting(nameResolver.Resolve(s)));
```
mathewc: Need to add tests. I believe there is already an attribute cloner test suite you can add to

PR #2049, opened 2018-12-07
Properly handling F# Task<Unit> results

fabiocav:


test/Microsoft.Azure.WebJobs.Host.FunctionalTests/DataBindingFunctionalTests.cs
```diff
@@ -51,6 +51,23 @@ public void BindNonStringableParameter_FailsIndexing()
                 exception.Message);
         }
 
+        [Fact]
+        public async Task FSharpTasks_AreCorrectlyIndexed()
```
mathewc: This test doesn't seem to belong in this test suite (DataBindingFunctionalTests). Seems like it belongs in FunctionIndexerTests

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/DataBindingFunctionalTests.cs
```diff
@@ -51,6 +51,23 @@ public void BindNonStringableParameter_FailsIndexing()
...
+                BindingFlags.Static | BindingFlags.Public);
+            Assert.NotNull(method); // Guard
+
+            FunctionIndexer indexer = FunctionIndexerFactory.Create(CloudStorageAccount.DevelopmentStorageAccount);
+            var indexCollector = new TestIndexCollector();
+            
+            // Act & Assert
+            await indexer.IndexMethodAsyncCore(method, indexCollector, CancellationToken.None);
+
+            Assert.NotEmpty(indexCollector.Functions);
```
mathewc: Can this be a stronger assert - the exact number you expect, name/parameters, etc.?

PR #2052, opened 2018-12-08
Fix TelemetryConfiguration.Active setup to fix rolename and instancename on heartbeats

lmolkova:
When we set up Active config, we  check that instrumentation key is set and do nothing if it is set.

However, ikey is read automatically from the environment variable, i.e. we don't set up WebJobsRoleEnvironmentTelemetryInititalizer. 

So telemetry (heartbeats) that is reported using Active config does not have rolename and instance info.


lmolkova:
@brettsam please review

PR #2054, opened 2018-12-12
Using function name instead of method name for logging

soninaren:


soninaren:
Used this as reference point https://github.com/Azure/azure-webjobs-sdk/pull/2051

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
```diff
@@ -310,7 +310,7 @@ internal async Task IndexMethodAsyncCore(MethodInfo method, IFunctionIndexCollec
 
             if (TypeUtility.IsAsyncVoid(method))
             {
-                string msg = $"Function '{method.Name}' is async but does not return a Task. Your function may not run correctly.";
+                string msg = $"Function '{Utility.GetFunctionName(method)}' is async but does not return a Task. Your function may not run correctly.";
```
brettsam: Can we introduce a `GetFunctionShortName()` that will return `method.GetShortName()` if there's no attribute? We really should have the class name in all of these, if the `FunctionNameAttribute` isn't there. But we can't change `GetFunctionName()` because it's already being used for generating the `Disabled` setting name.

It's a little weird b/c `GetFunctionShortName()` would be longer than `GetFunctionName()`, but it will be consistent with what we call "ShortName" elsewhere. Make sure it has some comments, and since it's internal, it should be okay.

soninaren: I thought of that, was kind of undecided between which way to go. GetFunctionShortName() sounds right.

src/Microsoft.Azure.WebJobs.Host/Utility.cs
```diff
@@ -35,6 +36,13 @@ internal static string GetFunctionName(MethodInfo methodInfo)
             return (functionNameAttribute != null) ? functionNameAttribute.Name : methodInfo.Name;
         }
 
+        internal static string GetFunctionShortName(MethodInfo methodInfo)
+        {
+            FunctionNameAttribute functionNameAttribute = TypeUtility.GetHierarchicalAttributeOrNull<FunctionNameAttribute>(methodInfo);
+            var methodName = String.Format(CultureInfo.InvariantCulture, "{0}.{1}", methodInfo.DeclaringType.Name, methodInfo.Name);
+            return (functionNameAttribute != null) ? functionNameAttribute.Name : methodName;
```
brettsam: Can you call `methodInfo.GetShortName()` here so we're guaranteed to be consistent? It'll prevent building that string unless it's required, too. Like:
```
return (functionNameAttribute != null) ? functionNameAttribute.Name : methodInfo.GetShortName();
```

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Indexers/FunctionIndexerTests.cs
```diff
@@ -114,7 +114,7 @@ public async Task IndexMethod_IfMethodReturnsAsyncVoid_Throws()
             // Act & Assert
             await product.IndexMethodAsync(typeof(FunctionIndexerTests).GetMethod("ReturnAsyncVoid"), index, CancellationToken.None);
 
-            string expectedMessage = "Function 'ReturnAsyncVoid' is async but does not return a Task. Your function may not run correctly.";
+            string expectedMessage = "Function 'FunctionIndexerTests.ReturnAsyncVoid' is async but does not return a Task. Your function may not run correctly.";
```
brettsam: This is an okay log change -- it's more descriptive and it should have been like this to begin with.

PR #2058, opened 2018-12-13
logging 'Executed...' before function marked as completed

### LINKED ISSUE #2057: [App Insights] "Executed..." logs do not contain correct operation id
See https://social.msdn.microsoft.com/Forums/en-US/c6e1c384-068a-4175-a272-29b4f8854ad4/function-app-having-invocation-id-in-app-insight-with-out-the-operationid?forum=AzureFunctions.

I believe this is an easy fix -- we're marking the function as complete (as far as App Insights is concerned) and *then* logging this message. Swap those two lines and things should work.

In v2.x, we explicitly set this: https://github.com/Azure/azure-webjobs-sdk/blob/v2.x/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/WebJobsTelemetryInitializer.cs#L38.

In v3.x, we let Application Insights track this automatically. Since the log isn't happening in the scope of the function, it is missing that id.

brettsam:
Fixes #2057.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
@@ -125,15 +125,15 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                     logCompletedCancellationToken = cancellationToken;
                 }
 
+                if (functionCompletedMessage != null)
+                {
+                    await _functionInstanceLogger.LogFunctionCompletedAsync(functionCompletedMessage, logCompletedCancellationToken);
+                }
+
                 if (instanceLogEntry != null)
                 {
                     await NotifyCompleteAsync(instanceLogEntry, functionCompletedMessage.Arguments, exceptionInfo);
                     _resultsLogger?.LogFunctionResult(instanceLogEntry);
-                }
-
-                if (functionCompletedMessage != null)
-                {
-                    await _functionInstanceLogger.LogFunctionCompletedAsync(functionCompletedMessage, logCompletedCancellationToken);
                 }
 
                 if (loggedStartedEvent)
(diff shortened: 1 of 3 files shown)


src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -125,15 +125,15 @@ public async Task<IDelayedException> TryExecuteAsync(IFunctionInstance functionI
...
+                    await _functionInstanceLogger.LogFunctionCompletedAsync(functionCompletedMessage, logCompletedCancellationToken);
                 }
 
-                if (functionCompletedMessage != null)
+                if (instanceLogEntry != null)
                 {
-                    await _functionInstanceLogger.LogFunctionCompletedAsync(functionCompletedMessage, logCompletedCancellationToken);
+                    await NotifyCompleteAsync(instanceLogEntry, functionCompletedMessage.Arguments, exceptionInfo);
+                    _resultsLogger?.LogFunctionResult(instanceLogEntry);
                 }
```
brettsam: The change swaps these two lines -- so we call `LogFunctionCompletedAsync` before we call `LogFunctionResult`, which marks the function as completed in App Insights.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/ApplicationInsightsEndToEndTests.cs
```diff
@@ -108,21 +108,22 @@ public async Task ApplicationInsights_SuccessfulFunction()
 
                 string expectedFunctionCategory = LogCategories.CreateFunctionCategory(testName);
                 string expectedFunctionUserCategory = LogCategories.CreateFunctionUserCategory(testName);
+                string expectedOperationId = request.Context.Operation.Id;
 
                 ValidateTrace(telemetries[0], "ApplicationInsightsLoggerOptions", "Microsoft.Azure.WebJobs.Hosting.OptionsLoggingService");
-                ValidateTrace(telemetries[1], "Executed ", expectedFunctionCategory, testName, invocationId);
-                ValidateTrace(telemetries[2], "Executing ", expectedFunctionCategory, testName, invocationId, request.Context.Operation.Id, request.Id);
+                ValidateTrace(telemetries[1], "Executed ", expectedFunctionCategory, testName, invocationId, expectedOperationId, request.Id);
```
brettsam: For some reason I wasn't verifying this. Now I am.

brettsam:
@fabiocav -- should be a quick review.

PR #2061, opened 2018-12-17
Telemetry is (wrongfully) filtered out

funkycoding:
When configuring the loggingbuilder to use applicationinsights, other types of telemetry are filtered out. For example, if i only want warnings to show up in ApplicationInsights i configure my appsettings like:

```
  "Logging": {
    "IncludeScopes": false,
    "LogLevel": {
      "Default": "Information"
    },
    "ApplicationInsights": {
      "LogLevel": {
        "Default": "Warning"
      }
    }
  }
```

When i now send custom metrics or events to application insights, they are assigned a default LogLevel in `FunctionExecutor.cs`:
```
using (logger.BeginScope(new Dictionary<string, object>
{
    [LogConstants.CategoryNameKey] = LogCategories.CreateFunctionCategory(instance.FunctionDescriptor.LogName),
    [LogConstants.LogLevelKey] = LogLevel.Information
}))
```

In `WebJobsTelemetryInitializer` it obtains the scope from:
```
IDictionary<string, object> scopeProps = DictionaryLoggerScope.GetMergedStateDictionary() ?? new Dictionary<string, object>();
```
which now holds the `LogLevel` set to `Information`, for all types of telemetry. It then updates the telemetry:

```
LogLevel? logLevel = scopeProps.GetValueOrDefault<LogLevel?>(LogConstants.LogLevelKey);
if (logLevel != null)
{
    telemetryProps[LogConstants.LogLevelKey] = logLevel.Value.ToString();
}
```
In the `FilteringTelemetryProcessor` it then filters out the telemetry due to the loglevel set to information...

Proposal in this PR is to only set the LogLevel property to `TraceTelemetry`. It can either be fixed here or we adjust the `FilteringTelemetryProcessor` to only filter out `TraceTelemetry`

Regards,
Mark

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
@@ -80,10 +80,14 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
                     telemetryProps[LogConstants.CategoryNameKey] = category;
                 }
 
-                LogLevel? logLevel = scopeProps.GetValueOrDefault<LogLevel?>(LogConstants.LogLevelKey);
-                if (logLevel != null)
+                // only set loglevel for filtering on trace telemetry
+                if (telemetry is TraceTelemetry)
                 {
-                    telemetryProps[LogConstants.LogLevelKey] = logLevel.Value.ToString();
+                    LogLevel? logLevel = scopeProps.GetValueOrDefault<LogLevel?>(LogConstants.LogLevelKey);
+                    if (logLevel != null)
+                    {
+                        telemetryProps[LogConstants.LogLevelKey] = logLevel.Value.ToString();
+                    }
                 }
 
                 int? eventId = scopeProps.GetValueOrDefault<int?>(LogConstants.EventIdKey);


funkycoding:
This PR is a quick fix. Would it be worth a discussion to refactor the logging? I think the implementation of `ILogger` => `ApplicationInsightsLogger` should only be concerned about LogXXX and not having all the magic enrichment that is going on in there.

They are working to seperate out the applicationinsights logger which lives under application insights asp.net core, see:
 - https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/749
 - https://github.com/Microsoft/ApplicationInsights-dotnet-logging/pull/239

I would suggest the following things:
 - Make the ApplicationInsightsLogger a 'simple' logger, swap to the logger mentioned above later on
 - Seperate logging from TrackMetric/TrackEvent/TrackDependency/TrackException
 - Use a function filter to start/end a logging scope and track result
 - clean up `FilteringTelemetryProcessor` (logging filtering should be on the logger)
 - clean up FunctionExecutor, removing logger scope stuff

funkycoding:
Tested this with the latest version available (3.0.6) - using a LogLevel higher than Information (.e.g Warning) filters out the request telemetry, event telemetry and dependency telemetry.

funkycoding:
Found an easier way to let events, requests and dependencies pass. In `IsEnabled` method in the `FilteringTelemetryProcessor i added this on the first line:

```
if (!(item is TraceTelemetry))
    return true;
```

which basically means it will pass through all other telemetry types apart from `TraceTelemetry`.

@lmolkova Also, i found that servicebus dependencies are filtered out by `OperationFilteringTelemetryProcessor`, which i suppose is not good?

lmolkova:
some telemetry items are expectedly filtered out (because they represent internal Function host operations) and considered as noise. For servicebus this is receive attempts that do not result in any function call.

funkycoding:
Closing, will make some tickets as that might be the better approach ;)

PR #2066, opened 2018-12-20
[Linux] Function still triggers despite being disabled.

alrod:
https://github.com/Azure/azure-functions-host/issues/3889


balag0:
Good to merge now. but won't take effect until Jan.

PR #2073, opened 2019-01-08
Http tracking and distributed tracing

lmolkova:
See https://github.com/Azure/azure-webjobs-sdk/pull/1965

~~This PR is work in progress and depends on prerelease version of AppInsights.~~


MHHenriksen:
FYI there's been another release of the AppInsights SDK, 2.6.1/2.9.1

lmolkova:
thanks! I'll update the PR or send another one if this is merged already. Hotfix applied in 2.9.1 is not relevant to functions and this change is not  a blocker

test/Microsoft.Azure.WebJobs.Host.UnitTests/Loggers/ApplicationInsightsConfigurationTests.cs
```diff
@@ -210,11 +251,12 @@ public void DependencyInjectionConfiguration_ConfiguresActive()
                 }).Build())
             {
                 // Verify Initializers
-                Assert.Equal(2, TelemetryConfiguration.Active.TelemetryInitializers.Count);
+                Assert.Equal(4, TelemetryConfiguration.Active.TelemetryInitializers.Count);
 
                 // These will throw if there are not exactly one
                 Assert.Single(TelemetryConfiguration.Active.TelemetryInitializers.OfType<OperationCorrelationTelemetryInitializer>());
                 Assert.Single(TelemetryConfiguration.Active.TelemetryInitializers.OfType<WebJobsRoleEnvironmentTelemetryInitializer>());
+                Assert.Single(TelemetryConfiguration.Active.TelemetryInitializers.OfType<WebJobsTelemetryInitializer>());
```
brettsam: What's the 4th initializer? Can we add it to the validation here?

test/Microsoft.Azure.WebJobs.Host.UnitTests/Loggers/ApplicationInsightsLoggerTests.cs
```diff
@@ -87,8 +88,8 @@ public async Task LogFunctionResult_Succeeded_SendsCorrectTelemetry()
             string expectedOperationId, expectedRequestId;
             using (logger.BeginFunctionScope(CreateFunctionInstance(_invocationId), _hostInstanceId))
             {
-                expectedRequestId = Activity.Current.Id;
-                expectedOperationId = Activity.Current.RootId;
+                expectedRequestId = $"|{Activity.Current.GetTraceId()}.{Activity.Current.GetSpanId()}.";
+                expectedOperationId = Activity.Current.GetTraceId();
```
brettsam: Curious: Did these change with the move to the new ApplicationInsights package?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
```diff
@@ -11,14 +11,39 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
...
+        /// (and turns on legacy correlation schema).  Enabled by default.
+        /// </summary>
+        public bool EnableW3CDistributedTracing { get; set; } = true;
+
+        /// <summary>
+        /// Gets or sets a flag that enables injection of multi-component correlation headers into responses.
+        /// This allows Application Insights to construct an Application Map to  when several
+        /// instrumentation keys are used.  Disabled by default.
+        /// </summary>
+        public bool EnableResponseHeaderInjection { get; set; } = true;
```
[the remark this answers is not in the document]
lmolkova: sure

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/WebJobs.Logging.ApplicationInsights.csproj
```diff
@@ -24,17 +24,18 @@
   </ItemGroup>
   
   <ItemGroup>
-  <PackageReference Include="Microsoft.ApplicationInsights.PerfCounterCollector" Version="2.7.2" />
+  <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.6.1" />
```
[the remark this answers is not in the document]
lmolkova: request listener stuff lives in AspNetCore SDK.
AspNetCore SDK is mostly an uber package and two things it does on its own is request tracking and configuration via dependency injection

test/Microsoft.Azure.WebJobs.Host.UnitTests/Loggers/ApplicationInsightsLoggerTests.cs
lmolkova: These are extension methods that implement W3C distributed tracing. They've been added in ApplicationInsights v2.8.
W3C support is being implemented in .NET Core and we'll consume them in ApplicaitonInsights after 3.0 releases instead of these extensions methods.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -426,9 +429,9 @@ private void StartTelemetryIfFunctionInvocation(IDictionary<string, object> stat
                 return;
             }
 
-            // Http and ServiceBus triggers are tracked automatically by the ApplicationInsights SDK
+            // HTTP and ServiceBus triggers are tracked automatically by the ApplicationInsights SDK
             // In such case a current Activity is present.
-            // We won't track and only stamp function specific details on the RequestTelemtery
+            // We won't track and only stamp function specific details on the RequestTelemetry
             // created by SDK via Activity when function ends
             if (Activity.Current == null)
```
brettsam: We'll still need to do something like:
```
if (Activity.Current == null ||
   DictionaryLoggerScope.GetMergedStateDictionary().ContainsKey("MS_IgnoreActivity"))
```

So that we can skip portal invocations. I'm going to do this locally so I can test fully E2E.

lmolkova: Did we need that because portal calls are not tracked atuomagically and we need to track them explicitly? I remember you mentioned it was fixed or I missed something.  Was it done with  'MS_IgnoreActivity'? If this workaround is needed, I'd love to understand why.

Let's return MS_IgnoreActivity for now and I'll check what happens with portal invocations separately.

brettsam: When we invoke (for example) a QueueTrigger via the portal, we do that with an Http request. And, more importantly, we actually invoke the function on a background Task and immediately return 202. 

So with the auto-Http tracking, App Insights sees that very fast response as the function execution and records it. When, in reality, we want to ignore that Http request and do the "non-Http" tracking pipeline for that invocation.

We'll need to continue doing this -- for a while it was broken b/c of something unrelated to Http tracking. All of my tests were failing b/c we weren't properly giving the "Executed..." TraceTelemetry the correct operation details. So it wouldn't show up. I fixed that in https://github.com/Azure/azure-webjobs-sdk/pull/2058. 

I've done some validation this afternoon with the "MS_IgnoreActivity" change and things are looking good. I'm setting up a private site extension now with some different trigger types and want to run through some more scenarios, but so far, so good.

lmolkova: thanks a lot for the explanation, it makes sense

lmolkova:
> In v1, we sanitized the URL to remove the QueryString. I believe Sergey recommended this. We probably need to do this again. Here's where we do that in v1, for reference: 

I noticed it, but we don't do this in ApplicaitonInsights for other web apps - we put a full URL.
Do we want to do this in functions to prevent auth key from being visible in the telemetry? Should we only strip the key and allow other parameters? Should we make it configurable?

> Should we see Client IP in Http requests?

This is tricky. I'll add `ClientIpHeaderTelemetryInitializer` into the list of telemetry initializers, so AI will populate it. This tricky part is that client IP is stripped on ingestion side because of GDPR, but geolocation info is still populated.

>  first HTTP request to an "asleep" site does not get treated like HTTP. It's likely somewhere in the invocation code in functions-host. I'm going to try to debug and figure it out... will keep you updated.

Ok, thanks. let me know if I can help

lmolkova:
Update:

1. I've removed query string.  We might consider to revisit it and strip only auth key

2. I've returned clientIp address. Even though AppInsights have special ClientIpTelemetryIntitializer, it requires additional configuration for http and is a bit hard to implement in functions. I've done it with MS_HttpContext scope, functions apply already.

brettsam:
I've got a repro for #3. The issue is that requests can come in before App Insights is initialized b/c it's initialized at the JobHost level. So:
1. Application is initialized and able to handle requests. JobHost is not ready yet.
2. Request comes in.
3. JobHost is still finalizing, but App Insights isn't wired up yet.
4. Application delays until JobHost can handle request.
5. Request flows through the JobHost, but because Activity.Current was null, it uses the "old" way without the auto-HTTP tracking.

My initial thought was that we just wire up App Insights with default settings at the Application level so that there's something watching for that "Request Started" event that creates the Activity. Would that be possible? I'm going to play with it a bit...

lmolkova:
Thanks for the repro.

>My initial thought was that we just wire up App Insights with default settings at the Application level so that there's something watching for that "Request Started" event that creates the Activity

I think this is possible but not a good long term solution. Diagnostics hook ASP.NET Core exposes and AI listeners are static singletons and it's hard to initialize them and then re-initialize with different settings. 
Configuring AppInsights twice with different settings might even work, but may create some nasty side effects, especially when multiple hosts are running in parallel. Or we won't be able to attach the WebJobs context to telemtery reported with 'default settings' 

I was playing with another problem yesterday (client ip and HttpContextAccessor) and had related issue: AspNetCore pipeline is built and ready when Function's AI is configured, so there is no way to customize the http pipeline in webjobs host startup code.
So I wonder, if there is a good long term solution, could it be early config of AI during AspNetCore startup?  

I hate to say that, but assuming this is hard and potentially breaking and RoI is very low, we can do a workaround that will initialize the first request from the MS_HttpRequest scope.

What do you think?

brettsam:
There's lots of subtle issues here that's making this tough to get right. We may need to sync with @fabiocav to see what he thinks. I'll start something internal and we can come back to this publicly once we have a design. Some responses to your questions:

> So I wonder, if there is a good long term solution, could it be early config of AI during AspNetCore startup?

Ideally, yes. But the `IWebJobsStartup` extensibility makes that hard since we don't call that until the JobHost starts up. At that point, its too late as the ASP.NET (and AI) services have all been built. That means you couldn't tweak any AI services anymore.

> I hate to say that, but assuming this is hard and potentially breaking and RoI is very low, we can do a workaround that will initialize the first request from the MS_HttpRequest scope.

The problem here is that an HttpTrigger function can return `IActionResult`. That result isn't evaluated until after the function exits. We don't call `ExecuteResultAsync` until the latest possible point in the ASP.NET pipeline. So if you check the response code from the place that we log RequestTelemetry today, it won't be correct (I just tested this...). Assuming we could somehow get the correct response, we would we lose some of the distributed tracing if we did something like this? Doesn't the RequestTrackingTelemetryModule wire-up parent values, etc, too?

lmolkova:
> So if you check the response code from the place that we log RequestTelemetry today, it won't be correct (I just tested this...).

got it, thanks for checking.

> Assuming we could somehow get the correct response, we would we lose some of the distributed tracing if we did something like this? Doesn't the RequestTrackingTelemetryModule wire-up parent values, etc, too?

the distributed tracing is wired up by AspNet Core and .NET, AppInsights is just listening to the hooks.
It also implements W3C for now, but it's temporary. If we have a response, we can still create valid request telemtery without RequestTrackingTelemetryModule help, but it makes me wonder if we need it at all and maybe we could just listen to the ASP.NET Core hook in functions for all requests.

Let me try to play with it.

brettsam:
I got something that appears to be working:
1. Wire up a RequestTrackingTelemetryModule with defaults when the app starts up.
2. As soon as we know we've initialized the first JobHost, dispose of the original module. This promotes the new Listener to "Active" (I was looking at some of the AI module/listener source).

That appeared to work in my testing. Even my custom ITelemetryChannel that I registered was working with the auto-tracked requests. I'll throw it into a PR tomorrow morning so you can see what I'm talking about.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
```diff
@@ -74,27 +74,29 @@ public static IServiceCollection AddApplicationInsights(this IServiceCollection
...
-                            EnableW3CDistributedTracing = options.EnableW3CDistributedTracing,
-                            InjectResponseHeaders = options.EnableResponseHeaderInjection
+                            EnableW3CDistributedTracing = options.HttpAutoCollectionOptions.EnableW3CDistributedTracing,
+                            InjectResponseHeaders = options.HttpAutoCollectionOptions.EnableResponseHeaderInjection
                         }
                     };
+                }
 
-                return requestCollector;
+                return new NoopTelemetryModule<RequestTrackingTelemetryModule>();
```
lmolkova: @brettsam, do you know a better way for it to avoid adding anything when `CollectExtendedHttpTriggerInformation` is false?

brettsam: You can return null here... but whenever you call GetServices<>(), you have to be prepared for one of the services in the collection to be null. But that feels a little messy -- I kinda wish if you returned null it'd just ignore that registered service completely.

We take this approach elsewhere (having some no-op/null service that we return instead). @fabiocav -- is there a "right" way to do this in a factory where you decide you don't want that service injected due to options? I went hunting through some of the .NET Core repos but couldn't find a good example of this pattern.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
```diff
@@ -160,19 +179,24 @@ private bool TryApplyProperty(RequestTelemetry request, KeyValuePair<string, str
...
                 request.Success = success;
                 wasPropertySet = true;
 
                 // Remove the Succeeded property as it's duplicated
                 request.Properties.Remove(LogConstants.SucceededKey);
             }
+            else if (property.Key == LoggingConstants.ClientIpKey)
+            {
+                request.Context.Location.Ip = property.Value;
+            }
```
brettsam: Let's remove this ClientIpKey property once we set it., otherwise it's showing up in the `customDimensions`. Your GDPR comment makes me nervous to include the raw IP like that. So we should just set it on the `Context.Location.Ip`, and let App Insights do the right thing on the server (stripping it out). I do see the regional details being populated which is great.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
fabiocav: This would be my recommended approach, to be honest. Otherwise (as @brettsam) pointed out, you put the burden of checking on the consumers and it makes constructor injection problematic (may not be the case with this if you're taking a collection)

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/NoopTelemetryModule.cs
```diff
@@ -0,0 +1,18 @@
...
+
+using Microsoft.ApplicationInsights.Extensibility;
+
+namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
+{
+    /// <summary>
+    /// Noop telemetry module that is added instead of another one, which is disabled by settings.
+    /// </summary>
+    /// <typeparam name="T">Type of the module substituted by this Noop module. Noop module is added to DI as singleton and generic type makes it unique.</typeparam>
+    class NoopTelemetryModule<T> : ITelemetryModule
```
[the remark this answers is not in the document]
lmolkova: > 2. If you use a singleton Instance (like NullLogger.Instance), can you remove the generic?

The problem is that we are going to use it like:

```csharp
services.AddSingleton<ITelemetryModule>(provider => {
  //....
  return new NullTelemetryModule<RequestTrackingTelemetryModule>();
});

services.AddSingleton<ITelemetryModule>(provider => {
  //....
  return new NullTelemetryModule<PerformanceCounters>();
});

services.AddSingleton<ITelemetryModule>(provider => {
  //....
  return new NullTelemetryModule<Dependencies>();
});

```

If we use NullTelemetryModule without generic, we'll get the exception because we are adding more than one singleton of the same service and implementation type.

If we use TryAddSingleton<ITelemetryModule>(factory), we'll not be able to add more than one ITelemetryModule of any type.

So generics is a feature :)
The way to avoid them is to return null and expect null in GetServices<ITelemetryModule>. I wanted to avoid it since users also have access to DI and we might create some confusion here.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
lmolkova: Good point, actually, this Activity.Tag and not the request property at the moment. I made sure it is not added into the request properties and added validation in the tests that we didn't add it

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/NoopTelemetryModule.cs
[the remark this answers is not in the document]
brettsam: What exception do you get? I just tested this with the .Instance approach and it looks like it worked:
```
services.AddSingleton<ITelemetryModule>(provider =>
{
    return NoopTelemetryModule.Instance;
});

services.AddSingleton<ITelemetryModule>(provider =>
{
    return NoopTelemetryModule.Instance;
});
```

And then when I run, I just see two (of the same) NoopTelemetryModules in the collection:
image

I thought we did this elsewhere with `NullLoggerProvider.Instance`, which is why I suggested it. But now I can't find it, so maybe I'm misremembering.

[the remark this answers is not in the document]
lmolkova: sorry, perhaps I was testing something else that threw on adding singleton. Changed this to the NullTelemetryModule.Instance approach and it worked. Thanks a lot

lmolkova:
@brettsam Ok, so I think I've done with all the comments. Let me know if I missed something.
Thanks a lot for your help, patience and expertize 🥇

lmolkova:
@brettsam done

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -397,6 +402,11 @@ private static void ApplyFunctionResultActivityTags(IEnumerable<KeyValuePair<str
                 {
                     currentActivity.AddTag(LogConstants.LogLevelKey, logLevel.ToString());
                 }
+
+                if (scope.TryGetValue(ApplicationInsightsScopeKeys.HttpRequest, out var request) && request is HttpRequest httpRequest)
+                {
+                    currentActivity.AddTag(LoggingConstants.ClientIpKey, GetIpAddress(httpRequest));
```
brettsam: Down in `GetIpAddress()` (I can't add a comment there b/c it's not part of your change) -- we need to change it to be...

` return httpRequest.HttpContext?.Connection?.RemoteIpAddress?.ToString() ?? LoggingConstants.ZeroIpAddress;`

...to allow for null `Connection` or `RemoteIpAddress`. The tests in Script that are using TestServer don't have values here so I'm getting a null ref. I'd only been testing this change e2e so hadn't bumped into this yet.

lmolkova:
@brettsam, I've fixed your suggestions regarding GetIPAddress. If you don't have other comments, could you please merge?

Thanks!

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -441,7 +441,12 @@ private void StartTelemetryIfFunctionInvocation(IDictionary<string, object> stat
             // We won't track and only stamp function specific details on the RequestTelemetry
             // created by SDK via Activity when function ends
             if (Activity.Current == null ||
-                DictionaryLoggerScope.GetMergedStateDictionary().ContainsKey("MS_IgnoreActivity"))
+                // Activity is tracked, but Functions wants to ignore it:
+                DictionaryLoggerScope.GetMergedStateDictionary().ContainsKey("MS_IgnoreActivity") ||
+                // Functions create another RequestTrackingTelemetryModule to make sure first request is tracked (as ASP.NET Core starts before web jobs)
+                // however at this point we may discover that RequestTrackingTelemetryModule is disabled by customer and even though Activity exists, request won't be tracked
+                // So, if we've got AspNetCore Activity and EnableHttpTriggerExtendedInfoCollection is false - track request here.
+                (!_loggerOptions.HttpAutoCollectionOptions.EnableHttpTriggerExtendedInfoCollection && Activity.Current?.OperationName == "Microsoft.AspNetCore.Hosting.HttpRequestIn"))
```
brettsam: Awesome! I wasn't sure how to differentiate the Activity between ServiceBus (and others) and Http -- but checking the OperationName is perfect.

I'm building this now and going to test in the private stamp. I was able to verify that it is, indeed, a bug -- we'd drop any cold start requests.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -480,6 +482,14 @@ private void StartTelemetryIfFunctionInvocation(IDictionary<string, object> stat
             }
         }
 
+        private bool IsHttpRequestActivity(Activity activity)
+        {
+            return activity != null &&
+                   (activity.OperationName == "Microsoft.AspNetCore.Hosting.HttpRequestIn" ||
+                    activity.OperationName == "ActivityCreatedByHostingDiagnosticListener");
+        }
```
brettsam: Could you add a comment about why we look a these two?

PR #2122, opened 2019-02-22
Indexing errors are not added to error collection (SDK)

alrod:


alrod:
https://github.com/Azure/azure-functions-host/issues/3970

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
```diff
@@ -114,6 +114,12 @@ public async Task IndexTypeAsync(Type type, IFunctionIndexCollector index, Cance
                     // If recoverable, continue to the rest of the methods.
                     // The method in error simply won't be running in the JobHost.
                     string msg = $"Function '{Utility.GetFunctionShortName(method)}' failed indexing and will be disabled.";
+
+                    FunctionDescriptor functionDescriptor = FromMethod(method);
```
mathewc: This is not what we want to do. There is an existing RecoverableException infrastructure as you can see above that is designed to make FunctionIndexErrors available to external processing/handling. In WebJobs SDK v2 this was via the TraceWriter/TraceMonitor loggers - custom loggers would receive the error stream and process them. That's how the v1 Functions host did it (code [here](https://github.com/Azure/azure-functions-host/blob/2a73a53260b55e8abc61c34da39be04244e5a957/src/WebJobs.Script/Host/ScriptHost.cs#L801)).

I think we should try to leverage the ILogger stream which provides access to the Exceptions. Host should have access to these and can mark them as handled as before and log them.

Recommend debugging through v1 to see how these were handled.

src/Microsoft.Azure.WebJobs.Protocols/FunctionDescriptor.cs
```diff
@@ -74,6 +74,9 @@ public class FunctionDescriptor
         /// </summary>
         [JsonIgnore]
         internal IEnumerable<IFunctionFilter> ClassLevelFilters { get; set; }
+
+        [JsonIgnore]
+        internal string ErrorMessage { get; set; }
```
mathewc: We don't want error messages on these description types

alrod:
Closed as not actual

PR #2129, opened 2019-02-27
Add SystemProperties dictionary to binding data

### LINKED ISSUE #2068: Cannot access IOT specific message properties in functions V2
In functions V1, a JS function with an eventhub trigger that was wired to an IOT hub would see a bunch of IOT specific metadata:

```
bindingData:
    …
    "systemPropertiesArray": [
      {
        "iothub-connection-device-id": "miye-device0",
        "iothub-connection-auth-method": "{\"scope\":\"device\",\"type\":\"sas\",\"issuer\":\"iothub\",\"acceptingIpFilterRule\":null}",
        "iothub-connection-auth-generation-id": "636640121350566756",
        "iothub-enqueuedtime": "2018-12-13T14:42:43.358Z",
        "iothub-message-source": "Telemetry",
        "x-opt-sequence-number": 6127,
        "x-opt-offset": "3015760",
        "x-opt-enqueued-time": "2018-12-13T14:42:43.403Z",
        "enqueuedTimeUtc": "2018-12-13T14:42:43.403Z",
        "sequenceNumber": 6127,
        "offset": "3015760"
      }
    ],
```

In V2, this metadata is not accessible:

```
bindingData:
   …
  "systemPropertiesArray": [
      {
        "SequenceNumber": 6128,
        "Offset": "3016216",
        "PartitionKey": null,
        "EnqueuedTimeUtc": "2018-12-13T14:55:43.94Z"
      }
    ],
   …
```

Some testing with the Event Hubs .NET Standard SDK indicates that this is not due to a change in event hubs SDK behavior and its most likely that these properties are somehow filtered out/hidden by functions. Further investigation is required.

pragnagopa:
Fixes #2068

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.EventHubs/Extensions/SystemPropertiesCollectionExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.EventHubs/Extensions/SystemPropertiesCollectionExtensions.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System.Collections.Generic;
-using Newtonsoft.Json.Linq;
 using static Microsoft.Azure.EventHubs.EventData;
 
 namespace Microsoft.Azure.WebJobs.EventHubs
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.WebJobs.Extensions.EventHubs/Extensions/SystemPropertiesCollectionExtensions.cs
```diff
@@ -11,13 +10,14 @@ static internal class SystemPropertiesCollectionExtensions
...
-            };
+            IDictionary<string, object> modifiedDictionary = collection;
+            
+            // Following is needed to maintain structure of bindingdata: https://github.com/Azure/azure-webjobs-sdk/pull/1849
+            modifiedDictionary["SequenceNumber"] = collection.SequenceNumber;
+            modifiedDictionary["Offset"] = collection.Offset;
+            modifiedDictionary["PartitionKey"] = collection.PartitionKey;
+            modifiedDictionary["EnqueuedTimeUtc"] = collection.EnqueuedTimeUtc;
+            return modifiedDictionary;
         }
```
paulbatum: Is it right to be modifying the SystemPropertiesCollection that was passed in here? Feels like we should be making a copy of that collection, and then modifying the copy. Let me know if I'm missing something.

pragnagopa: You are right. I did not intend to change the original collection. Will send follow up PR.

mathewc: Preferable, I agree. That way we aren't modifying the instance we got from EventData.SystemProperties.

jsaedtler:
I'm looking forward having this deployed and seeing the system properties in azure functions that are listening on event hubs that are populated by the iot hub routing.
When can I expect this to be online?

PR #2132, opened 2019-03-01
[QueueTrigger] adding logs if message is not fully processed

brettsam:
During a shutdown, we can occasionally have a message that fails to be deleted. The CancellationToken is canceled while the function is executing, which means we'll short-circuit when that function completes and not delete the message. It will then re-enter the queue when the visibility timeout occurs.

You end up with a message that the function completed successfully, and then inexplicably see the message re-tried 10 minutes later (by default).

We've gotten several customer reports on this and it's tough to debug, so adding an explicit message to make it clear what's happened.


azfuncgh:

PR #2145, opened 2019-03-16
Error message made more clear for the user.

vbenedichuk:


mhoeger:
@vbenedichuk - I changed the message to be `Storage account connection string '{name}' does not exist. Make sure that it a defined App Setting.` 

Does that still make sense to you? This will make sure that the error message is still applicable for azure functions scenarios

vbenedichuk:
Yes, thank you.

fabiocav:
@vbenedichuk thank you for the contribution!

PR #2163, opened 2019-04-03
version bump

brettsam:


alrod:
Are we bumping version because of:
https://github.com/Azure/azure-webjobs-sdk/commit/aa8a2d75c8404fc995f25fab9c955c3c3f394e92 ?

brettsam:
To prepare for a release, yeah. We really should bump right after we tag for a release (so that new builds are "newer" than current builds), but I didn't do that.

brettsam:
I'm trying to pull a build through to Host and it won't let me do it b/c technically our packages are older than the one it already references.

Issue #2158, opened 2019-04-01
VisibilityTimeout does not affect NextVisibleTime


jjasloot:
I have a webjob with queuetrigger running a short function that is somewhat time-critical.  If an execution of a queuemessage fails I want the webjob to try again quickly. To do this I set the visibilityTimeout to 60 seconds.  When logging the NextVisibleTime I however see that it is 10 minutes later, instead of 1 minute as I expected.

#### Repro steps
1. Create a webjob as per https://docs.microsoft.com/nl-nl/azure/app-service/webjobs-sdk-get-started

2. Set visibility timeout:
```c#
  builder.ConfigureWebJobs(b =>
            {
                b.AddAzureStorageCoreServices();
                b.AddAzureStorage(a => a.VisibilityTimeout = TimeSpan.FromSeconds(30));
           });
```
and/or
```c#
            builder.ConfigureServices((context, services) =>
            {
                services.Configure<QueuesOptions>(options =>
                {
                    options.MaxPollingInterval = TimeSpan.FromSeconds(2);
                    options.VisibilityTimeout = TimeSpan.FromSeconds(40);
                });
            });
```

3. Add logging for the NextVisibleTime in the function:
``` c#
public static void ProcessQueueMessage([QueueTrigger("queue")] string message, DateTimeOffset NextVisibleTime, ILogger logger)
        {
            logger.LogInformation(message);
            logger.LogInformation("Now:" + DateTime.UtcNow + " NextVisibleTime: " + NextVisibleTime.ToString());
        }
```

4. Add a message to the queue
5. Observe that the NextVisibleTime is 10 minutes later than now.

Please see the repo here: https://github.com/jjasloot/QueueVisibilityTimeoutPOC

#### Expected behavior

I expect the NextVisibleTime of a message to be either 30 or 40 seconds after now, as per my setting for VisibilityTimeout.

#### Actual behavior

The NextVisibleTime is 10 minutes (the default) later than now.

#### Known workarounds
None

#### Related information 
Packages:
* Microsoft.Azure.WebJobs 3.0.5
* Microsoft.Azure.WebJobs.Extensions 3.0.2
* Microsoft.Azure.WebJobs.Extensions.Storage 3.0.4
* Microsoft.Azure.WebJobs.Host.Storage 3.0.5

Source: https://github.com/jjasloot/QueueVisibilityTimeoutPOC

[...]

PR #2172, opened 2019-04-12
Set VisibilityTimeout on QueueListener when set in QueueOptions

jjasloot:
This PR solves Azure/azure-webjobs-sdk#2158. It includes a test to demonstrate the effect of the change.

In the QueueListener the VisibilityTimeout was always set to the default of 10 minutes, irrespective of the options set. This PR looks if the VisiblityTimeout in the options is not the default (TimeSpan.Zero) and if so updates the VisibilityTimeout to the value in the options.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Listeners/QueueListener.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Listeners/QueueListener.cs
@@ -74,7 +74,14 @@ namespace Microsoft.Azure.WebJobs.Host.Queues.Listeners
 
             // if the function runs longer than this, the invisibility will be updated
             // on a timer periodically for the duration of the function execution
-            _visibilityTimeout = TimeSpan.FromMinutes(10);
+            if (_queueOptions.VisibilityTimeout > TimeSpan.Zero)
+            {
+                _visibilityTimeout = _queueOptions.VisibilityTimeout;
+            }
+            else
+            {
+                _visibilityTimeout = TimeSpan.FromMinutes(10);
+            }
 
             if (sharedWatcher != null)
             {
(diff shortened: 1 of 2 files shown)


jjasloot:
There are some strange errors in the build log, which seem unrelated to the changes in code. Would it be possible to run the build again?

"The path 'bin' specified in the 'TestAdapterPath' is invalid. Error: The custom test adapter search path provided was not found, provide a valid path and try again."

jjasloot:
Any updates on how to proceed?

brettsam:
There are two visibility timeouts when we're running a queue trigger:
1. When we dequeue, we use one timeout, which is currently hardcoded to 10 minutes.
2. If there is a function execution failure, we release the message with the visibility timeout from the options (see here: https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/QueueProcessor.cs#L125). By default, this value is `TimeSpan.Zero`, which means that messages should be released immediately upon failure.

This addresses number 1 -- but that would really only matter if the process itself crashed. If a function execution failed, the message should be appearing immediately back in the queue. Is that the intention here -- to recover from crashes more quickly?

jjasloot:
We indeed experience crashes of the processes themselves. It appears the host sometimes just disappears. This PR is indeed aimed to recover more quickly from those crashes. In case the entire process is lost the Visibility Timeout is set to the default 10 minutes. This PR aims to replace that default by the user setting so a more predictable recovery can take place.

brettsam:
The process crashes seem like they should be investigated -- it should be very rare. Although even in a hosted environment, the host will occasionally restart for updates, so it can happen.

That being said, it seems like a good idea to let this "dequeue VisibilityTimeout" be controlled. I don't think it makes sense to re-use the existing timeout, though. You likely want this to be very short (as we default to zero) and if you have a short timer on the dequeue timeout, it means that we'll constantly be trying to renew the message as the function runs. The shorter the duration, the harder that is for us to guarantee and you can end up with duplicate message processing.

Looping in @mathewc as he has a lot of history here as well. Mathew -- what do you think about exposing a new timeout value that can be used for the dequeue timeout? Should we stick it directly on the QueueTriggerAttribute, or add something new to Options? I'd lean towards adding something to the attribute, myself.

jjasloot:
We have been in touch with Azure support about these crashes, no real concrete points came out, maybe we are just pushing a bit too much out of the hardware.

Glad to see you agree there is some benefit here

mathewc:
I'm thinking that rather than overload the existing VisibilityTimeout value, we introduce the ability to configure the InitialVisibility. See https://github.com/Azure/azure-sdk-for-net/issues/32635.

azfuncgh:

PR #2183, opened 2019-04-24
Prefix PerformanceCounter telemetry with f_, to indicate telemetry is from functions

cijothomas:
"azwapccore" is the default SDK Version reported for PerfCounters from App Service. This PR modifies the SDKVersion to prefix "f_", so that telemetry from Azure Functions can be distinguished from App Service for Web Apps.

Will update (https://github.com/Microsoft/ApplicationInsights-Home/blob/master/EndpointSpecs/SDK-VERSIONS.md) once this is merged.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/MetricSdkVersionTelemetryInitializer.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/MetricSdkVersionTelemetryInitializer.cs
@@ -9,7 +9,7 @@ using System;
 
 namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
 {
-    internal class PerfCounterSdkVersionTelemetryInitializer : ITelemetryInitializer
+    internal class MetricSdkVersionTelemetryInitializer : ITelemetryInitializer
     {
         private const string Prefix = "f_";
         public void Initialize(ITelemetry telemetry)
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/PerfCounterSdkVersionTelemetryInitializer.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/PerfCounterSdkVersionTelemetryInitializer.cs
@@ -1,0 +1,30 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.ApplicationInsights.Channel;
+using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.ApplicationInsights.Extensibility;
+using Microsoft.ApplicationInsights.Extensibility.Implementation;
+
+namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
+{
+    internal class PerfCounterSdkVersionTelemetryInitializer : ITelemetryInitializer
+    {
+        public void Initialize(ITelemetry telemetry)
+        {
+            if (telemetry == null)
+            {
+                return;
+            }
+            
+            if (telemetry is PerformanceCounterTelemetry)
+            {
+                var internalContext = telemetry.Context != null ? telemetry.Context.GetInternalContext() : null;
+                if (internalContext != null && internalContext.SdkVersion != null)
+                {
+                    internalContext.SdkVersion = "f_" + telemetry.Context.GetInternalContext().SdkVersion;
+                }
+            }
+        }
+    }
+}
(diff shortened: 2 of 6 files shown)


brettsam:
Is "f_azwapccore" the desired version? We already update with "webjobs" / "azurefunctions" / "azurefunctionscoretools" depending on where it is, via the registered `ISdkVersionProvider`. The intention was that this would make it easy to update that version by registering a new provider and it'd get picked up everywhere that needed it (obviously it's not getting picked up by PerfCounters).

https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs#L171
https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs#L135

We could alternatively add a step in the WebJobsTelemetryInitializer that sets the SdkVersion to this value explicitly for PerfCounters. I think we had to do this for RequestTelemetry due to the auto-tracking creating its own TelemetryClient (you or @lmolkova would know the details better than me). I'm guessing PerfCounter collection does something similar?

But that all depends on how you want these values tracked...

cijothomas:
@brettsam 
The 1st link is updating the version in TelemetryClient, so it won't have an effect of the auto collection modules, since they create own TelemetryClient.
The 2nd link does version for Requests alone.

The immediate goal is to have a way to distinguish PerfCounters from AppService vs Azure Functions. 
Once Liudmila is back from vacation, we'll create a plan to address the SDKVersion for entire telemetry from Azure Functions - those collected by host, user logging, auto-collected.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/PerfCounterSdkVersionTelemetryInitializer.cs
```diff
@@ -0,0 +1,32 @@
...
+        public void Initialize(ITelemetry telemetry)
+        {
+            if (telemetry == null)
+            {
+                return;
+            }
+            
+            if (telemetry is PerformanceCounterTelemetry)
+            {
+                var internalContext = telemetry.Context != null ? telemetry.Context.GetInternalContext() : null;
```
Mikhail-Pranovich: prefer telemetry.Context?.GetInternalContext() which is a bit more thread safe.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/PerfCounterSdkVersionTelemetryInitializer.cs
```diff
@@ -0,0 +1,32 @@
...
+                if (internalContext != null && internalContext.SdkVersion != null && !internalContext.SdkVersion.StartsWith(Prefix, StringComparison.OrdinalIgnoreCase))
+                {
+                    internalContext.SdkVersion = Prefix + telemetry.Context.GetInternalContext().SdkVersion;
```
Mikhail-Pranovich:
```suggestion
                    internalContext.SdkVersion = Prefix + internalContext.SdkVersion;
```

cijothomas: done

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/PerfCounterSdkVersionTelemetryInitializer.cs
cijothomas: done.

brettsam:
Once we verify e2e that it's working, I'll approve and we can get this in. I've got it deployed and am working with Cijo via chat to make sure it looks good.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/MetricSdkVersionTelemetryInitializer.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        private const string Prefix = "f_";
+        public void Initialize(ITelemetry telemetry)
+        {
+            if (telemetry == null)
+            {
+                return;
+            }
+            
+            if (telemetry is MetricTelemetry)
```
brettsam: right now we change the SdkVersion of any custom metrics that customers log with "webjobs" / "azurefunctions" / etc -- this will put an "f_" in front of those, too, won't it?

brettsam:
closing in favor of #2184

PR #2185, opened 2019-04-25
bumping versions; fixing test

brettsam:


test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/HttpDependencyCollectionTests.cs
```diff
@@ -155,7 +155,7 @@ await host.GetJobHost()
             Assert.Equal("Http", dependency.Type);
             Assert.Equal("www.microsoft.com", dependency.Target);
             Assert.Contains("https://www.microsoft.com", dependency.Data);
-            Assert.Equal("GET /", dependency.Name);
+            Assert.Equal("GET /en-us/", dependency.Name);
```
brettsam: Overnight, it looks like http://microsoft.com/ started redirecting to http://microsoft.com/en-us/. Going to update this test, but we may need to make this more reslient going forward if it keeps changing (i.e. -- pick a more stable Uri).

brettsam:
Gonna go ahead and merge this; it's fairly straightforward and unblocks us from getting the bits moved to host. 
/cc @fabiocav @mathewc

Issue #529, opened 2015-06-12
ServiceBus Session support


Zenuka:
I've posted this question yesterday on [StackOverflow](http://stackoverflow.com/questions/30773632/azure-webjob-servicebustrigger-for-sessions) but I guess this is a better place to ask; is it possible to have something like ServiceBusSessionTrigger for handling session through the webjobs sdk?

mathewc:
Can you detail more on the scenarios you're thinking of for it, and how you'd like it to work? The new extensibility model we're opening up would allow you to write one yourself if you wanted to. However, if this is something that many people would find useful, it would make sense for us to bake into our core ServiceBus assembly.

It's not clear to me whether this would need to be a completely new trigger binding, or whether on the trigger/receive side of things it couldn't just be an additional mode on the existing ServiceBusTrigger binding. Behind the scenes, it polls for and accepts sessions, and invokes your job function for session messages.

As far as sending session based messages, I think our ServiceBus binding (e.g. used when sending a BrokeredMessage via out param binding) would work as is? You just set your session ID on the message before sending.

Zenuka:
I'm going to use it for handling the messages of one session in a specific sequence and preferably multiple sessions at once. Since it's currently not possible, I'm going to use [this article](http://www.jayway.com/2014/08/28/routing-commands-using-azure-service-bus-sessions/) to process multiple sessions per webjob.
How should it work? Good question... I think it would be nice to have a method something like this:
```C#
public static void ProcessQueueMessage([ServiceBusTrigger(queueName:"inputqueue", useSessions: true)] List<string> messages) { ... }
```
Or whatever type of object you expect. Also you would need to specify somewhere in the config the timeout between sessions; after how many (milli)seconds will you assume the session is finished?

And you are correct, for sending messages you just need to set something in the SessionId.

JasonBSteele:
I'd love to see this feature as well. I need guaranteed my inbound messages to be in a guaranteed delivery order, and it seems using the Service Bus Session is the way to do this. However it is not at all clear how I would implement this in a WebJob.

isaacabraham:
@mathewc 

[...]

PR #2199, opened 2019-05-07
ServiceBusListener - MessageReceiver should be created no matter if sessions are enabled or not

brotaru89:
Fixed the issue when ServiceBusListener uses a queue with sessions; MessageReceiver should be created
Issue described at this link: https://github.com/Azure/azure-webjobs-sdk/issues/529#issuecomment-490029402


alrod:
MessageReceiver  does not support SB sessions. You should use ClientEntity and cast instead:
https://github.com/Azure/azure-webjobs-sdk/issues/529#issuecomment-487233051

PR #2190, opened 2019-04-26
ServiceBus sessions support

alrod:
PR from fork:
https://github.com/Azure/azure-webjobs-sdk/pull/2174

+ #1168 "EntityPath" from SB connection string is merged to this branch.

PR #2200, opened 2019-05-09
service bus binding does not accept EntityPath as connect string. Fixes #1168

### LINKED ISSUE #1168: service bus binding does not accept EntityPath as connect string
_From @globysdz on May 26, 2017 18:48_

{
  "disabled": false,
  "bindings": [
    {
      "name" : "myFuncname",
      "topicName" : "request",
      "subscriptionName" : "processor",
      "connection" : "IncomingSubscription"
      "accessRights" : "listen",
      "type" : "serviceBusTrigger",
      "direction" : "in"
    }
  ]
}

the "incomingSubscription" was set via ARM template: 
    {
              "name": "IncomingSubscription",
              "value": "[listkeys(resourceId('Microsoft.ServiceBus/namespaces/topics/authorizationRules', parameters('bus_name'), 'request', 'perm-r'), variables('sbVersion')).primaryConnectionString]" 
     }

the result "inccomingSubscription" string contains ";EntityPath=xxxx" at the end of the string which AzureFunction throws exception at startup time:

Error indexing method 'Functions.globyscadProcessor' ---> System.ArgumentException : ConnectionString used to create MessagingFactory shouldn't include EntityPath as MessagingFactory is at Namespace level.

I am using the same connection string for my webapi application, which works fine. but for function it doesn't work. and there is no good way to automatically set the connection string variable (unless there is a way to obtain the connection string from ARM without EntityPath=xxx part).

_Copied from original issue: Azure/Azure-Functions#315_

alrod:


paulbatum:
Can you provide a summary of how you handle potential conflicts, for example where the function.json has a queue name that does not match the entity path. Do we throw an error? A warning? Does it happen at execution time or indexing time?

bobvandevijver:
I would prefer to see that when an EntityPath is defined in the connection string, it takes precedence over the queue name defined in the trigger (as is the case with the event hub). In the event hub case, I actually set an empty name in the c# trigger definition, to ensure that the EntityPath must be provided with the connection string in the environment, making configuration errors solely based on the environment, and not on some static name defined in the source.

alrod:
@paulbatum, if function.json has a queue name that does not match then entity path there an exception will be thrown at indexing time

paulbatum:
@alrod Did you see the comments from @bobvandevijver above? I am thinking maybe we should allow the event hub name to be empty, and just honor the entity path in that case. What do you think?

alrod:
@bobvandevijver, @paulbatum 
EntityName constructor param for SB Trigger/Binding becomes optional in the PR. So you even don't  need to specify empty name.

alrod:
This branch is merged to session branch (PR: https://github.com/Azure/azure-webjobs-sdk/pull/2190)

PR #2202, opened 2019-05-09
[BlobTrigger] include blobs at previous timestamp to account for rounding

### LINKED ISSUE #2194: Blob Polling ScanInfo.LastScan Rounding Issue
The code [here](https://github.com/Azure/azure-webjobs-sdk/blob/25e3b0ed834f9ee64407ccd8825623e4961b100d/src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs#L155) describes a rounding issue that we have worked around on persistence of the scan info, by subtracting a millisecond. This code seems to indicate that blobs affected by this won't be picked up until the next restart. For a continuous WebJob (or Dedicated Function App) that might be a long time (days/weeks).

Because only the persisted value is decremented, the operating value being used is still susceptible to the issue. E.g. if a blob is added 10ms after the LastSweepCycleLatestModified value has been set, on the next scan it won't be picked up because the check done [here](https://github.com/Azure/azure-webjobs-sdk/blob/25e3b0ed834f9ee64407ccd8825623e4961b100d/src/Microsoft.Azure.WebJobs.Host/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs#L243) is strictly **greater than**. This blob won't be picked up until after the next host restart. 

I think we need to modify the in memory value of LastScan, not just the persisted value.

brettsam:
Fixes #2194.

This means we'll be hitting the blob receipts more than previously as we'll always be re-checking the previous timestamp. But the alternative is missing blobs, so it's worth it.

Tests needed tweaking b/c they didn't expect to see a blob a second time. Had to introduce some receipt logic in the unit tests to account for this.

I'm creating this as a draft because I want some early eyes on it -- but I want to add some more logging to this flow before committing this.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
@@ -238,7 +238,9 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
                     containerScanInfo.CurrentSweepCycleLatestModified = lastModifiedTimestamp;
                 }
 
-                if (lastModifiedTimestamp > containerScanInfo.LastSweepCycleLatestModified)
+                // Blob timestamps are rounded to the nearest second, so make sure we continue to check
+                // the previous timestamp to catch any blobs that came in slightly after our previous poll.
+                if (lastModifiedTimestamp >= containerScanInfo.LastSweepCycleLatestModified)
                 {
                     newBlobs.Add(currentBlob);
                 }
--- a/test/Microsoft.Azure.Webjobs.Extensions.Storage.UnitTests/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategyTests.cs
+++ b/test/Microsoft.Azure.Webjobs.Extensions.Storage.UnitTests/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategyTests.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.IO;
 using System.Linq;
 using System.Reflection;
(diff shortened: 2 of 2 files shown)


test/Microsoft.Azure.Webjobs.Extensions.Storage.UnitTests/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategyTests.cs
```diff
@@ -146,15 +147,60 @@ public async Task BlobPolling_IgnoresClockSkew()
...
+
+            // We should see the new item. We'll see 2 blobs, but only process 1 (due to receipt).
+            RunExecuterWithExpectedBlobs(expectedNames, product, executor, 1);
+
+            Assert.Equal(2, container.CallCount);
+            Assert.Equal(expectedNames, executor.BlobReceipts);
+        }
+
+        [Fact]
+        public async Task BlobPolling_IncludesPreviousBatch()
```
brettsam: This test failed before the change. We'd miss that second blob with the same timestamp.

src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -238,7 +238,7 @@ private void ThrowIfDisposed()
                     containerScanInfo.CurrentSweepCycleLatestModified = lastModifiedTimestamp;
                 }
 
-                if (lastModifiedTimestamp > containerScanInfo.LastSweepCycleLatestModified)
+                if (lastModifiedTimestamp >= containerScanInfo.LastSweepCycleLatestModified)
```
mathewc: Recommend adding a comment here on this logic, why it's needed, etc.

mathewc:
Looks good to me - thanks for taking this :)

brettsam:
Beefing up the logging was turning out to be a pretty big job (we don't flow function names or loggers to any of these classes today, for example) -- so rather than going down that rabbit hole, I want to get this PR in since it's pretty critical.

PR #2204, opened 2019-05-10
bumping version numbers

brettsam:


brettsam:
@alrod / @mathewc / @fabiocav -- just an FYI I'm bumping the version numbers.

Issue #2154, opened 2018-11-02
Deserializing ServiceBusMessage to POCO fails


ChristianWeyer:
My setup:
- macOS 10.13.6
- VS Code Version 1.28.2
- Azure Functions Core Tools (2.1.725 Commit hash: 68f448fe6a60e1cade88c2004bf6491af7e5f1df)
- Function Runtime Version: 2.0.12134.0

Repro is here: https://github.com/ChristianWeyer/azure-functions-servicebus-serialization-repro

This is the scenario:

C# Azure Function 1:
-HTTP Trigger -> creates a message and sends it into SB queue via output binding
=>
Java Azure function:
-SB queue trigger -> creates new message based on incoming message and sends it to another queue via SB queue output binding
=>
C# Azure Function 2:
-Service Bus queue trigger -> inspects received message... and does nothing ;-)

The message coming from the Java function into C# Function 2 (via SB) cannot be deserialized through model binding - so this fails:
```
public static class NotifyClientsAboutOrderShipmentFunctions
{
    [FunctionName("NotifyClientsAboutOrderShipment")]
    public static void Run(
        [ServiceBusTrigger("shippingsinitiated", Connection = "ServiceBus")]
        ShippingCreatedMessage msg, // Issue with deserializing incoming message
        ILogger log)
    {
        log.LogInformation($"NotifyClientsAboutOrderShipment SB queue trigger function processed message: {msg}");
        
        var messageToNotify = new { orderId = msg.OrderId };

        // TODO: Notify users' clients through SignalR...
    }
}
```
  
Exception:
```
[02/11/2018 09:06:21] Executing 'NotifyClientsAboutOrderShipment' (Reason='New ServiceBus message detected on 'shippingsinitiated'.', Id=9b058fc4-b60b-4e50-aa1b-c10f97135767)
[02/11/2018 09:06:24] Executed 'NotifyClientsAboutOrderShipment' (Failed, Id=9b058fc4-b60b-4e50-aa1b-c10f97135767)
[02/11/2018 09:06:24] System.Private.CoreLib: Exception while executing function: NotifyClientsAboutOrderShipment. Microsoft.Azure.WebJobs.Host: Exception binding parameter 'msg'. System.Private.DataContractSerialization: There was an error deserializing the object of type Serverless.Messages.ShippingCreatedMessage. The input source is not correctly formatted. System.Private.DataContractSerialization: The input source is not correctly formatted.
[...]

PR #2209, opened 2019-05-21
 Added error message for SB POCO xml deserialization

alrod:


src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Triggers/UserTypeArgumentBindingProvider.cs
```diff
@@ -96,7 +97,14 @@ private static TInput GetBody(Message message, ValueBindingContext context)
...
                 else
                 {
-                    return message.GetBody<TInput>();
+                    try
+                    {
+                        return message.GetBody<TInput>();
+                    }
+                    catch (SerializationException e)
+                    {
+                        throw new InvalidOperationException($"XML formatted message is expected for ContentType='{message.ContentType}'. For JSON formatted message set ContentType='application/json': {e.Message} ");
```
SeanFeldman: This is not XML formatted message, but DataContract XML serialized object. See https://github.com/Azure/azure-webjobs-sdk/issues/2154#issuecomment-494244856

PR #2211, opened 2019-05-24
ServiceBus POCO JSON deserialization for a"text/plain" message.

alrod:


src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Triggers/UserTypeArgumentBindingProvider.cs
```diff
@@ -71,32 +72,46 @@ public async Task<ITriggerData> BindAsync(Message value, ValueBindingContext con
...
+                try
                 {
-                    return message.GetBody<TInput>();
+                    return JsonConvert.DeserializeObject<TInput>(contents, Constants.JsonSerializerSettings);
+                }
+                catch (JsonException e)
+                {
+                    // Easy to have the queue payload not deserialize properly. So give a useful error. 
+                    string msg = string.Format(
+    @"Binding parameters to complex objects (such as '{0}') uses Json.NET serialization.
```
SeanFeldman: Missing explanation about `XmlObjectSerializer`.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Triggers/UserTypeArgumentBindingProvider.cs
```diff
@@ -71,32 +72,46 @@ public async Task<ITriggerData> BindAsync(Message value, ValueBindingContext con
...
                 {
-                    return message.GetBody<TInput>();
+                    return JsonConvert.DeserializeObject<TInput>(contents, Constants.JsonSerializerSettings);
+                }
+                catch (JsonException e)
+                {
+                    // Easy to have the queue payload not deserialize properly. So give a useful error. 
+                    string msg = string.Format(
+    @"Binding parameters to complex objects (such as '{0}') uses Json.NET serialization. 
+1. Bind the parameter type as 'string' instead of '{0}' to get the raw values and avoid JSON deserialization, or
```
SeanFeldman:
```suggestion
1. Bind the parameter type as 'string' instead of '{0}' to get the raw value and avoid JSON deserialization, or
```

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Triggers/UserTypeArgumentBindingProvider.cs
```diff
@@ -71,32 +72,46 @@ public async Task<ITriggerData> BindAsync(Message value, ValueBindingContext con
...
+2. Change the queue payload to be valid json. The JSON parser failed: {1}
```
SeanFeldman:
```suggestion
2. Change the queue payload to be valid JSON. The JSON parser failed: {1}
```

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Triggers/UserTypeArgumentBindingProvider.cs
```diff
@@ -71,32 +72,46 @@ public async Task<ITriggerData> BindAsync(Message value, ValueBindingContext con
...
-                    contents = StrictEncodings.Utf8.GetString(message.Body);
-
+                    return DeserializeJsonObject(message);
+                }
+                else
+                {
                     try
                     {
-                        return JsonConvert.DeserializeObject<TInput>(contents, Constants.JsonSerializerSettings);
+                        return message.GetBody<TInput>();
```
mathewc: I don't understand what has changed here. Isn't this what the existing code does in this case already - call `message.GetBody<TInput>()`?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Triggers/UserTypeArgumentBindingProvider.cs
```diff
@@ -71,32 +72,46 @@ public async Task<ITriggerData> BindAsync(Message value, ValueBindingContext con
                 return new TriggerData(provider, bindingData);
             }
 
-            private static TInput GetBody(Message message, ValueBindingContext context)
+            private static TInput GetBody(Message message)
             {
+                // 1. If ContentType is "application/json" deserialize as JSON.
+                // 2. If ContentType is not "application/json" deserialize the body of a message that was serialized using XmlObjectSerializer.
```
mathewc: How do you know in case 2 that XmlObjectSerializer was used - we don't, right? All we know is it isn't application/json.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Triggers/UserTypeArgumentBindingProvider.cs
alrod: In scenario "language worker" + "SB output" we send POCO json as "text/plain". Language worker execution result always treats as string. It requires more work if we want to change this behavior. 

C# SB output POCO -> C# SB Trigger POCO: works
Java SB output POCO -> C# SB Trigger POCO: does not work

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Triggers/UserTypeArgumentBindingProvider.cs
alrod: `message.GetBody<TInput>()` uses XmlObjectSerializer

SeanFeldman: @alrod you're correct that GetBody() works with XmlObjectSerializer serialized body, but the question was how do you know that's the case? Just because it wasn't JSON, doesn't mean it will be XmlObjectSerializer serialized or coming from the legacy ASB client. That's what I don't quite get as well.

alrod: I will add explanation about `XmlObjectSerializer` (your first comment above).

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Triggers/UserTypeArgumentBindingProvider.cs
mathewc: So the change you're making here is just an additional json deserialization attempt to cover cases where content type is specified incorrectly.

alrod: > So the change you're making here is just an additional json deserialization attempt to cover cases where content type is specified incorrectly.

exactly

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Triggers/UserTypeArgumentBindingProvider.cs
mathewc: Yeah, just recommend a comment like:

```
// 1. If ContentType is "application/json" deserialize as JSON.
// 2. If ContentType is not "application/json" attempt to deserialize using Message.GetBody, which will handle cases like XML serialization
// 3. If this deserialization fails, do a final attempt attempt at JSON deserialization to catch cases where the content type might be incorrect
```

PR #2213, opened 2019-05-30
updating AppInsights nugets

brettsam:
Updating App Insights nugets -- which also brought along a new extension method `AddApplicationInsights()`, which conflicts with ours. To make this more straightforward, I've marked this obsolete and renamed it `AddApplicationInsightsWebJobs`.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsLoggingBuilderExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsLoggingBuilderExtensions.cs
@@ -21,7 +21,7 @@ namespace Microsoft.Extensions.Logging
         public static ILoggingBuilder AddApplicationInsights(
             this ILoggingBuilder builder)
         {
-            return builder.AddApplicationInsights(null);
+            return AddApplicationInsights(builder, null);
         }
 
         /// <summary>
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsLoggingBuilderExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsLoggingBuilderExtensions.cs
@@ -15,19 +15,34 @@ namespace Microsoft.Extensions.Logging
     /// </summary>
     public static class ApplicationInsightsLoggingBuilderExtensions
     {
-        /// <summary>
-        /// Registers Application Insights and <see cref="ApplicationInsightsLoggerProvider"/> with an <see cref="ILoggingBuilder"/>.
-        /// </summary>        
+        [Obsolete("Use " + nameof(AddApplicationInsightsWebJobs) + " instead.", false)]
         public static ILoggingBuilder AddApplicationInsights(
             this ILoggingBuilder builder)
         {
-            return AddApplicationInsights(builder, null);
+            return AddApplicationInsightsWebJobs(builder);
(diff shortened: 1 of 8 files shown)


src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsLoggingBuilderExtensions.cs
```diff
@@ -21,7 +21,7 @@ public static class ApplicationInsightsLoggingBuilderExtensions
         public static ILoggingBuilder AddApplicationInsights(
             this ILoggingBuilder builder)
         {
-            return builder.AddApplicationInsights(null);
+            return AddApplicationInsights(builder, null);
```
brettsam: This was an ambiguous reference... I'm assuming the new AspNetCore package includes the `AddApplicationInsights()` extension. Wasn't sure if it was worth updating to that or not...

lmolkova:
@cijothomas could you please have a look?

@brettsam, there was some caveat around the ILogger provider. Now the WebJobs ILogger and AppInsihgts ILogger have the same name for the provider. We've been wondering if it affects anything and if it creates any confusion. 

@cijothomas wanted to look into this more precisely and see if we need to do anything else.

brettsam:
>We've been wondering if it affects anything and if it creates any confusion.

It does create some confusion... I'm not sure that we can change it on our end in v2, however. In v3 I will rename it to be more WebJobs/Functions-specific. We already call out that no one should ever call `AddApplicationInsights()` from Function code, so I'm hoping that it doesn't happen much... https://docs.microsoft.com/en-us/azure/azure-functions/functions-dotnet-dependency-injection#logging-services. 

Although that calls out `AddApplicationInsightsTelemetry()`... is that what changed recently?

cijothomas:
AddApplicationInsightsTelemetry() on IServiceCollection or UseApplicationInsights() on IWebHostBuilder - both internally does the same, and now enables Ilogger collection Warning and above. Since docs suggest customers to NOT do this, i think we are fine.

Due to conflicting name of the Provider alias, configuring logging level appsettings.json should be validated. In code, its possible to use fully qualified name, but not sure about config based approach.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsLoggingBuilderExtensions.cs
cijothomas: There is no need to update to our new loggerprovider. But please consider marking this method as obsoltete, and provide a new one with AppApplicationInsightsWebJob

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsLoggingBuilderExtensions.cs
```diff
@@ -15,19 +15,34 @@ namespace Microsoft.Extensions.Logging
...
+           Action<ApplicationInsightsLoggerOptions> configure)
+        {
+            return AddApplicationInsightsWebJobs(builder, configure);
+        }
+
         /// <summary>
         /// Registers Application Insights and <see cref="ApplicationInsightsLoggerProvider"/> with an <see cref="ILoggingBuilder"/>.
         /// </summary>        
-        public static ILoggingBuilder AddApplicationInsights(
+        public static ILoggingBuilder AddApplicationInsightsWebJobs(
```
mathewc: Technically a breaking change since the methods are public? We OK with this?

PR #2216, opened 2019-06-01
Expose PartitionManagerOptions. Fixed #2193

### LINKED ISSUE #2193: EventHubs: Expose PartitionManagerOptions
`PartitionManagerOptions` aren't exposed anywhere in our `EventHubOptions`. We should add these, which would allow them to be settable via host.json as well.

https://docs.microsoft.com/en-us/dotnet/api/microsoft.servicebus.messaging.partitionmanageroptions?view=azure-dotnet

alrod:


src/Microsoft.Azure.WebJobs.Extensions.EventHubs/Config/EventHubOptions.cs
```diff
@@ -37,6 +37,7 @@ public class EventHubOptions : IOptionsFormatter
         public EventHubOptions()
         {
             EventProcessorOptions = EventProcessorOptions.DefaultOptions;
+            PartitionManagerOptions = new PartitionManagerOptions();
```
mathewc: Does this change the default values we were running with before, or do those remain the same?

src/Microsoft.Azure.WebJobs.Extensions.EventHubs/Config/EventHubOptions.cs
```diff
@@ -368,10 +373,21 @@ public string Format()
                 };
             }
 
+            JObject partitionManagerOptions = null;
+            if (partitionManagerOptions != null)
+            {
+                eventProcessorOptions = new JObject
+                {
+                    { nameof(PartitionManagerOptions.LeaseDuration), EventProcessorOptions.EnableReceiverRuntimeMetric },
+                    { nameof(PartitionManagerOptions.RenewInterval), EventProcessorOptions.InvokeProcessorAfterReceiveTimeout },
```
mathewc: These values look wrong? Copy paste error? Need to update the test so it catches this.

src/Microsoft.Azure.WebJobs.Extensions.EventHubs/Config/EventHubOptions.cs
alrod: `PartitionManagerOptions` was null for `EventProcessorHost` previosly

test/Microsoft.Azure.WebJobs.Extensions.EventHubs.UnitTests/EventHubConfigurationTests.cs
```diff
@@ -34,6 +34,55 @@ public EventHubConfigurationTests()
...
+            EventHubOptions options = CreateOptions();
+
+            string format = options.Format();
+            IEnumerable<string> resultLines = format.Split(Environment.NewLine, StringSplitOptions.RemoveEmptyEntries)
+                .Select(x => x.Replace(" ", string.Empty))
+                .OrderBy(x => x);
+
+            IEnumerable<string> expectedLines = new string[]
+            {
+                "{",
```
mathewc: This test would be MUCH simpler to read/maintain/extend if you just did a JObject.Parse on the result, then simply verified the properties by name, rather than doing all this line parsing/formatting

PR #2225, opened 2019-06-11
Exposing EventHubClient to support sending Partition Key. Fixes #1643.

### LINKED ISSUE #1643: .NET Core v2 Azure Function EventHub PartitionKey doesn't seem to work
I'm unable to effectively set the PartitionKey property when using an Azure Function v2 (.NET Core).

Firstly, I am unsure *how* I am correctly supposed to set the PartitionKey property. I initially used the new Microsoft.Azure.WebJobs.Extensions.EventHubs package and its Microsoft.Azure.EventHubs.EventData class, however there is no PartitionKey property on that class. There is a PartitionKey property in the SystemProperties member on EventData, however SystemProperties and its PartitionKey property are both read-only, so I can't figure out how to set it from that package. Using the "old" Microsoft.ServiceBus.Messaging.EventData class, I can set PartitionKey on the object, however it doesn't seem to have any effect.

I'm assuming that I am doing something wrong but I can't figure out what. I am expecting that when I set the same PartitionKey value, it gets hashed and modded or something and should cause the message to be assigned to the same partition every time (as described in the documentation).

This is the Function code (using the "old" EventData class):

    public static class Function1
    {
        [FunctionName("Function1")]
        public static async Task Run(
            [TimerTrigger("0 */1 * * * *")] TimerInfo myTimer,
            [EventHub("events", Connection = "ConnectionString")] IAsyncCollector<EventData> results,
            CancellationToken token,
            TraceWriter log
        )
        {
            var ed1 = new EventData(Encoding.UTF8.GetBytes("test"));
            ed1.PartitionKey = "0";
            await results.AddAsync(ed1);
        }
    }

I've tried various partition keys (non-numeric, long, short, ...) but that didn't make a difference. The function runs every minute, es expected, but this is what I see in my client, which just dumps out the metadata:

untitled

This shows two Function invocations a minute apart. Note that the first line of each list item includes the partition on which the message was received, and that it alternates from partition 1 to partition 0, even though the PartitionKey property is always "0". If I change PartitionKey to "1", it updates in the data and stays consistent, but the reader-partition reporting the event keeps alternating the same as it does in the image. I would expect these partition IDs to remain the same if the messages were received on the same partition.

This is the code that displays the messages:

    public async Task ProcessEventsAsync(PartitionContext context, IEnumerable<EventData> messages) {
        await processor.ProcessEventsAsync(context, messages);

        foreach (EventData eventData in messages) {
            string data = Encoding.UTF8.GetString(eventData.GetBytes());
            var obj = JsonConvert.DeserializeObject(data);
            var c = JsonConvert.SerializeObject(obj, Formatting.Indented);
            await window.Dispatcher.InvokeAsync(() => window.viewModel.log.Add($"Message received (partition {context.Lease.PartitionId}): {c}"));
        }
    }

This is currently displaying the lease partition, but I later added the runtime partition and they were always the same (both partition IDs 1 or both 0) and kept alternating with subsequent Function invocations.

alrod:


src/Microsoft.Azure.WebJobs.Extensions.EventHubs/Config/EventHubExtensionConfigProvider.cs
```diff
@@ -72,6 +73,9 @@ public void Initialize(ExtensionConfigContext context)
             context.AddBindingRule<EventHubAttribute>()
                 .BindToCollector(BuildFromAttribute);
 
+            // register EventClient output binding
```
mathewc: nit: I'd just remove this comment. It's not technically an output binding, it's a client/service binding

src/Microsoft.Azure.WebJobs.Extensions.EventHubs/Config/EventHubExtensionConfigProvider.cs
```diff
@@ -142,5 +156,25 @@ private static Task<object> ConvertPocoToEventData(object arg, Attribute attrRes
         {
             return Task.FromResult<object>(ConvertString2EventData(JsonConvert.SerializeObject(arg)));
         }
+
+        private sealed class EventHubClientBinder : IValueBinder
+        {
+            private readonly object _messageSender;
+            private readonly Type _type;
+
+            public EventHubClientBinder(object messageSender, Type type)
```
mathewc: Wondering why this is called "messageSender" when it is an EventClient? Why is the type you use object, and not EventClient?

src/Microsoft.Azure.WebJobs.Extensions.EventHubs/Config/EventHubExtensionConfigProvider.cs
```diff
@@ -72,6 +73,9 @@ public void Initialize(ExtensionConfigContext context)
             context.AddBindingRule<EventHubAttribute>()
                 .BindToCollector(BuildFromAttribute);
 
+            // register EventClient output binding
+            context.AddBindingRule<EventHubAttribute>().BindToValueProvider(GetEventClient);
```
mathewc: I think you should be using BindToInput instead. See other examples in other extensions, which use this for returning "clients". Here's an example: https://github.com/Azure/azure-webjobs-sdk-extensions/blob/b7aa8f84c2b3c9065861fe955ce8180b9d19f939/src/WebJobs.Extensions.CosmosDB/Config/CosmosDBExtensionConfigProvider.cs#L58

mathewc: So you'd have something like:

```charp
context.AddBindingRule<EventHubAttribute>()
    .BindToInput<EventHubClient>(a =>
    {
        return _options.Value.GetEventHubClient(a.EventHubName, a.Connection);
    });
```

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/EventHubEndToEndTests.cs
```diff
@@ -127,41 +129,107 @@ public static void SendEvents_TestHub2(int numEvents, string input, [EventHub(Te
                 }
             }
         }
+    }
+
+    public class EventHubPartitionKeyEndToEndTests : IClassFixture<TestFixture<EventHubPartitionKeyEndToEndTests.EventHubTestJobs>>
```
mathewc: why are you creating a separate end to end test suite for this partition test? Did you run into problems? We shouldn't need a separate full test suite for each new test scenario we want to add.

alrod: yeah, I want to reuse a eventhub for the test and I do not want create a new eventhub. If I add everything in one class another function can be triggered on new event.

mathewc: Well if you consume all the events after each test that would solve the issue, right? Ensuring the checkpoint is moved to the "end". Even with your new test suite, it seems to me that you can get leakage across suites. E.g. if one suite wrote a bunch of events that didn't get fully processed (e.g. due to failures or something, or because the tests passed but didn't consume all events), the other suite will see them in the hub when it runs?

alrod: Yes, the other suite will see unprocessed messages. So you propose just to create the other eventhub for the test?

mathewc: We don’t want a new event hub per test, nor do we want a new test suite per test :) We need to find a way to clear/consume the messages before each test so each test starts clean.

alrod: I fixed the tests:
1. Use one test suite.
2. Use one eventhub.
3. Filter messages for the current test is using.

mvsmal:
Hello @alrod. This change is mentioned in release [3.0.10](https://github.com/Azure/azure-webjobs-sdk/releases/tag/v3.0.10), however that was a release of the `Microsoft.Azure.WebJobs` package and this change needs to be released in the `Microsoft.Azure.WebJobs.Extensions.EventHubs` package. And the latter still has version 3.0.5, which doesn't contain this fix. When do you plan to release it? Or am I missing something? Thanks.

alrod:
Hi @mvsmal,
Yes, I just published:
https://www.nuget.org/packages/Microsoft.Azure.WebJobs.Extensions.EventHubs/3.0.6

The package should have "EventHubClient" change.
Thanks for heads up!

PR #2230, opened 2019-06-13
Bumping up EH and SB

alrod:
I have some EventHub and ServiceBus fixes


brettsam:
Do we need to do this? I see both at 3.0.5 right now:
https://www.nuget.org/packages/Microsoft.Azure.WebJobs.Extensions.ServiceBus
https://www.nuget.org/packages/Microsoft.Azure.WebJobs.Extensions.EventHubs

PR #2238, opened 2019-06-20
adds support for method injection using services registered in the IServiceProvider

### LINKED ISSUE #4538: ILogger<T> not injected at function level but is Injected at Class level.
This might be as intended, but its annoying.

I am using DI in Azure functions, non static classes.  I can inject a Generic Logger at the class level e.g.

```C#
        public Functions(IDbConnection cn, ILogger<Functions> logger)
        {
            this.cn = cn;
            this.logger = logger;
        }
```

However, if I do it at the log level:

```
   [FunctionName("GetThings")]
        public async Task<IActionResult> GetThings(
            [HttpTrigger(AuthorizationLevel.Function, "get", Route = null)]
            HttpRequest req, ILogger<Functions> log)
        {
```

I get the following:

```
[2019-06-05 16:07:26] The following 1 functions are in error:
[2019-06-05 16:07:26] GetPrincipals: Microsoft.Azure.WebJobs.Host: Error indexing method 'GetStuff'. Microsoft.Azure.WebJobs.Host: Cannot bind parameter 'log' to type ILogger`1. Make sure the parameter Type is supported by the
binding. If you're using binding extensions (e.g. Azure Storage, ServiceBus, Timers, etc.) make sure you've called the registration method for the extension(s) in your startup code (e.g. builder.AddAzureStorage(), builder.AddServiceBus(), builder.AddTimers(), etc.).
```

#### Investigative information

Please provide the following:

I ran locally. will try to stand up a function service later.

- Timestamp:
- Function App version (1.0 or 2.0): 2
- Function App name:
- Function name(s) (as appropriate):
- Invocation ID:
- Region:

#### Repro steps

Provide the steps required to reproduce the problem:

#### Expected behavior

Provide a description of the expected behavior.

#### Actual behavior

Provide a description of the actual behavior observed.

#### Known workarounds

Provide a description of any known workarounds.

#### Related information 

* Programming language used 
* Links to source
* Bindings used

davidpeden3:
Fixes https://github.com/Azure/azure-functions-host/issues/4538.

Addresses method injection without having to explicitly mark method parameters with an attribute as referenced in https://github.com/Azure/azure-functions-host/issues/3736.

@fabiocav @brettsam

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/ServiceProvider/ServiceProviderBinding.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/ServiceProvider/ServiceProviderBinding.cs
@@ -1,0 +1,43 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Reflection;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Protocols;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Azure.WebJobs.Host.Bindings
+{
+    internal class ServiceProviderBinding : IBinding
+    {
+        private readonly ParameterInfo _parameter;
+        private readonly IServiceProvider _serviceProvider;
+
+        public ServiceProviderBinding(ParameterInfo parameter, IServiceProvider serviceProvider)
+        {
+            _parameter = parameter;
+            _serviceProvider = serviceProvider;
+        }
+
+        public Task<IValueProvider> BindAsync(object value, ValueBindingContext context)
+        {
+            return Task.FromResult<IValueProvider>(new ObjectValueProvider(value, _parameter.ParameterType));
+        }
+
+        public Task<IValueProvider> BindAsync(BindingContext context)
+        {
+            return BindAsync(_serviceProvider.GetRequiredService(_parameter.ParameterType), context.ValueContext);
+        }
+
+        public ParameterDescriptor ToParameterDescriptor()
+        {
+            return new ParameterDescriptor
+            {
+                Name = _parameter.Name
+            };
+        }
+
+        public bool FromAttribute => false;
+    }
+}
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/ServiceProvider/ServiceProviderBindingProvider.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/ServiceProvider/ServiceProviderBindingProvider.cs
@@ -1,0 +1,26 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Host.Bindings
+{
+    /// <summary>
+    /// Binding provider for registered services in <see cref="IServiceProvider"/>.
+    /// </summary>
+    internal class ServiceProviderBindingProvider : IBindingProvider
+    {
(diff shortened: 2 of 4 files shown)


fabiocav (commented):
Hey @davidpeden3, thank you for sending this!

Took an initial look at this and it looks great, but there are some issues we need to address to make sure this works as expected:
1 - The binding would conflict with other binding providers. While not having an attribute is OK in many cases (e.g. execution context, logger and HTTP request binders will do this), providers need to be careful about the types they support. If this binding was registered and executed before the execution context or logger binders, for example, you'd see that you'd start seeing failures in the binding process, as the current implementation unconditionally binds to any type, and since it treats that as a required service, it would fail. Binders without attributes have to use a collaborative model, allowing other binding providers to handle the binding request if unsuccessful. Our original design plans were to introduce a `FromServicesAttribute`, making this process explicit, which not only plays well with the binding pipeline, but has other advantages and is also consistent with the model used in ASP.NET
2 -  The service provider cannot be cached and reused across invocations. Each invocation has its own scope, and that is required to properly support scoped services. You need to retrieve the scoped service provider for the current invocation so you can properly resolve those services.

Having a test that mixes this with some of the other binding providers I mentioned would highlight issue number one.

Having a test to validate scoped, transient and singleton services work as appropriate would be good to validate the functionality and would highlight issue number 2.

Again, thanks for sending this!

davidpeden3:
@fabiocav good points, thanks for the feedback. between pro and being tied up at work, i haven't had a chance to come back to this yet. i'm out all week next week but should have some time after that. just wanted to drop a note here so you didn't think the pr was abandoned. ;)

src/Microsoft.Azure.WebJobs.Host/Bindings/ServiceProvider/ServiceProviderBinding.cs
```diff
@@ -0,0 +1,43 @@
...
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Azure.WebJobs.Host.Bindings
+{
+    internal class ServiceProviderBinding : IBinding
+    {
+        private readonly ParameterInfo _parameter;
+        private readonly IServiceProvider _serviceProvider;
+
+        public ServiceProviderBinding(ParameterInfo parameter, IServiceProvider serviceProvider)
```
KondzioSSJ4, line 17: Keep in mind that instance is `JobHostServiceProvider` so it will be a main instance of provider
Probably you want to use something else to receive provider from the current scope

src/Microsoft.Azure.WebJobs.Host/Bindings/ServiceProvider/ServiceProviderBinding.cs
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        public Task<IValueProvider> BindAsync(object value, ValueBindingContext context)
+        {
+            return Task.FromResult<IValueProvider>(new ObjectValueProvider(value, _parameter.ParameterType));
+        }
+
+        public Task<IValueProvider> BindAsync(BindingContext context)
+        {
+            return BindAsync(_serviceProvider.GetRequiredService(_parameter.ParameterType), context.ValueContext);
```
KondzioSSJ4, line 30: You could get scoped instance of `_serviceProvider` using:
```
context?.ValueContext?.FunctionContext?.CreateObjectInstance<ServiceProviderWrapper>()?.Provider
```
where `ServiceProviderWrapper` is:
```
public class ServiceProviderWrapper
    {
        public ServiceProviderWrapper(IServiceProvider provider)
        {
            Provider = provider;
        }

        public IServiceProvider Provider { get; }
    }
```
That would works because used method inject via ctor all required services
in this case, would be a scoped `IServiceProvider` what you probably want

azfuncgh:

PR #2240, opened 2019-06-23
Use container name as role instance name if machine name is not set

balag0:
Computername is not set in Linux consumption


balag0:
Thanks. Created an issue to track the unit test -https://github.com/Azure/azure-webjobs-sdk/issues/2258
Will get it in at a later time to unblock this change for now.

Issue #2123, opened 2019-02-22
Improve ApplicationInsightsOptions


brettsam:
We should have on/off switches in `ApplicationInsightsOptions` for different things like:
- QuickPulse
- Dependency tracking
- Perf counters (allow the list of counters here)

Anufriev:
Hi, do you have any estimations when the feature will be ready?

brettsam:
@Anufriev -- which part are you specifically looking for?

Anufriev:
As a minimum, I need a possibility to disable tracking of dependencies to reduce costs.

jeffhollan:
Fair request. @brettsam let me know if you have any aspect of the size of work required here but going to move this to triaged and P2 for now to help us track this.

brettsam:
It should be fairly small. We've already introduced this pattern to allow for disabling perf counters.

Anufriev:
Hi @brettsam,

I appreciate your assistance.

Thank you!

PR #2243, opened 2019-06-26
Add options to disable telemetry modules

ankitkumarr:
Helps with #2123 

I still need to test this out in a private site extensions.

The issue talks about allowing selection of perf counters to use. For now, I didn't make that change. As we should discuss if we want to allow users to have a black-list or a white-list. And counter names are like `"\\Process(??APP_WIN32_PROC??)\\% Processor Time Normalized"`. Not sure if we want users to have strings like that in the `host.json` to configure that. They seem too specific to do a string mapping.

@brettsam, thoughts?

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
@@ -30,10 +30,42 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
         /// <summary>
         /// Gets or sets authentication key for Quick Pulse (Live Metrics).
         /// </summary>
-        public string QuickPulseAuthenticationApiKey { get; set; }
+        [Obsolete("Use LiveMetricsAuthenticationApiKey instead.")]
+        public string QuickPulseAuthenticationApiKey
+        {
+            get
+            {
+                return LiveMetricsAuthenticationApiKey;
+            }
+            set
+            {
+                LiveMetricsAuthenticationApiKey = value;
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets authentication key for Live Metrics.
+        /// </summary>
+        public string LiveMetricsAuthenticationApiKey { get; set; }
 
         /// <summary>
         /// Gets or sets the time to delay Quick Pulse (Live Metrics) initialization. Delaying this initialization
+        /// can result in decreased application startup time. Default value is 15 seconds.
+        /// </summary>
+        [Obsolete("Use LiveMetricsInitializationDelay instead.")]
+        public TimeSpan QuickPulseInitializationDelay {
+            get
+            {
+                return LiveMetricsInitializationDelay;
+            }
+            set
+            {
+                LiveMetricsInitializationDelay = value;
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets the time to delay Live Metrics initialization. Delaying this initialization
         /// can result in decreased application startup time. Default value is 15 seconds.
         /// </summary>
         public TimeSpan LiveMetricsInitializationDelay { get; set; } = TimeSpan.FromSeconds(15);
@@ -46,7 +78,7 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
         public bool EnablePerformanceCountersCollection { get; set; } = true;
 
         /// <summary>
-        /// Gets or sets the flag that enables quick pulse collection.
+        /// Gets or sets the flag that enables live metrics collection.
         /// Enabled by default.
         /// </summary>
         public bool EnableLiveMetrics { get; set; } = true;
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
@@ -46,6 +46,18 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
         public bool EnablePerformanceCountersCollection { get; set; } = true;
 
         /// <summary>
+        /// Gets or sets the flag that enables quick pulse collection.
+        /// Enabled by default.
+        /// </summary>
+        public bool EnableQuickPulse { get; set; } = true;
+
+        /// <summary>
+        /// Gets or sets the flag that enables dependency tracking.
+        /// Enabled by default.
+        /// </summary>
+        public bool EnableDependencyTracking { get; set; } = true;
+
+        /// <summary>
         /// Gets or sets HTTP request collection options. 
         /// </summary>
         public HttpAutoCollectionOptions HttpAutoCollectionOptions { get; set; } = new HttpAutoCollectionOptions();
@@ -105,7 +117,9 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
                 { nameof(SnapshotConfiguration), snapshot },
                 { nameof(EnablePerformanceCountersCollection), EnablePerformanceCountersCollection},
                 { nameof(HttpAutoCollectionOptions), httpOptions},
-                { nameof(QuickPulseInitializationDelay), QuickPulseInitializationDelay }
+                { nameof(QuickPulseInitializationDelay), QuickPulseInitializationDelay },
+                { nameof(EnableQuickPulse), EnableQuickPulse },
+                { nameof(EnableDependencyTracking), EnableDependencyTracking }
             };
 
             return options.ToString(Formatting.Indented);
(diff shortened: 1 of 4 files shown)


src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
```diff
@@ -38,6 +38,18 @@ public class ApplicationInsightsLoggerOptions : IOptionsFormatter
         /// </summary>
         public bool EnablePerformanceCountersCollection { get; set; } = true;
 
+        /// <summary>
+        /// Gets or sets the flag that enables quick pulse collection.
+        /// Enabled by default.
+        /// </summary>
+        public bool EnableQuickPulse { get; set; } = true;
```
cijothomas: Suggest to use the name LiveMetrics as thats the name of public feature. (QuickPulse was an internal name)
https://docs.microsoft.com/en-us/azure/azure-monitor/app/live-stream

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
```diff
@@ -105,7 +117,9 @@ public string Format()
                 { nameof(SnapshotConfiguration), snapshot },
                 { nameof(EnablePerformanceCountersCollection), EnablePerformanceCountersCollection},
                 { nameof(HttpAutoCollectionOptions), httpOptions},
-                { nameof(QuickPulseInitializationDelay), QuickPulseInitializationDelay }
+                { nameof(QuickPulseInitializationDelay), QuickPulseInitializationDelay },
```
ankitkumarr: I need to update this name as well

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
```diff
@@ -36,7 +36,7 @@ public class ApplicationInsightsLoggerOptions : IOptionsFormatter
         /// Gets or sets the time to delay Quick Pulse (Live Metrics) initialization. Delaying this initialization
         /// can result in decreased application startup time. Default value is 15 seconds.
         /// </summary>
-        public TimeSpan QuickPulseInitializationDelay { get; set; } = TimeSpan.FromSeconds(15);
+        public TimeSpan LiveMetricsInitializationDelay { get; set; } = TimeSpan.FromSeconds(15);
```
brettsam, line 71: Thanks :-)

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
```diff
@@ -257,9 +271,9 @@ internal static LoggerFilterOptions CreateFilterOptions(LoggerFilterOptions regi
                     }
 
                     // QuickPulse can have a startup performance hit, so delay its initialization.
-                    delayer.ScheduleInitialization(() => module.Initialize(configuration), options.QuickPulseInitializationDelay);
+                    delayer.ScheduleInitialization(() => module.Initialize(configuration), options.LiveMetricsInitializationDelay);
                 }
-                else
+                else if (module != null)
```
brettsam, line 276: Can the module ever be null here? You can get the NullTelmetryModule.Instance, but that should just be a no-op.

brettsam (approved):
This looks good to me! 

FYI for @lmolkova and @cijothomas -- we're adding properties to allow people to turn a couple of modules off (via host.json in functions). Let us know if you have any objections...

brettsam:
Let's get these two new ones in, then we can come back to chat about perf counters. Feel free to split off the perf counters into a separate issue for tracking purposes going forward (make sure to note it in the existing issue).

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
```diff
@@ -45,6 +45,18 @@ public class ApplicationInsightsLoggerOptions : IOptionsFormatter
         /// </summary>
         public bool EnablePerformanceCountersCollection { get; set; } = true;
 
+        /// <summary>
+        /// Gets or sets the flag that enables quick pulse collection.
```
cijothomas: minor: comment also should say Live Metrics

cijothomas (approved):
Please modify the doc to advice customers of the new ability to turn off speific features.
https://docs.microsoft.com/en-us/azure/azure-functions/functions-monitoring#disable-built-in-logging

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
ankitkumarr: Oh right, my bad

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
ankitkumarr: Yes, at least in our tests they can be null. One of the instance in `telemetryModules` was null. I think it's because the Host setup test explicitly removes an `ITelemetryModule`. I figured it's a harmless check to have.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
brettsam: Shoot -- I don't think we can change this as this has been released already. So -- can you create a second property called `LiveMetricsInitializationDelay` -- and mark the `QuickPulse` one as deprecated? We just won't ever use `QuickPulse` one internally.

brettsam: ... it was never documented but it makes me nervous that someone could build locally against one version and deploy against another with a missing property.

ankitkumarr: Ohh I didn't realize it had been released. Fair, I will make that update. Working through a couple of release issue. May get delayed until Monday

ankitkumarr: Looks like there's also this setting `QuickPulseAuthenticationApiKey` that i need to update
But that's been there for a while. We probably need to keep supporting it as well.

ankitkumarr:
@brettsam, I updated the PR and added `[Deprecated]` in the comments. I didn't find any attribute or something that we use to deprecate flags. So figured, this would be ok. Lmk what you think

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
```diff
@@ -28,23 +28,48 @@ public class ApplicationInsightsLoggerOptions : IOptionsFormatter
         public SnapshotCollectorConfiguration SnapshotConfiguration { get; set; }
 
         /// <summary>
+        /// [Deprecated] Please use LiveMetricsAuthenticationApiKey.
```
brettsam: Use `[Obsolete]` here -- it lets you put in a string for a description and just say "Use ... instead."

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
```diff
@@ -28,23 +28,48 @@ public class ApplicationInsightsLoggerOptions : IOptionsFormatter
...
+        /// Gets or sets authentication key for Live Metrics.
+        /// </summary>
+        public string LiveMetricsAuthenticationApiKey { get; set; }
+
+        /// <summary>
+        /// [Deprecated] Please use LiveMetricsInitializationDelay.
         /// Gets or sets the time to delay Quick Pulse (Live Metrics) initialization. Delaying this initialization
         /// can result in decreased application startup time. Default value is 15 seconds.
         /// </summary>
         public TimeSpan QuickPulseInitializationDelay { get; set; } = TimeSpan.FromSeconds(15);
```
brettsam: Let's make this getter and setter just use LiveMetricsInitializationDelay directly -- so there's only one source of truth until we delete it.

Something like: 
`get { return LiveMetricsInitializationDelay; } set { LiveMetricsInitializationDelay = value; }`

Same goes for the auth api key...

brettsam (approved):
couple small comments...

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
ankitkumarr: Awesome, that's what I was looking for. Thanks :)

PR #2245, opened 2019-06-26
Removing ServiceBus extension

alrod:
* added "Microsoft.Azure.WebJobs.Source" nuget package. It will be used in SB extension in separate repo.
* added "Microsoft.Azure.WebJobs.Host.TestCommon" nuget package. It will be used for SB tests in separate repo.
* removed all SB stuff


src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/WebJobs.Extensions.ServiceBus.csproj
```diff
@@ -31,17 +31,6 @@
...
-  <ItemGroup>
-    <Compile Include="..\Microsoft.Azure.WebJobs.Host\Converters\AsyncConverter.cs" Link="Converters\AsyncConverter.cs" />
-    <Compile Include="..\Microsoft.Azure.WebJobs.Host\Converters\CompositeObjectToTypeConverter.cs" Link="Converters\CompositeObjectToTypeConverter.cs" />
-    <Compile Include="..\Microsoft.Azure.WebJobs.Host\Converters\ConversionResult.cs" Link="Converters\ConversionResult.cs" />
-    <Compile Include="..\Microsoft.Azure.WebJobs.Host\Converters\IAsyncObjectToTypeConverter.cs" Link="Converters\IAsyncObjectToTypeConverter.cs" />
-    <Compile Include="..\Microsoft.Azure.WebJobs.Host\Converters\IdentityConverter.cs" Link="Converters\IdentityConverter.cs" />
-    <Compile Include="..\Microsoft.Azure.WebJobs.Host\Converters\IObjectToTypeConverter.cs" Link="Converters\IObjectToTypeConverter.cs" />
-    <Compile Include="..\Microsoft.Azure.WebJobs.Host\Triggers\ITriggerDataArgumentBinding.cs" Link="Triggers\ITriggerDataArgumentBinding.cs" />
-    <Compile Include="..\Microsoft.Azure.WebJobs.Host\TypeUtility.cs" Link="TypeUtility.cs" />
-  </ItemGroup>
```
brettsam: Are you copying these into the ServiceBus repo? If so, I wonder if we want to create a separate nupkg for shared content files across the repos so we can keep them internal yet share the implementation...

https://medium.com/@attilah/source-code-only-nuget-packages-8f34a8fb4738.

I suspect we'll hit more-and-more of these. The ASP.NET team does this all over, sometimes for single files.

src/Microsoft.Azure.WebJobs.Host/Properties/AssemblyInfo.cs
```diff
@@ -14,3 +14,5 @@
...
+[assembly: InternalsVisibleTo("Microsoft.Azure.WebJobs.ServiceBus, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
+[assembly: InternalsVisibleTo("Microsoft.Azure.WebJobs.ServiceBus.UnitTests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
```
brettsam: It'd be nice to eventually remove these -- any sense for what internal pieces we need access to in these tests?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/WebJobs.Extensions.ServiceBus.csproj
mathewc: If you're going to make a nupkg for them, they'll be public types, so why not just make them public in the core SDK?

Some of these are OK candidates for promoting to public types. Others (e.g. ITriggerDataArgumentBinding) are historical relics and shouldn't be made public.

src/Microsoft.Azure.WebJobs.Host/Properties/AssemblyInfo.cs
mathewc: We can't add these - it's important that the SB extension not have internals access to the core SDK, since that's how we verify the extensibility model interfaces.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/WebJobs.Extensions.ServiceBus.csproj
brettsam: They're not public on nuget. It's a "source code only" package. In addition to the article I linked -- see here: https://twitter.com/davidfowl/status/1004237117608751104

It's useful for things like this -- Utility types or "legacy" types that you don't want to use to establish a public contract, but that you want to use internally across repos.

From the article:

> -The code you want to share is meant for setting up some internal things in your system and does not meant to be part of your publicly exposed types.
> -The code contains some useful helper, utility classes that’s used by other components within your system and/or can be useful for others as well, but only if consumed as source code, hence the type is internal.

mathewc: Ah, I see. Yes I agree then that that might be a good approach.

mathewc (approved):

PR #2270, opened 2019-07-17
Passing the scoped service provider to bindings and exposing a mechanism to instantiate objects using injected dependencies

fabiocav:
Pending some tests. Just want to get a build/packages so we can validate the changes for some scenarios we intend to support.


src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
```diff
@@ -26,12 +29,13 @@ public class FunctionBindingContext
         public FunctionBindingContext(
             Guid functionInstanceId,
             CancellationToken functionCancellationToken,
+            IServiceProvider functionInvocationServices = null,
             FunctionDescriptor functionDescriptor = null)
         {
```
brettsam: This is a public class - isn't this a breaking change to the ctor?

src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
```diff
@@ -48,5 +52,39 @@ public class FunctionBindingContext
         /// The short name of the current function. 
         /// </summary>
         public string MethodName { get; private set; }
+
+        /// <summary>
+        /// The service provider for the current function invocation scope.
+        /// </summary>
+        public IServiceProvider FunctionServices { get; set; }
```
brettsam: nit -- should this be `InvocationServices` or `InstanceServices` (which we use elsewhere)?

src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
```diff
@@ -48,5 +52,39 @@ public class FunctionBindingContext
...
+        }
+
+        /// <summary>
+        /// Creates an object instance with constructor arguments provided in the
+        /// method call and from the current function context scoped services.
+        /// </summary>
+        /// <typeparam name="T">The type to instantiate.</typeparam>
+        /// <param name="parameters">Parameters not provided by the function service provider.</param>
+        /// <returns>An instance of type T.</returns>
+        public T CreateObjectInstance<T>(params object[] parameters)
```
brettsam: Curious -- what's the scenario where you'd use these methods?

src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
fabiocav: Sorry, I should have marked this as a draft, but wanted to make sure that wasn't going to impact or packaging. This will be a proper overload.

src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
fabiocav: I debated. In this context, I ended up landing on this name, but we can use the same for consistency.

PR #2271, opened 2019-07-18
removing usage of FormattedLogValuesCollection

brettsam:
This type has gone internal in .NET Core 3. Our usage wasn't really necessary, so removing it completely in order to move Functions forward.


test/Microsoft.Azure.WebJobs.Logging.FunctionalTests/LoggerTest.cs
```diff
@@ -37,7 +37,23 @@ public void Dispose()
...
             }
 
-            Task.WaitAll(tasks.ToArray());
+            try
+            {
+                Task.WaitAll(tasks.ToArray());
+            }
+            catch (AggregateException aex)
+            {
+                if (!aex.InnerExceptions.All(e =>
```
fabiocav: nit: use `.Handle()` instead

brettsam: I somehow didn't know that existed. Sweet.

PR #2273, opened 2019-07-19
Add support for generating batch param type for generic bindings

### LINKED ISSUE #2287: Batch type is not supported for Bindings implemented using Fluent Pattern...
.. for out of proc languages.

Functions Host calls `JobHostMetadataProvider.GetDefaultType` method to figure out the right type to bind with ([src](https://github.com/Azure/azure-webjobs-sdk/blob/5df91e0e6a23b08db0ff37ecdf803c7a145bda1b/src/Microsoft.Azure.WebJobs.Host/Extensions/JobHostMetadataProvider.cs#L223)).
For fluent bindings, after nested series of calls, it ends up at `TriggerAdapterBindingProvider`'s `GetDefaultType`. [This function](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/TriggerAdapterBindingProvider.cs#L44-L54) ignores the type that was requested by the user and provides a default type regardless.

We need to do something like [this](https://github.com/Azure/azure-webjobs-sdk/blob/5df91e0e6a23b08db0ff37ecdf803c7a145bda1b/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/BindToInputBindingProvider.cs#L102) in `TriggerAdapterBindingProvider`.

cc: @fabiocav @mathewc 

EventGrid is implemented using Fluent Binding pattern and this issue is blocking the support to batch events for out of proc languages.

ankitkumarr:
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/2287

### DIFF
--- a/test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Indexers/FunctionIndexerTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Indexers/FunctionIndexerTests.cs
@@ -214,11 +214,11 @@ namespace Microsoft.Azure.WebJobs.Host.UnitTests.Indexers
 
         [AttributeUsage(AttributeTargets.Parameter)]
         [Binding]
-        public class ExtensionTrigger : Attribute
+        public class ExtensionTriggerAttribute : Attribute
         {
             private string _path;
 
-            public ExtensionTrigger(string path)
+            public ExtensionTriggerAttribute(string path)
             {
                 _path = path;
             }
@@ -231,11 +231,11 @@ namespace Microsoft.Azure.WebJobs.Host.UnitTests.Indexers
 
         [AttributeUsage(AttributeTargets.Parameter)]
         [Binding]
-        public class Extension : Attribute
+        public class ExtensionAttribute : Attribute
         {
             private string _path;
 
-            public Extension(string path)
+            public ExtensionAttribute(string path)
             {
                 _path = path;
             }
@@ -303,7 +303,7 @@ namespace Microsoft.Azure.WebJobs.Host.UnitTests.Indexers
             throw new NotImplementedException();
         }
 
-        public static void MethodWithExtensionJobParameterAttributes([ExtensionTrigger("path")] string input, [@Extension("path")] TextWriter writer)
+        public static void MethodWithExtensionJobParameterAttributes([ExtensionTrigger("path")] string input, [Extension("path")] TextWriter writer)
         {
             throw new NotImplementedException();
         }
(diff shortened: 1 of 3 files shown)


test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Indexers/FunctionIndexerTests.cs
```diff
@@ -214,11 +214,11 @@ public Task<ITriggerBinding> TryCreateAsync(TriggerBindingProviderContext contex
 
         [AttributeUsage(AttributeTargets.Parameter)]
         [Binding]
-        public class ExtensionTrigger : Attribute
+        public class ExtensionTriggerAttribute : Attribute
         {
             private string _path;
 
-            public ExtensionTrigger(string path)
+            public ExtensionTriggerAttribute(string path)
```
ankitkumarr, line 221: Had to make these changes because my tests were failing. This is due to https://github.com/Azure/azure-webjobs-sdk/blob/5df91e0e6a23b08db0ff37ecdf803c7a145bda1b/src/Microsoft.Azure.WebJobs.Host/Extensions/JobHostMetadataProvider.cs#L128

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/JobHostMetadataProviderTests.cs
```diff
@@ -139,6 +200,22 @@ public void DefaultTypeForQueue()
...
+            IHost host = new HostBuilder()
+                .ConfigureDefaultTestHost(b =>
+                {
+                    b.AddExtension<FluentExtensionConfig>();
+                })
+                .Build();
+
+            var metadataProvider = host.CreateMetadataProvider();
+
+            var dataType = metadataProvider.GetDefaultType(new TestAttribute(), FileAccess.Read, typeof(object[]));
```
mathewc: Should also work for SomeData[] and string[] right? Can you also add test cases for that?

mathewc (approved):
Change looks good to me, just one small possible test addition

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/JobHostMetadataProviderTests.cs
ankitkumarr: Good point! I think SomeData[] would never really be requested (if I understand correctly), given this is only called by out of proc languages. string[] is definitely a valid case. Either way, I will add tests for both because they should work.

Issue #2269, opened 2019-07-16
Queue Trigger Polling Instrumentation Request


FinVamp1:
There have been a number of reports of the QueueTrigger stop polling and a restart of the Web Job is needed.

This is on QueueTrigger jobs written using Web Jobs 3.x.

E.G.

[Queue trigger stops picking up messages after connection drops and restores}(https://github.com/Azure/azure-webjobs-sdk/issues/2072) 
[QueueTrigger stops triggering - how to troubleshoot?] (https://github.com/Azure/azure-webjobs-sdk/issues/2136) 

This is referenced in general at [Instrument Polling Algorithms with Trace statements](https://github.com/Azure/azure-webjobs-sdk/issues/1956)

The main idea would be to show if we're still polling the Queue or if it has stopped for some reason. 

Otherwise to troubleshoot this we would need more information.

1. Storage Account Logs for the Queue we are polling.
2. A Dump file of the Dotnet.exe process hosting the WebJob that is not responding.

FinVamp1:
Queue trigger stops picking up messages after connection drops and restores (https://github.com/Azure/azure-webjobs-sdk/issues/2072)
QueueTrigger stops triggering - how to troubleshoot? (https://github.com/Azure/azure-webjobs-sdk/issues/2136)

FinVamp1:
https://github.com/Azure/azure-webjobs-sdk/pull/2277 is going to help here.

brettsam:
I think #2277 addresses this completely -- I'm going to mark this one as closed. Note that customers will have to move their extensions forward to that release when it gets out (this won't happen automatically).

PR #2277, opened 2019-07-22
adding additional queue logging

brettsam:
We're getting a lot of CRIs that require us to figure out why queue messages are being missed or why they're taking longer than expected. We don't really have any insight into this today. I'm adding two new logs that will help us be able to diagnose these problems.

@FinVamp1 FYI... this may be useful for you.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Listeners/QueueListener.Logger.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Listeners/QueueListener.Logger.cs
@@ -1,0 +1,28 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Host.Queues.Listeners
+{
+    internal sealed partial class QueueListener
+    {
+        private static class Logger
+        {
+            private static readonly Action<ILogger, string, string, string, int, long, Exception> _getMessages =
+                LoggerMessage.Define<string, string, string, int, long>(LogLevel.Debug, new EventId(1, nameof(GetMessages)),
+                    "Poll for function '{functionName}' on queue '{queueName}' with ClientRequestId '{clientRequestId}' found {messageCount} messages in {pollLatency} ms.");
+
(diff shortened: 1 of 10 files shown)


src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Listeners/QueueListener.Logger.cs
```diff
@@ -0,0 +1,28 @@
...
+        private static class Logger
+        {
+            private static readonly Action<ILogger, string, string, string, int, long, Exception> _getMessages =
+                LoggerMessage.Define<string, string, string, int, long>(LogLevel.Debug, new EventId(1, nameof(GetMessages)),
+                    "Poll for function '{functionName}' on queue '{queueName}' with ClientRequestId '{clientRequestId}' found {messageCount} messages in {pollLatency} ms.");
+
+            public static void GetMessages(ILogger logger, string functionName, string queueName, string clientRequestId, int messageCount, long pollLatency) =>
+                _getMessages(logger, functionName, queueName, clientRequestId, messageCount, pollLatency, null);
+
+            private static readonly Action<ILogger, string, double, string, Exception> _backoffDelay =
```
alrod: Can you please move it up so private fields will be in the beginning.

alrod (approved):
Approved with a minor comment

src/Microsoft.Azure.WebJobs.Extensions.Storage/StorageLoadBalancerQueue.cs
```diff
@@ -111,6 +111,7 @@ private CloudQueue Convert(string queueMoniker)
                 sharedWatcher: _sharedWatcher,
                 queueOptions: _queueOptions,
                 queueProcessorFactory: _queueProcessorFactory,
+                functionDescriptor: new FunctionDescriptor(),
```
brettsam: I'm pretty sure that this isn't used anywhere anymore -- but even if so, I believe it was meant to be a "backplane" for queues across our system, so having (null) in the function name would make sense. I didn't want to branch the logging logic based on this, though -- so leaving as-is for now. If we end up seeing this as a problem, we can address it with a follow-up.

FinVamp1:
Thanks Brett, adding this to https://github.com/Azure/azure-webjobs-sdk/issues/2269 as that was my ask there.

PR #2282, opened 2019-07-25
adding lots of new blob processing logs

### LINKED ISSUE #2348: Improve blob processing logs
We're currently pretty blind when we're polling and processing blobs. Need to improve this so we can have some traceability through the system.

brettsam:
Fixes #2348

Note: see below for a summary of all the specific changes: https://github.com/Azure/azure-webjobs-sdk/pull/2282#issuecomment-548549213

LOTS of plumbing added here so we can start tracing the path of a blob through our system. 

Note that when we find blobs, we stick a message in a queue, and then a QueueTrigger takes over. This allows for better processing across instances.

The idea:
1. When we poll for new blobs, we use a generated request id to correlate with storage.
2. When we start processing each of the blobs we find, I write a log saying that blob 'a' from request id 'b' is now processing. This lets us correlate any invocation back to the storage request that found it. It's not perfect, but without ripping a lot of this apart, it should do.
3. If anything happens to short-circuit the trigger being called, log it (we've already seen it / it's disappeared / etc)
4. Log the queue message id and queue name when we write to the queue.
5. On the QueueTrigger side, log any time we may be abandoning that message and not triggering. Include the queue message in these messages for correlation.

This is a lot more verbose... but considering we had zero logs before, I'm welcoming it :-)

I've not tested every path here, but let's start looking at it.

@FinVamp1 as FYI


pragnagopa:
Thanks Brett! 
@katiecai - FYI please see this PR for guidance on adding logs

brettsam:
After shelving this for a few months, I want to push this over the finish line. Here's a layout of how blob processing works (and the corresponding logs that show up). A few initial things to note:

1. We have two blob listeners running at all time to find anything we may miss. 
    1. **Container scan.** This is looking for all new logs from the last log we found. Requires we pull down *all* logs (there's no way to filter) and process them in the host.
    2. **Log scan.** This scans the blob logs that tell us when new blobs have been added.
2. There is not a "listener-per-function" like some other listeners. There is a host-wide shared listener that listens on all containers that functions care about.
3. When a blob is found, it's not immediately executed by the function. Instead, it:
    1. Is checked against blob receipts (in case it's already been processed)
    2. Performs validation. For example, does the blob name pattern match what the function asks for (i.e. "{name}.txt"...)?
    3. Adds the blob details into a queue.
    4. Then, a separate QueueListener listens for those messages and executes the function. This allows a single blob listener to fan out to a bunch of host instances and also gives us the same poison-queue and invisibility behavior that we get from QueueTrigger.
    
So the idea is to be able to trace a blob from the scan all the way to execution. Or, if it was ignored, explicitly log why it was ignored.

**Container scan**
This is the primary scan mechanism.

1. The listener polls the registered container looking for any blobs that are newer than the last found blob. When we poll, we assign a `PollId`. This id is used as the ClientRequestId for the request we send to storage (to allow for correllation) and it is also written out in every log so we can track back when and how the blob was processed.

    `Poll for blobs newer than '2019-10-31T19:27:33.999Z' in container 'samples-output' with ClientRequestId '86b1430e-6f8d-4ecd-9136-1b0c24ad5e25' found 4 blobs in 768 ms. ContinuationToken: False.`

2. Each of these blobs is then sent to be validated.
    1. If valid and not yet processed

        `Blob '4659_Request.log' is ready for processing by function 'BlobFunc'. A message with id 'ebc839ef-eec2-45d0-8082-95ec46aec9ff' has been added to queue 'azure-webjobs-blobtrigger-functiondev-brettsam2'. This message will be dequeued and processed by the BlobTrigger. PollId: '86b1430e-6f8d-4ecd-9136-1b0c24ad5e25'. Source: 'ContainerScan'.`

    2. If it does not match the pattern: 

        `Blob 'display.txt' will be skipped for function 'BlobFunc' because it does not match the pattern 'samples-output/{name}.log'. PollId: '86b1430e-6f8d-4ecd-9136-1b0c24ad5e25'. Source: 'ContainerScan'.`

    3. If it has already been processed: 

        `Blob '4659_Request.log' will be skipped for function 'BlobFunc' because this blob with ETag '"0x8D75E38C6EB4A79"' has already been processed. PollId: '34132dbe-d3e6-4e53-9c90-0cace0f5eb6a'. Source: 'ContainerScan'.`

3. From there, the QueueTrigger logs take over and you can see the results. Note that the trigger details for a BlobQueueTrigger also include some details about the blob itself (not just the queue message): 

    `Trigger Details: MessageId: ebc839ef-eec2-45d0-8082-95ec46aec9ff, DequeueCount: 1, InsertionTime: 2019-10-31T19:33:53.000+00:00, BlobCreated: 2019-10-31T19:30:24.000+00:00, BlobLastModified: 2019-10-31T19:30:24.000+00:00`

**Log scan**
Log scan is the fallback. It behaves very similarly to the container scan, but step 1 will have a slightly different message (and the `Source:` for each log will be `LogScan`):

`Log scan for recent blob updates in container 'samples-output' with PollId 'd888a807-1b63-4913-8d2c-5d63ffeb1b53' found 4 blobs.`

**Misc**
There are various other error and informational logs added as well. Examples:
- `Initialized scan info for container 'samples-output'. Scan will start with blobs inserted or updated after '2019-10-31T19:30:23.999Z'.`
- `Container 'samples-output' does not exist.`
- `"Blob 'abc.log' will be skipped for function 'BlobFunc' because its ETag cannot be found. The blob may have been deleted. PollId: 'acc833ef-eec2-45d0-8022-95ec46aec9ff'. Source: 'ContainerScan'."`

src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/BlobTriggerExecutor.cs
```diff
@@ -49,18 +62,22 @@ public async Task<FunctionResult> ExecuteAsync(ICloudBlob value, CancellationTok
...
-            var receiptBlob = _receiptManager.CreateReference(_hostId, _functionId, value.Container.Name,
+            var receiptBlob = _receiptManager.CreateReference(_hostId, _functionDescriptor.Id, value.Container.Name,
                 value.Name, possibleETag);
 
             // Check for the completed receipt. If it's already there, noop.
             BlobReceipt unleasedReceipt = await _receiptManager.TryReadAsync(receiptBlob, cancellationToken);
 
             if (unleasedReceipt != null && unleasedReceipt.IsCompleted)
             {
+                Logger.BlobAlreadyProcessed(_logger, _functionDescriptor.LogName, value.Name, possibleETag, pollId, triggerSource);
```
mathewc, line 79: Does this mean that on every container scan, we'll be logging a line for all blobs that are already processed?

brettsam: No, we'll skip over any that are before the current timestamp -- those are ignored completely without any log. But ones that get past those checks and then make it this far will be logged. Since this is singleton, it seems to be fairly rare that we see blobs processed twice and hit this check (other than the "bug" I've mentioned elsewhere -- where we continue to check with `>=` to make sure we don't miss any late-coming blobs from that previous timestamp).

mathewc (approved):

src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/FailedBlobNotification.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.WindowsAzure.Storage.Blob;
+
+namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
+{
+    internal class BlobNotification
+    {
```
pragnagopa: Nit: Match class name to file name

src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/BlobTriggerExecutor.Logger.cs
```diff
@@ -0,0 +1,44 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
+{
+    internal partial class BlobTriggerExecutor
+    {
```
pragnagopa, line 10: Nit: should this be sealed class similar to other partial classes you added for loggers?

pragnagopa (approved):
Looks good. Thanks for the detailed description in the PR. Added couple of nits.

src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/BlobTriggerExecutor.Logger.cs
brettsam: No, I only did that when the other part of the class was also sealed. Although that wasn't strictly required: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/partial-classes-and-methods

> If any part is declared abstract, then the whole type is considered abstract. If any part is declared sealed, then the whole type is considered sealed. If any part declares a base type, then the whole type inherits that class.

PR #2291, opened 2019-08-06
[Storage] raise an unhandled exception if we detect a hang in a storage operation

brettsam:
We're seeing issues with network connections that can cause certain storage operations to deadlock. There's no easy fix as it happens much lower down than we can control. Attempting a fix here where we simply restart the process if this happens.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/BlobLogListener.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/BlobLogListener.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Timers;
 using Microsoft.WindowsAzure.Storage.Blob;
 using Microsoft.WindowsAzure.Storage.Shared.Protocol;
 
@@ -24,10 +25,12 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
         private readonly CloudBlobClient _blobClient;
         private readonly HashSet<string> _scannedBlobNames = new HashSet<string>();
         private readonly StorageAnalyticsLogParser _parser = new StorageAnalyticsLogParser();
+        private readonly IWebJobsExceptionHandler _exceptionHandler;
 
-        private BlobLogListener(CloudBlobClient blobClient)
+        private BlobLogListener(CloudBlobClient blobClient, IWebJobsExceptionHandler exceptionHandler)
         {
             _blobClient = blobClient;
+            _exceptionHandler = exceptionHandler ?? throw new ArgumentNullException(nameof(exceptionHandler));
         }
 
         public CloudBlobClient Client
@@ -37,10 +40,10 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
 
         // This will throw if the client credentials are not valid. 
         public static async Task<BlobLogListener> CreateAsync(CloudBlobClient blobClient,
-            CancellationToken cancellationToken)
+           IWebJobsExceptionHandler exceptionHandler, CancellationToken cancellationToken)
         {
             await EnableLoggingAsync(blobClient, cancellationToken);
-            return new BlobLogListener(blobClient);
+            return new BlobLogListener(blobClient, exceptionHandler);
         }
 
         public async Task<IEnumerable<ICloudBlob>> GetRecentBlobWritesAsync(CancellationToken cancellationToken,
@@ -49,7 +52,7 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
             List<ICloudBlob> blobs = new List<ICloudBlob>();
 
             var time = DateTime.UtcNow; // will scan back 2 hours, which is enough to deal with clock sqew
-            foreach (var blob in await ListRecentLogFilesAsync(_blobClient, time, cancellationToken, hoursWindow))
+            foreach (var blob in await ListRecentLogFilesAsync(_blobClient, time, cancellationToken, hoursWindow, _exceptionHandler))
             {
                 bool isAdded = _scannedBlobNames.Add(blob.Name);
                 if (!isAdded)
@@ -137,7 +140,7 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
         // Logs are about 6 an hour, so we're only scanning about 12 logs total. 
         // $$$ If logs are large, we can even have a cache of "already scanned" logs that we skip. 
         private static async Task<List<ICloudBlob>> ListRecentLogFilesAsync(CloudBlobClient blobClient,
-            DateTime startTimeForSearch, CancellationToken cancellationToken, int hoursWindow)
+            DateTime startTimeForSearch, CancellationToken cancellationToken, int hoursWindow, IWebJobsExceptionHandler exceptionHandler)
         {
             string serviceName = "blob";
 
@@ -147,7 +150,7 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
             for (int i = 0; i < hoursWindow; i++)
             {
                 var prefix = GetSearchPrefix(serviceName, lastHour, lastHour);
-                await GetLogsWithPrefixAsync(selectedLogs, blobClient, prefix, cancellationToken);
+                await GetLogsWithPrefixAsync(selectedLogs, blobClient, prefix, exceptionHandler, cancellationToken);
                 lastHour = lastHour.AddHours(-1);
             }
 
@@ -157,12 +160,17 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
         // Populate the List<> with blob logs for the given prefix. 
         // http://blogs.msdn.com/b/windowsazurestorage/archive/2011/08/03/windows-azure-storage-logging-using-logs-to-track-storage-requests.aspx
         private static async Task GetLogsWithPrefixAsync(List<ICloudBlob> selectedLogs, CloudBlobClient blobClient,
-            string prefix, CancellationToken cancellationToken)
+            string prefix, IWebJobsExceptionHandler exceptionHandler, CancellationToken cancellationToken)
         {
-            // List the blobs using the prefix
+            // List the blobs using the prefix            
             IEnumerable<IListBlobItem> blobs = await blobClient.ListBlobsAsync(prefix,
-                useFlatBlobListing: true, blobListingDetails: BlobListingDetails.Metadata,
-                cancellationToken: cancellationToken);
+                    useFlatBlobListing: true, blobListingDetails: BlobListingDetails.Metadata,
+                    operationName: "ScanLogs", exceptionHandler: exceptionHandler, cancellationToken: cancellationToken);
+
+            if (blobs == null)
+            {
+                return;
+            }
 
             // iterate through each blob and figure the start and end times in the metadata
             // Type cast to IStorageBlob is safe due to useFlatBlobListing: true above.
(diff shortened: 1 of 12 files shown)


paulbatum (commented):
Are we able to attach a client request ID to these operations and log the ID when we hit the timeout/deadlock?

brettsam:
Yes, we can -- I had that in my other branch with the more graceful (but more complicated) disposing and forgot to port it over. However in this case, the request is successful. The deadlock happens when parsing the Xml response. I suspect that to Storage these will look like very quick, successful requests. I'll still add the request ids, though, as that's useful and may help us identify something new.

paulbatum:
I'm a bit confused, if you're able to repro the issue by interfering with the network connection, then shouldn't storage see a failure on their side (the client disappeared before receiving the response payload)?

brettsam:
You may be right -- locally when looking at the code, I see we get a 200, and then as they are parsing the XML from the response stream, the network goes away and then comes back. I'm not sure what storage would see that as on their side... maybe it'd be a client error?

src/Microsoft.Azure.WebJobs.Extensions.Storage/DeadlockHandler.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Timers;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Storage
+{
+    internal static class DeadlockHandler
+    {
+        private static readonly TimeSpan DefaultDeadlockTimeout = TimeSpan.FromMinutes(2);
+
+        public static async Task<T> RestartIfDeadlocked<T>(string operationName, IWebJobsExceptionHandler exceptionHandler, Func<Task<T>> operation)
```
mathewc: This technically just implements a timeout pattern, not deadlock detection, right? You're doing the same thing here that you do for TimeoutAttribute in FunctionExecutor. So perhaps this helper should be renamed to ExecuteWithTimeout

src/Microsoft.Azure.WebJobs.Extensions.Storage/DeadlockHandler.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Runtime.ExceptionServices;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Timers;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Storage
+{
+    internal static class DeadlockHandler
+    {
+        private static readonly TimeSpan DefaultDeadlockTimeout = TimeSpan.FromMinutes(2);
```
mathewc: How'd you arrive at this number? I get that we're just trying to pick a number higher than any reasonable expected duration.

brettsam: Just a reasonable choice. From my testing, successes are very fast (maybe a couple seconds max, even with 5000 blobs) -- but I have no idea what the outliers would be. I figured 2 minutes of paused polling is better than the infinity minutes we see today. Open to other suggestions.

src/Microsoft.Azure.WebJobs.Extensions.Storage/DeadlockHandler.cs
brettsam: Yeah, I started renaming it with my other solution -- which was actually tracking and killing the deadlocked Response. I'll update this one to be more generically-named.

src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/BlobLogListener.cs
```diff
@@ -157,12 +161,20 @@ private static string GetSearchPrefix(string service, DateTime startTime, DateTi
...
-            string prefix, CancellationToken cancellationToken)
+            string prefix, IWebJobsExceptionHandler exceptionHandler, CancellationToken cancellationToken)
         {
             // List the blobs using the prefix
-            IEnumerable<IListBlobItem> blobs = await blobClient.ListBlobsAsync(prefix,
-                useFlatBlobListing: true, blobListingDetails: BlobListingDetails.Metadata,
-                cancellationToken: cancellationToken);
+            IEnumerable<IListBlobItem> blobs = await DeadlockHandler.RestartIfDeadlocked("ScanLogs", exceptionHandler, () =>
+            {
+                return blobClient.ListBlobsAsync(prefix,
```
brettsam: I didn't realize this was an extension we wrote to loop through all continuation tokens -- I'm going to move this timeout handling code down there...

src/Microsoft.Azure.WebJobs.Extensions.Storage/TimeoutHandler.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Runtime.ExceptionServices;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Timers;
+
+namespace Microsoft.Azure.WebJobs.Extensions.Storage
+{
+    internal static class TimeoutHandler
+    {
+        private static readonly TimeSpan DefaulTimeout = TimeSpan.FromMinutes(2);
```
paulbatum: typo

src/Microsoft.Azure.WebJobs.Extensions.Storage/TimeoutHandler.cs
```diff
@@ -0,0 +1,49 @@
...
+                Task<T> operationTask = operation();
+
+                Task completedTask = await Task.WhenAny(timeoutTask, operationTask);
+
+                if (Equals(timeoutTask, completedTask))
+                {
+                    ExceptionDispatchInfo exceptionDispatchInfo;
+                    try
+                    {
+                        throw new TimeoutException($"The operation '{operationName}' with id '{clientRequestId}' did not complete in '{DefaulTimeout}'. Restarting host to recover from potential deadlock.");
```
mathewc: nit: I'd remove the "Restarting host to recover from potential deadlock." bit since we're not restarting anything here. We're just propagating an exception - handlers may decide to restart.

mathewc (approved):
One more comment, otherwise I have no more feedback on this.

ManfredLange:
In case you are observing this with the Azure Storage Emulator while running in a local environment, then be aware that version 5.10.0.0 appears to have issues causing blocking calls in particular for queue operations. My recommendation is to use the new Azurite open-source emulator instead. See https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azurite

derekrprice:
Is there a ticket out there tracking the deadlock at the root of this issue?  We're not using durable functions, so it's not Azure/azure-functions-durable-extension#961.  After this workaround, we are a lot better off than before since we no longer need to detect stalls and restart in a wrapper, but we're still seeing the new exceptions and subsequent restarts about every 16 minutes.  Given the two minute timeout and a minute or two to restart the container, that means that our functions application is not receiving new blob notifications for about 3.5 minutes out of every 16, or almost a quarter of the time.  Again, better than before, but not ideal, and I was hoping to at least subscribe to an issue following this.  Thanks!

paulbatum:
@derekrprice its really interesting to hear that you are seeing this happen at a regular interval. The repro's that @brettsam was able to come up with were quite intermittent. Can you clarify for me, do you know specifically which operation and codepath is hitting this deadlock with such regularity?

I have a set of [changes in progress](https://github.com/Azure/azure-functions-host/pull/5218) that move the storage extension and the function host's direct storage dependency to v11.x which the storage team advised are likely to help with this issue. Still more testing pending to confirm this.

derekrprice:
@paulbatum, I may have implied consistent timing between exceptions, which we aren't seeing. 
 Scanning our logs, I see 116 occurrences of the `System.TimeoutException: The operation 'ScanContainer' with id 'XXXX' did not complete in '00:02:00'.` exception in the last 24 hours.  That AVERAGES to once every 12.4 minutes or so.  When I checked on the 7th, the same query showed an average of about once every 16 minutes.

That said, it never goes more than an hour before recurring, and you can do the math to see that it often happens 5 or more times in the same hour.

As for which operation or code path is causing the problem, we are just using what I believe is a standard azure-functions blob trigger defined via the function.json bindings, waiting on new blob events.  There is no stack trace in the logs accompanying the exception, so I can't really tell you more than that.

The only particularly odd thing about our environment is that we are running our own azure-functions container in an AKS cluster instead of using Functions directly.  I have a support request open with MS.  Most recently, one of the functions support people requested some network traces.  I am going to generate those for him today.

derekrprice:
@paulbatum, here's a recent full log entry from the exception, in case you want it.  The only line numbers are from the TimeoutHandler class:
```2020-01-09T13:09:03.018015463Z fail: Microsoft.Azure.WebJobs.Script.WebHost.WebScriptHostExceptionHandler[0]
2020-01-09T13:09:03.018052863Z       An unhandled exception has occurred. Host is shutting down.
2020-01-09T13:09:03.018075663Z System.TimeoutException: The operation 'ScanLogs' with id 'e1688b13-770f-4466-b45b-40b4c1ce21d0' did not complete in '00:02:00'.
2020-01-09T13:09:03.018081163Z    at Microsoft.Azure.WebJobs.Extensions.Storage.TimeoutHandler.ExecuteWithTimeout[T](String operationName, String clientRequestId, IWebJobsExceptionHandler exceptionHandler, Func`1 operation) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Extensions.Storage\TimeoutHandler.cs:line 30
2020-01-09T13:09:03.018094264Z fail: Microsoft.Azure.WebJobs.Script.WebHost.WebScriptHostExceptionHandler[0]
2020-01-09T13:09:03.018098464Z       An unhandled exception has occurred. Host is shutting down.
2020-01-09T13:09:03.018102464Z System.TimeoutException: The operation 'ScanContainer' with id 'fd5798ef-ff98-4ef8-b488-802e1aa97fda' did not complete in '00:02:00'.
2020-01-09T13:09:03.018106164Z    at Microsoft.Azure.WebJobs.Extensions.Storage.TimeoutHandler.ExecuteWithTimeout[T](String operationName, String clientRequestId, IWebJobsExceptionHandler exceptionHandler, Func`1 operation) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Extensions.Storage\TimeoutHandler.cs:line 30
2020-01-09T13:09:03.082965328Z info: Microsoft.Azure.WebJobs.Hosting.JobHostService[0]
2020-01-09T13:09:03.082998628Z       Stopping JobHost
2020-01-09T13:09:03.24790997Z info: Host.Startup[0]
2020-01-09T13:09:03.248384776Z       Job host stopped
2020-01-09T13:09:03.4627207Z Application is shutting down...```

batizar:
Hi guys, any updates on this? We see the same exception hosting WebJobs with BlobTrigger on AKS. Thanks.

brettsam:
Yes! Please try using Microsoft.Azure.WebJobs.Extensions.Storage 4.0.0-preview2, which has the latest fix for this. Release notes are here: https://github.com/Azure/azure-webjobs-sdk/releases/tag/storage-v4.0.0-preview2.

batizar:
Thanks @brettsam

batizar:
Hi @brettsam, we upgraded to the preview2 package and our pod in dev env started crashing every few seconds in K8S, it was crashing less frequently before this. The package definitely is not addressing the issue.

batizar:
@brettsam the only difference I see from our container logs is that there is no exceptions anymore, just restart logs.

brettsam:
Do you have any logs that may be useful? You don't see any exceptions being thrown? And going back to preview1 works again?

batizar:
@brettsam after a bit looking around and removing any reference to `WindowsAzure.Storage` or `Microsoft.WindowsAzure.Storage` from solution and replacing them with `Microsoft.Azure.Cosmos` everything started to work fine and it had no restart in dev env for 16 hours, thanks.

Any estimations on release date?

brettsam:
I'm glad you got it working. I need to write up a small migration guide (basically -- change your namespaces) so no one else runs into this.

We have no other known work required for this release, but I'd like to wait a few more days to get a little more usage under our belts and see if any other issues arise before we roll it out to everyone. Thank you for your feedback already, it is very helpful to know how things are going with these previews.

PR #2293, opened 2019-08-06
Passing the scoped service provider to bindings and exposing a mechanism to instantiate objects using injected dependencies

fabiocav:
@pragnagopa these are the changes from https://github.com/Azure/azure-webjobs-sdk/pull/2270, updated based on feedback. Tests will be coming in a separate PR, just need to get this in to flow into the host.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Threading;
 using Microsoft.Azure.WebJobs.Host.Protocols;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace Microsoft.Azure.WebJobs.Host.Bindings
 {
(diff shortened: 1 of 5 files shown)


src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
```diff
@@ -48,5 +66,39 @@ public class FunctionBindingContext
...
+        /// <summary>
+        /// Creates an object instance with constructor arguments provided in the
+        /// method call and from the current function context scoped services.
+        /// </summary>
+        /// <param name="type">The type to instantiate.</param>
+        /// <param name="parameters">Parameters not provided by the function service provider.</param>
+        /// <returns>An instance of the type specified.</returns>
+        public object CreateObjectInstance(Type type, params object[] parameters)
+        {
+            return ActivatorUtilities.CreateInstance(_functionInvocationServices, type, parameters);
```
pragnagopa, line 84: When is CreateObjectInstance called? Also, can you clarify what do you mean by `Parameters not provided by the function service provider.`

fabiocav: This is a new public API exposed to extension authors (and other components in the pipeline) to instantiate objects, which will leverage DI. This helps them avoid direct instantiation/`Activator` calls while enabling automatic service injection.

`Parameters not provided by the function service provider` are all parameters directly provided by the invoker, instead of the DI service provider.

pragnagopa (approved):
LGTM!

PR #2303, opened 2019-08-16
Added "samplingExcludedTypes" configuration to AppInsightsLoggers

### LINKED ISSUE #4753: Change behavior so that requests and metrics are never throttled
https://docs.microsoft.com/en-us/azure/azure-monitor/app/sampling#types-of-sampling
<ExcludedTypes>Trace;Exception</ExcludedTypes>
A semi-colon delimited list of types that you do not want to be sampled. Recognized types are: Dependency, Event, Exception, PageView, Request, Trace. All instances of the specified types are transmitted; the types that are not specified are sampled.
 

The setting is here https://github.com/Microsoft/ApplicationInsights-dotnet/blob/master/src/ServerTelemetryChannel/AdaptiveSamplingTelemetryProcessor.cs#L70, not on the SamplingSettings like I was hoping. So we’ll need to figure out where this goes and what it looks like in host.json, but should be pretty easy once we decide.

gzuber:
Resolves https://github.com/Azure/azure-functions-host/issues/4753

I made the initial decision to create a new field `samplingExcludedTypes` at the `applicationInsights` level in host.json (as opposed to inside the `samplingSettings` object). I did this because it more closely mimics the way ApplicationInsights has `ExcludedTypes` outside of `SamplingPercentageEstimatorSettings`, so anyone with experience with AppInsights won't be thrown for a loop. Alternatively, this can be moved to be inside our `samplingSettings` object. I feel this might be more intuitive for someone with no prior experience -- it has to do with sampling, so it's in the `samplingSettings`. The only drawback to that approach -- and one reason I didn't put it there to start -- is that it requires a bit of special case code in azure-functions-host that would just move the value to outside the `samplingSettings` object (coincidentally to right where I put it to begin with). We are special casing to an extent already with `isEnabled`, but it's not quite the same since `isEnabled` is more of a switch on/off than a setting that flows through.

TL;DR -- A decision has to be made as to whether the potential benefit to usability would outweigh the cost of special casing the code.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
@@ -23,6 +23,16 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
         public SamplingPercentageEstimatorSettings SamplingSettings { get; set; }
 
         /// <summary>
+        /// Gets or sets excluded types for sampling.
+        /// </summary>
+        public string SamplingExcludedTypes { get; set; }
+
+        /// <summary>
+        /// Gets or sets included types for sampling.
+        /// </summary>
+        public string SamplingIncludedTypes { get; set; }
+
+        /// <summary>
         /// Gets or sets snapshot collection options.
         /// </summary>
         public SnapshotCollectorConfiguration SnapshotConfiguration { get; set; }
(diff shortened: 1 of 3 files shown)


brettsam:
I think that's a fine choice to break the setting out. If anyone asks why we did it that way, we have a good answer :-)

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
```diff
@@ -22,6 +22,11 @@ public class ApplicationInsightsLoggerOptions : IOptionsFormatter
         /// </summary>
         public SamplingPercentageEstimatorSettings SamplingSettings { get; set; }
 
+        /// <summary>
+        /// Gets or sets excluded types for sampling.
+        /// </summary>
+        public string SamplingExcludedTypes { get; set; }
```
brettsam, line 28: While we're in here -- I think we should just add `IncludedTypes` as well to give people the full API.

brettsam (commented):
I think we should add IncludedTypes here as well... should be pretty straightforward.

brettsam:
Looking at their XML config, I actually think we *should* put this under SamplingSettings in host.json. It looks like App Insights puts these settings all under the same XML node in their config: https://docs.microsoft.com/en-us/azure/azure-monitor/app/sampling#configuring-adaptive-sampling-for-aspnet-applications. So we should get that working as well. But doing it this way here still works -- we can do that translation in the host repo.

gzuber:
I added IncludedTypes here, and I have the translation in the Host repo done here: https://github.com/Azure/azure-functions-host/pull/4843

brettsam (approved):
Looks great!

PR #2315, opened 2019-08-29
Adding support for runtime scale monitoring

mathewc:
Addresses https://github.com/Azure/azure-functions-host/issues/4657

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Listeners/QueueListener.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Listeners/QueueListener.cs
@@ -12,6 +12,7 @@ using Microsoft.Azure.WebJobs.Extensions.Storage;
 using Microsoft.Azure.WebJobs.Host.Executors;
 using Microsoft.Azure.WebJobs.Host.Listeners;
 using Microsoft.Azure.WebJobs.Host.Protocols;
+using Microsoft.Azure.WebJobs.Host.Scale;
 using Microsoft.Azure.WebJobs.Host.Timers;
 using Microsoft.Extensions.Logging;
 using Microsoft.WindowsAzure.Storage;
@@ -19,8 +20,10 @@ using Microsoft.WindowsAzure.Storage.Queue;
 
 namespace Microsoft.Azure.WebJobs.Host.Queues.Listeners
 {
-    internal sealed partial class QueueListener : IListener, ITaskSeriesCommand, INotificationCommand
+    internal sealed partial class QueueListener : IListener, ITaskSeriesCommand, INotificationCommand, IScaleMonitor<QueueTriggerMetrics>
     {
+        const int NumberOfSamplesToConsider = 5;
+
         private readonly ITaskSeriesTimer _timer;
         private readonly IDelayStrategy _delayStrategy;
         private readonly CloudQueue _queue;
@@ -35,10 +38,18 @@ namespace Microsoft.Azure.WebJobs.Host.Queues.Listeners
         private readonly TimeSpan _visibilityTimeout;
         private readonly ILogger<QueueListener> _logger;
         private readonly FunctionDescriptor _functionDescriptor;
+        private readonly string _functionId;
+        private readonly ScaleMonitorDescriptor _scaleMonitorDescriptor;
+
         private bool? _queueExists;
         private bool _foundMessageSinceLastDelay;
         private bool _disposed;
         private TaskCompletionSource<object> _stopWaitingTaskSource;
+
+        // for mock testing only
+        internal QueueListener()
+        {
+        }
 
         public QueueListener(CloudQueue queue,
             CloudQueue poisonQueue,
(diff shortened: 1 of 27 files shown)


build/common.props
```diff
@@ -1,9 +1,8 @@
 <Project>
   <PropertyGroup>
     <!-- Extensions can have independent versions and only increment when released -->
-    <Version>3.0.13$(VersionSuffix)</Version>
-    <ExtensionsStorageVersion>3.0.9$(VersionSuffix)</ExtensionsStorageVersion>
-
+    <Version>3.0.14$(VersionSuffix)</Version>
```
mathewc: This is just for my own local testing and won't be checked in

src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Listeners/QueueListener.cs
```diff
@@ -377,5 +387,184 @@ private void ThrowIfDisposed()
...
+        }
+
+        async Task<ScaleMetrics> IScaleMonitor.GetMetricsAsync()
+        {
+            return await GetMetricsAsync();
+        }
+
+        public async Task<QueueTriggerMetrics> GetMetricsAsync()
+        {
+            int queueLength = 0;
```
mathewc, line 411: The GetMetricsAsync/GetScaleStatus implementations are a direct port from ScaleController

src/Microsoft.Azure.WebJobs.Host/Listeners/HostListenerFactory.cs
```diff
@@ -26,9 +27,10 @@ internal class HostListenerFactory : IListenerFactory
         private readonly ILoggerFactory _loggerFactory;
         private readonly ILogger _logger;
         private readonly bool _allowPartialHostStartup;
+        private readonly IScaleMonitorManager _monitorManager;
 
         public HostListenerFactory(IEnumerable<IFunctionDefinition> functionDefinitions, SingletonManager singletonManager, IJobActivator activator,
-            INameResolver nameResolver, ILoggerFactory loggerFactory, bool allowPartialHostStartup = false)
+            INameResolver nameResolver, ILoggerFactory loggerFactory, IScaleMonitorManager monitorManager, bool allowPartialHostStartup = false)
```
mathewc, line 33: I introduced this IScaleMonitorManager as a registry for monitor instances, to solve the problem that these instances are discovered/registered at *runtime* below. This way, we only register the monitor for a listener if it isn't a disabled function. Open to better ways of doing this. I didn't think I should be taking and dynamically modifying the DI services collection.

src/Microsoft.Azure.WebJobs.Host/Scale/IScaleMonitor.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// </summary>
+        /// <param name="context">The <see cref="ScaleStatusContext"/> to use to determine
+        /// the scale status.</param>
+        /// <returns>The scale status.</returns>
+        ScaleStatus GetScaleStatus(ScaleStatusContext context);
+    }
+
+    public interface IScaleMonitor<TMetrics> : IScaleMonitor where TMetrics : ScaleMetrics
+    {
+        new Task<TMetrics> GetMetricsAsync();
```
mathewc, line 36: The non-generic/generic split in these interfaces is somewhat unfortunate for implementors, though necessary I think. The functions host needs to interact with collections of IScaleMonitors (non-generic), however it also needs access to the actual TMetrics type, for serialization/deserialization, which it does at runtime. Open to other ideas on how to handle this.

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleMonitorManager.cs
```diff
@@ -0,0 +1,44 @@
...
+        private readonly List<IScaleMonitor> _monitors = new List<IScaleMonitor>();
+        private object _syncRoot = new object();
+
+        public ScaleMonitorManager()
+        {
+        }
+
+        public ScaleMonitorManager(IEnumerable<IScaleMonitor> monitors)
+        {
+            // add any initial monitors coming from DI
```
mathewc, line 19: In addition to IListener based monitors mapped to functions, I've designed all this to support more general monitors, e.g. a host level CPU/Memory monitor. Such a monitor isn't tied to a function listener but could be registered with DI and would be picked up here.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/WebJobs.Host.EndToEndTests.csproj
```diff
@@ -32,7 +32,7 @@
     <PackageReference Include="Microsoft.Extensions.Logging" Version="2.1.0" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.5.0" />
     <PackageReference Include="Moq" Version="4.7.145" />
-    <PackageReference Include="WindowsAzure.Storage" Version="9.3.1" />
+    <PackageReference Include="WindowsAzure.Storage" Version="9.3.3" />
```
mathewc, line 35: Was getting a bunch of build errors, even when synced to dev w/o my changes. Not sure why - build was complaining that this package was being downgraded. This change fixes it.

Issue #2038, opened 2018-11-29
Enable Event Hubs dependency tracking in Application Insights


samneirinck:
#### Repro steps
1. Create a function which sends to event hubs using either output binding or using the event hub nuget packages directly.
2. Trigger the function and let it send to even thub

#### Expected behavior
A dependency is tracked in Application Insights

#### Actual behavior
No dependency is tracked in Application Insights

#### Known workarounds
Manually wrap your calls to event hubs, and track a dependency. Or listen to the `Microsoft.Azure.EventHubs` EventSource for the events, and track them as dependency.

#### Related information 
On https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs#L71 we can see the following:

```csharp
var includedActivities = dependencyCollector.IncludeDiagnosticSourceActivities;
includedActivities.Add("Microsoft.Azure.ServiceBus");
```

If I set a breakpoint there, and also add the `Microsoft.Azure.EventHubs` to the list, the dependency is tracked correctly.

Of course, I can't do this in Azure, so I'm not sure if there's another workaround, or if this can get added?

sachinarora87:
Hi  Is this issue fixed?  I am also facing similar issue. 

- Create a function app. Using EventHubTrigger (Microsoft.Azure.WebJobs.Extensions.EventHubs v5.0.0-beta.7). 
- Publish a message on the event hub. 
- We can see function is processing message successfully but we cannot see Event hub in dependency telemetry in application insights. 
 
Is it due to the latest Azure.Messaging.EventHubs SDK?

Here logging levels
```
"logging": {
    "logLevel": {
      "default": "Information",
      "Host": "Information",
      "Function": "Information",
      "EDX.Stock": "Information"
    },
    "applicationInsights": {
      "samplingSettings": {
        "isEnabled": true,
        "excludedTypes": "Dependency;Event;Exception",
        "includedTypes": "PageView;Trace;Request"
      },
      "enableLiveMetrics": true,
      "enableDependencyTracking": true,
      "enablePerformanceCountersCollection": true
    }
  }
```

Following are the packages installed on function app.
```
    <PackageReference Include="Microsoft.Azure.Functions.Extensions" Version="1.1.0" />
[...]

PR #2335, opened 2019-09-24
Enable distributed tracing for Event Hubs trigger

lmolkova:
Enabled distributed tracing for EventHubs.

This also needs some changes in EventHubs extension: https://github.com/Azure/azure-functions-eventhubs-extension/pull/20 for end-to-end to work. Without extensions changes, this PR will only enable tracking send and receive calls, but not correlation between producer and consumer.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
@@ -6,12 +6,15 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
+using System.Text;
 using Microsoft.ApplicationInsights;
 using Microsoft.ApplicationInsights.DataContracts;
 using Microsoft.ApplicationInsights.Extensibility;
 using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
+using Newtonsoft.Json;
 
 namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
 {
@@ -441,6 +444,7 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
                 return;
             }
 
+            var allScopes = DictionaryLoggerScope.GetMergedStateDictionary();
             // HTTP and ServiceBus triggers are tracked automatically by the ApplicationInsights SDK
             // In such case a current Activity is present.
             // We won't track and only stamp function specific details on the RequestTelemetry
@@ -449,7 +453,7 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
             var currentActivity = Activity.Current;
             if (currentActivity == null ||
                 // Activity is tracked, but Functions wants to ignore it:
-                DictionaryLoggerScope.GetMergedStateDictionary().ContainsKey("MS_IgnoreActivity") ||
+                allScopes.ContainsKey("MS_IgnoreActivity") ||
                 // Functions create another RequestTrackingTelemetryModule to make sure first request is tracked (as ASP.NET Core starts before web jobs)
                 // however at this point we may discover that RequestTrackingTelemetryModule is disabled by customer and even though Activity exists, request won't be tracked
                 // So, if we've got AspNetCore Activity and EnableHttpTriggerExtendedInfoCollection is false - track request here.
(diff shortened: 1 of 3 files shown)


lmolkova:
@brettsam please review

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -459,31 +463,99 @@ private void StartTelemetryIfFunctionInvocation(IDictionary<string, object> stat
...
+        {
+            if (activities == null)
+            {
+                return;
+            }
+
+            if (activities.Any(l => l.Recorded))
+            {
+                request.ProactiveSamplingDecision = SamplingDecision.SampledIn;
+            }
```
brettsam, line 525: What does Recorded mean?
Also, you can force something to be included even if Sampling is on?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -459,31 +463,99 @@ private void StartTelemetryIfFunctionInvocation(IDictionary<string, object> stat
...
+                    .Append('.')
+                    .Append('\"');
+
+                // we explicitly ignore sampling flag, tracestate and attributes at this point.
+                linksJson.Append("},");
+            }
+
+            linksJson.Append("]");
+
+            request.Properties["_MS.links"] = linksJson.ToString();
```
brettsam, line 560: is this a special property?

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/ApplicationInsightsEndToEndTests.cs
```diff
@@ -447,6 +455,236 @@ public async Task ApplicationInsights_HttpRequestTrackingByWebJobs(string testNa
...
+                Startup.Host = host;
+                await host.StartAsync();
+
+                var loggerProvider = host.Services.GetServices<ILoggerProvider>().OfType<ApplicationInsightsLoggerProvider>().Single();
+                var logger = loggerProvider.CreateLogger(LogCategories.Results);
+
+                string traceId = ActivityTraceId.CreateRandom().ToHexString();
+                string parentSpanId = ActivitySpanId.CreateRandom().ToHexString();
+                string traceFlags = sampledIn ? "01" : "00";
+                string parentId = $"00-{traceId}-{parentSpanId}-{traceFlags}";
```
brettsam, line 637: Oh, so sampling is part of the trace context id? cool :-)

edit -- looks like it: https://www.w3.org/TR/trace-context/#trace-flags

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/ApplicationInsightsEndToEndTests.cs
```diff
@@ -447,6 +455,236 @@ public async Task ApplicationInsights_HttpRequestTrackingByWebJobs(string testNa
...
+                        anySampledIn ? ActivityTraceFlags.Recorded : ActivityTraceFlags.None);
+                links[1] = new Activity("link1").SetParentId(
+                    ActivityTraceId.CreateRandom(), 
+                    ActivitySpanId.CreateRandom(),
+                    ActivityTraceFlags.None);
+
+                // simulate functions behavior to set request on the scope
+                using (var _ = logger.BeginScope(new Dictionary<string, object>
+                {
+                    ["Links"] = links
```
brettsam, line 702: So these can be two, separate Activities? Is this how you deal with batches?

brettsam (commented):
It looks good -- but had some questions to make sure I understand.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
lmolkova: Recoded flag on activity means that parent was recorded (sampled in).

In event hubs scenarios, we may get 10 different unrelated messages with different contexts. We know which of there were sampled in by upstream service.

If any of the messages is sampled in, we mark request telemtery (for trigger) to be proactively sampled in.

It is processed once again when request is being tracked (function completes) and sampling algorithm will give more priority to telemtery that was proactively sampled in (while still keeping the rate under configured one).

DONE: It actually reminded me that we need to set Activity.Recorded to true on the Activity we create for trigger so everything under it's scope will get higher priority  - I'll add it and more tests for it

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
lmolkova: Yeah, UX will use this property to discover and show many-to-one relationship (many messages are merging into one processing). 

'_MS' is prefix that indicates internal stuff and 'links' is a common name in tracing world to represent such kind of relationships in batching scenarios

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/ApplicationInsightsEndToEndTests.cs
lmolkova: so each message potentially carries a unique context. Each of the context in the message we call a 'link' and represent with Activity. This activity never starts (i.e. never becomes current)

When we track event hub trigger (a batch processing) we create a brand new Activity and request telemtery out of it.
We pass all these links to this new context. There are a few ways:
- from AppInsights point of view, this is done via custom _MS.links property on request telemetry - this is what we do here.
- links are also first-class citizen in the OpenTelemtery - OSS effort we heavily participate in and moving towards.
- we are also working with .NET folks to support links on Activity level

So the new distributed context is generated for batch, but knows about linked messages. UX will show them (on the Gantt chart first and eventually App-Map) and users will also be able to query them (this is advanced experience). UX experiences are in development now.

brettsam:
I know there's another piece to this in the EH repo -- but is this good to go in before that?

lmolkova:
> I know there's another piece to this in the EH repo -- but is this good to go in before that?

Yes, it's good to go.
Thanks for the review!

DanielLarsenNZ:
@lmolkova @brettsam will this fix #2038 once it is included in Microsoft.Azure.WebJobs.Extensions.EventHubs ?

PR #2337, opened 2019-09-25
Fix disabling W3C distributed tracing

lmolkova:


src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
```diff
@@ -139,6 +139,7 @@ public static IServiceCollection AddApplicationInsights(this IServiceCollection
                 Activity.DefaultIdFormat = options.HttpAutoCollectionOptions.EnableW3CDistributedTracing
                     ? ActivityIdFormat.W3C
                     : ActivityIdFormat.Hierarchical;
+                Activity.ForceDefaultIdFormat = true;
```
brettsam, line 142: What does this mean? Even if you set the default format it may not use it?

brettsam (approved):

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
lmolkova: yep, if the format is not enforced, Activity will figure it out based on parent's format. This is a way for .NET Activity to keep backward compatibility. AppInsights is back-compatible anyway, so we don't need/want .NET Activity to be flexible around format

PR #2338, opened 2019-09-25
removing FormattedLogValuesCollection tests

brettsam:
Finalizing the removal of this type. I had originally removed our usage in #2271 but missed these files. It was showing up as a warning in our ReadyToRun compilation for v3.


brettsam:
This is more an an FYI as I missed these files earlier in #2271 -- gonna merge as soon as I see tests pass.

Issue #2109, opened 2019-02-01
Feature Request: Support Managed Service Identity for Storage connections


PhoenixHe-NV:
Please provide a succinct description of the issue.

#### Repro steps

Currently Azure Storage supports [Managed Service Identity](https://docs.microsoft.com/en-us/azure/storage/common/storage-auth-aad-msi?toc=%2fazure%2fstorage%2fqueues%2ftoc.json). But azure webjob sdk only spports connection string for storage account. It would be great that webjob sdk support MSI.

#### Known workarounds

I've found some workarounds by inject storage account on registering. First I need to generate the CloudStorageAccount by MSI in advance. Then, there are two services I need to inject:

```csharp
var blobClient = storageAccount.CreateCloudBlobClient();
var container = blobClient.GetContainerReference("webjob-lock");

var builder = new HostBuilder()
   .ConfigureWebJobs(b =>
   {
     b.Services.AddSingleton(new DistributedLockManagerContainerProvider
     {
       InternalContainer = container
     });
     b.Services.AddSingleton<StorageAccountProvider>(new ManagedIdentityStorageAccountProvider(storageAccount));

     b.AddAzureStorageCoreServices();
     b.AddAzureStorage();
     b.AddTimers();
   })
```

```csharp
public class ManagedIdentityStorageAccountProvider : StorageAccountProvider
{
  private readonly CloudStorageAccount storageAccount;

  public ManagedIdentityStorageAccountProvider(CloudStorageAccount storageAccount) : base(null)
  {
    this.storageAccount = storageAccount;
  }

  public override StorageAccount Get(string name)
  {
    return StorageAccount.New(this.storageAccount);
  }
}
```

I'm still looking for official MSI support.

NullMDR:
Any update on this one?

ross-p-smith:
The above code fails for me as the StorageCredentails does not have an AccountName when created with a Token Credential. This would work if we could approve Pull Request #2000

mmusin:
A year had passed...

mdddev:
I want to try out the proposed workaround of @NullMDR but I am to inexperienced to actually apply it properly. Unfortunately the information in the code snippet provided, I don't know where to actually implement it. Is this program.cs or function.cs? Sorry for the lack of knowledge..

mdddev:
Nevermind the above, it appears I have not read thoroughly enough. 
[...]

PR #2344, opened 2019-10-09
Managed identity for Storage Accounts

ross-p-smith:
This PR is around Feature Request #2109. There is a new ManagedIdentityStorageAccountProvider that will create a StorageAccountClient using Managed Identity. The feature is toggled by passing in the tenantId in your HostBuilder.

b.AddAzureStorage(managedIdentityTenantId: "xxxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxxx");

The StorageAccount or Connection attribute can then be used to determinethe Storage Account Name that will be connected to.

eg    [StorageAccount("stor161")]

This has been tested reading from a BlobTrigger and writing to a Queue with no Credentials in config files.


ross-p-smith:
I'm getting this in the AppVeyor build - how do I go about fixing that up?

The path 'bin' specified in the 'TestAdapterPath' is invalid. Error: The custom test adapter search path provided was not found, provide a valid path and try again.

JSalter2013:
@ross-p-smith Any word on when this PR is likely to accepted please? It would be handy for a client project I'm working on - they'd prefer to use MSI rather than storage connection strings when wiring up config for an azure functions durable orchestration running in an ASE/service plan.

AdamCoulterOz:
hi @fabiocav - this is needed to be able to remove passing AzureWebJobsStorage with a secret (AccountKey) in the connection string, which is not the ideal security scenario, and then requires rotation. Can we get an ETA on when this will be considered?

DanniJuhl:
Is this still a possibility to have considered? Quite a few organisations that I speak with, are not fond of this current limitation, and with the public preview of restricting access to Azure Storage from SAS/Account Key, I suspect this will only increase over time.

azfuncgh:

PR #2346, opened 2019-10-11
Make QueueTrigger shutdown graceful.

### LINKED ISSUE #2345: QueueTrigger does not do graceful shutdown (leaves queue messages hidden)
When doing graceful shutdown of the Host, any queue messages being processed are abandoned, left invisible in the queue and not available until the lease timeout expires, rather than releasing the leases to make the messages visible again and allow someone else to process them. For apps that use only QueueTrigger, this is the main thing graceful shutdown can/should do.

#### Repro steps

1. Create a program that processes a message from a queue.

2. Have the processing function take a long time but respect the cancellation token (return immediately if cancellation is requested).

3. Gracefully shutdown the host (such as via Ctrl+C).

#### Expected behavior

The message the host was processing is released - visible in the queue and available to others to process.

#### Actual behavior

The message the host was processing is left with its lease intact but abandoned - it is not visible in the queue, and nothing else will be able to process it until the lease it had previously taken expires.

#### Known workarounds

None known.

#### Related information 

App.csproj:
```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp3.0</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.13" />
    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Storage" Version="3.0.8" />
    <PackageReference Include="Microsoft.Azure.WebJobs.Host.Storage" Version="3.0.13" />
    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="3.0.0" />
  </ItemGroup>
</Project>

```

Program.cs:
```c#
using Microsoft.Internal.WindowsAzure.Storage;
using System;

class Program
{
    static void Main()
    {
        new HostBuilder()
            .ConfigureWebJobs(b => {
                b.AddAzureStorageCoreServices();
                b.AddAzureStorage();
            })
            .ConfigureLogging(b => b.AddConsole())
            .UseConsoleLifetime()
            .Build()
            .Run();
    }
}
```

Processor.cs:
```c#
using Microsoft.Azure.WebJobs;
using Newtonsoft.Json.Linq;
using System;
using System.Threading;

public static class Processor
{
    public static void Process([QueueTrigger("test")] JObject message, CancellationToken cancellationToken,
        ILogger logger)
    {
        logger.LogInformation($"Received: {message.ToString()}");
        logger.LogInformation("Waiting up to 10 minutes for cancellation token.");
        cancellationToken.WaitHandle.WaitOne(TimeSpan.FromMinutes(10));

        if (cancellationToken.IsCancellationRequested)
        {
            logger.LogInformation("Cancellation requested.");
            cancellationToken.ThrowIfCancellationRequested();
        }

        logger.LogInformation("Done processing message.");
    }
}
```

davidmatson:
Fixes #2345.

Disclaimer: I haven't had a chance to test, but here's the overall idea for a fix. (The project doesn't just compile as-is on my box, so this may not even be in a fully-compiling state yet.)

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Listeners/QueueListener.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Listeners/QueueListener.cs
@@ -40,6 +40,7 @@ namespace Microsoft.Azure.WebJobs.Host.Queues.Listeners
         private readonly FunctionDescriptor _functionDescriptor;
         private readonly string _functionId;
         private readonly ScaleMonitorDescriptor _scaleMonitorDescriptor;
+        private readonly CancellationTokenSource _shutdownCancellationTokenSource;
 
         private bool? _queueExists;
         private bool _foundMessageSinceLastDelay;
@@ -122,6 +123,7 @@ namespace Microsoft.Azure.WebJobs.Host.Queues.Listeners
             _delayStrategy = new RandomizedExponentialBackoffStrategy(QueuePollingIntervals.Minimum, maximumInterval);
 
             _scaleMonitorDescriptor = new ScaleMonitorDescriptor($"{_functionId}-QueueTrigger-{_queue.Name}".ToLower());
+            _shutdownCancellationTokenSource = new CancellationTokenSource();
         }
 
         // for testing
@@ -142,10 +144,13 @@ namespace Microsoft.Azure.WebJobs.Host.Queues.Listeners
 
         public async Task StopAsync(CancellationToken cancellationToken)
         {
-            ThrowIfDisposed();
-            _timer.Cancel();
-            await Task.WhenAll(_processing);
-            await _timer.StopAsync(cancellationToken);
+            using (cancellationToken.Register(() => _shutdownCancellationTokenSource.Cancel()))
+            {
+                ThrowIfDisposed();
+                _timer.Cancel();
+                await Task.WhenAll(_processing);
+                await _timer.StopAsync(cancellationToken);
+            }
         }
 
         public void Dispose()
@@ -153,6 +158,7 @@ namespace Microsoft.Azure.WebJobs.Host.Queues.Listeners
             if (!_disposed)
             {
                 _timer.Dispose();
+                _shutdownCancellationTokenSource.Dispose();
                 _disposed = true;
             }
         }
@@ -325,7 +331,10 @@ namespace Microsoft.Azure.WebJobs.Host.Queues.Listeners
                     await timer.StopAsync(cancellationToken);
                 }
 
-                await _queueProcessor.CompleteProcessingMessageAsync(message, result, cancellationToken);
+                // Use a different cancellation token for shutdown to allow graceful shutdown.
+                // Specifically, don't cancel the completion or update of the message itself during graceful shutdown.
+                // Only cancel completion or update of the message if a non-graceful shutdown is requested via _shutdownCancellationTokenSource.
+                await _queueProcessor.CompleteProcessingMessageAsync(message, result, _shutdownCancellationTokenSource.Token);
             }
             catch (StorageException ex) when (ex.IsTaskCanceled())
             {
@@ -533,7 +542,7 @@ namespace Microsoft.Azure.WebJobs.Host.Queues.Listeners
                     return status;
                 }
             }
-            
+
             bool queueLengthDecreasing =
                 IsTrueForLastN(
                     metrics,
--- a/test/Microsoft.Azure.WebJobs.Host.UnitTests/Bindings/Path/BindingParameterResolverTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.UnitTests/Bindings/Path/BindingParameterResolverTests.cs
@@ -5,6 +5,7 @@ using Microsoft.Azure.WebJobs.Host.Bindings.Path;
 using System.Linq;
 using Xunit;
 using System;
+using System.Globalization;
 
 namespace Microsoft.Azure.WebJobs.Host.UnitTests.Bindings.Path
 {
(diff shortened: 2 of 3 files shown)


davidmatson:
I tested the change, and it did build and work fine. For aborting the graceful shutdown via the cancellation token to IHostedService.StopAsync, I had to plumb through one more part - IJobHost.StopAsync didn't take this parameter. To avoid a breaking change, I had JobHost implement IHostedService in addition to IJobHost and prefer calling the IHostedService version and passing in the cancellation token where possible.

davidmatson:
@mathewc or @brettsam, could you take a look at this PR? It's a fairly small change to fix this bug. Thanks!

src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Listeners/QueueListener.cs
```diff
@@ -40,6 +40,7 @@ internal sealed partial class QueueListener : IListener, ITaskSeriesCommand, INo
         private readonly FunctionDescriptor _functionDescriptor;
         private readonly string _functionId;
         private readonly ScaleMonitorDescriptor _scaleMonitorDescriptor;
+        private readonly CancellationTokenSource _shutdownCancellationTokenSounce;
```
cloverst: typo: _shutdownCancellationTokenSou**r**ce (source not sounce)

davidmatson: Thanks! It's fixed now.

src/Microsoft.Azure.WebJobs.Host/Hosting/JobHostService.cs
```diff
@@ -13,11 +13,13 @@ public class JobHostService : IHostedService
     {
         private readonly ILogger<JobHostService> _logger;
         private readonly IJobHost _jobHost;
+        private readonly IHostedService _jobHostHostedService;
 
         public JobHostService(IJobHost jobhost, ILogger<JobHostService> logger)
         {
             _logger = logger ?? throw new ArgumentNullException(nameof(logger));
             _jobHost = jobhost;
+            _jobHostHostedService = jobhost as IHostedService;
```
brettsam: Can you do this cast down in `StopAsync` so we don't need a separate field?

src/Microsoft.Azure.WebJobs.Host/JobHost.cs
```diff
@@ -24,7 +25,7 @@ namespace Microsoft.Azure.WebJobs
     /// A <see cref="JobHost"/> is the execution container for jobs. Once started, the
     /// <see cref="JobHost"/> will manage and run job functions when they are triggered.
     /// </summary>
-    public class JobHost : IJobHost, IDisposable, IJobInvoker
+    public class JobHost : IJobHost, IHostedService, IDisposable, IJobInvoker
```
brettsam: It feels odd to have this implement `IHostedService` when we don't actually register as an `IHostedService` anywhere... and in fact we host it inside of another `IHostedService`. 

@fabiocav -- why did we do it this way?

brettsam (commented):
My only concern is implementing `IHostedService` there as it implies this is a hosted service... yet we never actually use it as an `IHostedService`.

I'd like @fabiocav and @mathewc to weigh in as well to see if they have any concerns.

src/Microsoft.Azure.WebJobs.Host/Hosting/JobHostService.cs
davidmatson: Sure.

src/Microsoft.Azure.WebJobs.Host/JobHost.cs
davidmatson: Yeah, I didn't see any major reason to have JobHostService separate from JobHost, other than perhaps the missing CancellationToken on IJobHost.StopAsync.

JobHostService is a public class, or I'd suggest just removing it entirely. Another option would be to leave the class but change WebJobsHostBuilderExtensions to register JobHost directly rather than JobHostService as the IHostedService (also moving the ILogger logic that's the only thing currently left in JobHostService).

Those both seem like slightly larger changes, so my inclination would be to do what's in the current PR, but I'm open to options here.

davidmatson: Per offline discussion, I've split that commit into its own PR (#2372) so we can merge this part in the meantime.

brettsam (approved):

src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Listeners/QueueListener.cs
```diff
@@ -320,7 +326,7 @@ internal async Task ProcessMessageAsync(CloudQueueMessage message, TimeSpan visi
                     await timer.StopAsync(cancellationToken);
                 }
 
-                await _queueProcessor.CompleteProcessingMessageAsync(message, result, cancellationToken);
+                await _queueProcessor.CompleteProcessingMessageAsync(message, result, _shutdownCancellationTokenSource.Token);
```
brettsam, line 337: One nit -- can you add a comment above this explaining why we're using a different token here? I know we're all going to forget why we did this in about 2 weeks :-)

brettsam:
Just a suggestion about adding a comment -- then can you squash this down to a single commit? After that, I can merge.

src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Listeners/QueueListener.cs
davidmatson: Good idea - sure.

PR #2347, opened 2019-10-14
Added logger to UseWebjobsStartup and UseExternalStartup

madelinegordon:


src/Microsoft.Azure.WebJobs.Host/Hosting/ServiceCollectionDecorator.cs
```diff
@@ -0,0 +1,98 @@
...
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Azure.WebJobs.Host.Hosting
+{
+    public class ServiceCollectionDecorator : IServiceCollection, ITrackingServiceCollection
```
brettsam: Make this `internal`.

src/Microsoft.Azure.WebJobs.Host/Hosting/ITrackingServiceCollection.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Microsoft.Extensions.DependencyInjection;
+
+
+namespace Microsoft.Azure.WebJobs.Host.Hosting
+{
+    interface ITrackingServiceCollection
```
brettsam: Explicitly make this `internal`.

src/Microsoft.Azure.WebJobs.Host/Hosting/ITrackingServiceCollection.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        IServiceCollection ServiceCollection { get; set; }
+
+        List<ServiceDescriptor> Diff { get; }
```
brettsam: It looks like this is used in an iterator only, so make this `IEnumerable<ServiceDescriptor>`.

src/Microsoft.Azure.WebJobs.Host/Hosting/ServiceCollectionDecorator.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        public List<ServiceDescriptor> Diff { get; private set; } // tracks any service collection modification.
+
+        public ServiceCollectionDecorator(IServiceCollection serviceCollection)
+        {
+            ServiceCollection = serviceCollection;
+            Diff = new List<ServiceDescriptor>();
+        }
+
+        public ServiceDescriptor this[int index] { get => ServiceCollection[index]; set => throw new NotImplementedException(); }
```
brettsam: I think the setter here should do `ServiceCollection[index] = value`

src/Microsoft.Azure.WebJobs.Host/Hosting/ServiceCollectionDecorator.cs
```diff
@@ -0,0 +1,98 @@
...
+        {
+            Diff.Clear();
+        }
+
+        private void TryAdd(ServiceDescriptor item)
+        {
+            if (!Diff.Any(x => x.ServiceType == item.ServiceType)) {
+                Diff.Add(item);
+            }
+        }
```
brettsam: We probably shouldn't filter here -- it's legal to register multiple `ServiceDescriptor`s for the same `ServiceType`. The last one will "win", and it'd be good to have the true representation. Unless this hit a snag somewhere...

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsBuilder.cs
```diff
@@ -12,9 +13,10 @@ internal class WebJobsBuilder : IWebJobsBuilder
     {
         public WebJobsBuilder(IServiceCollection services)
         {
-            Services = services ?? throw new ArgumentNullException(nameof(services));
+            if (services == null) { throw new ArgumentNullException(nameof(services)); } 
+            Services = new ServiceCollectionDecorator(services);
```
brettsam: nit -- can you do `Services = new ServiceCollectionDecorator(services) ?? throw new ArgumentNullException(nameof(services));`

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsBuilderExtensions.cs
```diff
@@ -58,22 +61,72 @@ public static IWebJobsBuilder UseHostId(this IWebJobsBuilder builder, string hos
...
+                    {
+                        implementation = service.ImplementationType.FullName;
+                    }
+                    else if (service.ImplementationFactory != null)
+                    {
+                        implementation = "Using ImplementationFactoy";
+                    }
+                 
+                    string lifetime = service.Lifetime.ToString(); 
+                    logger.LogInformation("Service type: {serviceName}, Implementation: {implementation}, Registered with lifetime: {lifetime}", serviceName, implementation, lifetime);
```
brettsam: I would make this `LogDebug`. We'll see it in Kusto then, but customers won't see it by default. I've seen some people that register dozens of services, so logging as Info could be overload.

brettsam: Couple of suggestions on the message:
1. Instead of `Implementation: {implementation}`, you can do something like `{registeredServiceString}` and when you build it up, make the whole string be `Instance: {instanceType}` or `Implementation: {implemenationType}` or `Factory`.
2. My opinion -- I think we should compact this and format all of them into a single message using a `StringBuilder`. It'll be easier to find in the logs.

Something like: 
```
Services registered by external startup:
  {serviceName}, {registeredServiceString}, {lifetime}
  {serviceName}, {registeredServiceString}, {lifetime}
  {serviceName}, {registeredServiceString}, {lifetime}
```

brettsam: Oh another thing here -- we should log this with an EventId as that greatly improves querying in kusto. You should just be able to use the LogInformation overload that takes an event. and make it something like `new EventId(500, "ExternalStartupServices")`

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsBuilderExtensions.cs
```diff
@@ -58,22 +61,72 @@ public static IWebJobsBuilder UseHostId(this IWebJobsBuilder builder, string hos
...
+                startup.Configure(builder);
+            }
+            else
+            {
+                ConfigureAndLogUserConfiguredServices(startup, builder, logger);      
+            }
             return builder;
         }
 
+        private static void ConfigureAndLogUserConfiguredServices(IWebJobsStartup startup, IWebJobsBuilder builder, ILogger logger)
```
brettsam: Rather than receive an `ILogger`, I think we should take an `ILoggerFactory` (there's a `NullLoggerFactory.Instance` as well). That way, we control what the category of that logger is.

Then down here we can call `loggerFactory.CreateLogger<ServiceCollectionDecorator>` and we'll then be in control of exactly what the category is.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsBuilderExtensions.cs
```diff
@@ -58,22 +61,72 @@ public static IWebJobsBuilder UseHostId(this IWebJobsBuilder builder, string hos
...
+                    else if (service.ImplementationFactory != null)
+                    {
+                        implementation = "Using ImplementationFactoy";
+                    }
+                 
+                    string lifetime = service.Lifetime.ToString(); 
+                    logger.LogInformation("Service type: {serviceName}, Implementation: {implementation}, Registered with lifetime: {lifetime}", serviceName, implementation, lifetime);
+                }
+
+                tracker.ResetTracking();
```
brettsam: Do we need to do this here? If not, can we remove this API from the interface?

src/Microsoft.Azure.WebJobs.Host/Hosting/ITrackingServiceCollection.cs
brettsam: Let's call this something other than `Diff` -- maybe `ExternalServices`?

brettsam (commented):
Nice! I have a bunch of nitpicky comments -- let me know if you have questions on any of them.

src/Microsoft.Azure.WebJobs.Host/Hosting/ITrackingServiceCollection.cs
fabiocav: `TrackedServices` or `TrackedChanges` would be my recommendation

src/Microsoft.Azure.WebJobs.Host/Hosting/ServiceCollectionDecorator.cs
```diff
@@ -0,0 +1,98 @@
...
+using System.Linq;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Azure.WebJobs.Host.Hosting
+{
+    public class ServiceCollectionDecorator : IServiceCollection, ITrackingServiceCollection
+    {
+        public IServiceCollection ServiceCollection { get; set; }
+
+        public List<ServiceDescriptor> Diff { get; private set; } // tracks any service collection modification.
```
fabiocav: Same comment on naming (something like `TrackedChanges` would be appropriate here)

src/Microsoft.Azure.WebJobs.Host/Hosting/ServiceCollectionDecorator.cs
fabiocav: 👍

src/Microsoft.Azure.WebJobs.Host/Hosting/ServiceCollectionDecorator.cs
fabiocav: Also change the type name to `TrackedServiceCollection`

src/Microsoft.Azure.WebJobs.Host/Hosting/ServiceCollectionDecorator.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        public ServiceCollectionDecorator(IServiceCollection serviceCollection)
+        {
+            ServiceCollection = serviceCollection;
```
fabiocav: perform a `null` check here.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsBuilder.cs
fabiocav: Not sure the above would work... the `null` coalescing operator here wouldn't evaluate to `null` here (you'd just end up with the null exception from the constructor when it validates its arguments)

fabiocav: my nit here would be to match the following for consistency:
``` csharp
if (services == null) 
{ 
    throw new ArgumentNullException(nameof(services)); 
} 
```

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsBuilder.cs
```diff
@@ -12,9 +13,10 @@ internal class WebJobsBuilder : IWebJobsBuilder
...
         }
 
-        public IServiceCollection Services { get; }
+        public IServiceCollection Services { get; }
```
fabiocav: Empty trailing spaces

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsBuilderExtensions.cs
```diff
@@ -58,22 +61,72 @@ public static IWebJobsBuilder UseHostId(this IWebJobsBuilder builder, string hos
...
+                ConfigureAndLogUserConfiguredServices(startup, builder, logger);      
+            }
             return builder;
         }
 
+        private static void ConfigureAndLogUserConfiguredServices(IWebJobsStartup startup, IWebJobsBuilder builder, ILogger logger)
+        {
+            ITrackingServiceCollection tracker = builder.Services as ITrackingServiceCollection;
+
+            if (tracker != null)
```
fabiocav: Change to:
```
if (builder.Services is ITrackingServices tracker)
```

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsBuilderExtensions.cs
```diff
@@ -58,22 +61,72 @@ public static IWebJobsBuilder UseHostId(this IWebJobsBuilder builder, string hos
...
+            {
+                startup.Configure(builder);
+
+                foreach (ServiceDescriptor service in tracker.Diff)
+                {
+                    string serviceName = service.ServiceType.FullName;
+                    string implementation = "";
+                    if (service.ImplementationInstance != null)
+                    {
+                        implementation = service.ImplementationInstance.GetType().FullName + "[instance]";
```
fabiocav: nit: add a space before ` [instance]`

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsBuilderExtensions.cs
```diff
@@ -58,22 +61,72 @@ public static IWebJobsBuilder UseHostId(this IWebJobsBuilder builder, string hos
...
+                    }
+                    else if (service.ImplementationType != null)
```
fabiocav: check implementation type first (more common case)

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsBuilderExtensions.cs
fabiocav: I like the suggestion. The one thing I'd add is the external startup type name to the header.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsBuilderExtensions.cs
fabiocav: You need to be able to reset tracking with every iteration (every startup), correct?

src/Microsoft.Azure.WebJobs.Host/Hosting/ITrackingServiceCollection.cs
madelinegordon: > It looks like this is used in an iterator only, so make this `IEnumerable<ServiceDescriptor>`.

In ServiceCollectionDecorator, Add, Remove, and Clear are called on the object. Is this suggestion to make this an IEnumerable here in the interface and then cast to a List in the decorator?

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsBuilder.cs
madelinegordon: > my nit here would be to match the following for consistency:
> 
> ```cs
> if (services == null) 
> { 
>     throw new ArgumentNullException(nameof(services)); 
> } 
> ```
I am not sure I understand fully--is this a formatting nit? Or something else that I am missing?

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsBuilderExtensions.cs
madelinegordon: Yes, tracking should be reset so that only newly added services on startup are logged.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsBuilder.cs
brettsam: Oh right, duh. 

@madelinegordon -- yeah, it's formatting, we put curly braces on newlines by themselves. I think we even have a build step that will log warnings if you don't do that (but maybe not in this repo?)

madelinegordon: Ooh okay, good to know. Thank you!

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsBuilderExtensions.cs
madelinegordon: That makes sense--I have one question though. If the logger is created within the scope of UseWebjobsStartup, how will I be able to access it in order to test that the expected services were logged? Is there an accessor to a factory's loggers?

brettsam: We have a `TestLoggerProvider` in the test code. So you create a `LoggerFactory`, add the test logger provider, then pass that `LoggerFactory` into your test. That will track Loggers created, all log messages, etc. If you search in the test code for where we use `TestLoggerProvider` you should find some examples.

madelinegordon: Ooh okay, I had tried that but registered a logger with the provider instead of using provider.GetAllLogMessages. Thank you!

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsBuilderExtensions.cs
brettsam: Ahh, right.

madelinegordon:
@brettsam @fabiocav I have updated the PR and addressed all comments that were there so far--let me know if there is anything else you find to be updated. Otherwise, will just need your approval (although it looks like there isn't a policy requiring an official approval in this repo). Thanks!

src/Microsoft.Azure.WebJobs.Host/Hosting/ServiceCollectionDecorator.cs
```diff
@@ -0,0 +1,100 @@
...
+        public IServiceCollection ServiceCollection { get; set; }
+
+        public IEnumerable<ServiceDescriptor> TrackedCollectionChanges // tracks any service collection modification.
+        {
+            get
+            {
+                return _trackedChanges as IEnumerable<ServiceDescriptor>;
+            }
+            private set { }
+        }
```
brettsam: For this you should be able to just do (it just makes it shorter).

` public IEnumerable<ServiceDescriptor> TrackedCollectionChanges => _trackedChanges;`

src/Microsoft.Azure.WebJobs.Host/Hosting/ServiceCollectionDecorator.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        private List<ServiceDescriptor> _trackedChanges { get; set; }
```
brettsam: nit: move private fields up to the top of the class.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsBuilderExtensions.cs
```diff
@@ -58,22 +62,77 @@ public static IWebJobsBuilder UseHostId(this IWebJobsBuilder builder, string hos
...
+            {
+                if (tracker != null)
+                {
+                    startup.Configure(builder);
+                    StringBuilder sb = new StringBuilder("Services registered by external startup type " + startup.GetType().ToString() + ":");
+
+                    foreach (ServiceDescriptor service in tracker.TrackedCollectionChanges)
+                    {
+                        string serviceName = "Service name: " + service.ServiceType.FullName;
+                        StringBuilder registeredService = new StringBuilder(", Implementation: ");
```
brettsam: Some logging string nits :-)

1. Don't create a new `StringBuilder` here -- use the existing one `sb` from above.
2. Rather than storing a separate string here for `serviceName` -- just append it. It'll make it a little easier to follow. (see below)
3. Saying "Service name" on every line isn't required as we'll understand it -- and it's mostly for debugging so we can be a little short with it. (see below)
4. Rather than doing "[instance]" -- just replace "Implementation" with "Instance" on those lines. (see below)
5. Use `sb.AppendLine()` -- which I think will use the right line feed mechanics, depending on Linux/Windows. Similarly, I'd use explicit spaces rather than `\t`.

So you can start each line with with:
`sb.AppendLine($"  {service.ServiceType.FullName}: ");` (note just 2 spaces here)

Then depending on whether Implemention/Instance/Factory:
`sb.Append($"Implementation: {service.ImplementationType.FullName}");`
...
`sb.Append($"Instance: {service.ImplementationInstance.GetType().FullName}");`
...
`sb.Append("Factory");`

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsBuilderExtensions.cs
```diff
@@ -58,22 +62,77 @@ public static IWebJobsBuilder UseHostId(this IWebJobsBuilder builder, string hos
...
+                        }
+                        else if (service.ImplementationFactory != null)
+                        {
+                            registeredService.Append("Using ImplementationFactory");
+                        }
+
+                        string lifetime = ", Lifetime: " + service.Lifetime.ToString();
+                        sb.Append("\n\t" + serviceName + registeredService.ToString() + lifetime);
+                    }
+                    logger.LogDebug(sb.ToString(), new EventId(500, "ExternalStartupServices"));
```
brettsam: `EventId` needs to be the first parameter here. -- there's an overload that allows for `object` to be passed, which is why this is working, but the event wouldn't get picked up by kusto.

src/Microsoft.Azure.WebJobs.Host/Hosting/ServiceCollectionDecorator.cs
```diff
@@ -0,0 +1,100 @@
...
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Azure.WebJobs.Host.Hosting
+{
+    internal class ServiceCollectionDecorator : IServiceCollection, ITrackedServiceCollection
```
brettsam: Another tiny nit -- maybe call this `TrackedServiceCollection`? We typically don't use `Decorator`, even if the class is a decorator. Better to describe what it's being used for.

brettsam (commented):
Some naming/string nits -- functionally looks good!

src/Microsoft.Azure.WebJobs.Host/Hosting/ServiceCollectionDecorator.cs
fabiocav: I had that feedback, with the exact name, in a previous commend :)

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsBuilderExtensions.cs
madelinegordon: I think AppendLine adds a newline to the end of the string, so could use with the first and last strings. Otherwise, I have addressed the rest of the comments. Thanks Brett!

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsBuilderExtensions.cs
```diff
@@ -58,22 +61,76 @@ public static IWebJobsBuilder UseHostId(this IWebJobsBuilder builder, string hos
...
+            if (builder.Services is ITrackedServiceCollection tracker)
+            {
+                if (tracker != null)
+                {
+                    startup.Configure(builder);
+                    StringBuilder sb = new StringBuilder("Services registered by external startup type " + startup.GetType().ToString() + ":");
+
+                    foreach (ServiceDescriptor service in tracker.TrackedCollectionChanges)
+                    {
+                        sb.Append($"\n {service.ServiceType.FullName}: ");
```
brettsam, line 110: Does this appear properly formatted? Should it be "\r\n"? https://docs.microsoft.com/en-us/dotnet/api/system.environment.newline?view=netframework-4.8#property-value.

madelinegordon: Yes, you are right! I will create a bug fix.

PR #2351, opened 2019-10-17
setting defaults for queueOptions based on the sku

### LINKED ISSUE #2339: Storage Queues Binding: Apply defaults based on environment capabilities and SKU
Currently, the queue binding default configuration values are applied without taking the hosting environment into consideration. In order to correctly apply default configuration settings that take advantage of hosting capabilities like additional cores, memory, etc., we need to take those factors into consideration.

This should apply to all SKUs, except consumption, where we need to ensure we maintain the same behavior we have today.

soninaren:
closes #2339


src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Config/QueuesOptions.cs
```diff
@@ -34,6 +35,12 @@ public class QueuesOptions : IOptionsFormatter
         public QueuesOptions()
         {
             _newBatchThreshold = -1;
+
+            string skuValue = Environment.GetEnvironmentVariable(Constants.AzureWebsiteSku);
```
mathewc: My only feedback is to put this into a shared helper, e.g. Utility.GetProcessorCountForSku(), since the code is replicated a bunch below (mostly in tests).That way this code and the tests just call the helper

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ParallelExecutionTests.cs
```diff
@@ -52,6 +54,70 @@ public static async Task ParallelQueueTrigger([QueueTrigger(TestQueueName)] int
             }
         }
 
+        [Theory]
+        // One is special case (the old behaviour)
+        [InlineData(1)]
+        // Odd and even values
+        [InlineData(2)]
+        [InlineData(3)]
+        public async Task MaxDegreeOfParallelism_Queues_nonDynamicSku(int batchSize)
```
mathewc: nit: fix casing

mathewc: Also, rather than copy this test, can't you just add a new param to the existing test for sku? E.g. perhaps just a bool isDynamic

soninaren: done

src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Config/QueuesOptions.cs
soninaren: Added the utility function as suggested

src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Config/QueuesOptions.cs
```diff
@@ -68,8 +70,7 @@ public int NewBatchThreshold
             {
                 if (_newBatchThreshold == -1)
                 {
-                    // if this hasn't been set explicitly, default it
-                    return _batchSize / 2;
+                    return (_batchSize / 2) * _processorCount;
```
mathewc, line 75: nit: please leave the comment that was there

src/Microsoft.Azure.WebJobs.Host/Utility.cs
```diff
@@ -50,5 +50,16 @@ internal static string GetFunctionShortName(MethodInfo methodInfo)
 
             return attribute?.ConfigurationSection ?? GetExtensionAliasFromTypeName(type.Name);
         }
+
+        internal static int GetProcessorCount()
```
mathewc: Shouldn't need this copy - the tests have internals visible to the Storage extension. If they don't, you can add it https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host.Storage/Properties/AssemblyInfo.cs. You want the tests to be calling the same helper that the product code is

brettsam:
@soninaren -- I'm reviewing the changelist for releasing 3.0.11 and this one confuses me a little. Why are we only adjusting the `NewBatchThreshold` and not the `BatchSize`?

max-mulawa:
Hi @brettsam, this change looks weird, some hardcoded variables from App Services hosting and change of NewBatchThreshold that breaks parallel processing limits introduced by former implementation of NewBatchThreshold.

Correlating NewBatchThreshold with processor count causes side effects, e.g. default BatchSize is still 16, which mean that NewBatchThreshold used to be set to 8. 

Now, if code executes on server with multiple processors (e.g. 9 CPUs), then the component will keep peeking and processing messages from the queue until parallel processed messages will reach (16/2)*9 = 72 parallel tasks.

PR #2356, opened 2019-10-22
Storage upgrade from 9.x to 11.x

paulbatum:
Not ready for review. I just found out that there is another new set of packages that are v12. Sending this PR to get an idea of whether there are any test failures I missed.


cveld:
Any clue why the central docs documentation is not mentioning this whole version gala? v9 / v10 / v11 and since last month there is v12 of the Storage APIs. Also two times Microsoft switched from package name:
- until 9.3 it was: WindowsAzure.Storage;
- v9.4 and later, v10 and v11 were named: Microsoft.Azure.Storage.Queue; although this splitting into separate packages is documented with the 9.3 package nuget page: https://www.nuget.org/packages/WindowsAzure.Storage/;
- v12 is named: Azure.Storage.Queues. The rationale for v12 is documented in de changelog: https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/storage/Azure.Storage.Queues/Changelog.txt

paulbatum:
@cveld Sorry, I can't really comment on that. What I will say is that this batch of changes is oriented around 9.x -> 11.x because this is still a low impact update i.e. the namespaces changed, but most of the API surface area stayed the same. This will allow customers to get reliability improvements present in 11.x with minimal impact to their webjobs/function code. The move to 12.x is more involved because of how significantly the API surface changed, which is why its not covered in this batch of changes.

src/Microsoft.Azure.WebJobs.Extensions.Storage/StorageAccount.cs
```diff
@@ -57,13 +58,16 @@ public virtual CloudBlobClient CreateCloudBlobClient()
...
         }
 
-        public virtual CloudTableClient CreateCloudTableClient()
+        public virtual Cosmos.Table.CloudTableClient CreateCloudTableClient()
         {
-            return SdkObject.CreateCloudTableClient();
+            return Microsoft.Azure.Cosmos.Table.CloudStorageAccountExtensions.CreateCloudTableClient(TableSdkObject);
+            //TableMicrosoft.Azure.Cosmos.Table.CloudStorageAccount.Parse
+            //return new Cosmos.Table.CloudTableClient( SdkObject.TableEndpoint Cosmos.Table.StorageCredentials;
+            //return SdkObject.CreateCloudTableClient();
```
brettsam: Can you remove these comments?

src/Microsoft.Azure.WebJobs.Extensions.Storage/StorageExtensions.cs
```diff
@@ -78,10 +78,10 @@ public static TableOperation Retrieve<TElement>(string partitionKey, string rowK
             return sdkOperation;
         }
 
-        public static Task<IList<TableResult>> ExecuteBatchAsync(this CloudTable sdk, TableBatchOperation batch,
+        public static async Task<IList<TableResult>> ExecuteBatchAsync(this CloudTable sdk, TableBatchOperation batch,
     CancellationToken cancellationToken)
         {
-            return sdk.ExecuteBatchAsync(batch, requestOptions: null, operationContext: null, cancellationToken: cancellationToken);
+            return await sdk.ExecuteBatchAsync(batch, cancellationToken);
```
brettsam: Can you return without awaiting here (and remove the `async`). That's some feedback we've gotten from the .NET folks -- when possible, return without awaiting and you'll have fewer allocations.

src/Microsoft.Azure.WebJobs.Extensions.Storage/Tables/TableEntityWriter.cs
```diff
@@ -8,8 +8,8 @@
 using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host.Bindings;
 using Microsoft.Azure.WebJobs.Host.Protocols;
-using Microsoft.WindowsAzure.Storage;
-using Microsoft.WindowsAzure.Storage.Table;
+//using Microsoft.Azure.Storage;
```
brettsam, line 11: remove comment?

src/Microsoft.Azure.WebJobs.Extensions.Storage/Tables/TableExtensions.cs
```diff
@@ -0,0 +1,69 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.Azure.Cosmos.Table;
+
+namespace Microsoft.Azure.WebJobs.Host.Tables
+{
+    internal static class TableExtensions
```
brettsam, line 11: Is this basically a copy of the other extensions, but with different namespaces for the types?

src/Microsoft.Azure.WebJobs.Host.Storage/Utility.cs
```diff
@@ -1,9 +1,12 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+
```
brettsam: nit -- i think there's an extra newline here?

src/Microsoft.Azure.WebJobs.Host.Storage/WebJobs.Host.Storage.csproj
```diff
@@ -6,6 +6,7 @@
   <Import Project="..\..\build\common.props" />  
   
   <PropertyGroup>
+    <Version>4.0.0$(VersionSuffix)</Version>
```
brettsam, line 9: Can you elevate this up to common.props? Make a `HostStorageVersion` property and then use it here? It makes it easier whenever we need to bump versions that they're all in the same file.

src/Microsoft.Azure.WebJobs.Logging/WebJobs.Logging.csproj
```diff
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\..\build\common.props" />
   <PropertyGroup>
+    <Version>4.0.0$(VersionSuffix)</Version>
```
brettsam, line 4: Same here -- elevate up to common.props with some new name.

test/FakeStorage/Queue/FakeQueue.cs
```diff
@@ -162,7 +162,12 @@ public override Task DeleteMessageAsync(CloudQueueMessage message)
         public override Task DeleteMessageAsync(CloudQueueMessage message, QueueRequestOptions options, OperationContext operationContext)
         {
             // throw new NotImplementedException();
-            // return base.DeleteMessageAsync(message, options, operationContext);
+            return base.DeleteMessageAsync(message, options, operationContext);
+            
+        }
```
brettsam, line 166: Can you remove the comment above (I know it was there before)? and I think there's an extra newline here now.

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/TestDoubles/StorageExceptionFactory.cs
```diff
@@ -40,7 +40,8 @@ private static StorageException Create(int httpStatusCode, XElement extendedErro
 
             using (XmlReader reader = requestResultElement.CreateReader())
             {
-                result.ReadXml(reader);
+                // Paul - is this wait OK?
+                result.ReadXmlAsync(reader).Wait();
```
brettsam: I think this is fine for tests.

test/Microsoft.Azure.Webjobs.Extensions.Storage.UnitTests/Mocks/StorageExceptionFactory.cs
```diff
@@ -40,7 +40,8 @@ private static StorageException Create(int httpStatusCode, XElement extendedErro
 
             using (XmlReader reader = requestResultElement.CreateReader())
             {
-                result.ReadXml(reader);
+                // Paul - is this wait ok?
+                result.ReadXmlAsync(reader).Wait();
```
brettsam: Yeah, I think it's okay here.

src/Microsoft.Azure.WebJobs.Logging/Entities/ContainerActiveEntity.cs
```diff
@@ -78,7 +78,7 @@ public static async Task<ContainerActiveEntity> LookupAsync(CloudTable table, lo
                 TableResult retrievedResult = await table.ExecuteAsync(retrieveOperation);
                 result = (ContainerActiveEntity)retrievedResult.Result;                
             }
-            catch (StorageException e)
+            catch (Cosmos.Table.StorageException e)
```
brettsam: This seems like a subtle change. We had a test that caught this, I assume?

build/common.props
```diff
@@ -1,9 +1,8 @@
 <Project>
   <PropertyGroup>
     <!-- Extensions can have independent versions and only increment when released -->
-    <Version>3.0.15$(VersionSuffix)</Version>
-    <ExtensionsStorageVersion>3.0.10$(VersionSuffix)</ExtensionsStorageVersion>
-
+    <Version>3.0.14$(VersionSuffix)</Version>
```
brettsam: looks like this should be 3.0.15 now...

brettsam (approved):
I've got a few minor comments -- unless there's something specific that you want looked at more deeply, I think this is good-to-go when those are addressed.

src/Microsoft.Azure.WebJobs.Extensions.Storage/Tables/TableExtensions.cs
paulbatum: I moved the two table storage related extension methods into their own file so that I could reference the tables namespaces without confusion.

src/Microsoft.Azure.WebJobs.Logging/Entities/ContainerActiveEntity.cs
paulbatum: Yeah, this is probably the most annoying thing about the lack of unification between tables and the rest of storage. I know I had a few test failures that helped me find these, and I did a review of places we were catching StorageException to check for similar cases that might need fixing.

src/Microsoft.Azure.WebJobs.Host.Storage/WebJobs.Host.Storage.csproj
brettsam: I'll make a follow-up change to adjust the versions to the main file here.

juliusl:
> * v12 is named: Azure.Storage.Queues. The rationale for v12 is documented in de changelog: https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/storage/Azure.Storage.Queues/Changelog.txt

Non-broken link:
https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/storage/Azure.Storage.Queues/CHANGELOG.md

PR #2364, opened 2019-11-01
Updated package version and logging string nit

madelinegordon:
Updated release version so that subsequent prerelease versions are considered higher than the current version by build.


brettsam (approved):
Looks good. Just be sure to squash before merging.

PR #2375, opened 2019-11-20
Bump ApplicationInsights SDKs to latest stable patch

### LINKED ISSUE #2302: Correlation stopped working, dependency telemetry is no longer child to request (and where did Legacy Operation Id come from?)
I get this strange behavior where the correlation of a WebJob-execution (`RequestTelemetry`) do not align correctly with the `DependencyTelemetry`. Looking at the properties it looks like `Legacy Operation Id` and `Legacy Id` makes more sense than `Operation Id` and `Request/Dependency Id`. However nowhere the `DependencyTelemetry` seem to have the `RequestTelemetry` as parent. I log the dependency telemetry manually with `TelemetryClient` (to be able to log the request and response body, I use a processor to get rid of the auto-logged since I do want to log other dependencies, if there would be any). There is just one registered `TelemetryClient` and one registered `TelemetryConfiguration` so that looks all fine to me. Not sure where to even start looking to solve this, any suggestions?

image

image

image

The expected behavior is that the top most telemetry is parent to all of the telemetry below. The dependency calls are made one after the other, they are not nested as it looks in the timeline. Pay especially attention to the first dependency telemetry, in the timeline that one is actually logged as the last call but it's shown as the first anyway.

lmolkova:
AppInsights latest patch includes fixes for SQL calls tracing
- https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/1291
- https://github.com/Microsoft/ApplicationInsights-dotnet-server/issues/1277

Fixes #2302

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/WebJobs.Logging.ApplicationInsights.csproj
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/WebJobs.Logging.ApplicationInsights.csproj
@@ -24,10 +24,10 @@
   </ItemGroup>
   
   <ItemGroup>
-  <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.8.0" />
-  <PackageReference Include="Microsoft.ApplicationInsights.PerfCounterCollector" Version="2.11.0" />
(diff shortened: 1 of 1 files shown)


lmolkova:
@brettsam please review

brettsam (approved):

PR #2376, opened 2019-11-21
Exposing App Insights endpoint configs

### LINKED ISSUE #2263: Support for Application Insights in the Azure Government Cloud [v3.x]
AI with the government cloud requires a change to point to the government AI endpoint:
 
"ApplicationInsights": { "InstrumentationKey": "instrumentationkey", "TelemetryChannel": { "EndpointAddress": "https://dc.applicationinsights.us/v2/track" } }
https://docs.microsoft.com/en-us/azure/azure-government/documentation-government-services-monitoringandmanagement#application-insights
 

We tried the settings below but because the config doesn't allow changes to the TelemetryChannel+ the EndpointAddress this does not work. 

https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
```
{
  "version": "2.0",
   "logging": {
        "applicationInsights": {
            "TelemetryChannel": {
      "EndpointAddress": "https://dc.applicationinsights.us/v2/track"
    }

        }
    }
}
```

gzuber:
Resolves #2263 

Initially made the decision to not group the endpoints because they're each configs for different components of App Insights. Left them at the top "Application Insights" level because giving them each their own subcategory seemed cluttered.

This would also require changes to the host.json schema

### DIFF
--- a/test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/ApplicationInsightsEndToEndTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/ApplicationInsightsEndToEndTests.cs
@@ -98,7 +98,7 @@ namespace Microsoft.Azure.WebJobs.Host.EndToEndTests
                 })
                 .ConfigureServices(services =>
                 {
-                    ServiceDescriptor quickPulse = services.Single(s => s.ImplementationType == typeof(QuickPulseTelemetryModule));
+                    ServiceDescriptor quickPulse = services.Single(s => s.ServiceType.Name == nameof(QuickPulseTelemetryModule));
                     services.Remove(quickPulse);
                     services.AddSingleton<ITelemetryModule, QuickPulseTelemetryModule>(s => new QuickPulseTelemetryModule()
                     {
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
-using System.Net;
 using Microsoft.ApplicationInsights.SnapshotCollector;
 using Microsoft.ApplicationInsights.WindowsServer.Channel.Implementation;
 using Microsoft.Azure.WebJobs.Hosting;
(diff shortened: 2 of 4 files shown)


brettsam:
@cijothomas can you take a look at this as well? I know that App Insights is introducing (or has introduced?) connection strings with an endpoint suffix (https://github.com/microsoft/ApplicationInsights-dotnet/issues/1221). Should we simplify this approach in functions and just expose the single suffix, or is it better to have the three separate properties like we have here? 

Eventually we should just support the connection string itself, but in V1 that likely won't happen and we're looking to unblock gov cloud connections ASAP.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/ApplicationInsights/ApplicationInsightsEndToEndTests.cs
```diff
@@ -98,7 +98,7 @@ private IHost ConfigureHost(LogLevel minLevel = LogLevel.Information, HttpAutoCo
                 })
                 .ConfigureServices(services =>
                 {
-                    ServiceDescriptor quickPulse = services.Single(s => s.ImplementationType == typeof(QuickPulseTelemetryModule));
+                    ServiceDescriptor quickPulse = services.Single(s => s.ServiceType.Name == nameof(QuickPulseTelemetryModule));
```
gzuber, line 101: I couldn't find a way to add the 3 AppInsights singletons with an ImplementationType and I don't think I fully understand the ramifications of adding them without one. This is the only thing I've found in the code that was counting on these modules having an ImplementationType, but I'm not sure if there are others.

cijothomas:
@brettsam Can you sync with @TimothyMothra to see if using connectionstrings here make more sense than all the individual endpoints? 
We have beta sdks with con strings available now, expected to ship stable next week or two.
It'd be ideal if all application insights scenarios (including Functions) used connectionstrings so documentation can be streamlines as well.

TimothyMothra:
@brettsam, If you have any questions about Connection Strings, feel free to message me.
I've got a draft documentation about Connection Strings here: https://github.com/microsoft/ApplicationInsights-Home/blob/master/Samples/ConnectionStringDraft.md
We're still working on complete documentation for Azure Docs.

You may find it simpler to expose the Environment Variable `APPLICATIONINSIGHTS_CONNECTION_STRING` which maps to `TelemetryConfiguration.ConnectionString`: https://github.com/microsoft/ApplicationInsights-dotnet/blob/add45ceed35a817dc7202ec07d3df1672d1f610d/BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs#L271-L274

This is available now in our SDK 2.12-Beta4.
2.12.0 will ship before December 10th.

TimothyMothra:
@gzuber @brettsam 
Hello, I'm writing to remind you that we shipped version 2.12.0 of the Application Insights sdks this week.
https://www.nuget.org/packages/Microsoft.ApplicationInsights/2.12.0

I want to politely ask that you consider onboarding to this version for Connection String support.
I have sdk examples documented here: https://github.com/microsoft/ApplicationInsights-Home/blob/master/Samples/ConnectionStringDraft.md#net-sdk-example
Please ping me if you have any questions.

gzuber:
@TimothyMothra thanks for the reminder! We are planning to use connection string for this feature, we've just been discussing internally the best way to expose it. I'll update this PR soon 😃

gzuber:
I actually ended up making a new branch as the changes for ConnectionString were more divergent than I expected. New PR is #2385

PR #2385, opened 2019-12-13
Expose AppInsights connection string setting

### LINKED ISSUE #2263: Support for Application Insights in the Azure Government Cloud [v3.x]

gzuber:
Resolves #2263

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
@@ -18,6 +18,12 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
         public string InstrumentationKey { get; set; }
 
         /// <summary>
+        /// Gets or sets Application Insights connection string. If set, this will
+        /// take precedence over the InstrumentationKey and overwrite it.
+        /// </summary>
+        public string ConnectionString { get; set; }
+
+        /// <summary>
         /// Gets or sets sampling settings.
         /// </summary>
         public SamplingPercentageEstimatorSettings SamplingSettings { get; set; }
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
@@ -160,9 +160,18 @@ namespace Microsoft.Extensions.DependencyInjection
 
                 var activeConfig = TelemetryConfiguration.Active;
                 if (!string.IsNullOrEmpty(options.InstrumentationKey) &&
-                string.IsNullOrEmpty(activeConfig.InstrumentationKey))
+                    string.IsNullOrEmpty(activeConfig.InstrumentationKey))
                 {
                     activeConfig.InstrumentationKey = options.InstrumentationKey;
+                }
+
+                // Set ConnectionString second because it takes precedence and
+                // we don't want InstrumentationKey to overwrite the value
+                // ConnectionString sets
+                if (!string.IsNullOrEmpty(options.ConnectionString) &&
+                    string.IsNullOrEmpty(activeConfig.ConnectionString))
+                {
+                    activeConfig.ConnectionString = options.ConnectionString;
                 }
 
                 if (!activeConfig.TelemetryInitializers.OfType<WebJobsRoleEnvironmentTelemetryInitializer>().Any())
@@ -240,7 +249,11 @@ namespace Microsoft.Extensions.DependencyInjection
             IRoleInstanceProvider roleInstanceProvider,
             QuickPulseInitializationScheduler delayer)
         {
-            if (options.InstrumentationKey != null)
+            if (options.ConnectionString != null)
+            {
+                configuration.ConnectionString = options.ConnectionString;
+            }
+            else if (options.InstrumentationKey != null)
             {
                 configuration.InstrumentationKey = options.InstrumentationKey;
             }
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/WebJobs.Logging.ApplicationInsights.csproj
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/WebJobs.Logging.ApplicationInsights.csproj
@@ -24,18 +24,18 @@
   </ItemGroup>
   
   <ItemGroup>
-  <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.8.2" />
-  <PackageReference Include="Microsoft.ApplicationInsights.PerfCounterCollector" Version="2.11.2" />
+  <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.12.0" />
+  <PackageReference Include="Microsoft.ApplicationInsights.PerfCounterCollector" Version="2.12.0" />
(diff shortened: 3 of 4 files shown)


TimothyMothra:
Looks good to me

TimothyMothra (approved):

shellayers (approved):

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
```diff
@@ -17,6 +17,11 @@ public class ApplicationInsightsLoggerOptions : IOptionsFormatter
         /// </summary>
         public string InstrumentationKey { get; set; }
 
+        /// <summary>
+        /// Gets or sets Application Insights connection string
```
brettsam: Add a note that this setting takes precedence over InstrumentationKey if they both are set.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
```diff
@@ -159,10 +159,16 @@ public static IServiceCollection AddApplicationInsights(this IServiceCollection
...
-                    activeConfig.InstrumentationKey = options.InstrumentationKey;
+                    if (!string.IsNullOrEmpty(options.ConnectionString))
+                    {
+                        activeConfig.ConnectionString = options.ConnectionString;
+                    }
+                    else if (!string.IsNullOrEmpty(options.InstrumentationKey))
+                    {
+                        activeConfig.InstrumentationKey = options.InstrumentationKey;
+                    }
                 }
```
brettsam, line 175: The logic here seems off. I think we want to set active ConnectionString (if not already set AND it's set in options). If that doesn't happen, then do the same thing with InstrumentationKey.

Basically, we need to get the active.ConnectionString or InstrumentationKey to match what our options say.

brettsam (commented):
Couple of comments...

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
gzuber: Ah, good point. I think my reasoning was I wanted to make sure InstrumentationKey doesn't overwrite the value that ConnectionString might set if both are present, but as long as ConnectionString is set last it should be fine.

gzuber:
@TimothyMothra I'm noticing that when I test setting the ConnectionString **without** setting `APPINSIGHTS_INSTRUMENTATIONKEY`, I'm seeing this in the traces during start up: 

```
AI: TelemetryChannel found a telemetry item without an InstrumentationKey. This is a required field and must be set in either your config file or at application startup.
```

I'm still getting traces, is this something I should worry about? Or is this a sign that we're not setting the ConnectionString everywhere we should?

TimothyMothra:
**Regarding the trace message**
That message is emitted from the TelemetryChannel itself.
- If you're getting flooded with those messages, then yes there is an issue with your configuration.
- If you're getting one or two messages and then it goes away, then telemetry is being sent before the configuration is fully set.

**Regarding the environment variable `APPINSIGHTS_INSTRUMENTATIONKEY`**
The reason you don't see this message when the environment variable is in use is because the Ikey then gets set during SDK Initialization and before the TelemetryChannel is operating.

----------

Taking a closer look at your implementation and I found some discrepancies that I want to call attention to that I missed in my original review.
I will comment on those lines specifically.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/WebJobs.Logging.ApplicationInsights.csproj
```diff
@@ -25,17 +25,17 @@
   
   <ItemGroup>
   <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.8.2" />
```
TimothyMothra: I would recommend upgrading AspNetCore to v2.12.0 to remain in sync with the rest of the SDK.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
```diff
@@ -160,11 +160,20 @@ public static IServiceCollection AddApplicationInsights(this IServiceCollection
 
                 var activeConfig = TelemetryConfiguration.Active;
```
TimothyMothra, line 161: I see two local instances of the TelemetryConfiguration here;
 - line 150 `TelemetryConfiguration config = TelemetryConfiguration.CreateDefault();`
- line 161 `var activeConfig = TelemetryConfiguration.Active;` 

Note that this call to TelemetryConfiguration.Active is initializing the entire SDK. [see here for reference](https://github.com/microsoft/ApplicationInsights-dotnet/blob/7633ae849edc826a8547745b6bf9f3174715d4bd/BASE/src/Microsoft.ApplicationInsights/Extensibility/TelemetryConfiguration.cs#L108-L125). **I expect this is where your extra traces we discussed below are coming from**

----------
I see that the InstrumentationKey/ConnectionString are set on `activeConfig` and then set again on the `config` variable when it's sent to your `SetupTelemetryConfiguration()` method.
I see from the comments that this is intentional, but this is unintuitive **to me**. 
I would have expected you call to `services.AddSingleton<TelemetryConfiguration>(` to only initialize a TelemetryConfiguration object and not initialize our entire SDK.

I'm not recommending dramatic changes, because obviously this is working for you in production and this is likely out of scope for this PR.
But i'm calling out that this looks suspicious to me.

brettsam: @lmolkova can weigh in on some of the subtleties there -- I know we had some problems that she addressed with setting things both on our Config and the Active config.

lmolkova: @TimothyMothra historically we used Active config everywhere and users got used to things like `new TelemetryClient()` and `TC.Active`. It created a lot of problems and when we can (where dependency injection is available) we create another instance of the configuration. However we do not want to break users who used Active.

So we apply very minimum configuration to Active, just enough for it to allow sending custom telemetry,

With this context, connection strings should be applied to Active AND to the instance of config created by Functions. If we don't do this - we'll end up in the situation, when code works differently in public and non-public clouds (i,e, you can send custom telemetry using Active config in public cloud, but cannot in china-one).

brettsam (approved):

PR #2397, opened 2019-12-31
Adding storage account validation and fixing the error message

### LINKED ISSUE #2396: Provide an error message when AzureWebJobsStorage app setting is invalid/missing
If you use the CLI to create a Python app and don’t include a valid connection string it gives you an unhelpful error (below).  Ideally it should say “No connection string provided for setting ‘AzureWebJobsStorage.’  Please provide a setting in the App Settings configuration or local.settings.json if running locally.”

[7/31/19 5:02:04 AM] The following 1 functions are in error:
[7/31/19 5:02:04 AM] BlobTrigger: Microsoft.Azure.WebJobs.Host: Error indexing method 'Functions.BlobTrigger'. Microsoft.WindowsAzure.Storage: Settings must be of the form "name=value".

soninaren:
closes #2396

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.Storage/StorageAccountProvider.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.Storage/StorageAccountProvider.cs
@@ -4,6 +4,7 @@
 using System;
 using Microsoft.Azure.WebJobs.Host;
 using Microsoft.Extensions.Configuration;
+using Microsoft.WindowsAzure.Storage;
 
 namespace Microsoft.Azure.WebJobs
 {
@@ -40,10 +41,15 @@ namespace Microsoft.Azure.WebJobs
             if (connectionString == null)
             {
                 // Not found
-                throw new InvalidOperationException($"Storage account connection string '{name}' does not exist. Make sure that it is a defined App Setting.");
+                throw new InvalidOperationException($"Storage account connection string '{IConfigurationExtensions.GetPrefixedConnectionStringName(name)}' does not exist. Make sure that it is a defined App Setting.");
             }
 
-            return StorageAccount.NewFromConnectionString(connectionString);
+            if (!CloudStorageAccount.TryParse(connectionString, out CloudStorageAccount account))
+            {   
+                throw new InvalidOperationException($"Storage account connection string for '{IConfigurationExtensions.GetPrefixedConnectionStringName(name)}' is invalid");
+            }
+
+            return StorageAccount.New(account);
         }
 
         /// <summary>


src/Microsoft.Azure.WebJobs.Extensions.Storage/StorageAccountProvider.cs
```diff
@@ -40,10 +41,15 @@ public virtual StorageAccount Get(string name)
             if (connectionString == null)
             {
                 // Not found
-                throw new InvalidOperationException($"Storage account connection string '{name}' does not exist. Make sure that it is a defined App Setting.");
+                throw new InvalidOperationException($"Storage account connection string '{IConfigurationExtensions.GetPrefixedConnectionStringName(name)}' does not exist. Make sure that it is a defined App Setting.");
             }
 
-            return StorageAccount.NewFromConnectionString(connectionString);
```
brettsam, line 46: Where else is `NewFromConnectionString()` used? Should this logic be moved there so all callers can benefit?

brettsam (commented):
Question about where we should have this logic...

src/Microsoft.Azure.WebJobs.Extensions.Storage/StorageAccountProvider.cs
soninaren: The only other place i see it being used is in tests
- https://github.com/Azure/azure-webjobs-sdk/blob/dev/test/Microsoft.Azure.Webjobs.Extensions.Storage.UnitTests/HostCallTests.cs#L963
- https://github.com/Azure/azure-webjobs-sdk/blob/dev/test/Microsoft.Azure.WebJobs.Host.EndToEndTests/MultipleStorageAccountsEndToEndTests.cs#L267

brettsam: Is it easy to move the logic into there? Just for any future usage as well.

soninaren: @brettsam do you mean move the validation logic into NewFromConnectionString method or into the tests. I think i can move the logic to NewFromConnectionString method. But for that would have to introduce a new  parameter since the method does not know about the source of the connections string, only the connection string value.

brettsam: Oh, to improve the error message, of course. And unfortunately NewFromConnectionString is public so we can't delete it, etiher. Oh well, yeah I guess this works :-)

brettsam (approved):
As long as you ran through the scenario in the issue and verified this works, let's do it!

PR #2398, opened 2020-01-02
Make FakeAzureStorage Packable

bb-naveenkumar:
I have been trying to do webjob trigger unit testing and using FakeAzureStorage seems like a good approach to do end to end test. However like test common NuGet there is no NuGet for FakeAzureStorage. Added the code here to create the NuGet for FakeAzureStorage project.


bb-naveenkumar:
I am not sure how to get over the appveyor build fail issue, @alrod, @fabiocav  can you please help me with that. Thanks!

fabiocav:
@bb-naveenkumar how are you trying to leverage this? Outside of the WebJobs code?

bb-naveenkumar:
@fabiocav we will be using same framework to perform end to end test like it is done in the webjob sdk. The difference lies where the FakeAzureStorage is used. That is not readily available to use unless until one write the same code. Using the similar approach as in sdk lets us properly mock the triggers and associated service which can be injected via unit test startup. Also, this lets us do unit testing for the triggers without the use of emulator.

bb-naveenkumar:
Hye @fabiocav, let me know if you have any  suggestions here.

PR #2402, opened 2020-01-10
[Do not merge] Event Grid blob triggering using blobtrigger shared queue

alrod:


fabiocav:
Closing stale PR that was in draft

PR #2405, opened 2020-01-15
Enable Startup class ctors to receive IConfiguration

idg10:
In ASP.NET Core, startup classes can define a constructor that receives an `IConfiguration` argument, providing access to configuration settings before services are configured. This is useful because it enables the service configuration process to be driven by configuration settings.

The WebJobs startup class feature does not support this. And, as discussed in https://github.com/Azure/azure-functions-host/issues/4577 this means systems built on top of WebJobs such as the Azure Functions SDK for .NET, do not provide any documented way to get hold of the `IConfiguration` until after service configuration has finished.

This PR enables any startup class to define a constructor taking an `IConfiguration`, or an `IHostingEnvironment` (or both). This PR is loosely modelled on the way that ASP.NET Core does it: it relies on the builder having access to the `HostBuilderContext` (which provide the `IConfiguration`), along with a mechanism by which a builder can declare that it wants to participate in the startup class process.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Hosting/ISupportsStartupInstantiation.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Hosting/ISupportsStartupInstantiation.cs
@@ -1,0 +1,13 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Azure.WebJobs.Hosting;
+
+namespace Microsoft.Azure.WebJobs.Host.Hosting
+{
+    interface ISupportsStartupInstantiation
+    {
+        IWebJobsStartup CreateStartupInstance(Type startupType);
+    }
+}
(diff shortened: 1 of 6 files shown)


BrianVallelunga:
This would be very helpful to me too, though I think it would be useful to add the injection of ILoggerFactory as well. This needs to be exposed to add logging providers at configuration time.

espray:
This looks helpful to get at the `IConfiguration`. Could a configuration delegate `Action<WebHostBuilderContext,IConfigurationBuilder>` be exposed as well? This would allow registering Azure App Configuration.

https://docs.microsoft.com/en-us/azure/azure-app-configuration/quickstart-aspnet-core-app?tabs=core3x

https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.hosting.webhostbuilder.configureappconfiguration?view=aspnetcore-3.1

brettsam:
Hey @idg10 -- so sorry for the delay on this. I'm starting to take a look now...

brettsam:
Okay -- @fabiocav and I talked about this, and with the introduction of the new `IWebJobsConfigurationStartup` interface (#2407)  (which will allow you to use `IConfigurationBuilder`), doing this injection into the constructor didn't make as much sense as the timing would be a bit off. Instead, I've got a PR out that will flow a new `WebJobsBuilderContext` through both of the Startup interfaces that we support. (#2490) 

From a `FunctionsStartup` standpoint, this will allow you to get the `IConfiguration` as well as some other valuable properties (you can see them on the context).

Thank you for pushing this scenario -- hopefully this will solve the issues once the linked PR is released.

fabiocav:
Closing this as the capability has been added in the PR mentioned by @brettsam . Thank you again, @idg10

PR #2407, opened 2020-01-15
adding support for IWebJobsConfigurationStartup

brettsam:


src/Microsoft.Azure.WebJobs.Host/Hosting/DefaultStartupTypeLocator.cs
```diff
@@ -15,22 +15,30 @@ namespace Microsoft.Azure.WebJobs.Hosting
...
         {
             _entryAssembly = entryAssembly;
         }
 
         public Type[] GetStartupTypes()
         {
-            Assembly assembly = _entryAssembly ?? Assembly.GetEntryAssembly();
-            IEnumerable<WebJobsStartupAttribute> startupAttributes = assembly.GetCustomAttributes<WebJobsStartupAttribute>();
+            return _startupTypes.Value
+                .Where(t => typeof(IWebJobsConfigurationStartup).IsAssignableFrom(t) || typeof(IWebJobsConfigurationStartup).IsAssignableFrom(t))
```
fabiocav: This is using the same interface...

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsBuilderExtensions.cs
```diff
@@ -164,7 +168,8 @@ public static IWebJobsBuilder UseExternalStartup(this IWebJobsBuilder builder, I
         /// <returns>The updated <see cref="IHostBuilder"/> instance.</returns>
         public static IWebJobsBuilder UseExternalStartup(this IWebJobsBuilder builder, IWebJobsStartupTypeLocator startupTypeLocator, ILoggerFactory loggerFactory)
         {
-            Type[] types = startupTypeLocator.GetStartupTypes();
+            IEnumerable<Type> types = startupTypeLocator.GetStartupTypes()
+                .Where(t => typeof(IWebJobsConfigurationStartup).IsAssignableFrom(t));
```
fabiocav: I'm trying to remember the previous discussion, I recall talking about consolidating this, but this is filtering on the new interface only. Is this expected?

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsConfigurationBuilder.cs
```diff
@@ -0,0 +1,24 @@
...
+using Microsoft.Azure.WebJobs.Host.Hosting;
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.Azure.WebJobs.Hosting
+{
+    internal class WebJobsConfigurationBuilder : IWebJobsConfigurationBuilder
+    {
+        public WebJobsConfigurationBuilder(IConfigurationBuilder builder)
+        {
+            if (builder == null)
```
fabiocav, line 14: nit: throw expression

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsStartupAttribute.cs
```diff
@@ -6,7 +6,7 @@
 namespace Microsoft.Azure.WebJobs.Hosting
 {
     /// <summary>
-    /// Attribute used to declare <see cref="IWebJobsStartup"/> Types that should be registered and invoked
+    /// Attribute used to declare <see cref="IWebJobsConfigurationStartup"/> Types that should be registered and invoked
```
fabiocav: `IWebJobsStartup` will continue to be supported, correct?

src/Microsoft.Azure.WebJobs.Host/Hosting/DefaultStartupTypeLocator.cs
brettsam: Duh, that was a refactoring mistake after I got rid of the other locator.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsConfigurationBuilder.cs
brettsam: Is there a good way to do that in these cases where you're not using a class variable? I could do something like:

```
ConfigurationBuilder = new TrackedConfigurationBuilder(builder ?? throw new ArgumentNullException(nameof(builder)));
```

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsStartupAttribute.cs
brettsam: Yeah, the attribute can use either `IWebJobsStartup` or `IWebJobsConfigurationStartup`. I just messed up the comments.

src/Microsoft.Azure.WebJobs.Host/Hosting/DefaultStartupTypeLocator.cs
brettsam: In fact, I'm going to remove this filtering entirely... callers should filter for the interface they need.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsBuilderExtensions.cs
brettsam: No, this is wrong. I've got it fixed -- this is meant to use `IWebJobsStartup` while UseExternalConfigurationStartup uses `IWebJobsConfigurationStartup`. 

I also figured out why the tests passed -- I had added `IWebJobsConfigurationStartup` to the existing test implementation, so this all worked (since it implemented both interfaces). I've split it out now to have one that only does `IWebJobsStartup` and the tests failed until I fixed it.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsConfigurationBuilder.cs
fabiocav: I see.... just leave it as is here.

fabiocav (approved):

PR #2411, opened 2020-01-21
Extending Hooks with custom "Poco -> object" conversion. 

alrod:
Extending Hooks with custom "Poco -> object" conversion. It's needed for a message serialized in non-json format. For example:
https://docs.microsoft.com/en-us/dotnet/api/microsoft.servicebus.messaging.brokeredmessage.getbody?view=azure-dotnet

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Triggers/TriggerArgumentBinding/ICustomPocoConvertor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Triggers/TriggerArgumentBinding/ICustomPocoConvertor.cs
@@ -1,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs.Host.Triggers
+{
+
+    /// <summary>
+    /// Adds custom conversion for PocoTriggerArgumentBindings which does not use json format.
+    /// </summary>
+    /// <typeparam name="TMessage">The native message type. For ServiceBus, this would be Message.</typeparam>
+    public interface ICustomPocoConvertor<TMessage>
+    {
+        /// <summary>
+        /// Given message and type to object.
+        /// This is used if we want convertor message to POCO using other than json deserialization
+        /// </summary>
+        /// <param name="message"></param>
+        /// <param name="type"></param>
+        /// <returns></returns>
+        object Convert(TMessage message, Type type);
+    }
+}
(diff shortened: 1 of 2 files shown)


alrod:
Fixes: https://github.com/Azure/azure-functions-servicebus-extension/issues/44

src/Microsoft.Azure.WebJobs.Host/Triggers/TriggerArgumentBinding/SimpleTriggerArgumentBinding.cs
```diff
@@ -82,5 +83,36 @@ public virtual async Task<ITriggerData> BindAsync(TTriggerValue value, ValueBind
 
             return triggerData;
         }
+
+        protected void InitializeBindingDataProvider()
```
mathewc: This type already has BindingDataContract/Hooks members that are used for contracts. You're adding another IBindingDataProvider + methods on the side, which will confuse things. Which should I use in a derived class?

What's the diference between CustomTriggerArgumentBinding and PocoTriggerArgumentBinding?

src/Microsoft.Azure.WebJobs.Host/Triggers/TriggerArgumentBinding/CustomTriggerArgumentBinding.cs
```diff
@@ -35,6 +37,7 @@ internal class CustomTriggerArgumentBinding<TMessage, TTriggerValue, TUserType>
             ValueBindingContext context)
         {
             var obj = await _converter(value, null, context);
+            AddBinbdingData(obj, bindingData);
```
mathewc: spelling errors here and below

src/Microsoft.Azure.WebJobs.Host/Triggers/TriggerArgumentBinding/SimpleTriggerArgumentBinding.cs
alrod: `CustomTriggerArgumentBinding` is used if there are predefined convertors (Message -> string, Message -> byte[], etc.)
`PocoTriggerArgumentBinding` is used for PocoTypes.

It looks like changing base class `SimpleTriggerArgumentBinding` is a bad idea. I will update the PR with Hooks extension.

alrod:
Another approach here:
https://github.com/Azure/azure-functions-servicebus-extension/pull/57

src/Microsoft.Azure.WebJobs.Host/Triggers/TriggerArgumentBinding/SimpleTriggerArgumentBinding.cs
mathewc: Your initial PR was close I think. Anyhow, I've sent https://github.com/Azure/azure-webjobs-sdk/pull/2423 with some additional refactoring/cleanup

PR #2414, opened 2020-01-21
[BlobTrigger] fixing bug with log scans

### LINKED ISSUE #2413: Blob triggered function causes unhandled exception
I have a blob triggered Azure function created using the VS2019 (v16.4.0) template.

When running the function crashes with unhandled exception: "System.Private.CoreLib: The given key 'LogType' was not present in the dictionary."

image

Azure Functions Core Tools (2.7.1948 Commit hash: 29a0626ded3ae99c4111f66763f27bb9fb564103)
Function Runtime Version: 2.0.12888.0

brettsam:
Fixes #2413 and https://github.com/microsoft/vscode-azurefunctions/issues/1725

I'm guessing that there's some kind of race here where storage doesn't add this Metadata immediately so there's a chance it's missing. I see a lot of errors throughout kusto. Best thing to do here is just skip 
this file and we'll catch it during the next pass.

I'm a little nervous that people are trying Storage V2 and there'll be subtle issues: https://docs.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-known-issues#issues-and-limitations-with-using-blob-apis.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/BlobLogListener.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/BlobLogListener.cs
@@ -175,7 +175,7 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
             foreach (var item in blobs)
             {
                 ICloudBlob log = item as ICloudBlob;
-                if (log != null)
+                if (log != null && log.Metadata.ContainsKey(LogType))
                 {
                     // we will exclude the file if the file does not have log entries in the interested time range.
                     string logType = log.Metadata[LogType];


pragnagopa (approved):

brettsam:
Actually hang on... it looks like in DataLake the metadata is `"logtype"` rather than `"LogType"`... could that be what's happening here?

brettsam:
No that's not it -- this fix makes sense. I was able to repro this -- the issue is with the "ListBlobs" known difference in the link above. Because we don't explicitly send a `delimiter`, the call to List includes the parent directory, which does not have this metadata. So checking this works.

haugan:
I'm having issues as well running a template Function App triggering on Blob Storage using a (general purpose) StorageV2 with HNS enabled.

> An unhandled exception has occurred. Host is shutting down... System.Private.CoreLib: The given key 'LogType' was not present in the dictionary.

I'm on .NET Core 3.1 and I've updated Microsoft.Azure.WebJobs.Extension.Storage to 3.0.10 and Microsoft.NET.Sdk.Functions to 3.0.5. How can I get this fix?

kiingdavies:
I currently have a customer experiencing the same issue. When should we be expecting a fix for this?

brettsam:
What version of the storage package are they using? The fix has been released for some time now -- but I believe the bundles release that includes the fix is still being deployed.

PR #2418, opened 2020-01-24
[Storage] changing WebJobs ref to depend on a public nuget package rather than project

brettsam:
Having the storage extension use a project ref means that we cannot release it until the corresponding WebJobs nuget is public, which can take weeks. We rarely (if ever) need to depend on something that has been added to the host, so I'm severing that dependency and instead relying on the public nuget version of WebJobs like our other extensions do. This will allow us to release this as-needed.


brettsam:
@paulbatum -- I'm adding you to some of these PRs just as FYIs. You don't really need to review them.

PR #2424, opened 2020-02-04
Fixing "GetAwaiter().GetResult()" in BlobsExtensionConfigProvider whi…

### LINKED ISSUE #5416: Blobinput performance assessment
Simple Azure Function is created that just uses an HTTP Trigger with a blob input binding to return like 100kb worth of JSON data.  I published to a premium plan EP2.  I then slammed it with a load test of 1000 requests per second.  The app almost instantly locks up – stops sending logs and everything. I don’t even see live metrics.  

Running the queries shows that the host InvokeCore WorkerFunctionInvoker did not even invoked from the Job host.

This is tested for C# and Java.

alrod:
…ch causes deadlock.

Fixes: https://github.com/Azure/azure-functions-host/issues/5416

Language worker creates a binding on each execution:
https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs#L123

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Config/BlobsExtensionConfigProvider.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Config/BlobsExtensionConfigProvider.cs
@@ -20,9 +20,9 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs.Bindings
 {
     [Extension("AzureStorageBlobs", "Blobs")]
     internal class BlobsExtensionConfigProvider : IExtensionConfigProvider,
-        IAsyncConverter<BlobAttribute, CloudBlobContainer>,
-        IAsyncConverter<BlobAttribute, CloudBlobDirectory>,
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Config/BlobsExtensionConfigProvider.cs
```diff
@@ -331,8 +327,9 @@ private async Task<ICloudBlob> ConvertFromInvokeString(DirectInvokeString input,
         }
         private ParameterDescriptor ToBlobDescr(BlobAttribute attr, ParameterInfo parameter, INameResolver nameResolver)
         {
-            // Resolve the connection string to get an account name. 
-            var client = Task.Run(() => GetClientAsync(attr, CancellationToken.None)).GetAwaiter().GetResult();
```
alrod, line 335: This line caused deadlock during high load.

mathewc (approved):

PR #2426, opened 2020-02-06
Adding support for LogEvent to log the EventTelemetry

amigup:
Added implementation of Log Event to add the telemetry as EventTelemetry for defined event.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/LoggerExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/LoggerExtensions.cs
@@ -33,6 +33,19 @@ namespace Microsoft.Extensions.Logging
             logger?.Log(LogLevel.Information, LogConstants.MetricEventId, payload, null, (s, e) => null);
         }
 
+        /// <summary>
+        /// Logs a event.
+        /// </summary>
+        /// <param name="logger">The ILogger.</param>
+        /// <param name="eventId">The event to log.</param>
+        /// <param name="properties">Named string values for classifying and filtering metrics.</param>
+        public static void LogEvent(this ILogger logger, EventId eventId, IDictionary<string, string> properties = null)
+        {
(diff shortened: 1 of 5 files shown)


amigup:
It seems that AppVeyor build is failing due to configuration issue. @brettsam / @lmolkova - Can you please fix the build pipeline?

amigup:
Can this PR be reviewed?

lmolkova (commented):
Can you please share your thoughts on how it is going to be used and why do you need it?

This can be achieved without changing anything on WebJobs side similar to https://docs.microsoft.com/en-us/azure/azure-functions/functions-monitoring#version-2x-and-later-2


Events do not seem to be ILogger concept and introducing them just to support ApplicationInsights events (which already can be done) does not seem like a common scenario.

amigup:
@lmolkova - The logger should provide the capability of logging event telemetry, the way it's provide to log the metric telemetry currently. 
Currently, I'm porting lot of code to functions which requires to log events so raised this PR to incorporate the `LogEvent` implementation .

I've seen the link mentioned by you (https://docs.microsoft.com/en-us/azure/azure-functions/functions-monitoring#version-2x-and-later-2), but I don't want to expose two different mechanisms to log trace, exception, metric and event (via `ILogger` and `TelmetryClient` in the code base. The abstracted `ILogger` should provide all the relevant capabilities of logging to Application Insights); I believe that's why the **`LogMetric`** implementation was added (LogMetric - https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/LoggerExtensions.cs#L25  and  Application Insights - https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs#L116). based on LogMetric, I've added the implementation of `LogEvent`.

azfuncgh:

PR #2428, opened 2020-02-12
Populate extra EventHub data: time in queue and hub info

lmolkova:
This change enables custom metrics on time in queue for EventHubs messages and populates EventHub info on trigger request telemetry (this will allow to show AppMap properly and navigate to EventHub resource from AI views.

This relies on https://github.com/Azure/azure-functions-eventhubs-extension/pull/52 changes in eventhub extensions to populate necessary information. Please share your feeback on the extension as well.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Constants/LogConstants.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Constants/LogConstants.cs
@@ -158,5 +158,25 @@ namespace Microsoft.Azure.WebJobs.Logging
         ///  Get the name of the key to store the current process id.
         /// </summary>
         public const string ProcessIdKey = "ProcessId";
+
+        /// <summary>
+        /// Get the name of the key to store the time when queue message
+        /// was enqueued in <see cref="System.Diagnostics.Activity"/> tags.
+        /// </summary>
+        public const string MessageEnqueuedTimeKey = "EnqueuedTime";
+
+        /// <summary>
+        /// Get the name of the key to store the endpoint in trigger details
+        /// that represents endpoint or FQDN of the service (Storage account, EventHub,
+        /// or any other Azure or non-Azure service).
+        /// </summary>
+        public const string TriggerDetailsEndpointKey = "Endpoint";
+
+        /// <summary>
+        /// Get the name of the key to store the entity or path in trigger details
+        /// that represents name of the entity (EventHub entity, Azure Blob container or
+        /// Storage queue name).
+        /// </summary>
+        public const string TriggerDetailsEntityNameKey = "Entity";
     }
 }
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
@@ -11,10 +11,8 @@ using Microsoft.ApplicationInsights;
 using Microsoft.ApplicationInsights.DataContracts;
 using Microsoft.ApplicationInsights.Extensibility;
 using Microsoft.AspNetCore.Http;
-using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
-using Newtonsoft.Json;
 
 namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
 {
@@ -45,7 +43,8 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
                 ScopeKeys.FunctionInvocationId,
                 ScopeKeys.FunctionName,
                 ScopeKeys.HostInstanceId,
-                OperationContext
+                OperationContext,
+                ScopeKeys.TriggerDetails
             };
 
         public ApplicationInsightsLogger(TelemetryClient client, string categoryName, ApplicationInsightsLoggerOptions loggerOptions)
@@ -383,6 +382,10 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
                             // this is set in the WebJobs initializer,
                             // we will ignore it here
                             break;
+                        case LogConstants.MessageEnqueuedTimeKey:
+                            // this is populated when creating telemetry
+                            // we will ignore it here
+                            break;
                         case LogConstants.DurationKey:
                             if (prop.Value is TimeSpan duration)
                             {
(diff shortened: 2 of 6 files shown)


lmolkova:
@brettsam can you please have a look?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -478,47 +482,64 @@ private void StartTelemetryIfFunctionInvocation(IDictionary<string, object> stat
...
+                        operation = _telemetryClient.StartOperation<RequestTelemetry>(functionName);
+                    }
+
+                    var triggerDetails = stateValues.GetValueOrDefault<IDictionary<string, string>>(ScopeKeys.TriggerDetails);
+                    if (triggerDetails != null &&
+                        triggerDetails.TryGetValue("EventHubEndpoint", out var endpoint) &&
+                        triggerDetails.TryGetValue("EventHubName", out var hubName))
+                    {
+                        operation.Telemetry.Source = endpoint.EndsWith("/") ? string.Concat(endpoint, hubName) : string.Concat(endpoint, "/", hubName);
+                    }
```
brettsam, line 524: I don't like the logger having special knowledge about EventHubs properties here. Can these be added to the Activity's tags in the EH extension and then they will just get pushed to the RequestTelemetry automatically?

brettsam: Oh... I see you want this to go with the `operation.Telemetry.Source` rather than with the custom properties. Could we have some "well-known" property in `LogConstants` like "Source" that we look up special for this, and have EH set that instead?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -33,6 +31,7 @@ internal class ApplicationInsightsLogger : ILogger
         internal const string MetricMinKey = "min";
         internal const string MetricMaxKey = "max";
         internal const string MetricStandardDeviationKey = "standarddeviation";
+        internal const string MessageEnqueuedTime = "EnqueuedTime";
```
brettsam: Should this be bumped up to `LogConstants` so other triggers can take advantage?

brettsam (commented):
Some initial questions...

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
lmolkova: Source is a special field for AppInsights UX, this is the place that is intended to have information about caller. 
Basically using `Source` is critical part for the UX. EventGrid or ServiceBus or Storage Queues should use `Source` in the same way as  EventHub does

Custom dimensions are a property bag that, in general, UX do not use - they are more expensive to access and group by.

lmolkova: we can possibly standardize trigger details used to populate info about queue (not specific to the EventHub)

- `endpoint` will contain URI of the EventHub/ServiceBus/storage account/etc
- `entity` will contain entity name within account (hub/bus/queue/topic/container name).

brettsam: I think I'd feel better about that... having EH-specific stuff in here for logging feels wrong. Ideally the extension would handle this, but I can't think of any way for the EH extension to log this itself unless it registered its own `ITelemetryProcessor`.

lmolkova: I've done the trigger details attributes not-specific to EH. Registering `ITelemetryProcessor` would mean taking dependency on AI packages and I'd prefer to avoid it.

lmolkova:
@brettsam can you please have a look again? I implemented your suggestions

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -512,13 +529,28 @@ private void StartTelemetryIfFunctionInvocation(IDictionary<string, object> stat
...
+            PopulateLinks(activities, request);
+
+            // If any of the links is sampled in (on upstream) we also preliminary sample in 
+            // request telemetry and everything that happens in this request scope. 
+            // There will be additional level of sampling applied to limit rate to one 
+            // configured on this Function/WebJob
+            if (request.ProactiveSamplingDecision == SamplingDecision.SampledIn)
             {
-                return;
+                Activity.Current.ActivityTraceFlags |= ActivityTraceFlags.Recorded;
```
brettsam, line 546: What does this do? I feel like i asked this before, sorry :-)

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Constants/LogConstants.cs
```diff
@@ -158,5 +158,25 @@ public static class LogConstants
         ///  Get the name of the key to store the current process id.
         /// </summary>
         public const string ProcessIdKey = "ProcessId";
+
+        /// <summary>
+        /// Get the name of the key to store the time when queue message
+        /// was enqueued in <see cref="System.Diagnostics.Activity"/> tags.
+        /// </summary>
+        public const string MessageEnqueuedTime = "EnqueuedTime";
```
brettsam: Can you make these new consts end in `Key`?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -478,31 +481,45 @@ private void StartTelemetryIfFunctionInvocation(IDictionary<string, object> stat
...
+                            operation.Telemetry.Name = functionName;
+                        }
+                        else
+                        {
+                            operation = CreateRequestFromLinks(activities, functionName);
+                        }
+
+                        if (this.TryGetAverageTimeInQueueForBatch(activities, operation.Telemetry.Timestamp, out long enqueuedTime))
+                        {
+                            operation.Telemetry.Metrics["timeSinceEnqueued"] = enqueuedTime;
```
brettsam, line 498: is there anything special about this key? Just noticed you're using `EnqueuedTime` for the key above and `timeSinceEnqueued` here...

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Constants/LogConstants.cs
```diff
@@ -158,5 +158,25 @@ public static class LogConstants
...
+        /// or any other Azure or non-Azure service).
+        /// </summary>
+        public const string TriggerEndpoint = "Endpoint";
+
+        /// <summary>
+        /// Get the name of the key to store the entity or path in trigger details
+        /// that represents name of the entity (EventHub entity, Azure Blob container or
+        /// Storage queue name).
+        /// </summary>
+        public const string TriggerEntityName = "Entity";
```
brettsam: Since these are meant to be for TriggerDetails -- maybe we call these `TriggerDetailsEndpointKey` and `TriggerDetailsEntityKey`?

brettsam (approved):
A few minor comments and questions but in general this is good.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
lmolkova: this is a magical string for Application Insights UX. It will be exposed as a metric with this name + shown on the AppMap for EventHubs. `timeSinceEnqueued` is relative time and can be a metric. while `EnqueuedTime` is absolute time.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
lmolkova: this solves problem of sampling for batched execution. Normally, when no batching is involved, we do best effort to make consistent sampling decisions based on the operation_Id, but when batching is involved, we generate a new one.

E.g. if there are 10 messages received and context in one of messages indicates that it was sampled in on the upstream service, we want to increase chances of  the batched request to be sampled in. 

So this flag tells AI SDK to try its best to sample in everything under this Activity, at the same time SDK does some post-processing that will keep telemetry rate at the configured threshold.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
brettsam: Got it -- I assumed it was something special but just wanted to double-check.

lmolkova:
@brettsam I believe I resolved all the comments, can you please merge it?

PR #2438, opened 2020-02-25
BlobTrigger uses Event Grid subscription

alrod:
Using:
1. EventGrid subscription with HTTP endpoint should be pre-created as we do not have a tool for the EventGrid subscription creation yet.
https://[function app uri]/runtime/webhooks/blobs?functionId=[function id in format namespace.class.method]&code=[blob_extension key]
Also you can specify a filter to get the blobs only from specified container/path.
2. Blob trigger definition:
`public static async Task Run([BlobTrigger("test/{name2}", UseEventGrid = true, Connection = "Storage1")]string myBlob, ILogger log)`

How it works:
Blob uploaded/created -> EventGrid sends a HTTP request to function app extension endpoint -> Function runtime receives the HTTP request and adds the message to internal queue -> internal queue listener processes new messages (poison queue supported)

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.Storage/StorageLoadBalancerQueue.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.Storage/StorageLoadBalancerQueue.cs
@@ -68,11 +68,7 @@ namespace WebJobs.Extensions.Storage
 
             public async Task AddAsync(T item, CancellationToken cancellationToken = default(CancellationToken))
             {
-                string contents = JsonConvert.SerializeObject(
-                    item,
-                    JsonSerialization.Settings);
-
-                var msg = new CloudQueueMessage(contents);
+                var msg = new CloudQueueMessage(item.ToString());
                 await _queue.AddMessageAndCreateIfNotExistsAsync(msg, cancellationToken);
 
                 _parent._sharedWatcher.Notify(_queue.Name);
@@ -87,7 +83,7 @@ namespace WebJobs.Extensions.Storage
         private CloudQueue Convert(string queueMoniker)
         {
             // $$$ Review
-            var account = _storageAccountProvider.Get(ConnectionStringNames.Dashboard);
+            var account = _storageAccountProvider.Get(ConnectionStringNames.Storage);
             var queue = account.CreateCloudQueueClient().GetQueueReference(queueMoniker);
             return queue;
         }
--- a/src/Microsoft.Azure.WebJobs.Extensions.Storage/WebJobs.Extensions.Storage.csproj
+++ b/src/Microsoft.Azure.WebJobs.Extensions.Storage/WebJobs.Extensions.Storage.csproj
@@ -96,12 +96,9 @@
     <PackageReference Include="Microsoft.Azure.Cosmos.Table" Version="1.0.5" />
     <PackageReference Include="Microsoft.Azure.Storage.Blob" Version="11.1.0" />
     <PackageReference Include="Microsoft.Azure.Storage.Queue" Version="11.1.0" />
+    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.18" />
     <PackageReference Include="StyleCop.Analyzers" Version="1.1.0-beta004">
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
-  </ItemGroup>
-
-  <ItemGroup>
-    <ProjectReference Include="..\Microsoft.Azure.WebJobs.Host\WebJobs.Host.csproj" />
   </ItemGroup>
 </Project>
--- a/src/Microsoft.Azure.WebJobs.Extensions.Storage/WebJobs.Extensions.Storage.csproj
+++ b/src/Microsoft.Azure.WebJobs.Extensions.Storage/WebJobs.Extensions.Storage.csproj
@@ -96,9 +96,12 @@
     <PackageReference Include="Microsoft.Azure.Cosmos.Table" Version="1.0.5" />
     <PackageReference Include="Microsoft.Azure.Storage.Blob" Version="11.1.0" />
     <PackageReference Include="Microsoft.Azure.Storage.Queue" Version="11.1.0" />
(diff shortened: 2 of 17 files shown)


src/Microsoft.Azure.WebJobs.Extensions.Storage/WebJobs.Extensions.Storage.csproj
```diff
@@ -96,9 +96,12 @@
...
     <PackageReference Include="Microsoft.Azure.Storage.Blob" Version="11.1.0" />
     <PackageReference Include="Microsoft.Azure.Storage.Queue" Version="11.1.0" />
-    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.14" />
     <PackageReference Include="StyleCop.Analyzers" Version="1.1.0-beta004">
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
   </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.Azure.WebJobs.Host\WebJobs.Host.csproj" />
```
fabiocav: Was this an intentional change, or something you were doing while developing? We should be able to continue depending on the package only here, right?

src/Microsoft.Azure.WebJobs.Extensions.Storage/StorageLoadBalancerQueue.cs
```diff
@@ -87,7 +83,7 @@ public Task FlushAsync(CancellationToken cancellationToken = default(Cancellatio
         private CloudQueue Convert(string queueMoniker)
         {
             // $$$ Review
-            var account = _storageAccountProvider.Get(ConnectionStringNames.Dashboard);
+            var account = _storageAccountProvider.Get(ConnectionStringNames.Storage);
```
fabiocav, line 86: Was this a bug?

src/Microsoft.Azure.WebJobs.Host/Dispatch/HttpRequestProcessor.cs
```diff
@@ -0,0 +1,121 @@
...
+    {
+        void RegisterHttpEnpoint(ExtensionConfigContext context);
+
+        Task<HttpResponseMessage> ProcessHttpRequestAsync(HttpRequestMessage req, CancellationToken cancellationToken);
+    }
+
+    /// <summary>
+    /// Hadles HttpRequests that comes to extension http endpoint.
+    /// </summary>
+    internal class HttpEndpointManager : IHttpEndpointManager
```
fabiocav, line 33: Is this a new class?  and interface?

src/Microsoft.Azure.WebJobs.Host/Dispatch/HttpRequestProcessor.cs
```diff
@@ -0,0 +1,121 @@
...
+                    // cloudevent schema
+                    events = new JArray
+                    {
+                        token
+                    };
+                }
+
+                foreach (JObject jo in events)
+                {
+                    // TODO ALROD - Azure Queues do not suport batch sending. Sending in parallel? ServiceBus?
```
fabiocav: It would indeed be good to send them in parallel.

What is the desired behavior if enqueuing one of the messages fails? Right now, we'd end up in a partial number of messages enqueued and an exception bubbling up, correct?

src/Microsoft.Azure.WebJobs.Host/Dispatch/IMessageHandler.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.Azure.WebJobs.Host.Dispatch
     /// Defines the contract for executing a triggered function using a dispatch queue.
     /// <see cref="ListenerFactoryContext.GetDispatchQueue(IMessageHandler)"/>
     /// </summary>
-    internal interface IMessageHandler
+    public interface IMessageHandler
```
fabiocav, line 17: Is the visibility of these interfaces changing for to allow the extension to consume them?

@mathewc can you please comment here as well?

fabiocav (commented):
Some small comments, mostly questions. But this is looking good. A couple of spots that would be good to get @mathewc to chime in as well.

src/Microsoft.Azure.WebJobs.Extensions.Storage/WebJobs.Extensions.Storage.csproj
alrod: Fixed.

src/Microsoft.Azure.WebJobs.Extensions.Storage/StorageLoadBalancerQueue.cs
alrod: Yes. Actually we did not use `StorageLoadBalancerQueue` class before but the class was checked in in the repository.

src/Microsoft.Azure.WebJobs.Host/Dispatch/HttpRequestProcessor.cs
alrod: Yes, we use `HttpEndpointManager` to register http enpoint for an extension and handle to http requests for the extension.

Interface is introduced for DI as implementation is used some internal stuff.

src/Microsoft.Azure.WebJobs.Host/Dispatch/IMessageHandler.cs
alrod: we did not use the interface before (there was only E2E test). `IMessageHandle`r must be public in order to use in an extension.

src/Microsoft.Azure.WebJobs.Host/Dispatch/HttpRequestProcessor.cs
alrod: Added sending in parallel.

>Right now, we'd end up in a partial number of messages enqueued and an exception bubbling up, correct?

Correct.

alrod:
the change was done in track2

PR #2444, opened 2020-02-28
Revert "Remove dashboard signing"

paulbatum:
This reverts commit 304707a656114faceaaa94991a71534d4b340cca.

It turns out that dealing with dashboard signing is more complicated than I thought. We'll need to either update the signing job, or perhaps change the dashboard to use the webjobs nugets instead of project references. Either way, this change is no good as is so I'm reverting.


mathewc (approved):

fabiocav (approved):

fabiocav:
@paulbatum still planning on merging this? Or should this be closed?

Issue #2445, opened 2020-02-28
Improve re-throwing exceptions by using ExceptionDispatchInfo throughout


macrogreg:
At many places the SDK uses the following pattern:

```
try
{
    // Do stuff
}
catch (Exception ex)
{
    if (SomeCondition())
    {
        // Handle ex
    }
    else
    {
        throw;
    }
}
```

The way the exception being re thrown is problematic. The resulting stack trace only shows the like on the respective method that contains the throw statement. A better way to re-thrown exceptions exists since NetFx version 4.5: `ExceptionDispatchInfo.Capture(ex).Throw()`. That approach will include richer stack trace information: bith,m the line where the exception originally occured, and where it was re-thrown. 

We should move to using the ExceptionDispatchInfo-based approach throughout.  Unless someone is parsing the exception stack trace, this is _not_ a breaking change.

macrogreg:
I have a PR for this locally, but I am having trouble pushing a private branch. Perhaps I am missing some permissions issue?

macrogreg:
PR ready for review. I am working thought learning how to deal with the tests. Seems to be an infra issue. :)

macrogreg:
@brettsam asked to provide a demo of how `ExceptionDispatchInfo.Capture(ex).Throw();` improves things compared to `throw;`. Here is an example:

I provide some source code below as an image, so that you can easily see the line numbers. Then I provide the output. It should make things clear.

![image](https://user-images.githubusercontent.com/20691812/75601194-b1d9d680-5a6d-11ea-8d80-6360c77525fe.png)

First we run this with line 54 uncommented and line 55 commented. **So we use throw.**
The output is:

```
Sum: 0.
Executing DoStuff1.
Value: 1, Sum: 0.
Executing DoStuff2.

Caught exception in DoAllTheStuff:
System.Exception: Throwing an exception from within DoStuff2
   at ExceptionDispatchInfoDemo.Program.DoStuff2() in C:\00\Code\Sandbox\ExceptionDispatchInfoDemo\Program.cs:line 17
   at ExceptionDispatchInfoDemo.Program.DoAllTheStuff() in C:\00\Code\Sandbox\ExceptionDispatchInfoDemo\Program.cs:line 40
Rethrowing..

Caught exception in Exec:
System.Exception: Throwing an exception from within DoStuff2
[...]

PR #2448, opened 2020-02-29
Improve re-throwing exceptions by using ExceptionDispatchInfo throughout (Issue #2445)

macrogreg:
Addresses https://github.com/Azure/azure-webjobs-sdk/issues/2445
Please see the description there.


brettsam:
For external PRs we need to manually add the `full-ci` label to get CI to run correctly... I've just added it.

brettsam:
@mathewc / @fabiocav -- this seems like general goodness, especially in some of our more complicated `catch` blocks. I think we do some fiddling with stack traces (in Functions?) to make errors look nicer -- any worries that this may affect that? I don't think there is but want to call it out.

src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Bindings/StorageBlobContainerExtensions.cs
```diff
@@ -48,7 +49,8 @@ internal static class StorageBlobContainerExtensions
 
                 if (result == null || result.HttpStatusCode != 404)
                 {
-                    throw;
+                    ExceptionDispatchInfo.Capture(exception).Throw();
```
mathewc, line 52: Here and in all the other instances below you're doing an explicit recapture/throw for exceptions that we're already using the "throw" keyword for. My understanding is that as long as we're not doing a rethrow of the form "throw ex", the stack trace is maintained already without a recapture.

It seems we're already doing this correctly in most places. E.g. in places in our execution pipeline where we catch an invocation exception and don't throw immediately, we're capturing and throwing the ExceptionDispatchInfo and using it to rethrow later, e.g. [here](https://github.com/Azure/azure-webjobs-sdk/blob/f7da27d133b9ce0aadf9918770f6ed45748f3c8a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs#L134).

macrogreg: Thanks for your response, @mathewc .  This statement is not quite correct: 

> as long as we're not doing a rethrow of the form "throw ex", the stack trace is maintained already without a recapture.

Please see my example in the issue https://github.com/Azure/azure-webjobs-sdk/issues/2445 .
The stack you get with `throw` shows the line where the exception came up in the `try`-block, bit not the line where it was re-thrown in the `catch`-block. The stack you get with `ExceptionDispatchInfo` shows both. Please see the above-linked post for an example.

azfuncgh:

PR #2458, opened 2020-03-19
Passing timeout token down to host

yojagad:
As part of the workitem here https://github.com/Azure/azure-functions-host/issues/5391 , one of the things we wanted to leverage for out-of-proc workers was the timeout token set in the webjobs . More specifically, to use the timeout cancellationToken here (https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs#L79)

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/ExecutionContext/ExecutionContext.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/ExecutionContext/ExecutionContext.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Threading;
 
 namespace Microsoft.Azure.WebJobs
 {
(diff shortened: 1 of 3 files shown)


pragnagopa:
@fabiocav / @brettsam - Passing the timeout token source will enable us to pass the information about the timeout to the Rpc layer. Do let us know if you have any concerns

brettsam:
What information do you want to pass to Rpc layer? There are other causes for cancellation as well (host shutting down, for example) -- do you need those details as well?

pragnagopa:
@brettsam - Ideally we would like to pass all the reasons for shutting down to Worker Processes so that a worker process can shutdown gracefully.  Are there other reasons for shutdown coming from webjobs sdk ?

brettsam:
If the host is shutting down we'll cancel those tokens as well. Is the Rpc layer going to do something special if it's a timeout vs a host shutdown? I think I'm just curious how this is going to be used.

We already wrap those details up in the CancellationToken we pass to .NET functions. Can that be pulled out in host today without any changes to Sdk? I think we already pass that token as a param.

yojagad:
@brettsam Right now what gets passed down to host are these
image
as part of invoking InvokeAsync and none of those seem to have the cancellationToken. Is there any other place where the timeout cancellationToken is being passed that we might be missing? 

And yes, one of the intentions is to not cause a full host shutdown if timeout happens for an out-of-proc function but rather just recycle the worker process as detailed here https://github.com/Azure/azure-functions-host/issues/5391

yojagad:
Pinging again for reviews. Let me know if there are concerns, please.

PR #2463, opened 2020-03-27
Add function execution retry support

### LINKED ISSUE #1597: Provide a configurable retry policy for function execution failures
The issue was created from discussion [here](https://github.com/Azure/azure-functions-host/issues/2439#issuecomment-368730514).

The current design of how event hub triggered functions work is optimized around making sure that bad messages or incorrect function code do not prevent the system from processing new messages and do not create infinite processing loops that generate out of control bills in the consumption plan. However this behavior has a downside in that it makes it difficult for the application developer to ensure that their function successfully processes ALL messages written to the event hub. 

In particular, even if the application developer implements their function with general error handling that  catches a processing failure and attempts to store the message for later processing, this error handling won't help them if their function failed to run in the first place, perhaps due to some sort of assembly loading issue (this can happen due to user error or due to platform issues).

This issue tracks adding a configurable retry. If dispatching an execution to the function fails (either within the function or before the user code even runs), then the execution is retried up to the specified number of times, with some backoff. These retries would only occur within the currently executing process - they are not persisted anywhere and if the process crashes or shuts down then the retry count is essentially reset. A checkpoint would not be written until all executions for a given batch have completed successfully or the retry count is reached. The retry count can be set to a special "infinite" value to indicate that a batch should ONLY be checkpointed when all function invocations for the batch completed successfully.

Turning this feature on has some implications that are worth spelling out. Say you have function code that does not handle all errors and there is a message on the event hub that causes the function to throw an exception and fail. If this function app was configured with an "infinite" retry policy then processing of the partition that this message was on would essentially get "stuck". Any data on that partition that was written after the bad message would not be processed until the bad message falls off the event hub due to its retention policy or the function code is updated by the developer to handle the bad message. **In the meantime, the other events in that batch might be reprocessed many times**.  Every time that partition gets a new owner, it would resume from the "stuck" batch and reprocess all the messages in that batch. This could theoretically lead to a given message on the event hub being processed by the function code hundreds or thousands of times.

This feature has significant parallels with a current PR open against the cosmosdb binding extension:
https://github.com/Azure/azure-webjobs-sdk-extensions/pull/349

More review is required to determine whether this feature would be useful as described, whether it has other side effects that need to be discussed, and whether this feature can land in functions V2 only or if it would also have to be backported to V1.

/cc @tobiasb

alrod:
Fixes #1597

Initial design:
https://github.com/jeffhollan/retry-design

Changes in this PR:
- Added `ExponentialBackoffRetryAttribute` - Retry options with `RandomizedExponentialBackoffStrategy`. Can be applied at class level or function level
- Added `FixedDelayRetryAttribute` - Retry options with `FixedDelayStrategy`. Can be applied at class level or function level
- Added `RetryAttribute` - Abstract base class for retries. Users can provide custom implementations of retry stratagies.
- Added `IRetryStrategy` as part of `FunctionDescriptor` - retries failed invocations upto `MaxRetryCount` using retry strategy built from `RetryAttribute`

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
@@ -95,7 +95,6 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                     logger.LogDebug($"Function execution cancelled.");
                     break;
                 }
-                logger.LogFunctionRetryAttempt(attempt, retryStrategy.MaxRetryCount);
                 functionResult = await TryExecuteAsyncCore(functionInstance, token, logger);
                 if (functionResult == null)
                 {
@@ -103,22 +102,22 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                     break;
                 }
 
-                // Build retry context
-                var retryContext = new RetryContext
-                {
-                    RetryCount = ++attempt,
-                    Exception = functionResult.Exception,
-                    Instance = functionInstance
-                };
-
-                if (attempt >= retryStrategy.MaxRetryCount)
+                if (++attempt >= retryStrategy.MaxRetryCount)
                 {
                     // no.of retries exceeded
                     break;
                 }
+                // Build retry context
+                var retryContext = new RetryContext
+                {
+                    RetryCount = attempt,
+                    Exception = functionResult.Exception,
+                    Instance = functionInstance
+                };
                 nextDelay = retryStrategy.GetNextDelay(retryContext);
-                logger.LogDebug($"Waiting for {nextDelay} before retrying function execution.");
+                logger.LogFunctionRetryDelay(nextDelay);
                 await Task.Delay(nextDelay);
+                logger.LogFunctionRetryAttempt(attempt, retryStrategy.MaxRetryCount);
             } while (attempt < retryStrategy.MaxRetryCount || retryStrategy.MaxRetryCount == -1);
 
             return functionResult;
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
@@ -58,8 +58,8 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
             _functionOutputLogger = functionOutputLogger;
             _exceptionHandler = exceptionHandler ?? throw new ArgumentNullException(nameof(exceptionHandler));
             _functionEventCollector = functionEventCollector ?? throw new ArgumentNullException(nameof(functionEventCollector));
-            _loggerFactory = loggerFactory;
-            _resultsLogger = _loggerFactory?.CreateLogger(LogCategories.Results);
+            _loggerFactory = loggerFactory ?? throw new ArgumentNullException(nameof(loggerFactory));
+            _resultsLogger = _loggerFactory.CreateLogger(LogCategories.Results);
             _globalFunctionFilters = globalFunctionFilters ?? Enumerable.Empty<IFunctionFilter>();
             _drainModeManager = drainModeManager;
         }
@@ -73,7 +73,59 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
         public async Task<IDelayedException> TryExecuteAsync(IFunctionInstance functionInstance, CancellationToken cancellationToken)
         {
             var functionInstanceEx = (IFunctionInstanceEx)functionInstance;
-            var logger = _loggerFactory?.CreateLogger(LogCategories.CreateFunctionCategory(functionInstance.FunctionDescriptor.LogName));
+            var logger = _loggerFactory.CreateLogger(LogCategories.CreateFunctionCategory(functionInstanceEx.FunctionDescriptor.LogName));
+
+            if (functionInstance.FunctionDescriptor.RetryStrategy != null)
+            {
+                return await ExecuteWithRetriesAsync(functionInstanceEx, cancellationToken, logger, functionInstance.FunctionDescriptor.RetryStrategy);
+            }
+
+            return await TryExecuteAsyncCore(functionInstanceEx, cancellationToken, logger);
+        }
+
+        internal async Task<IDelayedException> ExecuteWithRetriesAsync(IFunctionInstanceEx functionInstance, CancellationToken token, ILogger logger, IRetryStrategy retryStrategy)
+        {
+            var attempt = 0;
+            IDelayedException functionResult = null;
+            TimeSpan nextDelay;
+            do
+            {
+                if (token.IsCancellationRequested)
+                {
+                    logger.LogDebug($"Function execution cancelled.");
+                    break;
+                }
+                logger.LogFunctionRetryAttempt(attempt, retryStrategy.MaxRetryCount);
+                functionResult = await TryExecuteAsyncCore(functionInstance, token, logger);
+                if (functionResult == null)
+                {
+                    // function invocation succeeded.
+                    break;
+                }
+
+                // Build retry context
+                var retryContext = new RetryContext
+                {
+                    RetryCount = ++attempt,
+                    Exception = functionResult.Exception,
+                    Instance = functionInstance
+                };
+
+                if (attempt >= retryStrategy.MaxRetryCount)
+                {
+                    // no.of retries exceeded
+                    break;
+                }
+                nextDelay = retryStrategy.GetNextDelay(retryContext);
+                logger.LogDebug($"Waiting for {nextDelay} before retrying function execution.");
+                await Task.Delay(nextDelay);
+            } while (attempt < retryStrategy.MaxRetryCount || retryStrategy.MaxRetryCount == -1);
+
+            return functionResult;
+        }
+
+        private async Task<IDelayedException> TryExecuteAsyncCore(IFunctionInstanceEx functionInstanceEx, CancellationToken cancellationToken, ILogger logger)
+        {
             var functionStartedMessage = CreateStartedMessage(functionInstanceEx);
             var parameterHelper = new ParameterHelper(functionInstanceEx);
             ExceptionDispatchInfo exceptionInfo = null;
@@ -133,7 +185,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
             }
             finally
             {
-                ((IDisposable)functionInstance)?.Dispose();
+                ((IDisposable)functionInstanceEx)?.Dispose();
             }
 
             return exceptionInfo != null ? new ExceptionDispatchInfoDelayedException(exceptionInfo) : null;
@@ -221,6 +273,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                         functionCancellationTokenSource.Token,
                         instance.InstanceServices,
                         instance.FunctionDescriptor);
+
                     var valueBindingContext = new ValueBindingContext(functionContext, cancellationToken);
 
                     using (logger.BeginScope(_inputBindingScope))
@@ -567,13 +620,13 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
             }
 
             // Next, any class level filters are added
-            foreach (var filter in functionDescriptor.ClassLevelFilters.OfType<TFilter>())
+            foreach (var filter in functionDescriptor.ClassLevelFilters?.OfType<TFilter>())
             {
                 yield return filter;
             }
 
             // Finally, any method level filters are added
-            foreach (var filter in functionDescriptor.MethodLevelFilters.OfType<TFilter>())
+            foreach (var filter in functionDescriptor.MethodLevelFilters?.OfType<TFilter>())
             {
                 yield return filter;
             }
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
@@ -620,13 +620,13 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
             }
 
             // Next, any class level filters are added
-            foreach (var filter in functionDescriptor.ClassLevelFilters?.OfType<TFilter>())
+            foreach (var filter in functionDescriptor.ClassLevelFilters.OfType<TFilter>())
             {
                 yield return filter;
             }
 
             // Finally, any method level filters are added
-            foreach (var filter in functionDescriptor.MethodLevelFilters?.OfType<TFilter>())
+            foreach (var filter in functionDescriptor.MethodLevelFilters.OfType<TFilter>())
             {
                 yield return filter;
             }
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
@@ -115,11 +115,9 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                     Instance = functionInstance
                 };
                 nextDelay = retryStrategy.GetNextDelay(retryContext);
-                logger.LogFunctionRetryDelay(nextDelay);
+                logger.LogFunctionRetryAttempt(nextDelay, attempt, retryStrategy.MaxRetryCount);
                 await Task.Delay(nextDelay);
-                logger.LogFunctionRetryAttempt(attempt, retryStrategy.MaxRetryCount);
             } while (attempt < retryStrategy.MaxRetryCount || retryStrategy.MaxRetryCount == -1);
-
             return functionResult;
         }
 
--- a/src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
@@ -39,10 +39,10 @@ namespace Microsoft.Azure.WebJobs.Host.Indexers
         private readonly ILogger _logger;
         private readonly SharedQueueHandler _sharedQueue;
         private readonly TimeoutAttribute _defaultTimeout;
+        private readonly IRetryStrategy _defaultRetryStrategy;
         private readonly bool _allowPartialHostStartup;
         private readonly IConfiguration _configuration;
         private readonly IServiceScopeFactory _serviceScopeFactory;
-        private readonly ILoggerFactory _loggerFactory;
 
         public FunctionIndexer(
             ITriggerBindingProvider triggerBindingProvider,
@@ -56,7 +56,8 @@ namespace Microsoft.Azure.WebJobs.Host.Indexers
             INameResolver nameResolver = null,
             SharedQueueHandler sharedQueue = null,
             TimeoutAttribute defaultTimeout = null,
-            bool allowPartialHostStartup = false)
+            bool allowPartialHostStartup = false,
+            IRetryStrategy defaultRetryStrategy = null)
         {
             _triggerBindingProvider = triggerBindingProvider ?? throw new ArgumentNullException(nameof(triggerBindingProvider));
             _bindingProvider = bindingProvider ?? throw new ArgumentNullException(nameof(bindingProvider));
@@ -68,9 +69,9 @@ namespace Microsoft.Azure.WebJobs.Host.Indexers
             _sharedQueue = sharedQueue;
             _defaultTimeout = defaultTimeout;
             _allowPartialHostStartup = allowPartialHostStartup;
+            _defaultRetryStrategy = defaultRetryStrategy;
             _configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
             _serviceScopeFactory = serviceScopeFactory;
-            _loggerFactory = loggerFactory;
         }
 
         public async Task IndexTypeAsync(Type type, IFunctionIndexCollector index, CancellationToken cancellationToken)
@@ -354,7 +355,8 @@ namespace Microsoft.Azure.WebJobs.Host.Indexers
             IConfiguration configuration,
             IJobActivator jobActivator = null,
             INameResolver nameResolver = null,
-            TimeoutAttribute defaultTimeout = null)
+            TimeoutAttribute defaultTimeout = null,
+            IRetryStrategy defaultRetryStrategy = null)
         {
             var disabled = HostListenerFactory.IsDisabled(method, nameResolver, jobActivator, configuration);
 
@@ -373,6 +375,8 @@ namespace Microsoft.Azure.WebJobs.Host.Indexers
                 }
             }
 
+            IRetryStrategy retryStrategy = TypeUtility.GetHierarchicalAttributeOrNull<RetryAttribute>(method) ?? defaultRetryStrategy;
+
             return new FunctionDescriptor
             {
                 Id = method.GetFullName(),
@@ -382,6 +386,7 @@ namespace Microsoft.Azure.WebJobs.Host.Indexers
                 IsDisabled = disabled,
                 HasCancellationToken = hasCancellationToken,
                 TimeoutAttribute = TypeUtility.GetHierarchicalAttributeOrNull<TimeoutAttribute>(method) ?? defaultTimeout,
+                RetryStrategy = retryStrategy,
                 SingletonAttributes = method.GetCustomAttributes<SingletonAttribute>().ToArray(),
                 MethodLevelFilters = method.GetCustomAttributes().OfType<IFunctionFilter>().ToArray(),
                 ClassLevelFilters = method.DeclaringType.GetCustomAttributes().OfType<IFunctionFilter>().ToArray()
@@ -391,7 +396,7 @@ namespace Microsoft.Azure.WebJobs.Host.Indexers
         private FunctionDescriptor CreateFunctionDescriptor(MethodInfo method, string triggerParameterName,
             ITriggerBinding triggerBinding, IReadOnlyDictionary<string, IBinding> nonTriggerBindings)
         {
-            var descr = FromMethod(method, _configuration, this._activator, _nameResolver, _defaultTimeout);
+            var descr = FromMethod(method, _configuration, this._activator, _nameResolver, _defaultTimeout, _defaultRetryStrategy);
 
             List<ParameterDescriptor> parameters = new List<ParameterDescriptor>();
 
--- a/sample/SampleHost/Functions.cs
+++ b/sample/SampleHost/Functions.cs
@@ -4,6 +4,7 @@
 using System.Threading.Tasks;
 using Microsoft.Azure.EventHubs;
 using Microsoft.Azure.WebJobs;
+using Microsoft.Azure.WebJobs.Host;
 using Microsoft.Extensions.Logging;
 using Newtonsoft.Json.Linq;
 using SampleHost.Filters;
@@ -61,6 +62,12 @@ namespace SampleHost
             log.LogInformation($"Message complete (Id={messageId})");
         }
 
+        /// <summary>
+        /// If there is an error during function invocation, function execution will be retried up to 3 times waiting for "00:00:20" between each retry
+        /// </summary>
+        /// <param name="events"></param>
+        /// <param name="log"></param>
+        [FixedDelayRetryAttribute(3, "00:00:20")]
         public void ProcessEvents([EventHubTrigger("testhub2", Connection = "TestEventHubConnection")] EventData[] events,
             ILogger log)
         {
--- a/sample/SampleHost/Functions.cs
+++ b/sample/SampleHost/Functions.cs
@@ -67,9 +67,8 @@ namespace SampleHost
         /// </summary>
         /// <param name="events"></param>
         /// <param name="log"></param>
-        [FixedDelayRetryAttribute(3, "00:00:20")]
-        public void ProcessEvents([EventHubTrigger("testhub2", Connection = "TestEventHubConnection")] EventData[] events,
-            ILogger log)
+        [FixedDelayRetry(3, "00:00:03")]
+        public void ProcessEvents([EventHubTrigger("testhub2", Connection = "TestEventHubConnection")] EventData[] events, ILogger log)
         {
             foreach (var evt in events)
             {
--- a/sample/SampleHost/SampleHost.csproj
+++ b/sample/SampleHost/SampleHost.csproj
@@ -13,7 +13,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.EventHubs" Version="3.0.5" />
+    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.EventHubs" Version="3.0.6" />
     <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.ServiceBus" Version="3.1.1" />
     <PackageReference Include="Microsoft.Extensions.Configuration.CommandLine" Version="2.1.0" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="2.1.0" />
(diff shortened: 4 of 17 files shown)


alrod:
EventHubs change:
https://github.com/Azure/azure-functions-eventhubs-extension/pull/56

src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
```diff
@@ -40,15 +41,18 @@ public class FunctionBindingContext
         /// <param name="functionCancellationToken">The <see cref="CancellationToken"/> to use.</param>
         /// <param name="functionInvocationServices">The user logger.</param>
         /// <param name="functionDescriptor">Current function being executed. </param>
+        /// <param name="retryCount">The retry count.</param>
         public FunctionBindingContext(
```
fabiocav: Breaking change on this constructor. We need to introduce an overload.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsServiceCollectionExtensions.cs
```diff
@@ -124,6 +125,13 @@ public static IWebJobsBuilder AddWebJobs(this IServiceCollection services, Actio
                     section.Bind(options);
                 });
 
+            services.AddOptions<RetryPolicyOptions>()
+                .Configure<IConfiguration>((options, config) =>
+                {
+                    var section = config.GetWebJobsRootConfiguration().GetSection(CheckpointRetryConfigSectionName);
```
fabiocav: nit: `var` -> type name

I know this changed over time, but can you also validate that a missing section defined will still return an empty section and not `null`?

src/Microsoft.Azure.WebJobs.Host/Retry/RetryException.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        public RetryException(RetryResult retryResult)
+        {
+            RetryResult = retryResult;
+        }
+
+        /// <summary>
+        /// Gets RestryResult for RetryException.
+        /// </summary>
+        public RetryResult RetryResult { get; private set; }
```
fabiocav: nit: drop private set.

src/Microsoft.Azure.WebJobs.Host/Retry/RetryManager.cs
```diff
@@ -0,0 +1,159 @@
...
+            { 
+                if (retry != null)
+                {
+                    AsyncFallbackPolicy fallbackPolicy = Policy
+                           .Handle<FunctionInvocationException>()
+                           .FallbackAsync(async (cancellationToken) =>
+                           {
+                               // Adding empty fallback to prevent throwing an exception after all retries have been exhausted 
+                               LogWithId("All retries have been exhausted");
+                               await Task.CompletedTask;
```
fabiocav: drop the `async` and `await` here and just return the completed task.

src/Microsoft.Azure.WebJobs.Host/Retry/RetryManager.cs
```diff
@@ -0,0 +1,159 @@
...
+                           });
+                    Context context = new Context { { "retryCount", 0 }, {"retry", retry} };
+                    await fallbackPolicy
+                        .WrapAsync(GetRetryPolicy(retry))
+                        .ExecuteAsync(async (ctx, tkn) =>
+                        {
+                            int retryCount = (int)ctx["retryCount"];
```
fabiocav: are we guaranteed to have this set? Should we be a bit more defensive here?

src/Microsoft.Azure.WebJobs.Host/Retry/RetryManager.cs
```diff
@@ -0,0 +1,159 @@
...
+            }
+        }
+
+        private void RethrowExceptionIfNeeded(IEnumerable<FunctionResult> results, bool throwInvocationException)
+        {
+            foreach (var result in results)
+            {
+                if (!result.Succeeded && (result.Exception is RetryException || throwInvocationException))
+                {
+                    throw result.Exception;
```
fabiocav: Throw a new exception with this set as the inner exception.

src/Microsoft.Azure.WebJobs.Host/Retry/RetryManager.cs
```diff
@@ -0,0 +1,159 @@
...
+        private AsyncRetryPolicy GetRetryPolicy(RetryResult retry)
+        {
+            PolicyBuilder builder = Policy.Handle<FunctionInvocationException>();
+            if (retry.RetryCount == -1)
+            {
+                if (retry.ExponentialBackoff)
+                {
+                    return builder.WaitAndRetryForeverAsync(GetDurationExponential, OnRetryExponentialAsync);
+                }
+                else
```
fabiocav: nit: esle here and below may be dropped

src/Microsoft.Azure.WebJobs.Host/Retry/RetryManager.cs
```diff
@@ -0,0 +1,159 @@
...
+                {
+                    return builder.WaitAndRetryAsync(retry.RetryCount, GetDuration, OnRetryAsync);
+                }
+            }
+        }
+
+        private TimeSpan GetDuration(int retryCount, Context context)
+        {
+            return ((RetryResult)context["retry"]).SleepDuration;
```
fabiocav: are we guaranteed to have this key in the context?

src/Microsoft.Azure.WebJobs.Host/Retry/RetryManager.cs
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        private TimeSpan GetDurationExponential(int retryCount, Context context)
+        {
+            return TimeSpan.FromSeconds(Math.Pow(2, retryCount));
+        }
+
+        private Task OnRetryAsync(Exception exception, TimeSpan timeSpan, int retryCount, Context context)
+        {
+            LogWithId($"New linar retry attempt, Exception: '{exception}', Timestamp: '{timeSpan}', RetryCount: {retryCount}");
```
fabiocav: nit: spelling -> `linear`

src/Microsoft.Azure.WebJobs.Host/Retry/RetryManager.cs
```diff
@@ -0,0 +1,159 @@
...
+        private Task OnRetryExponentialAsync(Exception exception, int retryCount, TimeSpan timeSpan, Context context)
+        {
+            LogWithId($"New exponential retry attempt, Exception: '{exception}', Timestamp: '{timeSpan}', RetryCount: {retryCount}");
+            context["retryCount"] = retryCount;
+            return Task.CompletedTask;
+        }
+
+        private void LogWithId(string message)
+        {
+            _logger.LogDebug($"Id: '{_id}', {message}");
```
fabiocav: Why is this being called `id` and not `message`?

src/Microsoft.Azure.WebJobs.Host/Retry/RetryPolicyOptions.cs
```diff
@@ -0,0 +1,28 @@
...
+        }
+
+        /// <summary>
+        /// Gets a value indicating whether the retry options are defined.
+        /// </summary>
+        public bool IsDefined
+        {
+            get
+            {
+                return RetryCount != 0;
```
fabiocav: Does this map to what is defined in the context? If so, isn't -1 the value that indicates it isn't defined?

src/Microsoft.Azure.WebJobs.Host/Triggers/TriggerArgumentBinding/SimpleTriggerArgumentBinding.cs
```diff
@@ -125,5 +129,30 @@ protected void AddToBindingData(IBindingDataProvider provider, Dictionary<string
...
+            {
+                return Task.FromResult(Value);
+            }
+            public string ToInvokeString()
+            {
+                return Value.ToString();
+            }
+        }
+
+        private class ReturnValueBinder : ReturnValueProvider, IValueBinder
```
fabiocav: Why is this defined as a separate class? Can we just implement `IValueBinder` above?

fabiocav (changes requested):
Initia change requests, questions and small comments.

src/Microsoft.Azure.WebJobs.Host/WebJobs.Host.csproj
```diff
@@ -29,6 +29,7 @@
   <PackageReference Include="Microsoft.Extensions.Hosting" Version="2.1.0" />
   <PackageReference Include="Microsoft.Extensions.Logging" Version="2.1.0" />
   <PackageReference Include="Microsoft.Extensions.Logging.Configuration" Version="2.1.0" />
+  <PackageReference Include="Polly" Version="7.2.0" />
```
paulbatum: How do we feel about this dependency from both the perspective of:
a) the surface area change for WebJobs users - are we planning on releasing this as a minor version bump? 
b) any existing functions users that are already referencing Polly

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsServiceCollectionExtensions.cs
```diff
@@ -31,6 +31,7 @@ namespace Microsoft.Azure.WebJobs
     public static class WebJobsServiceCollectionExtensions
     {
         private const string SingletonConfigSectionName = "Singleton";
+        private const string CheckpointRetryConfigSectionName = "CheckpointRetryPolicy";
```
mathewc: We don't want extension specific logic/config in the core SDK. "CheckPoint" is only for EH, right?

src/Microsoft.Azure.WebJobs.Host/Retry/RetryResult.cs
```diff
@@ -0,0 +1,56 @@
...
+using Microsoft.Azure.WebJobs.Hosting;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// Represents retry settings.
+    /// </summary>
+    public class RetryResult : IOptionsFormatter
```
mathewc: If this is an Options class, I think it should be named with an "Options" suffix like all our other classes

src/Microsoft.Azure.WebJobs.Host/Bindings/ExecutionContext/ExecutionContext.cs
```diff
@@ -32,5 +32,14 @@ public class ExecutionContext
...
         public string FunctionAppDirectory { get; set; }
+
+        /// <summary>
+        /// Gets the function current retry count.
+        /// -1 - no retry policy is configured.
+        /// 0 - the function with retry policy runs first time. 
+        /// 1 - first retry attempt.
+        /// (n) - (n) retry attempt.
+        /// </summary>
+        public int RetryCount { get; internal set; }
```
mathewc: I wonder if people will be confused by this, given that triggers already have their own level of inherent retries. E.g. for a failed queue message, it might be invocation #3 of the function, and this retry will be 0. I.e. there are now have 2 layers of retries - from this new policy, and from the trigger itself.

src/Microsoft.Azure.WebJobs.Host/Retry/RetryManager.cs
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        /// <summary>
+        /// Executes a specified function with retries.
+        /// </summary>
+        /// <param name="execute">The action to perform.</param>
+        /// <param name="executeParams">Inpput arameters for the action.</param>
+        /// <param name="token">A cancellation token which can be used to cancel the action. When a retry policy
+        /// in use, also cancels any further retries.</param>
+        public async Task ExecuteWithRetriesAsync(Func<int, object[], Task<IEnumerable<FunctionResult>>> execute, object[] executeParams, CancellationToken token)
```
mathewc: I don't see where this is being called? I'd expect to see FunctionExecutor have an "execute with retries" function, similar to the existing "execute with timeout", and handle failures/retries there. As it stands, it looks like this isn't yet a general function retry mechanism? E.g. Be able to apply a [RetryPolicy] attribute to a function, with corresponding support in Azure Functions function.json. That's what I expected when I saw your PR title.

However, maybe you're not trying to provide that general feature for invocation retries, but instead something targeted only for EH checkpointing retries?  Then I'd expect this core SDK to just expose some general RetryPolicy types, and the EH extension simply uses them, and handles all checkpointing retries internally, without any changes to FunctionExecutor. In the future when we want to add a [RetryPolicy] attribute for function retries, we'd leverage that RetryPolicy type.

src/Microsoft.Azure.WebJobs.Host/WebJobs.Host.csproj
mathewc: We might consider not taking a dependency on Polly, and instead provide our own simple RetryPolicy type(s), and make it pluggable via DI. Customers can then use what they want.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -250,12 +256,21 @@ internal static async Task HandleExceptionAsync(TimeoutAttribute timeout, Except
                     // TODO: Refactor all of this logging to be implemented as a separate ILogger.
                     FunctionOutputLogger.SetOutput(outputLog);
 
+                    // Getting retry count from TriggerDetails
+                    int retryCount = 0;
```
mathewc: Not sure I understand what is going on here.

src/Microsoft.Azure.WebJobs.Host/WebJobs.Host.csproj
fabiocav: An implementation based on Polly could be broken off into its own package and outside of the SDK. That's the approach taken by the HTTP client factory to avoid this situation.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsServiceCollectionExtensions.cs
alrod: I verified that the missing section still returns empty section

src/Microsoft.Azure.WebJobs.Host/Retry/RetryManager.cs
alrod: yes, the context is set above with 0:
`new Context { { RetryCountString, 0 }, { RertyString, retry} }`

src/Microsoft.Azure.WebJobs.Host/Retry/RetryManager.cs
alrod: We do not want to wrap the exception as we need `FunctionInvocationException`/ `RetryException `types for retry policy:

```
AsyncFallbackPolicy fallbackPolicy = Policy                       
     .Handle<FunctionInvocationException>()
...
catch (RetryException retryException)
```

src/Microsoft.Azure.WebJobs.Host/Retry/RetryManager.cs
alrod: yes, all contexts are created with "retry" property

src/Microsoft.Azure.WebJobs.Host/Retry/RetryManager.cs
alrod: because "id" is added to each log

src/Microsoft.Azure.WebJobs.Host/Retry/RetryPolicyOptions.cs
alrod: Actually we do not need the `IsDefined` property. I removed it.

src/Microsoft.Azure.WebJobs.Host/WebJobs.Host.csproj
alrod: I added WebJobs.Extensions.Polly with the reference to Polly which can be added to an extension via DI.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsServiceCollectionExtensions.cs
alrod: we also want to checkpoint CosmosDB and Kafka

src/Microsoft.Azure.WebJobs.Host/Retry/RetryResult.cs
alrod: Fixed.

src/Microsoft.Azure.WebJobs.Host/Bindings/ExecutionContext/ExecutionContext.cs
alrod: Yes, I see how this can be confusing. 
But we need the property in `ExecutionContext` to indicate retry attempt number. I think we can describe the behavior of the property in documentation and point that it only make sense for EventHub/CosmosDB/Kafka triggers.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
alrod: `TriggerDetails` is used to pass current retry count to `ExecutionContext`

https://github.com/Azure/azure-webjobs-sdk/pull/2463/files#diff-d4edc2f1f55dd03bebe0b9fa640a1656R407

https://github.com/Azure/azure-functions-eventhubs-extension/pull/56/files#diff-5e1bd09eff27cb8f09152894835d29b6R88 (load RetryManagerTests.cs in the PR)

src/Microsoft.Azure.WebJobs.Host/Retry/RetryManager.cs
alrod: The calls:
https://github.com/Azure/azure-webjobs-sdk/pull/2463/files#diff-d4edc2f1f55dd03bebe0b9fa640a1656R391 (load RetryManagerTests.cs in the PR)

https://github.com/Azure/azure-functions-eventhubs-extension/pull/56/files#diff-195d2c9ccd88ec324d536317b700333aR204

The idea of `RetryManager` is to execute with retries code which returns `IEnumerable<FunctionResult>` so we can use it for EH/ComsosDB/Kafka listeners. We do not want `[RetryPolicy]` attribute for any trigger.

Design of the feature is here:
https://github.com/jeffhollan/retry-design

src/Microsoft.Azure.WebJobs.Host/Bindings/ExecutionContext/ExecutionContext.cs
mathewc: I'd add a cref here to RetryAttribute, basically clarifying that this retry count only applies if the retry annotation is used.

src/Microsoft.Azure.WebJobs.Host/Config/JobHostRetryOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        /// <summary>
+        /// The retry count.
+        /// </summary>
+        public int RetryCount { get; set; }
+
+        /// <summary>
+        /// The duration to wait for for a particular retry attemp
+        /// </summary>
+        public TimeSpan SleepDuration { get; set; } = TimeSpan.Zero;
```
mathewc: I'd call this "Delay" - the delay between retries.

src/Microsoft.Azure.WebJobs.Host/Config/JobHostRetryOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    /// <summary>
+    /// Configuration options for controlling function execution retry behavior.
+    /// </summary>
+    public class JobHostRetryOptions
+    {
+        /// <summary>
+        /// The retry count.
+        /// </summary>
+        public int RetryCount { get; set; }
```
mathewc: I'd call this "MaxRetries"?

src/Microsoft.Azure.WebJobs.Host/Config/JobHostRetryOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        /// <summary>
+        /// The duration to wait for for a particular retry attemp
+        /// </summary>
+        public TimeSpan SleepDuration { get; set; } = TimeSpan.Zero;
+
+        /// <summary>
+        /// Gets a value indicating whether the expbnential backoff is used.
+        /// </summary>
+        public bool ExponentialBackoff { get; set; }
```
mathewc: It's confusing to have both linear and exponential stategies encoded in this single options. E.g. what if Exponetial is true and a Delay is also specified? The delay doesn't apply.

I wonder if instead we shouldn't have an options hierarchy:

 ```csharp
JobHostRetryOptions
{
    int MaxRetries
}

JobHostLinearBackoffRetryOptions : JobHostRetryOptions
{
    TimeSpan Delay
}

JobHostExponentialBackoffRetryOptions : JobHostRetryOptions
{
    TimeSpan MinimumDelay
    TimeSpan MaximumDelay
}

```

For the exponential, I propose a min/max interval based on how our existing RandomizedExponentialBackoffStrategy works.

This would then translate to also having a hierarchy in the attributes: RetryAttribute, LinearBackoffRetryAttribute, ExponentialBackoffRetryAttribute.

@fabiocav what do you think?

File: src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs [hunk 22]
```diff
@@ -175,6 +175,12 @@ private async Task<IDelayedException> TryExecuteAsyncCore(IFunctionInstanceEx fu
                 await HandleExceptionAsync(functionInstance.FunctionDescriptor.TimeoutAttribute, exceptionInfo, _exceptionHandler);
             }
 
+            if (parameterHelper.ReturnValue is RetryResult retry)
```
mathewc: The HandleException logic above ends up invoking the IWebJobsExceptionHandler, which in Functions we have an implementation that you should look at. Looks like your retry exceptions will pass through that handler and ultimately cause the Functions host to shut down.

I think for retries you need to handle them before and not let them flow through there, since they're not fatal exceptions.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
mathewc: Since the executor fully controls the retries, I don't see why you need to be fishing this out of TriggerDetails like this. Instead, you should just pass the retry count in from wherever it is being performed upstream

src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionExecutor.cs
```diff
@@ -14,12 +14,14 @@ internal class TriggeredFunctionExecutor<TTriggerValue> : ITriggeredFunctionExec
         private FunctionDescriptor _descriptor;
         private ITriggeredFunctionInstanceFactory<TTriggerValue> _instanceFactory;
         private IFunctionExecutor _executor;
+        private readonly IRetryManagerProvider _retryManagerProvider;
```
mathewc: This will only provide retries for triggered functions. I think this means that you won't currently get retries for JobHost.CallAsync direct invocations, which happen for manually triggered functions as well as for http functions in AzureFunctions. I think you need to hoist the retry stuff up higher, into FunctionExecutor itself. Also add tests verifying this.

src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionExecutor.cs
```diff
@@ -31,6 +33,20 @@ public FunctionDescriptor Function
         }
 
         public async Task<FunctionResult> TryExecuteAsync(TriggeredFunctionData input, CancellationToken cancellationToken)
+        {
+            if (_retryManagerProvider != null)
+            {
+                IRetryManager retryManager = _retryManagerProvider.Create(_descriptor.RetryAttribute);
```
mathewc: Ideally on the execution hotpath, we wouldn't have to create a new manager for each invocation. Instead, can't you create it once on construction (or get it from DI) and have the ExecuteWithRetries handle state internally (i.e. the manager instance itself is stateless)?

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
```diff
@@ -376,6 +383,7 @@ internal async Task IndexMethodAsyncCore(MethodInfo method, IFunctionIndexCollec
                 IsDisabled = disabled,
                 HasCancellationToken = hasCancellationToken,
                 TimeoutAttribute = TypeUtility.GetHierarchicalAttributeOrNull<TimeoutAttribute>(method) ?? defaultTimeout,
+                RetryAttribute = TypeUtility.GetHierarchicalAttributeOrNull<RetryAttribute>(method) ?? defaultRetry,
```
mathewc: This means that the default host level retry might be linear, while individual functions can have an exponential. That's good, just calling it out

src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionExecutor.cs
mathewc: Actually, why do we need an IRetryManagerProvider then at that point? It can just be a single IRetryManager interface, with the instance created only once.

src/Microsoft.Azure.WebJobs.Host/Retry/RetryResult.cs
```diff
@@ -0,0 +1,32 @@
...
+
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// Represents retry settings.
+    /// </summary>
+    public class RetryResult : RetryAttribute
```
mathewc: I don't understand this Type at all. I don't think it's needed.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -175,6 +175,12 @@ private async Task<IDelayedException> TryExecuteAsyncCore(IFunctionInstanceEx fu
...
+            {
+                // Set RetryException here so the execution is not logged as failed
+                exceptionInfo = ExceptionDispatchInfo.Capture(new RetryException(retry));
```
mathewc: I don't understand why we need the RetryException/RetryResult types at all. It seems RetryManager can just handle retries internally, handling all exceptions and state internally. If the final retry fails, it just lets the error bubble out naturally

src/Microsoft.Azure.WebJobs.Host/Triggers/TriggerArgumentBinding/SimpleTriggerArgumentBinding.cs
```diff
@@ -88,7 +89,10 @@ public virtual async Task<ITriggerData> BindAsync(TTriggerValue value, ValueBind
             string invokeString = await ConvertToStringAsync(eventData);
             var valueProvider = new ConstantValueProvider(userValue, this.ElementType, invokeString);
 
-            return new TriggerData(valueProvider, bindingData);
+            return new TriggerData(valueProvider, bindingData)
+            {
+                ReturnValueProvider = new ReturnValueProvider() { Value = "return" }
```
mathewc: Why is this needed? We shouldn't have to make any changes here.

src/Microsoft.Azure.WebJobs.Host/Retry/RetryResult.cs
gabrieljoelc: Is it a copy from @jeffhollan 's prototype [here](https://github.com/jeffhollan/retry-design/blob/36d3a4d6704e2dc21b212dae7b9ea6175e87e38b/csharp/EventHubTrigger.cs#L54)?

gabrieljoelc:
_I've been following the Event Hub resiliency story for about 3 years and so am excited about this pull request. Forgive my ignorance in this post._ 😅 

I realize that this PR is in the `azure-webjobs-sdk` repo and not the [azure-functions-eventhubs-extension](https://github.com/Azure/azure-functions-eventhubs-extension) repo, but I'm interested in how they interact. Specifically these comments in `EventProcessor#ProcessEventsAsync()` [here](https://github.com/Azure/azure-functions-eventhubs-extension/blob/e263c0c09b360d7b8ab31c4767e58bc99a1e36ad/src/Microsoft.Azure.WebJobs.Extensions.EventHubs/Listeners/EventHubListener.cs#L218-L222):
```csharp
                // Note: we intentionally checkpoint the batch regardless of function 
                // success/failure. EventHub doesn't support any sort "poison event" model,
                // so that is the responsibility of the user's function currently. E.g.
                // the function should have try/catch handling around all event processing
                // code, and capture/log/persist failed events, since they won't be retried.
```
does that comment/behavior have any bearing on this PR (again forgive my ignorance)?

pragnagopa:
- Pushed original changes in the PR to : `pgopa/alrod-retry`
- Rebased on dev
- Will ping here after addressing the feedback

src/Microsoft.Azure.WebJobs.Host/Retry/RetryResult.cs
pragnagopa: @jeffhollan -I agree with @mathewc . Do you have any concerns if we do not add this type and let the function code throw exception if the intention of the user is to retry execution?

src/Microsoft.Azure.WebJobs.Host/Bindings/ExecutionContext/ExecutionContext.cs
pragnagopa: Removed `RetryCount` from execution context.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
pragnagopa: Udpated!

src/Microsoft.Azure.WebJobs.Host/Config/JobHostRetryOptions.cs
pragnagopa: Updated!

src/Microsoft.Azure.WebJobs.Host/Config/JobHostRetryOptions.cs
pragnagopa: Updated to `MaxRetryCount` ?

src/Microsoft.Azure.WebJobs.Host/Config/JobHostRetryOptions.cs
pragnagopa: Added this. Please take a look.

src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionExecutor.cs
pragnagopa: Moved to FucntionExecutor. Please take a look.

src/Microsoft.Azure.WebJobs.Host/Retry/RetryResult.cs
pragnagopa: Removed this type for now.

src/Microsoft.Azure.WebJobs.Host/Triggers/TriggerArgumentBinding/SimpleTriggerArgumentBinding.cs
pragnagopa: deleted.

pragnagopa:
@mathewc  - Addressed most of the feedback. I still have to do more testing. Can you please take a look at the `DelayStrategy` changes?

pragnagopa:
PR is ready for review. Please take a look.

mathewc (approved):

ditodo:
Is there any official documentation on how to use the retry feature?
Is Jeff's spec still valid or did the solution divert from that proposal?
https://github.com/jeffhollan/retry-design

pragnagopa:
@ditodo - please see https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-error-pages?tabs=csharp#retry-policies-preview for details documentation.

jack0fshad0ws:
Is this planned for isolated functions? Net7 is isolated only..

PR #2473, opened 2020-04-10
To have nuget package for FakeAzureStorage

bb-naveenkumar:
I have been trying to do webjob trigger unit testing and using FakeAzureStorage seems like a good approach to do end to end test. However like test common NuGet there is no NuGet for FakeAzureStorage. Added the code here to create the NuGet for FakeAzureStorage project.

### DIFF
--- a/Build.ps1
+++ b/Build.ps1
@@ -22,7 +22,8 @@ $projects =
   "src\Microsoft.Azure.WebJobs.Logging.ApplicationInsights\WebJobs.Logging.ApplicationInsights.csproj",
   "src\Microsoft.Azure.WebJobs.Extensions.Storage\WebJobs.Extensions.Storage.csproj",
   "src\Microsoft.Azure.WebJobs.Host.Storage\WebJobs.Host.Storage.csproj",
-  "test\Microsoft.Azure.WebJobs.Host.TestCommon\WebJobs.Host.TestCommon.csproj"
+  "test\Microsoft.Azure.WebJobs.Host.TestCommon\WebJobs.Host.TestCommon.csproj",
+  "test\FakeStorage\FakeAzureStorage.csproj"
 
 foreach ($project in $projects)
 {
--- a/Build.ps1
+++ b/Build.ps1
@@ -22,8 +22,7 @@ $projects =
   "src\Microsoft.Azure.WebJobs.Logging.ApplicationInsights\WebJobs.Logging.ApplicationInsights.csproj",
   "src\Microsoft.Azure.WebJobs.Extensions.Storage\WebJobs.Extensions.Storage.csproj",
   "src\Microsoft.Azure.WebJobs.Host.Storage\WebJobs.Host.Storage.csproj",
-  "test\Microsoft.Azure.WebJobs.Host.TestCommon\WebJobs.Host.TestCommon.csproj",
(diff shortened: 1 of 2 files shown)


bb-naveenkumar:
Testing appveyor integrations

bb-naveenkumar:
#2473 , the fork was deleted so had to close that PR.

brettsam:
How are you planning on consuming this? Are you building it yourself or are you taking it from our CI builds? I want to make it clear that this isn't something we'd support externally -- not every scenario is implemented so there may be APIs that don't work as expected.

bb-naveenkumar:
@brettsam we will be building the host builder our self meaning adding the mock services and handling, however since we don't want to enable the emulator as a testing requirement we will need to create storage accounts like done here in this project and use that to verify the test logs and scenarios for the triggers. 

Also it does not need to have every scenario covered in the APIs since the methods could be overridden which are for specific scenarios like some of the methods are already there, which I believe are sufficient for end to end test for triggers based on basic testing. 

Also, I am already using this approach in my unit testing since last October and is quite a help, we just need to provide the input and mock out the services and execute the trigger like RunTrigger methods, the thing is some of the code I used is repetitive which is already part of sdk but not available like getting to add a fake storage. Also, it makes hard for others projects to use similar thing unless they do some copy pasting of code or create specific usage nugets which I don't want, cause it is already here in sdk though not available to use and it will also make the testing strategies used by others be in sync with what sdk is doing. 

I hope this answers your query, let me know if there is still something you need to clarify the usage.

There are still little improvements that could be added to this project like getting to add the fake storage with creds that we specify or something like that, but they could be improved once people starts using this approach in their unit testing one done in sdk rather than doing traditional mocks of creating mock and adding in the constructor or method to verify tests.

Few host extensions and service extensions for injection service mocks and reading those mocks from webjob host that are not here, though could be added in projects separately as well.

bb-naveenkumar:
Also, do point to some scenarios that are not handled in this and are known, may be I can contribute in handling those as well, either get them to work or skip which are not needed to do a test.

brettsam:
I don't think we would publish this package to nuget -- how are you intending to consume it? Are you building WebJobs yourself?

bb-naveenkumar:
Hi @brettsam, sorry for late reply. Are we saying we would not publish it due to pipeline not being configured or is it not recommended to do testing for webjob triggers like it is done in azure sdk https://github.com/Azure/azure-webjobs-sdk/blob/4130350327c6d637d48456222de7e658c6cf729a/test/Microsoft.Azure.Webjobs.Extensions.Storage.UnitTests/Queues/QueueTriggerTests.cs in this file.

Way to consume is simple, like here in above file it is reference once nuget is available everyone can reference that in their unit test projects and use it. Line of code being used `var account = CreateFakeStorageAccount();`

Or are we saying that the way it is tested in the sdk is not recommended to be done by others?

bb-naveenkumar:
@brettsam any updates?

bb-naveenkumar:
@brettsam any updates?

fabiocav:
@bb-naveenkumar that project is really not built for and designed for public consumption. If you need that code, the recommendation would be to publish the package to a private NuGet feed for reuse. Thanks!

PR #2474, opened 2020-04-14
Track original activity at the Application Insights Logger

TsuyoshiUshio:
Hi @brettsam 

I use to submit the PR that is enabling suppression of the telemetry, however, we realized that we don't need suppression, but we need to track the activity that is generated from an extension with adding custom property of the functions host (e.g. invocation id) 

We can do it if we specify "MS_IgnoreActivity" however, it generate a new telemetry that is a child of the original telemetry. 

To enable it, I create this PR. Could you have a look? 

I'll going to close the 
https://github.com/Azure/azure-webjobs-sdk/pull/2467
Since It is not needed anymore. 

## Usage 

```csharp
using (logger.BeginScope(new Dictionary<string, object> { ["MS_TrackActivity"] = null }))
{
     // functions invocation
 }
```

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
@@ -526,7 +526,10 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
                     // We'll need to store this operation context so we can stop it when the function completes
                     stateValues[OperationContext] = operation;
                 }
-            } // Track Activity if the key is the TrackActivity Scope is specified. 
+            } 
+            // If there is a current activity, it is assumed that Application Insights will track it so we do not start an operation. 
+            // However, in some cases (such as Durable functions), this is not the case. This allows the scope to decide whether
+            // an operation should be started, even when the current activity is not null. 
             else if (allScopes.ContainsKey("MS_TrackActivity"))
             {
                 var operation = _telemetryClient.StartOperation<RequestTelemetry>(currentActivity);
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
@@ -526,6 +526,11 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
                     // We'll need to store this operation context so we can stop it when the function completes
                     stateValues[OperationContext] = operation;
                 }
+            } // Track Activity if the key is the TrackActivity Scope is specified. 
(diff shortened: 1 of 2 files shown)


reyang (approved):
LGTM.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -526,6 +526,11 @@ private void StartTelemetryIfFunctionInvocation(IDictionary<string, object> stat
                     // We'll need to store this operation context so we can stop it when the function completes
                     stateValues[OperationContext] = operation;
                 }
+            } // Track Activity if the key is the TrackActivity Scope is specified. 
+            else if (allScopes.ContainsKey("MS_TrackActivity"))
```
brettsam, line 533: Should you also check that `currentActivity != null` here? I think it'd be more clear what is happening.

Also change the comment to be:

```
// If there is a current activity, it is assumed that Application Insights will track it so we do not start an operation. 
// However, in some cases (such as Durable functions), this is not the case. This allows the scope to decide whether
// an operation should be started, even when the current activity is not null. 
```

brettsam (changes requested):
Minor comments...

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
TsuyoshiUshio: Hi @brettsam ,
I update the comment!  I also tried to `currentActivity != null`, however, VS19 complaint it, since the first if clause is 

```csharp
            var currentActivity = Activity.Current;
            if (currentActivity == null ||
                // Activity is tracked, but Functions wants to ignore it:
                allScopes.ContainsKey("MS_IgnoreActivity") ||
                // Functions create another RequestTrackingTelemetryModule to make sure first request is tracked (as ASP.NET Core starts before web jobs)
                // however at this point we may discover that RequestTrackingTelemetryModule is disabled by customer and even though Activity exists, request won't be tracked
                // So, if we've got AspNetCore Activity and EnableHttpTriggerExtendedInfoCollection is false - track request here.
                (!_loggerOptions.HttpAutoCollectionOptions.EnableHttpTriggerExtendedInfoCollection && IsHttpRequestActivity(currentActivity)))
            {
```

brettsam: Oh interesting -- that's okay to leave it off, I was just trying to make it more explicit. I guess it must not be null if the first check failed.

brettsam (approved):

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
TsuyoshiUshio: Thanks!

Issue #2420, opened 2020-01-27
CancellationToken never cancelled in Triggered WebJob


NickSevens:
I'm having a triggered webjob which executes some `async` actions (Graph API calls) in a loop.
When trying to use the `cancellationToken` to handle application shutdown, it seems it's never set to cancelled.

My webjob is initiated with:

```
using var host = hostBuilder.Build();

await host.StartAsync();

var jobHost = host.Services.GetService<IJobHost>();
await jobHost.CallAsync(nameof(Functions.RunBoxSync));

await host.StopAsync();
```
and Functions.RunBoxSync is defined as such:
```
public async Task RunBoxSync(ILogger log, CancellationToken cancellationToken)
```
But, as said, the cancellationToken is never cancelled.

How do I pass the **correct** cancellationToken to the Function?

divinci:
See issue #2088 only a year old!

gidziu:
same issue, is anyone working on this? Probably it is better to stay away from thsi webjob business

mathewc:
In your code above, you're awaiting function completion BEFORE calling stop. So the function has completed before the token is cancelled by stop, so it's expected that the token your function receives isn't cancelled. What does your function body look like?

v-anvari:
Hi @NickSevens , Apologies for the delayed response. Let us know if you are still facing the issue, so that we can look into it further

PR #2478, opened 2020-04-19
Throughput improvements

mathewc:
For Functions work item: https://github.com/Azure/azure-functions-host/issues/5601

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/ExecutionContext/ExecutionContextBindingProvider.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/ExecutionContext/ExecutionContextBindingProvider.cs
@@ -42,6 +42,7 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
         {
             private readonly ParameterInfo _parameter;
             private readonly IOptions<ExecutionContextOptions> _options;
+            private static ParameterDisplayHints _displayHints = new ParameterDisplayHints { Description = "Function ExecutionContext" };
 
             public ExecutionContextBinding(ParameterInfo parameter, IOptions<ExecutionContextOptions> options)
             {
@@ -61,7 +62,7 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
                     throw new ArgumentNullException("context");
                 }
 
-                return BindInternalAsync(CreateContext(context.ValueContext));
+                return Task.FromResult<IValueProvider>(new ExecutionContextValueProvider(context.ValueContext, _options));
             }
 
             public Task<IValueProvider> BindAsync(object value, ValueBindingContext context)
@@ -71,29 +72,7 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
                     throw new ArgumentNullException("context");
                 }
 
-                return BindInternalAsync(CreateContext(context));
-            }
-
-            private ExecutionContext CreateContext(ValueBindingContext context)
-            {
-                var result = new ExecutionContext
-                {
-                    InvocationId = context.FunctionInstanceId,
-                    FunctionName = context.FunctionContext.MethodName,
-                    FunctionDirectory = Environment.CurrentDirectory,
-                    FunctionAppDirectory = _options.Value.AppDirectory
-                };
-
-                if (result.FunctionAppDirectory != null)
-                {
-                    result.FunctionDirectory = System.IO.Path.Combine(result.FunctionAppDirectory, result.FunctionName);
-                }
-                return result;
-            }
-
-            private static Task<IValueProvider> BindInternalAsync(ExecutionContext executionContext)
-            {
-                return Task.FromResult<IValueProvider>(new ExecutionContextValueProvider(executionContext));
+                return Task.FromResult<IValueProvider>(new ExecutionContextValueProvider(context, _options));
             }
 
             public ParameterDescriptor ToParameterDescriptor()
@@ -101,20 +80,19 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
                 return new ParameterDescriptor
                 {
                     Name = _parameter.Name,
-                    DisplayHints = new ParameterDisplayHints
-                    {
-                        Description = "Function ExecutionContext"
-                    }
+                    DisplayHints = _displayHints
                 };
             }
 
             private class ExecutionContextValueProvider : IValueProvider
             {
-                private readonly ExecutionContext _context;
+                private readonly ValueBindingContext _context;
+                private readonly IOptions<ExecutionContextOptions> _options;
 
-                public ExecutionContextValueProvider(ExecutionContext context)
+                public ExecutionContextValueProvider(ValueBindingContext context, IOptions<ExecutionContextOptions> options)
                 {
                     _context = context;
+                    _options = options;
                 }
 
                 public Type Type
@@ -124,12 +102,30 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
 
                 public Task<object> GetValueAsync()
                 {
-                    return Task.FromResult<object>(_context);
+                    return Task.FromResult<object>(CreateContext());
                 }
 
                 public string ToInvokeString()
                 {
-                    return _context.InvocationId.ToString();
+                    return _context.FunctionInstanceId.ToString();
+                }
+
+                private ExecutionContext CreateContext()
+                {
+                    var result = new ExecutionContext
+                    {
+                        InvocationId = _context.FunctionInstanceId,
+                        FunctionName = _context.FunctionContext.MethodName,
+                        FunctionDirectory = Environment.CurrentDirectory,
+                        FunctionAppDirectory = _options.Value.AppDirectory
+                    };
+
+                    if (result.FunctionAppDirectory != null)
+                    {
+                        result.FunctionDirectory = System.IO.Path.Combine(result.FunctionAppDirectory, result.FunctionName);
+                    }
+
+                    return result;
                 }
             }
         }
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
@@ -8,6 +8,7 @@ using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using System.Runtime.ExceptionServices;
 using System.Threading;
 using System.Threading.Tasks;
@@ -97,7 +98,6 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
 
         private async Task<IDelayedException> TryExecuteAsyncCore(IFunctionInstanceEx functionInstance, CancellationToken cancellationToken)
         {
-
             ILogger logger = _loggerFactory?.CreateLogger(LogCategories.CreateFunctionCategory(functionInstance.FunctionDescriptor.LogName));
 
             FunctionStartedMessage functionStartedMessage = CreateStartedMessageWithoutArguments(functionInstance);
@@ -227,16 +227,9 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
         private async Task<string> ExecuteWithLoggingAsync(IFunctionInstanceEx instance, FunctionStartedMessage message,
             FunctionInstanceLogEntry instanceLogEntry, ParameterHelper parameterHelper, ILogger logger, CancellationToken cancellationToken)
         {
-            IFunctionOutputDefinition outputDefinition = null;
-            IFunctionOutput outputLog = null;
-            ITaskSeriesTimer updateOutputLogTimer = null;
-            TextWriter functionOutputTextWriter = null;
-
-            IFunctionOutputLogger outputLogger = _functionOutputLogger;
-            outputDefinition = await outputLogger.CreateAsync(instance, cancellationToken);
-            outputLog = outputDefinition.CreateOutput();
-            functionOutputTextWriter = outputLog.Output;
-            updateOutputLogTimer = StartOutputTimer(outputLog.UpdateCommand, _exceptionHandler);
+            var outputDefinition = await _functionOutputLogger.CreateAsync(instance, cancellationToken);
+            var outputLog = outputDefinition.CreateOutput();
+            var updateOutputLogTimer = StartOutputTimer(outputLog.UpdateCommand, _exceptionHandler);
 
             try
             {
@@ -268,9 +261,8 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                     string startedMessageId = null;
                     using (parameterHelper)
                     {
+                        // Log started events
                         startedMessageId = await LogFunctionStartedAsync(message, outputDefinition, parameterHelper, cancellationToken);
-
-                        // Log started
                         await NotifyPostBindAsync(instanceLogEntry, message.Arguments);
 
                         try
@@ -352,7 +344,6 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
             }
 
             TimeSpan timeout = attribute.Timeout;
-
             bool usingCancellationToken = instance.FunctionDescriptor.HasCancellationToken;
             if (!usingCancellationToken && !attribute.ThrowOnTimeout)
             {
@@ -371,8 +362,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
 
             timer.Elapsed += (o, e) =>
             {
-                OnFunctionTimeout(timer, instance.FunctionDescriptor, instance.Id, timeout, attribute.TimeoutWhileDebugging, logger, cancellationTokenSource,
-                    () => Debugger.IsAttached);
+                OnFunctionTimeout(timer, instance.FunctionDescriptor, instance.Id, timeout, attribute.TimeoutWhileDebugging, logger, cancellationTokenSource, () => Debugger.IsAttached);
             };
 
             timer.Start();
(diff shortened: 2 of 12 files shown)


src/Microsoft.Azure.WebJobs.Host/Bindings/BindingContext.cs
```diff
@@ -110,5 +133,14 @@ public ValueBindingContext ValueContext
                 return _valueContext;
             }
         }
+
+        public async Task<IReadOnlyDictionary<string, object>> GetBindingDataAsync()
```
mathewc, line 137: One of the big perf improvements I made for Http was to make resolution of binding data lazy, so it doesn't happen if not needed. E.g. in most http scenarios, binding data isn't needed since no binding expressions or in/out bindings are used. So many of the changes in this PR are about plumbing async lazy resolution of binding data through the SDK.

src/Microsoft.Azure.WebJobs.Host/Bindings/ExecutionContext/ExecutionContextBindingProvider.cs
```diff
@@ -61,7 +62,7 @@ public Task<IValueProvider> BindAsync(BindingContext context)
                     throw new ArgumentNullException("context");
                 }
 
-                return BindInternalAsync(CreateContext(context.ValueContext));
+                return Task.FromResult<IValueProvider>(new ExecutionContextValueProvider(context.ValueContext, _options));
```
mathewc, line 65: delay creation of the context until actually needed (when value is accessed)

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -455,13 +445,13 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
             CancellationTokenSource functionCancellationTokenSource)
         {
-            IFunctionInvoker invoker = instance.Invoker;
-
             ITaskSeriesTimer updateParameterLogTimer = null;
-
-            var parameterWatchers = parameterHelper.CreateParameterWatchers();
-            IRecurrentCommand updateParameterLogCommand = outputDefinition.CreateParameterLogUpdateCommand(parameterWatchers, logger);
-            updateParameterLogTimer = StartParameterLogTimer(updateParameterLogCommand, _exceptionHandler);
+            if (outputDefinition != NullFunctionOutputDefinition.Instance)
```
mathewc, line 449: Avoid a bunch of unnecessary work when not needed

src/Microsoft.Azure.WebJobs.Host/Extensions/ITriggerDataExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+            }
+            else
+            {
+                return Task.FromResult(triggerData.BindingData);
+            }
+        }
+
+        public static IValueBinder GetReturnValueBinderOrNull(this ITriggerData triggerData)
+        {
+            // TODO: we should NOT be casting to a specific type like this, since the binding interfaces are based
```
mathewc, line 27: In the next major version, we need to clean up the ITriggerData/TriggerData APIs

src/Microsoft.Azure.WebJobs.Host/Loggers/FunctionInstanceLogger.cs
```diff
@@ -24,12 +24,17 @@ public FunctionInstanceLogger(ILoggerFactory loggerFactory)
 
         public Task<string> LogFunctionStartedAsync(FunctionStartedMessage message, CancellationToken cancellationToken)
         {
-            string traceMessage = string.Format(CultureInfo.InvariantCulture, "Executing '{0}' (Reason='{1}', Id={2})", message.Function.ShortName, message.FormatReason(), message.FunctionInstanceId);
-            Log(LogLevel.Information, message.Function, message.FunctionInstanceId, traceMessage);
+            string name = message.Function.ShortName;
+            string reason = message.ReasonDetails ?? message.FormatReason();
+            var id = message.FunctionInstanceId;
+
+            Log(LogLevel.Information, message.Function, (s, e) => $"Executing '{name}' (Reason='{reason}', Id={id})");
```
mathewc: Not really an expensive log statement, but happens on every request, so I've moved it to the formatter func pattern

src/Microsoft.Azure.WebJobs.Host/Triggers/TriggerData.cs
```diff
@@ -3,25 +3,26 @@
...
+using System.Threading.Tasks;
 using Microsoft.Azure.WebJobs.Host.Bindings;
 
 namespace Microsoft.Azure.WebJobs.Host.Triggers
 {
     /// <summary>
     /// Represents data returned after a trigger parameter value is bound.
     /// </summary>
-    public class TriggerData : ITriggerData
+    public class TriggerData : ITriggerData, IBindingDataFactory
```
mathewc, line 14: Because the binding extension APIs are based on ITriggerData and not this concrete class, I needed a pattern that worked even if an extension returned its own ITriggerData implementation. Hence the new interface.

src/Microsoft.Azure.WebJobs.Host/Triggers/TriggerData.cs
```diff
@@ -45,13 +57,31 @@ public IValueProvider ValueProvider
         /// <inheritdoc/>
         public IReadOnlyDictionary<string, object> BindingData
         {
-            get { return _bindingData; }
+            get 
+            { 
+                if (_bindingData == null && _bindingDataFactory != null)
+                {
+                    // This is here for back compat. All framework consumers should be using GetBindingDataAsync
```
mathewc, line 64: All the internal calls in the SDK now use the below async method

src/Microsoft.Azure.WebJobs.Host/Triggers/TriggeredFunctionBinding.cs
```diff
@@ -48,16 +48,15 @@ internal class TriggeredFunctionBinding<TTriggerValue> : ITriggeredFunctionBindi
...
-            IReadOnlyDictionary<string, object> bindingData;
 
             IValueBinder triggerReturnValueProvider= null;
             try
             {
-                ITriggerData triggerData = await _triggerBinding.BindAsync(value, context);
+                triggerData = await _triggerBinding.BindAsync(value, context);
                 triggerProvider = triggerData.ValueProvider;
-                bindingData = triggerData.BindingData;
-                triggerReturnValueProvider = (triggerData as TriggerData)?.ReturnValueProvider;
```
mathewc, line 60: ReturnValueProvider should have been added as a separate interface, not a property on TriggerData, given that the extensibility interfaces aren't based on the TriggerData type

src/Microsoft.Azure.WebJobs.Host/Loggers/FunctionInstanceLogger.cs
brettsam: Did you look at changing this to the `LoggerMessage.Define` pattern at all? https://github.com/Azure/azure-webjobs-sdk/pull/2235.

It looks like there's some allocation savings doing it this way (esp for such an often-called message) -- but I clearly dropped the ball and never pushed this over the finish line.

mathewc: I've changed the executing/executed logs to use a defined delegate. I've also added logger caching.

I've left the templatized event details log as is for now, since that's more complicated. We can address that later. That's not logged on http functions anyways.

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostContextFactory.cs
```diff
@@ -126,7 +126,7 @@ public async Task<JobHostContext> Create(JobHost host, CancellationToken shutdow
                     executor.HostOutputMessage = hostOutputMessage;
                 }
 
-                IEnumerable<FunctionDescriptor> descriptors = functions.ReadAllDescriptors();
+                var descriptors = functions.ReadAllDescriptors();
```
brettsam: Any reason for this? Other than formatting, it's the only change in this file.

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
```diff
@@ -372,9 +372,9 @@ internal async Task IndexMethodAsyncCore(MethodInfo method, IFunctionIndexCollec
                 IsDisabled = disabled,
                 HasCancellationToken = hasCancellationToken,
                 TimeoutAttribute = TypeUtility.GetHierarchicalAttributeOrNull<TimeoutAttribute>(method) ?? defaultTimeout,
-                SingletonAttributes = method.GetCustomAttributes<SingletonAttribute>(),
-                MethodLevelFilters = method.GetCustomAttributes().OfType<IFunctionFilter>(),
-                ClassLevelFilters = method.DeclaringType.GetCustomAttributes().OfType<IFunctionFilter>()
+                SingletonAttributes = method.GetCustomAttributes<SingletonAttribute>().ToArray(),
+                MethodLevelFilters = method.GetCustomAttributes().OfType<IFunctionFilter>().ToArray(),
+                ClassLevelFilters = method.DeclaringType.GetCustomAttributes().OfType<IFunctionFilter>().ToArray()
```
brettsam, line 377: Is there a perf gain from enumerating these immediately?

src/Microsoft.Azure.WebJobs.Host/Executors/ShutdownFunctionExecutor.cs
```diff
@@ -19,8 +19,7 @@ public ShutdownFunctionExecutor(CancellationToken shutdownToken, IFunctionExecut
 
         public async Task<IDelayedException> TryExecuteAsync(IFunctionInstance instance, CancellationToken cancellationToken)
         {
-            using (CancellationTokenSource callCancellationSource = CancellationTokenSource.CreateLinkedTokenSource(
-                _shutdownToken, cancellationToken))
+            using (CancellationTokenSource callCancellationSource = CancellationTokenSource.CreateLinkedTokenSource(_shutdownToken, cancellationToken))
```
brettsam: Just formatting here?

src/Microsoft.Azure.WebJobs.Host/Loggers/FunctionInstanceLogger.cs
```diff
@@ -24,54 +24,79 @@ public FunctionInstanceLogger(ILoggerFactory loggerFactory)
...
+            private static readonly Action<ILogger, string, string, Guid, Exception> _functionStarted =
+                LoggerMessage.Define<string, string, Guid>(
+                    LogLevel.Information,
+                    new EventId(1, nameof(FunctionStarted)),
+                    "Executing '{name}' (Reason='{reason}', Id={id})");
+
+            private static readonly Action<ILogger, string, string, Guid, Exception> _functionCompleted =
+                LoggerMessage.Define<string, string, Guid>(
+                    LogLevel.Information,
+                    new EventId(2, nameof(FunctionStarted)),
```
brettsam: Should be `nameof(FunctionCompleted)`.

src/Microsoft.Azure.WebJobs.Host/Loggers/FunctionInstanceLogger.cs
```diff
@@ -24,54 +24,79 @@ public FunctionInstanceLogger(ILoggerFactory loggerFactory)
...
+            });
+        }
+
+        private static class Log
+        {
+            private static readonly Action<ILogger, string, string, Guid, Exception> _functionStarted =
+                LoggerMessage.Define<string, string, Guid>(
+                    LogLevel.Information,
+                    new EventId(1, nameof(FunctionStarted)),
+                    "Executing '{name}' (Reason='{reason}', Id={id})");
```
brettsam: I use `functionName` elsewhere when doing structured logging -- let's make that consistent with these. And maybe make it `invocationId`? These keys are available to loggers now so making them descriptive can be helpful.

src/Microsoft.Azure.WebJobs.Host/JobHost.cs
```diff
@@ -232,13 +232,14 @@ public async Task CallAsync(string name, IDictionary<string, object> arguments =
...
-
-            IDelayedException exception = null;
+            if (function == null)
+            {
+                throw new InvalidOperationException($"'{functionKey}' can't be invoked from Azure WebJobs SDK. Is it missing Azure WebJobs SDK attributes?");
+            }
 
-            exception = await _context.Executor.TryExecuteAsync(instance, cancellationToken);
+            var instance = CreateFunctionInstance(function, arguments);
+            var callCancellationSource = CancellationTokenSource.CreateLinkedTokenSource(_shutdownTokenSource.Token, cancellationToken);
```
brettsam: Did you just happen to notice that we weren't looking at the shutdown token here? Or was this an issue you hit?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -227,16 +227,9 @@ internal static async Task HandleExceptionAsync(TimeoutAttribute timeout, Except
...
-            TextWriter functionOutputTextWriter = null;
-
-            IFunctionOutputLogger outputLogger = _functionOutputLogger;
-            outputDefinition = await outputLogger.CreateAsync(instance, cancellationToken);
-            outputLog = outputDefinition.CreateOutput();
-            functionOutputTextWriter = outputLog.Output;
-            updateOutputLogTimer = StartOutputTimer(outputLog.UpdateCommand, _exceptionHandler);
+            var outputDefinition = await _functionOutputLogger.CreateAsync(instance, cancellationToken);
+            var outputLog = outputDefinition.CreateOutput();
+            var updateOutputLogTimer = StartOutputTimer(outputLog.UpdateCommand, _exceptionHandler);
```
brettsam, line 232: There's a lot of changes to `var` throughout the PR -- is that an auto-formatting thing you run or was there a perf reason for it?

src/Microsoft.Azure.WebJobs.Host/Loggers/FunctionInstanceLogger.cs
```diff
@@ -15,6 +14,7 @@ namespace Microsoft.Azure.WebJobs.Host.Loggers
 {
     internal class FunctionInstanceLogger : IFunctionInstanceLogger
     {
+        private readonly ConcurrentDictionary<string, ILogger> _cachedLoggers = new ConcurrentDictionary<string, ILogger>();
```
brettsam: We don't actually need this (I called it out in the other PR too) -- the logger factory already caches these so it's better just to use their implementation. I think some of the tests are failing because of this, too.

src/Microsoft.Azure.WebJobs.Host/Loggers/FunctionInstanceLogger.cs
```diff
@@ -24,54 +24,79 @@ public FunctionInstanceLogger(ILoggerFactory loggerFactory)
...
+        {
+            private static readonly Action<ILogger, string, string, Guid, Exception> _functionStarted =
+                LoggerMessage.Define<string, string, Guid>(
+                    LogLevel.Information,
+                    new EventId(1, nameof(FunctionStarted)),
+                    "Executing '{name}' (Reason='{reason}', Id={id})");
+
+            private static readonly Action<ILogger, string, string, Guid, Exception> _functionCompleted =
+                LoggerMessage.Define<string, string, Guid>(
+                    LogLevel.Information,
```
brettsam, line 82: If this has failed we want this to be logged as Error. So I think we'll need two templates -- one for success and one for failure. I think they should look exactly the same except for the log level.

src/Microsoft.Azure.WebJobs.Host/Loggers/FunctionInstanceLogger.cs
```diff
@@ -24,54 +24,79 @@ public FunctionInstanceLogger(ILoggerFactory loggerFactory)
...
+                    new EventId(2, nameof(FunctionStarted)),
+                    "Executed '{name}' ({status}, Id={id})");
```
brettsam: Make these `functionName`, `functionResult`, and `invocationId`? Now that we pass the function result as structured data here, we could pull it out into a separate column for the SystemLogger as well for easier querying. @sidkri is already looking into this for adding Duration to this message.

brettsam (approved):
Good stuff! A couple of changes requested on the new logging piece (thank you!) and a couple of general questions as well. But once done, it's good to get in.

src/Microsoft.Azure.WebJobs.Host/Executors/JobHostContextFactory.cs
mathewc: Initially there were other changed in this file which I ended up scaling back. Will revert.

src/Microsoft.Azure.WebJobs.Host/Executors/ShutdownFunctionExecutor.cs
mathewc: Same as above. Will revert

src/Microsoft.Azure.WebJobs.Host/JobHost.cs
mathewc: I was investigating https://github.com/Azure/azure-webjobs-sdk/issues/2420 and other recent reports while doing this work and I think I made an inadvertent change. I've reverted this.

src/Microsoft.Azure.WebJobs.Host/Loggers/FunctionInstanceLogger.cs
mathewc: Done

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
mathewc: Just code readability. As part of this work I was examining the entire pipeline and making readability tweaks as I went.

src/Microsoft.Azure.WebJobs.Host/Loggers/FunctionInstanceLogger.cs
mathewc: Fixed

src/Microsoft.Azure.WebJobs.Host/Loggers/FunctionInstanceLogger.cs
mathewc: Fixed

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
mathewc: Likely not a big one, but just like we did already for TimeoutAttribute, since this metadata is static once indexed, I'm just setting it up front. Don't need to requery for it on each invocation.

src/Microsoft.Azure.WebJobs.Host/Loggers/FunctionInstanceLogger.cs
mathewc: Removed

PR #2482, opened 2020-04-29
[Storage] disposing request upon timeout

brettsam:
Trying a different approach to the deadlock by disposing the in-flight request if we hit the timeout limit.


brettsam:
Closing in favor of another approach.

PR #2484, opened 2020-04-30
[Storage] adding connection limit for Consumption function plan

brettsam:


src/Microsoft.Azure.WebJobs.Extensions.Storage/StorageWebJobsBuilderExtensions.cs
```diff
@@ -45,6 +46,11 @@ public static IWebJobsBuilder AddAzureStorage(this IWebJobsBuilder builder, Acti
 
             builder.Services.TryAddSingleton<StorageAccountProvider>();
 
+            if (CommonUtility.IsDynamicSku)
+            {
```
fabiocav: Is there a default service implementation that is used on non-dynamic? `StorageAccountProvider` has a required dependency on it, correct?

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AzureStorageEndToEndTests.cs
```diff
@@ -399,7 +397,7 @@ public async Task BadQueueMessageE2ETests()
                         // The message is in the second queue
                         var queue2 = queueClient.GetQueueReference(_resolver.ResolveInString(BadMessageQueue2));
 
-                        Storage.StorageException ex = await Assert.ThrowsAsync<Storage.StorageException>(
+                        Azure.Storage.StorageException ex = await Assert.ThrowsAsync<Azure.Storage.StorageException>(
```
fabiocav, line 400: Should this namespace using be kept?

src/Microsoft.Azure.WebJobs.Extensions.Storage/StorageWebJobsBuilderExtensions.cs
brettsam: No.... but there should be!

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AzureStorageEndToEndTests.cs
brettsam: Yeah -- my new classes and interfaces use the `Microsoft.Azure.WebJobs.Host.Storage` namespace (which... surprisingly, nothing else uses). So this needed further disambiguation as it was resolving to our namespace rather than Storage's.

The namespaces in the Host.Storage project are all over the place... and none of them actually match the project name. Some use "...Extensions.Storage", some just use "Microsoft.Azure.WebJobs.Host" and some start with "WebJobs...". I thought I should make these use the "right" namespace since there's no consistency today...

fabiocav (approved):

PR #2486, opened 2020-05-04
[Storage] updating to 11.1.4; handling storage network timeouts

brettsam:
Storage 11.1.4 now cancels network timeouts after 100 seconds by default: https://github.com/Azure/azure-storage-net/pull/985 

Our TimeoutHandler workaround is likely not needed at all anymore, however, I wanted to leave it in just to make sure. So I changed this to:

1. Increase the timeout of our TimeoutHandler from 2 minutes to 3 minutes.
2. Check and log if we get a network timeout to make sure it's behaving as we expect (and also for the added data).

If things go well and we never see the TimeoutHandler fire anymore, we can remove it completely and simplify these calls in a future release.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.Storage/TimeoutHandler.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.Storage/TimeoutHandler.cs
@@ -2,21 +2,27 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Diagnostics;
 using System.Runtime.ExceptionServices;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Azure.Storage;
 using Microsoft.Azure.WebJobs.Host.Timers;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.WebJobs.Extensions.Storage
 {
     internal static class TimeoutHandler
     {
-        private static readonly TimeSpan DefaultTimeout = TimeSpan.FromMinutes(2);
+        private static readonly TimeSpan DefaultTimeout = TimeSpan.FromMinutes(3);
 
-        public static async Task<T> ExecuteWithTimeout<T>(string operationName, string clientRequestId, IWebJobsExceptionHandler exceptionHandler, Func<Task<T>> operation)
+        public static async Task<T> ExecuteWithTimeout<T>(string operationName, string clientRequestId,
+            IWebJobsExceptionHandler exceptionHandler, ILogger logger, CancellationToken cancellationToken, Func<Task<T>> operation)
         {
             using (var cts = new CancellationTokenSource())
             {
+                Stopwatch sw = Stopwatch.StartNew();
+
                 Task timeoutTask = Task.Delay(DefaultTimeout, cts.Token);
                 Task<T> operationTask = operation();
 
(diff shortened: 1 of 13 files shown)


src/Microsoft.Azure.WebJobs.Extensions.Storage/TimeoutHandler.cs
```diff
@@ -42,7 +48,30 @@ public static async Task<T> ExecuteWithTimeout<T>(string operationName, string c
...
+                try
+                {
+                    // Determine if this was a deadlock. If so, log it and rethrow. Use the passed-in cancellationToken
+                    // to detemrine of this operation was canceled explicitly or was due to an internal network timeout.
+
+                    return await operationTask;
+                }
+                catch (StorageException ex) when (ex.InnerException is OperationCanceledException && !cancellationToken.IsCancellationRequested) // network timeout
+                {
+                    Logger.StorageTimeout(logger, operationName, clientRequestId, sw.ElapsedMilliseconds);
```
brettsam, line 60: Most of the other changes were to allow me to plumb a logger through to here so all of this was centralized.

src/Microsoft.Azure.WebJobs.Extensions.Storage/TimeoutHandler.cs
```diff
@@ -2,21 +2,27 @@
...
 namespace Microsoft.Azure.WebJobs.Extensions.Storage
 {
     internal static class TimeoutHandler
     {
-        private static readonly TimeSpan DefaultTimeout = TimeSpan.FromMinutes(2);
+        private static readonly TimeSpan DefaultTimeout = TimeSpan.FromMinutes(3);
 
-        public static async Task<T> ExecuteWithTimeout<T>(string operationName, string clientRequestId, IWebJobsExceptionHandler exceptionHandler, Func<Task<T>> operation)
+        public static async Task<T> ExecuteWithTimeout<T>(string operationName, string clientRequestId,
+            IWebJobsExceptionHandler exceptionHandler, ILogger logger, CancellationToken cancellationToken, Func<Task<T>> operation)
```
brettsam, line 20: This is kind of clunky, but this `cancellationToken` isn't used to cancel anything. But in order to determine if it was a network timeout (for logging only), I need to know if someone canceled the operation directly. Since this is hopefully a short-lived API, I'm okay with it being odd for now and this whole class will go away in a future update.

src/Microsoft.Azure.WebJobs.Host/Timers/TaskSeriesTimer.cs
```diff
@@ -147,9 +147,9 @@ private async Task RunAsync(CancellationToken cancellationToken)
                         TaskSeriesCommandResult result = await _command.ExecuteAsync(cancellationToken);
                         wait = result.Wait;
                     }
-                    catch (Exception ex) when (ex.InnerException is TaskCanceledException)
+                    catch (Exception ex) when (ex.InnerException is OperationCanceledException)
```
brettsam, line 150: Took me a while to realize that moving this to `OperationCanceledException` did everything I needed. The new behavior throws an `OperationCanceledException` wrapped in a `StorageException` if the network hangs for 100 seconds. Now, that exception will bubble up to here and the loop will start over, just as we'd want it to. (and it behaves the same as before as `TaskCanceledException : OperationCanceledException`).

paulbatum (approved):
Looks like there are some host changes here. I assume you are OK with updating both at the same time. LGTM.

brettsam:
Thanks! Yeah those host changes are fine. That TaskSeriesTimer is source-shared across a couple projects so they'll all get that update internally (which is how I can ship this in the extension without needing a host deployment).

Issue #2488, opened 2020-05-12
All kinds of convenience methods for creating bindings in FluentBindingRule should have an overload that accepts a factory that can process a ValueBindingContext


macrogreg:
The `FluentBindingRule` class contains a number of `BindToXxx` methods that represent convenience APIs for creating attribute bindings for function parameters and for return values.

Two of such methods (`BindToInput(..)` and `BindToStream(..)`) have overloads that allow the value factory to inspect the `ValueBindingContext` while creating the value bound by an attribute. This enables a number of scenarios: For example, it provides access to the name of the function being invoked, and to the id of the particular invocation instance. These values are otherwise not accessible.

Unfortunately, other `BindToXxx` methods do not such overloads (`BindToCollector(..)` and `BindToValueProvider(..)`). 
This proposal is to add corresponding overloads.

We currently require the respective `BindToCollector(..)` overload for scenarios in the [Availability Monitoring Extension](https://github.com/Azure/azure-functions-availability-monitoring-extension). The `BindToValueProvider(..)` overload should be added for consistency. 

A candidate PR is readily available (https://github.com/Azure/azure-webjobs-sdk/pull/2489).

macrogreg:
Heads-up @mathewc : This is the change we discussed previously.
Also heads-up @brettsam FYI.

zmarty:
@macrogreg @mathewc  Any idea when this can be checked in and shipped? We would like to use ValueBindingContext inside BindToCollector for an internal MSFT project.

zmarty:
@fabiocav We still need this in an internal Microsoft project :) Thanks!

zmarty:
@fabiocav  We still need this :)

mathewc:
@zmarty Sorry this got lost in the shuffle. Assuming you guys still need these overloads so you can remove the temp reflection hack I put together for Greg?

zmarty:
Hi @mathewc. Yes, we definitely still need this. Also, out of curiosity can you point me to this: "temp reflection hack I put together for Greg"? You can send me an internal mail if you wish, just look for my name.

v-anvari:
Hi @zmarty , Can you confirm if this still requires a triage/ follow up or can this issue be closed.

zmarty:
> Hi @zmarty , Can you confirm if this still requires a triage/ follow up or can this issue be closed.

Yes, we need this. No need to keep asking, the answer is **Yes**.

PR #2489, opened 2020-05-12
All kinds of convenience methods for creating bindings in FluentBindingRule should have an overload that accepts a factory that can process a ValueBindingContext

macrogreg:
This PR addresses https://github.com/Azure/azure-webjobs-sdk/issues/2488 - please see the description there.
There is is also minor code clean-up: Adding mission `#region` statements and fixing some doc comments in `FluentBindingRule`.


macrogreg:
I think this PR needs a full-ci label, but I may not have permissions to assign it.
@mathewc  / @brettsam ,  could you please tag this PR accordingly?

fabiocav:
@mathewc are you planning on looking at this? Would it help to have it assigned to a sprint/individual?

liliankasem:
@mathewc do we still wanna review?

mathewc:
Based on the last comment from @zmarty in issue https://github.com/Azure/azure-webjobs-sdk/issues/2488 this is still needed. I believe @macrogreg may no longer be working on this (correct me if I'm wrong!). The changes were needed for the Availability Monitoring extension but in that repo https://github.com/Azure/azure-functions-availability-monitoring-extension I don't see any source code?

Work remaining in my mind for this would be to add tests for the new proposed overloads, and to test this out via a private nuget with the extension before committing. There was if I remember a hacky workaround they had to use, so we'd test that they can remove that and use the new overloads successfully.

Issue #2454, opened 2020-03-12
SingletonManager - No retry for blob lease renewal failure and eventually causes host crash 


yuvalpeled1:
We experience Azure Functions host crashes due to intermittent blob lease failures.

            public async Task<TaskSeriesCommandResult> ExecuteAsync(CancellationToken cancellationToken)
            {
                // Exceptions wil propagate 
                bool executionSucceeded = await _lockManager.RenewAsync(_lock, cancellationToken);

                TimeSpan delay = _speedupStrategy.GetNextDelay(executionSucceeded: true);
                return new TaskSeriesCommandResult(wait: Task.Delay(delay));
            }

You can see that the GetNextDelay always gets a '**true**' parameter even when the renew method failed. Therefore, the next retry will occur after the lease has expired [because the default retry interval is half of the time of the blob lease period].

#### Expected behavior

A proper retry will be applied if the code will be changed to:
                
`TimeSpan delay = _speedupStrategy.GetNextDelay(executionSucceeded);`

#### Actual behavior

After a single renewAsync failure the host crashes and the following log is emitted:

2020-03-08 11:54:20.7790944     Host.Singleton   2             Log: Singleton lock renewal failed for blob 'sqlfunctions-neu-prd/SqlFunctions.Functions.TiExportAsync.Listener' with error code 409: LeaseIdMismatchWithLeaseOperation. The last successful renewal completed at 2020-03-08T11:51:37.264Z (163514 milliseconds ago) with a duration of 3 milliseconds. The lease period was 60000 milliseconds.
2020-03-08 11:54:20.7837092     Host.Singleton   2             Log: Singleton lock renewal failed for blob 'sqlfunctions-neu-prd/SqlFunctions.Functions.TiIndicatorsStatisticsAsync.Listener' with error code 409: LeaseIdMismatchWithLeaseOperation. The last successful renewal completed at 2020-03-08T11:51:37.267Z (163516 milliseconds ago) with a duration of 6 milliseconds. The lease period was 60000 milliseconds.

Log: An unhandled exception has occurred. Host is shutting down., Exception: 
 
Exception Details
===================================
Exception Type: Microsoft.WindowsAzure.Storage.StorageException
Request Information: Microsoft.WindowsAzure.Storage.RequestResult
Target Site: Void MoveNext()
Message: The lease ID specified did not match the lease ID for the blob.
Help Link: 
Source: Microsoft.WindowsAzure.Storage
HResult: -2146233088
 
[...]

PR #2492, opened 2020-05-18
Singleton lease renewal fix (#2454)

mathewc:
Fix for https://github.com/Azure/azure-webjobs-sdk/issues/2454. @yuvalpeled1 is correct that this was regressed by [this commit ](https://github.com/Azure/azure-webjobs-sdk/commit/1d6e30fd719b85b0b0c6cf9e8a0b07e1e0a26bef?diff=split#diff-03975c56f40121be67c2ba5ff90cafd5).


yojagad (approved):

brettsam (commented):
Holy cow this is a great find. This may significantly reduce the number of 409s we see during lock renewal.

mathewc:
Yes it was an unfortunate regression. It will no doubt address MANY of the issues we've been seeing in this area.

PR #2501, opened 2020-05-28
Function execution log message now includes execution duration.

### LINKED ISSUE #5340: We no longer log the duration of a run for Kusto
When investigating customer issues we no longer include in the summary of the execution the duration of the execution.  Not sure when or why it disappeared.

sidkri:
#### Summary of changes
This change resolves Azure/azure-functions-host#5340 by adding the function execution duration to the log event written by the webjobs sdk each time a function execution completes (successfully or unsuccessfully).  In Functions v1, this duration was logged in the azure-functions-host code but that log line was removed from Functions v2 onwards, however a similar event is logged in the azure-webjobs-sdk code where I have added the duration.

This change also takes the opportunity to modify the placeholder names in the log message string in this file to Pascal case to meet the recommendation [here](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/loggermessage?view=aspnetcore-3.1).  The placeholder names show up as property names in the structured logs.

Sample log message before this change:
"Executed 'Functions.BlobTrigger' (Succeeded, Id=c3004a96-5aa3-4bbe-a31a-284623cae9e8)"

Sample log message after this change:
"Executed 'Functions.BlobTrigger' (Succeeded, Id=c3004a96-5aa3-4bbe-a31a-284623cae9e8**, Duration=596ms**)"

#### Validation
Code was tested locally and validated for both successful and unsuccessful function execution by checking console logs and App Insights logs.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Loggers/FunctionInstanceLogger.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Loggers/FunctionInstanceLogger.cs
@@ -50,9 +50,10 @@ namespace Microsoft.Azure.WebJobs.Host.Loggers
         {
             var logger = GetLogger(message.Function);
             Log.FunctionCompleted(
-                logger, 
-                message.Function.ShortName, 
+                logger,
+                message.Function.ShortName,
                 message.FunctionInstanceId,
+                message.EndTime.Subtract(message.StartTime),
                 message.Succeeded,
                 message.Failure);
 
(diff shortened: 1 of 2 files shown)


brettsam (approved):
looks good to me!

brettsam (approved):
Just one comment -- if you look at our structured logs elsewhere, we use camelCase. We should keep these consistent and also use camelCase. (ex: https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/BlobTriggerExecutor.Logger.cs)

sidkri:
> Just one comment -- if you look at our structured logs elsewhere, we use camelCase. We should keep these consistent and also use camelCase. (ex: https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/BlobTriggerExecutor.Logger.cs)

Reverted back to camelCase.

brettsam (approved):

PR #2505, opened 2020-06-03
Less allocatey logging for trigger details

Scooletz:
This PR reduces by half amount of memory allocated for logging the templatized trigger details.

### Bechmarking

For rough benchmarking purposes I simulated an Azure Queue trigger and its trigger details and used sharplab to verify the PR, see [this gist](https://gist.github.com/Scooletz/e7e9e91584e3123fee57b8a38770d6a7)

The amount of reduced memory footprint is quite high.

PR #2510, opened 2020-06-05
Allocating a lot less in ExceptionFormatter.GetFormattedException

Scooletz:
This PR reduces the amount of memory used by `ExceptionFormatter.GetFormattedException` by passing using all the way up/down a single `StringBuilder`. The amount of allocated memory shown by dotMemory have been reduced and for the following tests is following

|  Test name | Before (bytes)  | After (bytes)  |  
|---|---|---|
| FormatException_OutputsExpectedAsyncMethodParameters  | 294383 	  | 271439|
| FormatException_OutputsMethodParameters | 78014			  | 53774|
| FormatException_RemovesAsyncFrames  | 74614			  | 50450|
| FormatException_ResolvesAsyncMethodNames |94052			|69418|

This PR includes several commits that were put at the beginning and reverted at the end. 

The first one is related to `dotMemory` as I failed to find a benchmarking for this. I didn't want to blow it up out proportion so used locally dotMemory for finding the gaps.

The other commit is related to assertions for the formatted exceptions. I found the number of assertions no big enough to make me sure about the outcome, so I captured the initial output and used it through the whole process.

Even if this PR seems flaky, the implementation of `ExceptionFormatter.GetFormattedException` is good to be copied if/when the benchmarking and stronger assertions are added to the suite. Again, I'm not prescriptive, just trying to bring the reasoning and the work behind the PR

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Diagnostics/ExceptionFormatter.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Diagnostics/ExceptionFormatter.cs
@@ -209,7 +209,8 @@ namespace Microsoft.Azure.WebJobs.Host.Diagnostics
                 var end = fullName.LastIndexOf('>');
                 if (start >= 0 && end >= 0)
                 {
-                    stringBuilder.Append(fullName.Remove(start, 1).Substring(0, end - 1));
+                    stringBuilder.Append(fullName, 0, start);
+                    stringBuilder.Append(fullName, start + 1, end - start - 1);
                 }
                 else
                 {
--- a/src/Microsoft.Azure.WebJobs.Host/Diagnostics/ExceptionFormatter.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Diagnostics/ExceptionFormatter.cs
@@ -121,7 +121,7 @@ namespace Microsoft.Azure.WebJobs.Host.Diagnostics
             var method = frame.GetMethod();
             var declaringType = method?.DeclaringType;
             
-            if (ShouldShowFrame(declaringType) == false)
+            if (!ShouldShowFrame(declaringType))
             {
                 return;
             }
--- a/src/Microsoft.Azure.WebJobs.Host/Diagnostics/ExceptionFormatter.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Diagnostics/ExceptionFormatter.cs
@@ -27,34 +27,46 @@ namespace Microsoft.Azure.WebJobs.Host.Diagnostics
         /// <returns>A string representation of the exception with a sanitized stack trace.</returns>
         public static string GetFormattedException(Exception exception)
         {
+            var sb = new StringBuilder();
+            GetFormattedException(sb, exception);
+            return sb.ToString();
+        }
+
+        static void GetFormattedException(StringBuilder sb, Exception exception)
+        {
             try
             {
-                var aggregate = exception as AggregateException;
-                if (aggregate != null)
+                if (exception is AggregateException aggregate)
                 {
-                    return GetStackForAggregateException(exception, aggregate);
+                    GetStackForAggregateException(sb, exception, aggregate);
                 }
-
-                return GetStackForException(exception, includeMessageOnly: false);
+                else
+                {
+                    GetStackForException(sb, exception, includeMessageOnly: false);
+                }
             }
             catch (Exception)
             {
-                return exception?.ToString();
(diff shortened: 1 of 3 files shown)


Scooletz:
I think the failure might be related to https://github.com/Azure/azure-webjobs-sdk/pull/2505#issuecomment-638818228

Soft ping for @brettsam  😉

Scooletz:
After fixup with the tuple naming, the build fails with the error as below. Is it possible for any from the team to take a look and validate what is failing in here?

```
dotnet.exe : [xUnit.net 00:06:22.0680998]     Microsoft.Azure.WebJobs.Host.EndToEndTests.SingletonEndToEndTests.SingletonTriggerFunction_MultipleConcurrentInvocations_InvocationsAreSerialized [FAIL]
At C:\projects\azure-webjobs-sdk-rqm4t\runtests.ps1:15 char:3
+   & dotnet $cmd
+   ~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: ([xUnit.net 00:0...rialized [FAIL]:String) [], RemoteException
    + Full
```

Scooletz:
@brettsam Sorry for pinging you, but if you or anyone help me with the https://github.com/Azure/azure-webjobs-sdk/pull/2510#issuecomment-696613493 maybe I could move on and finalize this work :-)

src/Microsoft.Azure.WebJobs.Host/Diagnostics/ExceptionFormatter.cs
```diff
@@ -26,92 +26,106 @@ public static class ExceptionFormatter
...
+                return;
             }
 
-            var stackFrameLines = from frame in stackFrames
+            var toProcess = from frame in stackFrames
                                   let method = frame.GetMethod()
                                   let declaringType = method?.DeclaringType
                                   where ShouldShowFrame(declaringType)
-                                  select FormatFrame(method, declaringType, frame);
+                                  select (method: method, declaringType: declaringType, frame: frame);
```
brettsam: What if you kept FormatFrame here instead of creating the tuples and then iterating through them?

brettsam:
It may just be a flaky test -- I've re-run it to see. I also added one more question...

src/Microsoft.Azure.WebJobs.Host/Diagnostics/ExceptionFormatter.cs
Scooletz: Oh my 😅  I simplified it by pushing it down to `FormatFrame`

Scooletz:
@brettsam I amended it. And it's 🟢

src/Microsoft.Azure.WebJobs.Host/Diagnostics/ExceptionFormatter.cs
```diff
@@ -26,92 +26,108 @@ public static class ExceptionFormatter
...
             !(declaringType != null && typeof(INotifyCompletion).IsAssignableFrom(declaringType));
 
-        private static string FormatFrame(MethodBase method, Type declaringType, StackFrame frame)
+        private static void FormatFrame(StringBuilder stringBuilder, StackFrame frame)
         {
-            var stringBuilder = new StringBuilder();
+            var method = frame.GetMethod();
+            var declaringType = method?.DeclaringType;
+            
+            if (ShouldShowFrame(declaringType) == false)
```
brettsam: This is a really nitpicky comment :-) -- but can you make this `if (!ShouldShowFrame(declaringType))`?

Scooletz: Fixed in https://github.com/Azure/azure-webjobs-sdk/pull/2510/commits/4b6d6d71dbaf5894c1f17ac6598173ab5ef0e5fb

Scooletz:
I've fixed the negation style and removed two more stringy allocs in https://github.com/Azure/azure-webjobs-sdk/pull/2510/commits/ccecaae49f97a78da0affc78c895d429181c55e7

I thinks this is it for this PR. I'd propose some kind of caching for all the calculated values that are based on `MethodInfo` and `Type` but this is a separate PR I think. I'll leave this one as an alloc-related only, without caching applied.

Soooo, are we ok with this @brettsam ? 😄

brettsam (approved):

brettsam:
Thanks @Scooletz!

Scooletz:
👍 @brettsam

PR #2512, opened 2020-06-08
Porting disable/enable functions using configuration

alrod:


src/Microsoft.Azure.WebJobs.Host/Extensions/IJobHostMetadataProvider.cs
```diff
@@ -61,5 +61,12 @@ public interface IJobHostMetadataProvider
         /// </summary>
         /// <param name="output">The <see cref="TextWriter"/> to write the graph to.</param>
         void DebugDumpGraph(TextWriter output);
+
+        /// <summary>
+        /// Gets function metadata.
+        /// </summary>
+        /// <param name="functionName">Name of function to return metadata for.</param>
+        /// <returns></returns>
+        FunctionMetadata GetFunctionMetadata(string functionName);
```
mathewc: This unfortunately would be a breaking change to this public interface. Perhaps we don't need this though. The v1 functions host already has the settings logic for looking for these disable settings. That should be enough, meaning we don't have to flow this info via the metadata provider to the host.

src/Microsoft.Azure.WebJobs.Host/Listeners/HostListenerFactory.cs
```diff
@@ -64,6 +58,12 @@ public async Task<IListener> CreateAsync(CancellationToken cancellationToken)
                     continue;
                 }
 
+                IListenerFactory listenerFactory = functionDefinition.ListenerFactory;
```
mathewc, line 61: Why this move?

alrod: we want to log the disabled function if there is no listener

src/Microsoft.Azure.WebJobs.Host/ConfigurationUtility.cs
```diff
@@ -42,5 +42,20 @@ public static string GetConnectionFromConfigOrEnvironment(string connectionName)
 
             return Environment.GetEnvironmentVariable(connectionName) ?? configValue;
         }
+
+        public static bool IsSettingEnabled(string settingName)
```
mathewc, line 46: See Brett's PR https://github.com/Azure/azure-webjobs-sdk/pull/2513. This Type is going away

mathewc (approved):

PR #2513, opened 2020-06-09
removing ConfigurationUtility

### LINKED ISSUE #1841: Usage of ConfigurationUtility should be discontinued
The `ConfigurationUtility` class does not work with the configuration system and doesn't take the current configuration into consideration. Code using the type should be updated and the class should be removed.

brettsam:
Fixes #1841. 

See here for why this is more than just an internal issue: https://github.com/Azure/azure-webjobs-sdk/issues/1841#issuecomment-611719621


src/Microsoft.Azure.WebJobs.Host/Extensions/IConfigurationExtensions.cs
```diff
@@ -77,5 +78,20 @@ public static string GetPrefixedConnectionStringName(string connectionStringName
         /// <returns></returns>
         public static string GetConnectionStringOrSetting(this IConfiguration configuration, string connectionName) =>
             configuration.GetConnectionString(connectionName) ?? configuration[connectionName];
+
+        public static bool IsSettingEnabled(this IConfiguration configuration, string settingName)
```
brettsam, line 82: Copied this directly from the `ConfigurationUtilty` so the logic remains

mathewc (approved):

brettsam:
@mathewc -- removed a ConfigurationSource that we were no longer using (and were previously only using to check Disabled).

mathewc (approved):

PR #2519, opened 2020-06-16
QueueListener now uses cancellable delays

Scooletz:
This PR changes `QueueListener` to use the overload of `Task.Delay` with a `CancellationToken` support. This ensures that for any scenario when a queue is notified, it'll cancel an unneeded delay.

Additionally, two fields are made volatile as I found no strict rule of accessing them and making sure that eventually last value is visible.


Scooletz:
I think I might have another take on it. Closing for now.

PR #2526, opened 2020-06-24
Implementing drainmode for function host

soninaren:
When the host is put in drain mode, It should:
- Stop watching / listening for new traffic. 
- Call cancel on the cancellation token passed as a parameter to the function invocation

The idea behind implementing this mode is that Antares would put the host in drain mode. Wait a certain amount of time for the existing / pending invocations to complete. Finally perform a scale-in operation.

Currently scale in operation happens arbitrarily causing the existing function host to lose in-flight and pending invocations.

**Pending tests**


soninaren:
closes #2527

src/Microsoft.Azure.WebJobs.Host/Hosting/DrainModeManager.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    internal class DrainModeManager : IDrainModeManager
+    {
+        private readonly ILoggerFactory _loggerFactory;
+        private readonly ILogger _logger;
+
+        public DrainModeManager(ILoggerFactory loggerFactory = null)
```
brettsam: Use an `ILoggerFactory<DrainModeManager>` here.

src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
```diff
@@ -138,7 +138,11 @@ public async Task StopAsync(CancellationToken cancellationToken)
                 {
                     if (_started)
                     {
+                        _logger?.LogInformation($"Stopping the listener for function '{_descriptor.ShortName}'");
```
brettsam: Could you write out the type of the Listener here as well? That could be helpful for debugging.

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Constants/LogCategories.cs
```diff
@@ -44,6 +44,11 @@ public static class LogCategories
         /// </summary>
         public const string Bindings = "Host.Bindings";
 
+        /// <summary>
+        /// The category for all logs written by the function host during drain mode operations.
+        /// </summary>
+        public const string Drain = "Host.Drain";
```
brettsam: You won't need this once you use `ILogger<DrainModeManager>`

src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
```diff
@@ -138,7 +138,11 @@ public async Task StopAsync(CancellationToken cancellationToken)
...
+
                         await _listener.StopAsync(cancellationToken);
+
+                        _logger?.LogInformation($"Stopped the listener for function '{_descriptor.ShortName}'");
```
brettsam: use `_descriptor.LogName` here.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -236,6 +239,13 @@ internal static async Task HandleExceptionAsync(TimeoutAttribute timeout, Except
                 // Create a linked token source that will allow us to signal function cancellation
                 // (e.g. Based on TimeoutAttribute, etc.)                
                 CancellationTokenSource functionCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+
+                if (_drainModeManager.DrainModeEnabled)
+                {
+                    functionCancellationTokenSource.Cancel();
+                    logger?.LogInformation($"Requesting cancellation for function invocation:{instance.Id}");
+                }
```
brettsam, line 257: So this will cancel any new functions before they're invoked. But what about functions that are currently running? Do we need to cancel those as well?

soninaren: For currently running executions the data role / dwas would wait longer that the function timeout period for that SKU before scaling in. This would give currently running functions enough time to finish execution. So we shouldn't need to cancel currently running functions.

Build.ps1
```diff
@@ -5,7 +5,8 @@
 )
 
 if ($isLocal){
-  $packageSuffix = "dev" + [datetime]::UtcNow.Ticks.ToString()
+  $dstring = Get-Date -Format "ddHHmmss"
+  $packageSuffix = "dev" + $dstring
```
brettsam: I'm assuming this is b/c it's hard to mentally parse the Ticks string? :-) 

Can you add `yyMM` to the front of this just to make it clearer exactly when it's from?

src/Microsoft.Azure.WebJobs.Host/Hosting/DrainModeManager.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            _loggerFactory = loggerFactory;
+            _logger = _loggerFactory.CreateLogger<DrainModeManager>();
+        }
+
+        public bool DrainModeEnabled { get; private set; } = false;
```
brettsam: nit -- can you make this `IsDrainModeEnabled`?

src/Microsoft.Azure.WebJobs.Host/Hosting/IDrainModeManager.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    public interface IDrainModeManager
+    {
+        bool DrainModeEnabled { get; }
+
+        ICollection<IListener> Listeners { get; }
```
brettsam: Does Listeners need to be in the interface? I may have missed it -- but do we access this anywhere other than tests?

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
```diff
@@ -180,6 +186,8 @@ public async Task AsyncChainEndToEnd()
                     "}",
                 }.OrderBy(p => p).ToArray();
 
+
+
```
brettsam: nit -- extra newlines here...

brettsam (commented):
Looking good -- some nits and rename suggestions that I missed last time.

Build.ps1
soninaren: yes, Was hard to decipher a long number every time.

src/Microsoft.Azure.WebJobs.Host/Hosting/IDrainModeManager.cs
soninaren: Its only used in tests, have removed it from the interface

src/Microsoft.Azure.WebJobs.Host/Hosting/DrainModeManager.cs
soninaren: done

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
soninaren: done

soninaren:
@brettsam Have addressed the changes from the last iteration. Let me know if this is good to merge.

Build.ps1
fabiocav: :) shouldn't need to be human readable, with the update flow, this shouldn't be an issue.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -236,6 +239,13 @@ internal static async Task HandleExceptionAsync(TimeoutAttribute timeout, Except
                 // Create a linked token source that will allow us to signal function cancellation
                 // (e.g. Based on TimeoutAttribute, etc.)                
                 CancellationTokenSource functionCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+
+                if (_drainModeManager.IsDrainModeEnabled)
```
fabiocav: `_drainModeManager` could be `null` here

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
fabiocav: We can't rely on that infrastructure being present in standalone WebJobs scenarios. We need to ensure those are also signaled.

src/Microsoft.Azure.WebJobs.Host/Hosting/DrainModeManager.cs
fabiocav: Why do we need the factory? Can't you just use `ILogger<DrainModeManager>` here?

src/Microsoft.Azure.WebJobs.Host/Hosting/DrainModeManager.cs
```diff
@@ -0,0 +1,56 @@
...
+            {
+                IsDrainModeEnabled = true;
+                _logger.LogInformation($"DrainMode is set to {IsDrainModeEnabled}");
+
+                List<Task> tasks = new List<Task>();
+
+                _logger.LogInformation($"Calling StopAsync on the registered listeners");
+                foreach (IListener listener in Listeners)
+                {
+                    tasks.Add(listener.StopAsync(CancellationToken.None));
```
fabiocav: Why passing `CancellationToken.None`? The API should take a cancellation token and pass that into the calls to stop

src/Microsoft.Azure.WebJobs.Host/Hosting/DrainModeManager.cs
```diff
@@ -0,0 +1,56 @@
...
+        public async Task EnableDrainModeAsync()
+        {
+            if (!IsDrainModeEnabled)
+            {
+                IsDrainModeEnabled = true;
+                _logger.LogInformation($"DrainMode is set to {IsDrainModeEnabled}");
+
+                List<Task> tasks = new List<Task>();
+
+                _logger.LogInformation($"Calling StopAsync on the registered listeners");
```
fabiocav: This string and the one below don't need to use interpolation

src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
```diff
@@ -138,7 +138,11 @@ public async Task StopAsync(CancellationToken cancellationToken)
                 {
                     if (_started)
                     {
+                        _logger?.LogInformation($"Stopping the listener {_listener.GetType().FullName} for function '{_descriptor.LogName}'");
```
fabiocav: Let's at least cache the type name here. These calls tend to have a measurable impact.

src/Microsoft.Azure.WebJobs.Host/Listeners/HostListenerFactory.cs
```diff
@@ -83,7 +86,9 @@ public async Task<IListener> CreateAsync(CancellationToken cancellationToken)
 
             _listenersCreatedCallback?.Invoke();
 
-            return new CompositeListener(listeners);
+            var compositeListener = new CompositeListener(listeners);
+            _drainModeManager.RegisterListener(compositeListener);
```
fabiocav: This could be `null`

src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs
```diff
@@ -140,11 +140,11 @@ public async Task StopAsync(CancellationToken cancellationToken)
                     {
                         string listenerName = _listener.GetType().FullName;
                         string functionName = _descriptor.LogName;
-                        _logger?.LogInformation($"Stopping the {listenerName} listener for function '{functionName}'");
+                        _logger?.LogInformation($"Stopping the listener:{listenerName} for function:{functionName}");
 
                         await _listener.StopAsync(cancellationToken);
 
-                        _logger?.LogInformation($"Stopped the {listenerName} listener for function '{functionName}'");
+                        _logger?.LogInformation($"Stopped the listener:{listenerName} for function:{functionName}");
```
brettsam: Two things with this log message:
1. Use structured logging instead of string interpolation. The SystemLogger will find `functionName` and use it in the correct column, for example.
2. Instead of a colon, wrap the listenerName and functionName in `' '` -- it makes it easier to see where the name begins and ends. 

so...
` _logger?.LogInformation("Stopped the listener '{listenerName}' for function '{functionName}'", listenerName, functionName);`

src/Microsoft.Azure.WebJobs.Host/Hosting/DrainModeManager.cs
```diff
@@ -85,5 +69,26 @@ public async Task EnableDrainModeAsync(CancellationToken cancellationToken)
...
         }
+
+        public void CancelFunctionInvocations()
+        {
+            foreach (var keyValuePair in _cancellationTokenSources)
+            {
+                string invocationId = keyValuePair.Key.ToString();
+                try
+                {
+                    _logger?.LogInformation($"Requesting cancellation for function invocation:{invocationId}");
```
brettsam: Same comments as the other logging one -- and call the key `functionInvocationId`

Build.ps1
fabiocav: This is flagged as resolved, but shouldn't be done within the scope of this changes. Please revert and we can discuss.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -236,6 +242,17 @@ internal static async Task HandleExceptionAsync(TimeoutAttribute timeout, Except
...
                 CancellationTokenSource functionCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+
+                if (_drainModeManager?.IsDrainModeEnabled == true)
+                {
+                    functionCancellationTokenSource.Cancel();
+                    logger?.LogInformation($"Requesting cancellation for function invocation:{instance.Id}");
+                }
+                else
+                {
+                    _drainModeManager.RegisterTokenSource(instance.Id, functionCancellationTokenSource);
```
fabiocav: We still have a problem with the potential `null` exception here (if `null`, you'll fall into this block)

fabiocav: You might want to have a test with that to ensure this is covered.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -314,6 +331,7 @@ internal static async Task HandleExceptionAsync(TimeoutAttribute timeout, Except
                         exceptionInfo.Throw();
                     }
 
+                    _drainModeManager.UnRegisterTokenSource(instance.Id);
```
fabiocav: Potential `null` reference

src/Microsoft.Azure.WebJobs.Host/Hosting/DrainModeManager.cs
fabiocav: The comment was also marked as resolved but I don't see the change here. Can you please just update to use ILogger<T>

src/Microsoft.Azure.WebJobs.Host/Hosting/DrainModeManager.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            _cancellationTokenSources.RemoveIfContainsKey(guid);
+        }
+
+        public async Task EnableDrainModeAsync(CancellationToken cancellationToken)
+        {
+            if (!IsDrainModeEnabled)
+            {
+                IsDrainModeEnabled = true;
+                _logger.LogInformation($"DrainMode is set to {IsDrainModeEnabled}");
```
fabiocav: Won't this always be true? We don't need the interpolation here. Changing the message to `Drain mode enabled.` is a better approach.

src/Microsoft.Azure.WebJobs.Host/Hosting/DrainModeManager.cs
```diff
@@ -0,0 +1,94 @@
...
+                    _logger?.LogInformation("Requesting cancellation for function invocation '{invocationId}'", invocationId);
+                    keyValuePair.Value.Cancel();
+                }
+                catch (ObjectDisposedException)
+                {
+                    _logger?.LogInformation("Cancellation token for function invocation '{invocationId}' already disposed. No action required", invocationId);
+                }
+                catch (Exception e)
+                {
+                    _logger?.LogInformation("Exception occured when attempting to request cancellation for function invocation '{invocationId}', '{e.GetType().Name}:{e.Message}'", invocationId, e.GetType().Name, e.Message);
```
fabiocav: Why not log error and pass the exception here?

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
```diff
@@ -98,7 +98,7 @@ public async Task AsyncChainEndToEnd()
             using (_functionCompletedEvent = new ManualResetEvent(initialState: false))
             {
                 TestLoggerProvider loggerProvider = await AsyncChainEndToEndInternal(_hostBuilder);
-
+
```
fabiocav: nit: empty space

src/Microsoft.Azure.WebJobs.Host/Hosting/DrainModeManager.cs
soninaren: I remember making that change. I am not sure where i missed the commit. Will make the change again.

src/Microsoft.Azure.WebJobs.Host/Hosting/DrainModeManager.cs
soninaren: pass the exception where, Do you mean throw the exception?

fabiocav: Pass into the LogError call.

Build.ps1
soninaren: This is now reverted

src/Microsoft.Azure.WebJobs.Host/Hosting/DrainModeManager.cs
soninaren: Ok. this is now done in the new commit.
https://github.com/Azure/azure-webjobs-sdk/pull/2526/files#diff-ca6075feb78589c3cff858c1176209c0R18-R27

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
soninaren: using null-conditional-operators now

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
soninaren: Have added a test. Found an issue as a result and have fixed it.

src/Microsoft.Azure.WebJobs.Host/Hosting/DrainModeManager.cs
```diff
@@ -0,0 +1,92 @@
...
+        {
+            _cancellationTokenSources.RemoveIfContainsKey(guid);
+        }
+
+        public async Task EnableDrainModeAsync(CancellationToken cancellationToken)
+        {
+            if (!IsDrainModeEnabled)
+            {
+                IsDrainModeEnabled = true;
+                _logger.LogInformation($"DrainMode mode enabled");
```
fabiocav: nit: interpolation no longer required.

fabiocav (approved):
Small nit comment, if you're rebasing, might be good to just address that, but no need to block.

src/Microsoft.Azure.WebJobs.Host/Hosting/DrainModeManager.cs
soninaren: fixed

PR #2550, opened 2020-08-16
More performance improvements

mathewc:
See associated Functions Host PR: https://github.com/Azure/azure-functions-host/pull/6513


src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -79,94 +70,52 @@ public HostOutputMessage HostOutputMessage
...
         }
 
-        public async Task<IDelayedException> TryExecuteAsync(IFunctionInstance instance, CancellationToken cancellationToken)
+        public async Task<IDelayedException> TryExecuteAsync(IFunctionInstance functionInstance, CancellationToken cancellationToken)
         {
-            bool ownsInstance = false;
-
-            if (!(instance is IFunctionInstanceEx functionInstance))
-            {
-                functionInstance = new FunctionInstanceWrapper(instance, _serviceScopeFactory);
```
mathewc, line 88: Removed this wrapper type completely. One less object creation on the invoke path.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -79,94 +70,52 @@ public HostOutputMessage HostOutputMessage
...
                     parameterHelper.Initialize();
-                    functionStartedMessageId = await ExecuteWithLoggingAsync(functionInstance, functionStartedMessage, instanceLogEntry, parameterHelper, logger, cancellationToken);
-                    functionCompletedMessage = CreateCompletedMessage(functionStartedMessage);
+                    functionStartedMessageId = await ExecuteWithLoggingAsync(functionInstanceEx, functionStartedMessage, instanceLogEntry, parameterHelper, logger, cancellationToken);
                 }
                 catch (Exception exception)
                 {
-                    if (functionCompletedMessage == null)
-                    {
-                        functionCompletedMessage = CreateCompletedMessage(functionStartedMessage);
```
mathewc, line 130: Removed this object creation, which was just a copy of all the data that was already in the started message. A FunctionCompletedMessage "isa" FunctionStartedMessage, so above when creating the started message I just create an instance of the latter.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -258,9 +207,6 @@ internal static async Task HandleExceptionAsync(TimeoutAttribute timeout, Except
 
                 using (functionCancellationTokenSource)
                 {
-                    // This allows the FunctionOutputLogger to grab the TextWriter created by the IFunctionOutput. This enables user ILogger
```
mathewc, line 261: Just moved this comment into the method for readability

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -276,29 +222,26 @@ internal static async Task HandleExceptionAsync(TimeoutAttribute timeout, Except
                         await parameterHelper.BindAsync(instance.BindingSource, valueBindingContext);
                     }
 
-                    Exception invocationException = null;
-                    ExceptionDispatchInfo exceptionInfo = null;
+                    // Log started events
                     string startedMessageId = null;
-                    using (parameterHelper)
```
mathewc, line 282: There's already a using statement for the parameter helper earlier in the call stack

src/Microsoft.Azure.WebJobs.Host/Executors/IMethodInvoker.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Microsoft.Azure.WebJobs.Host.Executors
 {
-    internal interface IMethodInvoker<TReflected, TReturnValue>
+    public interface IMethodInvoker<TReflected, TReturnValue>
```
mathewc, line 8: Exposing the method invoker factory publically so Functions Host can use compiled lambdas rather than reflection to invoke methods, just like the SDK does. See the associated Host PR

src/Microsoft.Azure.WebJobs.Host/Executors/TaskMethodInvoker.cs
```diff
@@ -18,55 +17,7 @@ public TaskMethodInvoker(Func<TReflected, object[], Task<TReturnType>> lambda)
 
         public Task<TReturnType> InvokeAsync(TReflected instance, object[] arguments)
         {
-            Task<TReturnType> task = _lambda.Invoke(instance, arguments);
-            ThrowIfWrappedTaskInstance(task);
-            return task;
-        }
-
-        private static void ThrowIfWrappedTaskInstance(Task task)
```
mathewc, line 26: If someone were to return a nested task, it just won't work like they expect. I don't think checking for this on every invocation is warranted.

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndex.cs
```diff
@@ -12,65 +12,62 @@ internal class FunctionIndex : IFunctionIndex, IFunctionIndexCollector
...
             }
-
-            return _functionsById[functionId];
+            return function;
         }
 
         public IFunctionDefinition LookupByName(string name)
         {
-            // For compat, accept either the short name ("Class.Name") or log name (just "Name")
-            foreach (var items in _functionDescriptors)
```
mathewc, line 52: This is the overload used by Functions Host, and it resulted in a linear lookup each time.

paulbatum (changes requested):
Can you please break this up into separate PRs? It looks like there are several separate changes here.

PR #2553, opened 2020-08-18
Fix client information population of ApplicationInsights telemetry

### LINKED ISSUE #2552: Client location information is not populated on ApplicationInsights telemetry
Client Ip address is not captured by ApplicationInsights SDK and client location information is not populated on the telemetry inside Azure Functions, but populated when ApplicationInsights SDK is user in ASP.NET or ASP.NET Core service.
Client IPs are protected GDPR so they do not appear by default on the portal, instead Application Inisghts ingestion service populates client location (city, state, country).

#### Repro steps

1. Create HTTP-trigger Function, enable application insights and deploy it

2. Call Function

3. Check client location on the telemetry

#### Expected behavior

All telemetry reported in the scope of the user request should have client location populated (city, state, country)

#### Actual behavior

Telemetry does not have a client location.

#### Known workarounds

- Configure [Functions.Startup](https://docs.microsoft.com/en-us/azure/azure-functions/functions-dotnet-dependency-injection)
- Add Microsoft.Azure.WebJobs.Logging.ApplicationInsights nuget package 
- Copy [ClientIpHeaderTelemetryInitializerCopy](https://gist.github.com/lmolkova/b5fe5a1a48144f0caf95addcb6f86be6)
- Add this to startup:
 
```csharp
    {
        {
            var clientIpInitializerAdded = builder.Services.Any(i =>
                i.ServiceType == typeof(ITelemetryInitializer) && i.ImplementationType == typeof(ClientIpHeaderTelemetryInitializer));

            if (!clientIpInitializerAdded)
            {
                builder.Services.AddSingleton<ITelemetryInitializer, ClientIpHeaderTelemetryInitializerCopy>();
            }
        }
    }
```
 
#### Related information 

`ClientIpHeaderTelemetryInitializer` needs to be added in DI container by WebJobs: 

https://github.com/Azure/azure-webjobs-sdk/blob/1052ca70db52ff54261c33428f082c25b24c9b16/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs#L54-L58

WebJobs should **not** set IPs to zero

https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs#L50-L60

* Package version

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>netcoreapp3.1</TargetFramework>
    <AzureFunctionsVersion>v3</AzureFunctionsVersion>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.Azure.Functions.Extensions" Version="1.0.0" />
    <PackageReference Include="Microsoft.Azure.WebJobs.Logging.ApplicationInsights" Version="3.0.18" />
    <PackageReference Include="Microsoft.NET.Sdk.Functions" Version="3.0.3" />
  </ItemGroup>
</Project>
```

* Links to source

lmolkova:
Fix #2552

This PR adds default ApplicaitonInsights way to populate client IP. Because of GDPR, client IPs are removed at ingestion endpoint and instead location info is populated.

The default behavior (available for ASP.NET  and ASP.NET Core apps) populates client Ip from  `X-forwarded-For` (first value) on request telemetry and all nested telemetry items.

Functions/WebJobs behaves differently: webjobs parse `X-Forwarded-For` and falling back to `HttpContext.Connection.RemoteIpAddress` and set it on request only. WebJobs also set ip to 0.0.0.0 on all other items.

This PR enables default AppInsights way to populate clientIp by adding TelemetryInitializer. 
1. It relies on presence of `IHttpContextAccessor` enabled by [functions host](https://github.com/Azure/azure-functions-host/blob/826c0b520c1e81d2f3106d08711c08f1bce4aa95/src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs#L67)
   - it uses this mechanism only if extended HTTP trigger info is configured (`HttpAutoCollectionOptions.EnableHttpTriggerExtendedInfoCollection`, true by default)
2. for the case when `EnableHttpTriggerExtendedInfoCollection` is false, we keep the old behavior (use  `X-Forwarded-For` and fall back to `HttpContext.Connection.RemoteIpAddress` and set it on request only. WebJobs also set ip to 0.0.0.0 on all other items)

I'm also happy to change old behavior. I'm not sure what were the reasons to set IPs to `0.0.0.0` and how useful it is to fallback  `HttpContext.Connection.RemoteIpAddress`.  
@brettsam what do you think?

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
@@ -355,7 +355,7 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
         /// </summary>
         /// <param name="state"></param>
         /// <param name="scope"></param>
-        private static void ApplyFunctionResultActivityTags(IEnumerable<KeyValuePair<string, object>> state, IDictionary<string, object> scope)
+        private void ApplyFunctionResultActivityTags(IEnumerable<KeyValuePair<string, object>> state, IDictionary<string, object> scope)
         {
             // Activity carries tracing context. It is managed by instrumented library (e.g. ServiceBus or Asp.Net Core)
             // and consumed by ApplicationInsights.
@@ -416,7 +416,9 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
                     currentActivity.AddTag(LogConstants.LogLevelKey, logLevel.ToString());
                 }
 
-                if (scope.TryGetValue(ApplicationInsightsScopeKeys.HttpRequest, out var request) && request is HttpRequest httpRequest)
+                if (!_loggerOptions.HttpAutoCollectionOptions.EnableHttpTriggerExtendedInfoCollection && 
+                    scope.TryGetValue(ApplicationInsightsScopeKeys.HttpRequest, out var request) &&
+                    request is HttpRequest httpRequest)
                 {
                     currentActivity.AddTag(LoggingConstants.ClientIpKey, GetIpAddress(httpRequest));
                 }
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using Microsoft.ApplicationInsights;
 using Microsoft.ApplicationInsights.AspNetCore;
 using Microsoft.ApplicationInsights.AspNetCore.Extensions;
+using Microsoft.ApplicationInsights.AspNetCore.TelemetryInitializers;
 using Microsoft.ApplicationInsights.Channel;
 using Microsoft.ApplicationInsights.DependencyCollector;
 using Microsoft.ApplicationInsights.Extensibility;
@@ -18,6 +19,7 @@ using Microsoft.ApplicationInsights.Extensibility.PerfCounterCollector.QuickPuls
 using Microsoft.ApplicationInsights.SnapshotCollector;
 using Microsoft.ApplicationInsights.WindowsServer;
 using Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel;
+using Microsoft.AspNetCore.Http;
 using Microsoft.Azure.WebJobs.Logging.ApplicationInsights;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection.Extensions;
@@ -52,6 +54,21 @@ namespace Microsoft.Extensions.DependencyInjection
                 });
 
             services.AddSingleton<ITelemetryInitializer, HttpDependenciesParsingTelemetryInitializer>();
+            services.AddSingleton<ITelemetryInitializer>(provider =>
+            {
+                ApplicationInsightsLoggerOptions options = provider.GetService<IOptions<ApplicationInsightsLoggerOptions>>().Value;
+                if (options.HttpAutoCollectionOptions.EnableHttpTriggerExtendedInfoCollection)
+                {
+                    var httpContextAccessor = provider.GetService<IHttpContextAccessor>();
+                    if (httpContextAccessor != null)
+                    {
+                        return new ClientIpHeaderTelemetryInitializer(httpContextAccessor);
+                    }
+                }
+
+                return NullTelemetryInitializer.Instance;
+            });
+
             services.AddSingleton<ITelemetryInitializer, WebJobsRoleEnvironmentTelemetryInitializer>();
             services.AddSingleton<ITelemetryInitializer, WebJobsSanitizingInitializer>();
             services.AddSingleton<ITelemetryInitializer, WebJobsTelemetryInitializer>();
@@ -262,7 +279,10 @@ namespace Microsoft.Extensions.DependencyInjection
 
             foreach (ITelemetryInitializer initializer in telemetryInitializers)
             {
-                configuration.TelemetryInitializers.Add(initializer);
+                if (!(initializer is NullTelemetryInitializer))
+                {
+                    configuration.TelemetryInitializers.Add(initializer);
+                }
             }
 
             (channel as ServerTelemetryChannel)?.Initialize(configuration);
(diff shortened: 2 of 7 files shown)


src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
```diff
@@ -52,10 +54,23 @@ public static IServiceCollection AddApplicationInsights(this IServiceCollection
...
+
+                return NullTelemetryInitializer.Instance;
+            });
+
             services.AddSingleton<ITelemetryInitializer, WebJobsRoleEnvironmentTelemetryInitializer>();
             services.AddSingleton<ITelemetryInitializer, WebJobsSanitizingInitializer>();
             services.AddSingleton<ITelemetryInitializer, WebJobsTelemetryInitializer>();
             services.AddSingleton<ITelemetryInitializer, MetricSdkVersionTelemetryInitializer>();
+
+
```
brettsam: nit -- extra line here?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
```diff
@@ -52,10 +54,23 @@ public static IServiceCollection AddApplicationInsights(this IServiceCollection
                 });
 
             services.AddSingleton<ITelemetryInitializer, HttpDependenciesParsingTelemetryInitializer>();
+            services.AddSingleton<ITelemetryInitializer>(provider =>
+            {
+                var httpContextAccessor = provider.GetService<IHttpContextAccessor>();
```
brettsam: If `EnableHttpTriggerExtendedInfoCollection` is false, can we skip this initializer as well?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -416,7 +416,9 @@ private static void ApplyFunctionResultActivityTags(IEnumerable<KeyValuePair<str
                     currentActivity.AddTag(LogConstants.LogLevelKey, logLevel.ToString());
                 }
 
-                if (scope.TryGetValue(ApplicationInsightsScopeKeys.HttpRequest, out var request) && request is HttpRequest httpRequest)
+                if (!_loggerOptions.HttpAutoCollectionOptions.EnableHttpTriggerExtendedInfoCollection && 
+                    scope.TryGetValue(ApplicationInsightsScopeKeys.HttpRequest, out var request) &&
+                    request is HttpRequest httpRequest)
                 {
                     currentActivity.AddTag(LoggingConstants.ClientIpKey, GetIpAddress(httpRequest));
```
brettsam, line 423: So is `GetIpAddress()` just not working anymore?

lmolkova: it's only working if `EnableHttpTriggerExtendedInfoCollection` is disabled (explicitly) - in this case AppInsights won't be able to collect any extra info about requests.  So, I'd say it's there for back-compat. 

What do you think, should we keep it? 
`

brettsam: Yeah, I think we should just keep it. I thought this originally worked when we did the Http tracking (without this PR) -- I was curious if it stopped somehow.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
```diff
@@ -43,21 +43,12 @@ public void Initialize(ITelemetry telemetry)
...
 
             telemetry.Context.Cloud.RoleInstance = _roleInstanceName;
-
-            RequestTelemetry request = telemetry as RequestTelemetry;
-
-            // Zero out all IP addresses other than Requests
-            if (request == null)
+            if (telemetry.Context.Location.Ip == null)
             {
                 telemetry.Context.Location.Ip = LoggingConstants.ZeroIpAddress;
```
lmolkova, line 48: @brettsam I was also wondering if we can drop this? I don't remember what's the reason for the overriding null IP with 0.0.0.0 ?

lmolkova:
@brettsam can you please have a look? It's squashed and green!

Also, not sure if we should keep populating 0.0.0.0 when IP is not set - I'd rather remove it unless you expect it can break something.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
brettsam: I also don't remember why we did that, unfortunately. At one point, was it required? But I'd rather not risk removing it -- just in case someone depends on it.

brettsam:
It looks good. But before we merge, were you able to confirm this works end-to-end in a private site extension? If you don't remember how to set that up, ping me and I can help.

lmolkova:
tried in private extension. everything seems to work as expected: http-trigger function produces telemetry with IP addresses on custom and standard telemetry 

**Before**:


image


**After**:


image


Function code (no startup)
```csharp
        private readonly TelemetryClient client;

        public Function1(TelemetryConfiguration tc)
        {
            client = new TelemetryClient(tc);
        }

        [FunctionName("Function1")]
        public  async Task<IActionResult> Run(
            [HttpTrigger(AuthorizationLevel.Function, "get", "post", Route = null)] HttpRequest req,
            ILogger log)
        {
            client.TrackEvent("foo");
            log.LogInformation("C# HTTP trigger function processed a request.");

            string name = req.Query["name"];

            string requestBody = await new StreamReader(req.Body).ReadToEndAsync();
            dynamic data = JsonConvert.DeserializeObject(requestBody);
            name = name ?? data?.name;

            string responseMessage = string.IsNullOrEmpty(name)
                ? "This HTTP triggered function executed successfully. Pass a name in the query string or in the request body for a personalized response."
                : $"Hello, {name}. This HTTP triggered function executed successfully.";

            return new OkObjectResult(responseMessage);
        }
    }
```

lmolkova:
@brettsam can you merge? Or let me know if you want me to test some specific scenarios with private extension. Thanks!

brettsam:
All good! Thanks!

PR #2554, opened 2020-08-21
Making IMethodInvoker/MethodInvokerFactory public

mathewc:


paulbatum:
I assume this is so the functions host can use compiled lambdas?

brettsam (approved):

PR #2555, opened 2020-08-21
Removing FunctionInstanceWrapper

mathewc:
Removing this unecessary wrapper type to improve performance on the invoke path by removing unnecessary object allocations.


src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -79,94 +70,52 @@ public HostOutputMessage HostOutputMessage
...
                     parameterHelper.Initialize();
-                    functionStartedMessageId = await ExecuteWithLoggingAsync(functionInstance, functionStartedMessage, instanceLogEntry, parameterHelper, logger, cancellationToken);
-                    functionCompletedMessage = CreateCompletedMessage(functionStartedMessage);
+                    functionStartedMessageId = await ExecuteWithLoggingAsync(functionInstanceEx, functionStartedMessage, instanceLogEntry, parameterHelper, logger, cancellationToken);
                 }
                 catch (Exception exception)
                 {
-                    if (functionCompletedMessage == null)
-                    {
-                        functionCompletedMessage = CreateCompletedMessage(functionStartedMessage);
```
mathewc, line 130: Removed this object creation, which was just a copy of all the data that was already in the started message. A FunctionCompletedMessage "isa" FunctionStartedMessage, so above when creating the started message I just create an instance of the latter.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -258,9 +207,6 @@ internal static async Task HandleExceptionAsync(TimeoutAttribute timeout, Except
 
                 using (functionCancellationTokenSource)
                 {
-                    // This allows the FunctionOutputLogger to grab the TextWriter created by the IFunctionOutput. This enables user ILogger
```
mathewc, line 261: Just moved this comment into the method for readability

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -276,29 +222,26 @@ internal static async Task HandleExceptionAsync(TimeoutAttribute timeout, Except
                         await parameterHelper.BindAsync(instance.BindingSource, valueBindingContext);
                     }
 
-                    Exception invocationException = null;
-                    ExceptionDispatchInfo exceptionInfo = null;
+                    // Log started events
                     string startedMessageId = null;
-                    using (parameterHelper)
```
mathewc, line 282: There's already a using statement for the parameter helper earlier in the call stack

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -562,28 +505,29 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
 
-            // Combine #1 and #2 with a timeout task (handled by this method).
-            // functionCancellationTokenSource.Token is passed to each function that requests it, so we need to call Cancel() on it
-            // if there is a timeout.
-            bool isTimeout = await TryHandleTimeoutAsync(invokeTask, functionCancellationTokenSource.Token, throwOnTimeout, timeoutTokenSource.Token,
-                timerInterval, instance, () => functionCancellationTokenSource.Cancel());
-
-            // #2 occurred. If we're going to throwOnTimeout, watch for a timeout while we wait for invokeTask to complete.
-            if (throwOnTimeout && !isTimeout && functionCancellationTokenSource.IsCancellationRequested)
+            if (timerInterval > TimeSpan.Zero)
```
mathewc, line 516: Can skip timeout handling if no timeout is specified

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -1027,25 +962,14 @@ private int GetParameterIndex(string name)
...
-                    parameterNames[index] = parameterName;
-                    index++;
-                }
-
-                IValueProvider[] parameterValues = new IValueProvider[_valueProviders.Count];
-                index = 0;
+                string[] parameterNames = _valueProviders.Keys.ToArray();
 
-                foreach (IValueProvider parameterValue in _valueProviders.Values)
+                if (parameterNames.Length > 1)
```
mathewc, line 977: Only need to run sort if there is > 1 parameters :)

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -79,94 +70,52 @@ public HostOutputMessage HostOutputMessage
...
                 }
-
-                if (functionCompletedMessage != null)
+                finally
                 {
-                    functionCompletedMessage.ParameterLogs = parameterHelper.ParameterLogCollector;
-                    functionCompletedMessage.EndTime = DateTimeOffset.UtcNow;
+                    functionStartedMessage.ParameterLogs = parameterHelper.ParameterLogCollector;
+                    functionStartedMessage.EndTime = DateTimeOffset.UtcNow;
+                    ((IDisposable)functionInstance)?.Dispose();
```
brettsam: This lifetime has changed now, hasn't it? I think this should happen after everything else -- note that this gets passed into `HandleExceptionAsync()`. Not sure if it matters but it'd likely be safer to make sure we dispose at the end of all of this.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -668,9 +612,9 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
             return false;
         }
 
-        private FunctionStartedMessage CreateStartedMessageWithoutArguments(IFunctionInstance instance)
+        private FunctionCompletedMessage CreateStartedMessage(IFunctionInstance instance)
```
brettsam, line 621: Can you add a comment here so someone in the future knows why `CreateStartedMessage` actually returns a `CompletedMessage` :-)

brettsam (approved):
Couple of comments, but looks good.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
mathewc: Yes, good point - the lifetime had changed slightly. I've moved the disposal to the very end.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
mathewc: Done

PR #2556, opened 2020-08-21
Function lookup improvement + other minor cleanup

mathewc:
Remove the linear lookup the host was doing in LookupByName, and a few other minor cleanups.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/TaskMethodInvoker.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/TaskMethodInvoker.cs
@@ -37,8 +37,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
 
             if (innerTaskType != null && typeof(Task).IsAssignableFrom(innerTaskType))
             {
-                throw new InvalidOperationException("Returning a nested Task is not supported. Did you mean to await " +
-                    "or Unwrap the task instead of returning it?");
+                throw new InvalidOperationException("Returning a nested Task is not supported. Did you mean to await or Unwrap the Task instead of returning it?");
             }
         }
 
--- a/src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndex.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndex.cs
@@ -12,65 +12,62 @@ namespace Microsoft.Azure.WebJobs.Host.Indexers
     {
         private readonly IDictionary<string, IFunctionDefinition> _functionsById;
         private readonly IDictionary<MethodInfo, IFunctionDefinition> _functionsByMethod;
+        private readonly IDictionary<string, IFunctionDefinition> _functionsByName;
         private readonly ICollection<FunctionDescriptor> _functionDescriptors;
 
         public FunctionIndex()
         {
-            _functionsById = new Dictionary<string, IFunctionDefinition>();
+            _functionsById = new Dictionary<string, IFunctionDefinition>(StringComparer.OrdinalIgnoreCase);
             _functionsByMethod = new Dictionary<MethodInfo, IFunctionDefinition>();
+            _functionsByName = new Dictionary<string, IFunctionDefinition>(StringComparer.OrdinalIgnoreCase);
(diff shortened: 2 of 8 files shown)


src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndex.cs
```diff
@@ -12,65 +12,62 @@ internal class FunctionIndex : IFunctionIndex, IFunctionIndexCollector
...
                 return null;
             }
-
-            return _functionsById[functionId];
+            return function;
         }
 
         public IFunctionDefinition LookupByName(string name)
         {
-            // For compat, accept either the short name ("Class.Name") or log name (just "Name")
```
mathewc, line 51: This is the overload used by Functions Host, and it resulted in a linear lookup each time.

src/Microsoft.Azure.WebJobs.Host/Executors/TaskMethodInvoker.cs
```diff
@@ -18,55 +17,7 @@ public TaskMethodInvoker(Func<TReflected, object[], Task<TReturnType>> lambda)
 
         public Task<TReturnType> InvokeAsync(TReflected instance, object[] arguments)
         {
-            Task<TReturnType> task = _lambda.Invoke(instance, arguments);
-            ThrowIfWrappedTaskInstance(task);
```
mathewc: If someone were to return a nested task, it just won't work like they expect. I don't think checking for this on every invocation is warranted.

src/Microsoft.Azure.WebJobs.Host/JobHost.cs
```diff
@@ -320,13 +320,16 @@ private Task EnsureHostInitializedAsync(CancellationToken cancellationToken)
 
             TaskCompletionSource<bool> tsc = null;
 
-            lock (_lock)
+            if (_hostInitializationTask == null)
```
mathewc, line 319: Didn't discover a problem here, but in reviewing the code realized that we could do a null check prior to lock to skip an unnecessary lock

src/Microsoft.Azure.WebJobs.Host/Executors/TaskMethodInvoker.cs
paulbatum: I agree that checking this every invocation seems excessive. I worry somewhat that not having the check at all might result in error messages that are less clear when customers do make this mistake. I'm ok with removing it for now but if this comes back to bite us we might need to consider an implementation where we check it on the first call.

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndex.cs
paulbatum: Nice find! Does the by-name dictionary need to be configured to use OrdinalIgnoreCase to be consistent with the comparison being done below?

paulbatum (approved):

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndex.cs
mathewc: Good catch. I made this case insensitive and added a test. I made the other "by ID" dictionary case insensitive as well. That overload (lookup by function ID) is only used by Dashboard host message execution.

src/Microsoft.Azure.WebJobs.Host/Executors/TaskMethodInvoker.cs
mathewc: Bind time static checks verify that the return type is valid for the target output binding. E.g. there won't be a binding that will allow Task<Task<string>> - it would lead to an indexing error. So this check was all about a method that declares it returns Task, but at runtime returns a nested Task. Given that the value can vary at runtime, a one time check wouldn't catch everything, if the method is varying what it returns.

mathewc: All that said, I've chosen to leave this in place for now. I did add a unit test for the error though - it was missing before.

PR #2579, opened 2020-09-15
Added stringified stack trace as an optional parameter.

yojagad:
In case of errors, there are some logs which do not have the full stack trace. Adding an optional parameter as a workaround that would aid in investigations. One example is this https://github.com/Azure/azure-webjobs-sdk/issues/2576  

image

The above screenshot just has the exception without the stack trace.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
@@ -117,7 +117,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                     if (instanceLogEntry != null)
                     {
                         await NotifyCompleteAsync(instanceLogEntry, functionStartedMessage.Arguments, exceptionInfo);
-                        _resultsLogger?.LogFunctionResult(instanceLogEntry);
+                        _resultsLogger?.LogFunctionResult(instanceLogEntry, exceptionInfo?.SourceException.ToString());
                     }
 
                     if (loggedStartedEvent)
--- a/test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
@@ -537,7 +537,6 @@ namespace Microsoft.Azure.WebJobs.Host.EndToEndTests
 
             // Results logs the exception with no message
             Assert.NotNull(resultLogger.GetLogMessages().Single().Exception);
-            Assert.Null(resultLogger.GetLogMessages().Single().FormattedMessage);
 
             // It logs Executed/Executing messages and the timeout message.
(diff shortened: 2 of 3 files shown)


pragnagopa (commented):
Thanks for the PR! LGTM. @brettsam  - is there a better way to capture full exception stack trace?

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/AsyncChainEndToEndTests.cs
```diff
@@ -537,7 +537,6 @@ private async Task RunTimeoutTest(IHost host, Type expectedExceptionType, string
 
             // Results logs the exception with no message
             Assert.NotNull(resultLogger.GetLogMessages().Single().Exception);
-            Assert.Null(resultLogger.GetLogMessages().Single().FormattedMessage);
```
yojagad, line 540: How bad are the consequences of having a formatted message now?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -117,7 +117,7 @@ public async Task<IDelayedException> TryExecuteAsync(IFunctionInstance functionI
                     if (instanceLogEntry != null)
                     {
                         await NotifyCompleteAsync(instanceLogEntry, functionStartedMessage.Arguments, exceptionInfo);
-                        _resultsLogger?.LogFunctionResult(instanceLogEntry);
+                        _resultsLogger?.LogFunctionResult(instanceLogEntry, exceptionInfo?.SourceException.ToString());
```
brettsam, line 120: This is the wrong place for this -- the FunctionResult is meant to have just data, no strings. `_functionInstanceLogger.LogFunctionCompletedAsync ()`is where we write out the Completed message in Kusto, etc.

But that should have the exception logged to it -- maybe we're losing the stack somewhere in this? 

```
functionStartedMessage.Failure = new FunctionFailure
{
    Exception = exception,
    ExceptionType = exception.GetType().FullName,
    ExceptionDetails = exception.ToDetails(),
};
```

brettsam (changes requested):

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
yojagad: That was where I started my investigation with. Whether exception.ToDetails() was losing the stack while prettifying it somewhere. But turns out that's not the case. I verified the stack trace being present all the way in https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Loggers/FunctionInstanceLogger.cs#L119

I also noticed it doesn't do anything with failure.ExceptionDetails in the above FunctionInstanceLogger. Maybe pass it as another string parameter there ? But that wouldn't look too great in the Function Completed message :(

yojagad:
Closing this PR in favor of https://github.com/Azure/azure-functions-host/pull/6680

PR #2580, opened 2020-09-15
[pack] Added support for overriding app insights telemetry and quick pulse endpoints

### LINKED ISSUE #2284: Support for Application Insights in the Azure Government Cloud [v2.x]
Creating a separate copy of #2263 to explicitly do this for v2.x (and functions v1). The changes will be very different from the v3.x changes tracked in the other issue.

gzuber:
Resolves #2284 

It doesn't look like there's a way to override Profile Query endpoints with the version of App Insights we're using (2.4.0). I believe it _may_ have been introduced around 2.6.0.

[Related Host PR](https://github.com/Azure/azure-functions-host/pull/6652)

The user would specify an "endpointAddress" and "quickPulseServiceEndpoint" property in the "applicationInsights" section of their host.json. These values are passed to webjobs where they overwrite the appropriate configs.

Testing this in the government cloud, I was able to get Live Metrics (quick pulse) and request logs working in app insights (which I couldn't get working without these changes).


test/Microsoft.Azure.WebJobs.Host.UnitTests/Loggers/DefaultTelemetryClientFactoryTests.cs
```diff
@@ -26,5 +26,15 @@ public void InitializeConfiguguration_Configures()
...
         }
+
+        [Fact]
+        public void InitializeConfiguration_ConfiguresWithIngestionEndpoint()
+        {
+            var testIngestionEndpoint = "https://testIngestion.Endpoint/";
+            var factory = new DefaultTelemetryClientFactory(string.Empty, null, testIngestionEndpoint, null, null);
+            var config = factory.InitializeConfiguration();
+
+            Assert.Equal(testIngestionEndpoint.ToLower(), config.TelemetryChannel.EndpointAddress.ToLower());
```
brettsam, line 37: Just one nit -- can you also verify for live metrics?

brettsam (approved):

test/Microsoft.Azure.WebJobs.Host.UnitTests/Loggers/DefaultTelemetryClientFactoryTests.cs
brettsam: Talked with @gzuber -- this isn't completely straightforward so we'll validate with e2e testing.

PR #2593, opened 2020-09-30
SharedQueueWatcher simplified

Scooletz:
This PR changes the way `SharedQueueWatcher` is written, by replacing the costly `ConcurrentBag` and the call to `.ToArray` when notifing. The replacement is provided with a simple array `INotificationCommand[]`. The update function for `AddOrUpdate` is changed to operate on the input and return a new array. It ensures that the value, if swapped, in the concurrent dictionary will be properly updated.

This removes the `.ToArray` and heaviness of `ConcurrentBag` from the hot `Notify` path. 

### Benchmarks

After introducing this change I understood that it's highly likely that there will be a single notification command and run a bechmark that is in history of these changed, but is removed as it's something that won't be useful in the future.

#### Before:

|                    Method |     Mean |   Error |  StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------------------- |---------:|--------:|--------:|-------:|------:|------:|----------:|
| SharedQueueWatcher_Notify | 150.8 ns | 2.78 ns | 5.15 ns | 0.0100 |     - |     - |      32 B |

#### After:

|                    Method |     Mean |    Error |   StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------------------- |---------:|---------:|---------:|------:|------:|------:|----------:|
| SharedQueueWatcher_Notify | 44.40 ns | 1.757 ns | 4.897 ns |     - |     - |     - |         - |

It's a small gain but makes the code a bit clearer imho, so I'll still propose it as a change.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Listeners/SharedQueueWatcher.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Listeners/SharedQueueWatcher.cs
@@ -2,21 +2,23 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
 
 namespace Microsoft.Azure.WebJobs.Host.Queues.Listeners
 {
     internal class SharedQueueWatcher : IMessageEnqueuedWatcher
     {
-        private readonly ConcurrentDictionary<string, ConcurrentBag<INotificationCommand>> _registrations =
-            new ConcurrentDictionary<string, ConcurrentBag<INotificationCommand>>();
+        private readonly ConcurrentDictionary<string, INotificationCommand[]> _registrations =
+            new ConcurrentDictionary<string, INotificationCommand[]>();
 
         public void Notify(string enqueuedInQueueName)
         {
-            ConcurrentBag<INotificationCommand> queueRegistrations;
+            INotificationCommand[] queueRegistrations;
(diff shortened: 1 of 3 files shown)


Scooletz:
@brettsam Is this possibly lack of the `full-ci`? If yes, can I ask for applying the label and kicking the build?

Scooletz:
It's 🟢 @brettsam . 

I'm thinking now, that if I'm too pushy with these PRs, just let me know.

brettsam:
Thanks again @Scooletz -- do you have benchmarks to show the improvements on this one? Adding @mathewc as he's been doing a bunch of perf work, specifically in the HttpTrigger pipeline.

mathewc:
Thanks @Scooletz - these changes look good to me. This is code that hasn't been changed forever, and I see there aren't any unit tests for SharedQueueWatcher. I'd feel safer if along with these changes we had a Mock unit test or two that verified this. E.g. just a test that Registers some notifications, and verifies the correct number are called when Notify is invoked.

Scooletz:
Give me one Earth spin and I'll cover it @mathewc @brettsam . It's 9PM now 😅

Scooletz:
I rewrote the history and pushed with `-force` to make it look like:

1. add a benchmark to provide some numbers for this PR
1. implementation changed
1. the benchmark removed as it's not vital for future development
1. a concurrent Thread-based test

Looking forward for a review and comments 😄

mathewc (approved):

Issue #2589, opened 2020-09-23
Release Storage extension to nuget


pragnagopa:
cc @gohar94 

Once the deployment for https://github.com/Azure/app-service-announcements/issues/288 completes, release storage extension to nuget and create release work item to update bundles.

brettsam:
https://www.nuget.org/packages/Microsoft.Azure.WebJobs.Extensions.Storage/4.0.3 is now released.

PR #2596, opened 2020-09-30
moving storage to 4.0.3

brettsam:
- Moving storage to 4.0.3
- Locking storage extension to reference WebJobs 3.0.22

This helps with #2589

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.Storage/WebJobs.Extensions.Storage.csproj
+++ b/src/Microsoft.Azure.WebJobs.Extensions.Storage/WebJobs.Extensions.Storage.csproj
@@ -96,12 +96,10 @@
     <PackageReference Include="Microsoft.Azure.Cosmos.Table" Version="1.0.7" />
     <PackageReference Include="Microsoft.Azure.Storage.Blob" Version="11.1.7" />
     <PackageReference Include="Microsoft.Azure.Storage.Queue" Version="11.1.7" />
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.WebJobs.Extensions.Storage/WebJobs.Extensions.Storage.csproj
```diff
@@ -96,12 +96,10 @@
     <PackageReference Include="Microsoft.Azure.Cosmos.Table" Version="1.0.7" />
     <PackageReference Include="Microsoft.Azure.Storage.Blob" Version="11.1.7" />
     <PackageReference Include="Microsoft.Azure.Storage.Queue" Version="11.1.7" />
+    <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.22" />
     <PackageReference Include="StyleCop.Analyzers" Version="1.1.0-beta004">
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
   </ItemGroup>
 
-  <ItemGroup>
```
pragnagopa, line 104: intentionally removing project reference and relying in nuget?

brettsam: Yes, this repo has already moved forward to 3.0.23 -- so any build we generate now will need to wait (or i'll need to roll that back, which is confusing). This let's us release "out-of-band" by directly targeting 3.0.22, which is already out in nuget.

pragnagopa (approved):

PR #2597, opened 2020-10-01
Create new FunctionInstance on execution retry

pragnagopa:


src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionExecutor.cs
```diff
@@ -51,15 +55,41 @@ public async Task<FunctionResult> TryExecuteAsync(TriggeredFunctionData input, C
...
+            }
+
             FunctionResult result = exception != null ?
                 new FunctionResult(exception.Exception)
                 : new FunctionResult(true);
 
             return result;
         }
+
+        internal async Task<IDelayedException> TryExecuteWithRetriesAsync(FunctionInstanceFactoryContext<TTriggerValue> context, CancellationToken cancellationToken)
```
mathewc: To remove this code duplication, I was thinking that we could have a single static extension method for IFunctionExecutor with signature like `TryExecuteAsync(Func<IFunctionInstance> instanceFactory, ILogger logger, CancellationToken)`. You can then call this same method in all places needed as a one liner replacement.

Note - there's also one more place where we need to do this: https://github.com/Azure/azure-webjobs-sdk/blob/b798412ad74ba97cf2d85487ae8479f277bdd85c/src/Microsoft.Azure.WebJobs.Host/Executors/HostMessageExecutor.cs#L127. Now this path isn't really used in v3, but we should fix it up for completeness. It's for Dashboard integration, when someone hits "Replay" in the dashboard, and an internal queue message is used to trigger the invocation.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutorExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Triggers;
+
+namespace Microsoft.Azure.WebJobs.Host.Executors
+{
+    /// <summary>Provides extension methods for the <see cref="IFunctionExecutor"/> interface.</summary>
+    public static class FunctionExecutorExtensions
```
mathewc: Make this internal, then the method below can be public. We don't want this extension in our public surface area.

src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionExecutor.cs
```diff
@@ -51,10 +55,18 @@ public async Task<FunctionResult> TryExecuteAsync(TriggeredFunctionData input, C
                 };
             }
 
+            // initial attempt
```
mathewc: For all 3 of these call sites, I'm wondering why the initial invocation is outside of the TryExecuteAsync extension method. That method should be able to handle everything, including the first invocation, so we don't have to repeat the code?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutorExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+        internal static async Task<IDelayedException> TryExecuteAsync(this IFunctionExecutor executor, Func<IFunctionInstance> instanceFactory, IRetryStrategy retryStrategy, ILogger logger, CancellationToken cancellationToken)
+        {
+            var attempt = 0;
+            IDelayedException functionResult = null;
+            bool retriesExceeded = false;
+            while (!retriesExceeded)
+            {
+                IFunctionInstance functionInstance = instanceFactory.Invoke();
+                functionResult = await executor.TryExecuteAsync(functionInstance, cancellationToken);
+                retriesExceeded = await Utility.WaitForNextExecutionAttempt(functionInstance, functionResult, retryStrategy, logger, attempt);
```
mathewc: Can you just inline the retry logic here like it was previously, and remove this WaitForNextExecutionAttempt? That way we don't have to go async for successful invocations, as you're doing now.

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/ILoggerTests.cs
```diff
@@ -198,6 +232,18 @@ public void ILogger(ILogger log)
                 log.LogError(0, ex, "Log {other} keys {and} values", "3", "4");
             }
 
+            [NoAutomaticTrigger]
+            [FixedDelayRetry(3, "00:00:01")]
```
mathewc: Since the actual duration doesn't matter, can you use "00:00:00.500" or something so it doesn't run for 3 seconds unecessarily?

src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionExecutor.cs
```diff
@@ -51,9 +55,9 @@ public async Task<FunctionResult> TryExecuteAsync(TriggeredFunctionData input, C
                 };
             }
 
-            IFunctionInstance instance = _instanceFactory.Create(context);
-
-            IDelayedException exception = await _executor.TryExecuteAsync(instance, cancellationToken);
+            Func<IFunctionInstance> instanceFactory = () => _instanceFactory.Create(context);
+            var logger = _loggerFactory.CreateLogger(LogCategories.CreateFunctionCategory(_descriptor.LogName));
```
mathewc: Not used right? Remove.

src/Microsoft.Azure.WebJobs.Host/Utility.cs
```diff
@@ -1,12 +1,19 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using System;
 using System.Reflection;
+using System.Threading;
```
mathewc: Undo changes to this file

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutorExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+            {
+                IFunctionInstance functionInstance = instanceFactory.Invoke();
+                if (logger == null)
+                {
+                    logger = loggerFactory.CreateLogger(LogCategories.CreateFunctionCategory(functionInstance.FunctionDescriptor.LogName));
+                }
+                functionResult = await executor.TryExecuteAsync(functionInstance, cancellationToken);
+                if (functionResult == null || functionInstance.FunctionDescriptor.RetryStrategy == null)
+                {
+                    // function invocation succeeded
```
mathewc: Not necessarily succeeded - might have failed, but no retry strategy configured? You could move the RetryStrategy check to the if check below.

src/Microsoft.Azure.WebJobs.Host/Executors/HostMessageExecutor.cs
```diff
@@ -124,7 +128,8 @@ private async Task ProcessCallAndOverrideMessage(CallAndOverrideMessage message,
 
             if (instance != null)
             {
-                await _innerExecutor.TryExecuteAsync(instance, cancellationToken);
+                Func<IFunctionInstance> instanceFactory = () => CreateFunctionInstance(message);
```
mathewc: Need to remove the CreateFunctionInstance call above. To do so, you'll hoist the _functionLookup.Lookup check out of that method, and change the if (instance != null) check to instead check that the function lookup didn't return null.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutorExtensions.cs
pragnagopa: Moved it down to be more explicit!

src/Microsoft.Azure.WebJobs.Host/Utility.cs
pragnagopa: done!

src/Microsoft.Azure.WebJobs.Host/Executors/TriggeredFunctionExecutor.cs
pragnagopa: cleaned.

mathewc (approved):

PR #2599, opened 2020-10-02
Lighter QueueCausalityManager for Azure Storage Queues

Scooletz:
Currently `QueueCausalityManager` does a lot of not needed work when trying to `.GetOwner`. It deserializes whole message just to get one property. To measure the overhead I added `benchmarks` and measure some test cases (the bigger the message, the worse)

|                                  Method |     Mean |     Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------------------------------------- |---------:|----------:|----------:|-------:|------:|------:|----------:|
|  QueueCausalityManager_GetOwner_Minimal | 2.216 us | 0.0423 us | 0.0396 us | 1.0605 |     - |     - |   3.27 KB |
| QueueCausalityManager_GetOwner_Moderate | 5.873 us | 0.1099 us | 0.1028 us | 1.4572 |     - |     - |   4.48 KB |

Then, I rewrote probing for the `$AzureWebJobsParentId` with `JsonTextReader` and I got a constant memory overhead + a bit better time (which is not important here).

|                                  Method |     Mean |     Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------------------------------------- |---------:|----------:|----------:|-------:|------:|------:|----------:|
|  QueueCausalityManager_GetOwner_Minimal | 1.361 us | 0.0269 us | 0.0492 us | 0.8202 |     - |     - |   2.52 KB |
| QueueCausalityManager_GetOwner_Moderate | 1.310 us | 0.0134 us | 0.0126 us | 0.8202 |     - |     - |   2.52 KB |

As you can see, the initial current overhead is the worse, the bigger message is as it will deserialize everything to `J*`. The fix provides constant overhead.

I'm not sure what are the plans for benchmarking this repo. Did I put benchmarks in a right place? Do you have another set of benchmarks?

Nevertheless, the change removes a lot of allocations for big messages.


Scooletz:
Probably `full-ci` required 😅

WebJobs.sln
```diff
@@ -51,6 +51,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "TestProjects", "TestProject
 EndProject
 Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "FSharpFunctions", "test\TestProjects\FSharpFunctions\FSharpFunctions.fsproj", "{11702A4B-8402-4082-BE38-4F0C2CBBF61D}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Benchmarks", "test\Benchmarks\Benchmarks.csproj", "{78F8086D-8313-477A-B24F-E475A690880A}"
```
brettsam, line 54: I was just writing in one of your other PRs how it'd be great if we had a Benchmarks project so you can add some of these to it for future runs... and here it is :-)

brettsam (approved):
This looks great to me. I'd still like @mathewc to comment on this just in case he sees anything else we'd need to change. Thanks for adding that Benchmarks project! That gives us a place to add more in the future.

WebJobs.sln
mathewc: Yes, this is great. I'll be adding one for the Functions Host as well most likely.

mathewc (approved):

PR #2602, opened 2020-10-08
TaskSeriesTimer loop slimmed a little bit

Scooletz:
This PR moves the creation of `TaskCompletionSource<object>` and its registration to the `cancellationToken` before the `while` loop. As far as I understand, once the token is cancelled the loop break and the method returns so there's no need to recreate TCS with its registration for every single execution of `command.ExecuteAsync`.

I was considering using the `CancellationToken.Register(Action<object> callback, object state)` to remove the closure allocation, but now it isn't that harmful and it can be done in a separate PR if one considers it a cost.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Timers/TaskSeriesTimer.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Timers/TaskSeriesTimer.cs
@@ -120,12 +120,12 @@ namespace Microsoft.Azure.WebJobs.Host.Timers
 
                 Task wait = _initialWait;
 
-                // Execute tasks one at a time (in a series) until stopped.
-                while (!cancellationToken.IsCancellationRequested)
-                {
-                    TaskCompletionSource<object> cancellationTaskSource = new TaskCompletionSource<object>();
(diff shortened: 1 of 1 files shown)


Scooletz:
Due to some problems with running a full suite, I only run `TaskSeriesTimerTests` and they are all green.

The PR might need a `full-ci` label.

NickCraver:
cc @mathewc this hit my radar as well, was checking for existing work and found this PR. Note that we can improve this further by not capturing via the state overload, like this:
```cs
using (cancellationToken.Register((tcs) => ((TaskCompletionSource<object>)tcs).SetCanceled(), cancellationTaskSource))
```
In .NET 6 we can improve further with `await myTask.WaitAsync(cancellationToken);`, but that requires adding a `net6.0` build which I need to check reasonability on (not sure how much maintenance such a thing adds).

Both optimizations should be applied up in `.StopAsync()` overall - happy to do that in another PR proposal or can help advise here.

Scooletz:
Hey @NickCraver Nice to see you looking at these places 😍 As this PR is once year old, I'll close it as is leaving space for other ones

PR #2603, opened 2020-10-09
Application Insights logger allocates less

Scooletz:
This PR changes the logging with Application Insights. It does it in the following way.

1. `DictionaryLoggerScope` captures and persists the merged dictionary of values. With the changes in this PR it caches the value and reuses it, reducing allocations whenever one or more `.Log` operations are performed in a single `Scope`. A rough estimate is 2-3KB per call. 
1. The internal seam of `DictionaryLoggerScope` is changed to return `IReadOnlyDictionary<string,object>` to ensure that the caller does not interfere with the cached value.
1. A new `ReadOnlyScopeDictionary` is provided for using with `.BeginScope` where the keys are known and the only difference is in the array of values provided. For reasonably small number of keys brute-force compare should be more than enough.

The following changes were considered but not applied because of the lack of evidence of perf improvements

- prunning one LINQ call
- replacing an array `.Contains` with a `HashSet.Contains`

One thing to notice is that this PR makes the evaluation lazy and keeps the `IEnumerable<KeyValuePair<string,object>>` in the scope until it's used for the first time. On one side it has a positive aspect of lazy evaluation. If nothing is used, scope doesn't have to be evaluated. 

This PR does not change the underlying implementation to support `ISupportExternalScope` and `IExternalScopeProvider`. The external scope provided is controlled... externally 😅 and can have an impact on sematics of scoping. This should be considered as a separate PR.

### Benchmarks

Benchmarking two cases: a call to `.LogFunctionResult` and 3 subsequent calls to `.LogInformation`. The benchmarks were included in this PR as they test a broad area of API and might be a base for future improvements.

#### Before

|       Method |     Mean |    Error |   StdDev |   Median |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|------------- |---------:|---------:|---------:|---------:|-------:|-------:|------:|----------:|
|  OnException | 69.62 us | 0.813 us | 0.761 us | 69.53 us | 6.9580 | 0.1221 |     - |   21.3 KB |
| MultipleLogs | 44.36 us | 1.726 us | 4.979 us | 42.08 us | 5.1880 |      - |     - |  16.07 KB |

#### After

|       Method |     Mean |    Error |   StdDev |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|------------- |---------:|---------:|---------:|-------:|-------:|------:|----------:|
|  OnException | 62.56 us | 1.209 us | 1.241 us | 4.8828 |      - |     - |  15.34 KB |
| MultipleLogs | 34.17 us | 0.656 us | 0.701 us | 3.2959 | 0.0610 |     - |  10.17 KB |

Again, the time improvements are low, but trimming a few KB per log statement looks quite good.


Scooletz:
Softly pinging @mathewc @brettsam . This might require `full-ci`.

We might be running out of allocations soon 😅

brettsam:
Nice! This is definitely a hot area for perf improvements. This code is called a *lot* so any gains are multiplied across many invocations.

Is it possible to try a benchmark using the built-in `LoggerExternalScopeProvider`? This didn't exist when I first wrote this code, but I've moved to using it just about everywhere else (https://github.com/Azure/azure-functions-host/pull/5114).

Ideally I'd like to remove the `DictionaryLoggerScope` type completely but I took a dependency on it [here](https://github.com/Azure/azure-webjobs-sdk/blob/1dcece0045ff4de5fdddce6c587f047ad85eedef/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs#L58) that we'd have to work through.

Side note -- I'm on-call right now so a fully review probably won't happen until next week :-)

Scooletz:
This was my aim. Logging is called a lot and even when I can't reduce the price of App Insights, I could definitely make it cheaper on the execution side. 

Let me give it a spin in the next week then. I was not aware of the history behind it. I observed moving to `IExternalScopeProvider` in other repos but didn't consider it here. I think that it may a bit more complex, but will give it a try.

Scooletz:
☝️  This will break a build. I'm in the middle of adding `IExternalScopeProvider` but due to external control of the scope I have some problems with making it "just build (TM)". Will try to fix it soon.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
```diff
@@ -54,8 +54,8 @@ public void Initialize(ITelemetry telemetry)
             telemetryProps[LogConstants.ProcessIdKey] = _currentProcessId;
 
             // Apply our special scope properties
-            IDictionary<string, object> scopeProps =
-                DictionaryLoggerScope.GetMergedStateDictionary() ?? new Dictionary<string, object>();
+            IReadOnlyDictionary<string, object> scopeProps =
+                DictionaryLoggerScope.BuildLazy();
```
Scooletz: @brettsam This line is problematic for me. As far as I see it, this code requires access to scopes. `WebJobsTelemetryInitializer` is injected as a singleton so potentially it could accept the `IExternalScopeProvider` as a parameter, but...  the `LoggerFactoryScopeProvider` is initialized and kept only in the `LoggerFactory` giving no access to it beside the provider code.

https://github.com/dotnet/runtime/blob/b447b8efba94711d10296e27a35a1ef7f33ff4b5/src/libraries/Microsoft.Extensions.Logging/src/LoggerFactory.cs#L183-L190

Scooletz:
I'm not sure if I can provide a meaningful implementation based on `IExternalScopeProvider` as the scopes as described in a comment are needed in additional places. Additionally, with the externally owned scopes I can't do some of the opitmizations as scope potentially could be injected by something else. What should I do with it? 

1. Undo `IExternalScopeProvider` related work, provide just a perf optimization and leave the change to `IExternalScopeProvider`  for the future? 
1. Separate scopes used to pass metadata (log level) from the actual scopes used for logging and move to `IExternalScopeProvider` but leave additional scoping mechanism for steering the execution?
1. Something else?

Some input will be highly appreciated.

Scooletz:
Hey. Any input on https://github.com/Azure/azure-webjobs-sdk/pull/2603#issuecomment-706564583 @brettsam ?

Scooletz:
FYI @mathewc @brettsam I removed the unfinished work with `IExternalScopeProvider` and `ISupportExternalScope` and focused on improving the existing one. As you can see, the amount of memory being saved in here is significant. Additionally, there should be some gain in `FunctionExecutor` as it uses the `ReadOnlyScopeDictionary` now as well.

I did my best to describe the reason for abandoning the `IExternalScopeProvider` work. This would require a huge discussion about owning scopes, using them as a merge-dictionary for Application Insights logger etc., something that would take much longer and would require multiple roundtrips IMO.

I'm looking forward to hearing from you, what do you think about the results.

Scooletz:
I needed to rewrite the history. The reason for this is that I made a huge mistake in the initial benchmark as I took the call creating `FunctionInstance` as a part of the measurement. To fix it I amended the first commit with the benchmark, rerun it, rebased all my work on top of it and then rerun it again. Now it compares apples to apples.

brettsam:
Thanks for the diligence here. Yes, I agree that removing that ScopeProvider will take more work. I'm really regretting that I took that dependency the way I did. 

Do the measurements in the initial PR comment now have the updated benchmark results?

Scooletz:
> Yes, I agree that removing that ScopeProvider will take more work. I'm really regretting that I took that dependency the way I did.

I'm very opinionated in regards to the `IExternalScopeProvider` and who should be responsible for keeping scopes so I do not share the regret. At the same time, I do understand the overall approach for moving toward using it.

> Do the measurements in the initial PR comment now have the updated benchmark results?

Yes they do. I found that the allocs for the improved versions might vary (by 500bytes or so), but haven't found the reason.  They are left in the `Benchmarks` project so they may be rerun for the first commit as well as the last.

Sorry, for the red build. I'm running a subset of tests and this was the reason. It looks like it's failing on `UserCodeHttpCallsAreReportedOnceWhenMultipleHostsAreActive`. Is there a description how to set it up locally?

brettsam:
I'm not sure why that test is failing -- I've kicked it back off to see if perhaps there's some flakiness with it. Thanks for the patience on this one, my goal is to get this review done by end-of-week (plenty of time to get into our next release).

Scooletz:
Thank you @brettsam 🤗 Ok, it's green now.
🆗 Looking forward to hearing from you :-)

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -481,11 +487,8 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
 
                 try
                 {
-                    using (logger.BeginScope(new Dictionary<string, object>
-                    {
-                        [LogConstants.CategoryNameKey] = LogCategories.CreateFunctionCategory(instance.FunctionDescriptor.LogName),
-                        [LogConstants.LogLevelKey] = LogLevel.Information
-                    }))
+                    using (logger.BeginScope(new ReadOnlyScopeDictionary(ExecuteWatcherScopeKeys, 
+                        new object[] {LogCategories.CreateFunctionCategory(instance.FunctionDescriptor.LogName), LogLevel.Information})))
```
brettsam, line 491: Side note for a future improvement (@mathewc) -- would caching these category strings would give us a boost?

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ReadOnlyScopeDictionary.cs
```diff
@@ -0,0 +1,109 @@
...
+
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Microsoft.Extensions.Logging
+{
+    /// <summary>
+    /// The base class for all readonly dictionaries used for scopes
+    /// </summary>
+    public class ReadOnlyScopeDictionary : IReadOnlyDictionary<string, object>
```
brettsam, line 12: Is there a way we can use a `ReadOnlyDictionary` as a backing store for this rather than implementing `IReadOnlyDictionary`? Code like this makes me nervous unless there's a whole bunch of tests validating it.

brettsam (commented):
One initial comment -- and I've asked @mathewc to take a look here as well.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
Scooletz: Maybe we could push it to the `.FunctionDescriptor` and have a property in there?

src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/ReadOnlyScopeDictionary.cs
Scooletz: I didn't try it. My assumption was to reduce allocation on the repeated array of keys. With scopes of 8 keys, we have an overhead for storing 8 keys + length + other fields in a dictionary that for a scope might be simply unneeded. [ReadOnlyDictionary](https://github.com/dotnet/runtime/blob/master/src/coreclr/src/System.Private.CoreLib/src/System/Collections/ObjectModel/ReadOnlyDictionary.cs#L24) suffers from the same overhead as it wraps the dictionary.

We have benchmarks for some scenarios now. I understand that you might be unsure with this implementation. Let me give it a spin. Maybe I can replace it with a smaller alternative that isn't as untested as this one.

Scooletz: A fast try out breaks on `StartTelemetryIfFunctionInvocation` which requires the state to implement `IReadOnlyDictionary` for retrieval of the data.

I see two choices: 

1. add tests
1. undo the work with the scope as the new dictionary and rerun benchmarks to see how much it impacts the performance gain. Maybe with initializing capacity properly it would just work.

Scooletz:
I think I'll need to think a bit more about the approach. Maybe it's really about splitting the responsibilities of scopes? Some of them are used to navigate execution of the `ApplincationInsightsLogger` when other deliver properties for logging, not to mention the merge-able behavior.

I think I'll need to pause for a while and think more about it.

Scooletz:
I'll issue other PRs. I think this has been blown out of proportion.

PR #2608, opened 2020-10-18
Adding EnableSqlCommandTextInstrumentation to ApplicationInsightsLoggerOptions

### LINKED ISSUE #2548: Add ability to turn on SQL query logging for Azure Functions
Before 28 july 2020 we were able to see the sql queries in appinsight dependencies logs, but now we are only seeing server names and no sql queries are logged from azure functions consumption plan.  We have other code in aks cluster and sql queries are logged for calls done for app hosted in aks.

#### Expected behavior

Sql queries not logged in appinsight for azure function consumption plan.

#### Actual behavior

Sql queries should be logged in appinsight for azure function consumption plan.

#### Known workarounds
i tried this link but it didnt help
https://github.com/microsoft/ApplicationInsights-dotnet/issues/1425
Also adding reference to this dll brings aspnetcore things in azurefuctions project and is not making sense 

#### Related information 

* Package version
* Links to source
image

soninaren:
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/2548

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
@@ -69,7 +69,8 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
         /// can result in decreased application startup time. Default value is 15 seconds.
         /// </summary>
         [Obsolete("Use LiveMetricsInitializationDelay instead.")]
-        public TimeSpan QuickPulseInitializationDelay {
+        public TimeSpan QuickPulseInitializationDelay
+        {
             get
             {
                 return LiveMetricsInitializationDelay;
@@ -104,6 +105,8 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
         /// Enabled by default.
         /// </summary>
         public bool EnableDependencyTracking { get; set; } = true;
+
+        public DependencyTrackingOptions DependencyTrackingOptions { get; set; }
 
         /// <summary>
         /// Gets or sets HTTP request collection options. 
@@ -159,6 +162,18 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
                 { nameof(HttpAutoCollectionOptions.EnableResponseHeaderInjection), HttpAutoCollectionOptions.EnableResponseHeaderInjection }
             };
 
+            JObject dependencyTrackingOptions = new JObject
+            {
+                { nameof(DependencyTrackingOptions.DisableRuntimeInstrumentation), DependencyTrackingOptions.DisableRuntimeInstrumentation },
+                { nameof(DependencyTrackingOptions.DisableDiagnosticSourceInstrumentation), DependencyTrackingOptions.DisableDiagnosticSourceInstrumentation},
+                { nameof(DependencyTrackingOptions.EnableLegacyCorrelationHeadersInjection), DependencyTrackingOptions.EnableLegacyCorrelationHeadersInjection},
+                { nameof(DependencyTrackingOptions.EnableRequestIdHeaderInjectionInW3CMode), DependencyTrackingOptions.EnableRequestIdHeaderInjectionInW3CMode},
+                { nameof(DependencyTrackingOptions.EnableSqlCommandTextInstrumentation), DependencyTrackingOptions.EnableSqlCommandTextInstrumentation},
+                { nameof(DependencyTrackingOptions.SetComponentCorrelationHttpHeaders), DependencyTrackingOptions.SetComponentCorrelationHttpHeaders},
+                { nameof(DependencyTrackingOptions.EnableAzureSdkTelemetryListener), DependencyTrackingOptions.EnableAzureSdkTelemetryListener}
+            };
+
+
             JObject options = new JObject
             {
                 { nameof(SamplingSettings), sampling },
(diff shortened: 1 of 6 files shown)


src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
```diff
@@ -105,6 +105,9 @@ public string QuickPulseAuthenticationApiKey
         /// </summary>
         public bool EnableDependencyTracking { get; set; } = true;
 
+
+        public bool EnableSqlCommandTextInstrumentation { get; set; } = false;
```
brettsam: I think we should put this into a new  `DependencyTrackingOptions` type, which then lives here. It looks like there's a handful of other Enable/Disable properties on the module that we can also add to this type and just cover them all at once..

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/DependencyTrackingOptions.cs
```diff
@@ -0,0 +1,48 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
+{
+    public class DependencyTrackingOptions
```
brettsam, line 8: We'll need to initialize some of these that have defaults of `true` to make sure they match if someone doesn't set them. https://github.com/microsoft/ApplicationInsights-dotnet/blob/develop/WEB/Src/DependencyCollector/DependencyCollector/DependencyTrackingTelemetryModule.cs#L47-L96

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
```diff
@@ -159,6 +162,18 @@ public string Format()
                 { nameof(HttpAutoCollectionOptions.EnableResponseHeaderInjection), HttpAutoCollectionOptions.EnableResponseHeaderInjection }
             };
 
+            JObject dependencyTrackingOptions = new JObject
```
brettsam: do we need to check if `DependencyTrackingOptions` is null here?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
```diff
@@ -105,6 +106,8 @@ public string QuickPulseAuthenticationApiKey
         /// </summary>
         public bool EnableDependencyTracking { get; set; } = true;
 
+        public DependencyTrackingOptions DependencyTrackingOptions { get; set; }
```
brettsam, line 112: Add a summary comment here. And be sure to mention that this doesn't apply if `EnableDependencyTracking` is `false`.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
soninaren: fixed

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
soninaren: Fixed

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
```diff
@@ -105,6 +106,11 @@ public string QuickPulseAuthenticationApiKey
         /// </summary>
         public bool EnableDependencyTracking { get; set; } = true;
 
+        /// <summary>
+        /// Configuration for dependency tracking. The dependecny tracking configuration only takes effect if EnableDependencyTracking is set to true
```
brettsam, line 110: nit -- spelling of `dependecny`

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/DependencyTrackingOptions.cs
brettsam: Let's add a comment that these defaults are all coming from the link above. But let's link directly to a commit like: https://github.com/microsoft/ApplicationInsights-dotnet/blob/3933341dc9759a388a0601b47b6108f8570d15b1/WEB/Src/DependencyCollector/DependencyCollector/DependencyTrackingTelemetryModule.cs

brettsam (approved):

brettsam:
@soninaren -- we'll need to pull this into the host and make sure it binds as expected to host.json

Kapkan-Lesion:
@brettsam @soninaren 
Hi! Why should we expect this to become publicly available? Am desperately looking for a way to turn SQL commands back on in my Azure Function's App Insights!

brettsam:
It will be part of the next Functions release. Likely worldwide in a few weeks.

Kapkan-Lesion:
@brettsam Hey Brett, apologies for chasing this feature, when will this feature become available? I understand that you guys have a lot going on especially during this pandemic, just wanted to say thanks for the hard work!

soninaren:
@Kapkan-Lesion This should be available for function runtime version 3.0.15027 or higher.

PR #2611, opened 2020-10-22
Allow binding data to be lazily evaluated

mathewc:
Not all triggered functions actually use binding data when they're invoked. The binding pipeline will only access BindingContext.BindingData when it needs to bind input/output bindings. So for many functions that are simple trigger only, we don't need to build the binding data. However, currently we always build the dictionary eagerly. This can be expensive, depending on the specific trigger.

The PR for updating HttpTrigger to work with this is here: https://github.com/Azure/azure-webjobs-sdk-extensions/pull/675.


brettsam (approved):
Good stuff!

PR #2614, opened 2020-10-23
Reducing the number of async transitions in the core execution pipeline

mathewc:
Crank runs below (5 iterations for each, each iteration 15 seconds) show small improvements from this in RPS/Latency (about 5%).

Before:

| load                   |         |
| ---------------------- | ------- |
| CPU Usage (%)          | 14      |
| Raw CPU Usage (%)      | 28.10   |
| Working Set (MB)       | 32      |
| Build Time (ms)        | 5,779   |
| Start Time (ms)        | 1,296   |
| Published Size (KB)    | 68,222  |
| First Request (ms)     | 256     |
| Requests               | 31,129  |
| Bad responses          | 0       |
| Mean latency (us)      | 123,910 |
| Max latency (us)       | 636,788 |
| Requests/sec           | 2,071   |
| Requests/sec (max)     | 15,055  |
| Read throughput (MB/s) | 0.18    |

After:

| load                   |         |
| ---------------------- | ------- |
| CPU Usage (%)          | 14      |
| Raw CPU Usage (%)      | 27.86   |
| Working Set (MB)       | 32      |
| Build Time (ms)        | 5,933   |
| Start Time (ms)        | 1,221   |
| Published Size (KB)    | 68,222  |
| First Request (ms)     | 278     |
| Requests               | 32,585  |
| Bad responses          | 0       |
| Mean latency (us)      | 118,439 |
| Max latency (us)       | 772,088 |
| Requests/sec           | 2,165   |
| Requests/sec (max)     | 14,842  |
| Read throughput (MB/s) | 0.19    |

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
@@ -74,11 +74,6 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
         {
             var functionInstanceEx = (IFunctionInstanceEx)functionInstance;
             var logger = _loggerFactory.CreateLogger(LogCategories.CreateFunctionCategory(functionInstanceEx.FunctionDescriptor.LogName));
-            return await TryExecuteAsyncCore(functionInstanceEx, cancellationToken, logger);
-        }
-        
-        private async Task<IDelayedException> TryExecuteAsyncCore(IFunctionInstanceEx functionInstanceEx, CancellationToken cancellationToken, ILogger logger)
-        {
             var functionStartedMessage = CreateStartedMessage(functionInstanceEx);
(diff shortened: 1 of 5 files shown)


src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -92,19 +87,15 @@ private async Task<IDelayedException> TryExecuteAsyncCore(IFunctionInstanceEx fu
                 {
                     try
                     {
-                        instanceLogEntry = await NotifyPreBindAsync(functionStartedMessage);
```
mathewc, line 95: The pattern repeated many times below is that we have a small async helper function that does some non-async work then goes async. So you go async twice for each of those. Where it makes sense and keeps the code readable, I've broken these up.

brettsam (approved):
🚢

PR #2621, opened 2020-10-29
Fixing lazy BindingData bug in AmbientBindingContext

mathewc:
Discovered a bug when trying to pull the updated SDK version into Functions Host. The issue was only exposed in out of proc language tests.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingContext.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingContext.cs
@@ -110,7 +110,7 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
             {
                 if (_ambientContext == null)
                 {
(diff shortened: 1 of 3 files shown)


src/Microsoft.Azure.WebJobs.Host/Bindings/BindingContext.cs
```diff
@@ -110,7 +110,7 @@ internal AmbientBindingContext AmbientContext
             {
                 if (_ambientContext == null)
                 {
-                    _ambientContext = new AmbientBindingContext(_functionContext, _bindingData);
+                    _ambientContext = new AmbientBindingContext(_functionContext, () => BindingData);
```
mathewc, line 113: This was the bug - while we were correctly handling lazy evaluation when accessing the BindingData property, this property was not doing so when creating the AmbientBindingContext. This context is used in OOP language Binder scenarios.

brettsam (approved):

PR #2637, opened 2020-11-11
Fixing typo in ExponentialBackoffRetryAttribute

pragnagopa:


src/Microsoft.Azure.WebJobs.Host/ExponentialBackoffRetryAttribute.cs
```diff
@@ -32,7 +32,7 @@ public ExponentialBackoffRetryAttribute(int maxRetryCount, string minimumInterva
                 throw new ArgumentOutOfRangeException(nameof(maximumInterval));
             }
             MinimumInterval = minimumInterval;
-            MaxmumInterval = maximumInterval;
+            MaximumInterval = maximumInterval;
```
pragnagopa, line 35: @mathewc -  @TsuyoshiUshio reported this typo.  This should not effect host as RetryOptions exposed in host are ok. Breaking change would be minimal as on getter is exposed.

mathewc (approved):

PR #2657, opened 2021-01-06
Load ITelemetryProcessors when setting up the TelemetryConfigration

### LINKED ISSUE #3741: ITelemetryProcessor does not appear to be supported in Function Apps
_From @StevenTaylor on October 5, 2018 5:50_

Documentation (https://docs.microsoft.com/en-us/azure/application-insights/app-insights-api-filtering-sampling) suggests that ITelemetryProcessor can be used to filter\sample the telemetry before it is sent App Insights.

But the implementation code is never called in our Function App.

The code to set up the builder is called, which looks like:

`                var builder = TelemetryConfiguration.Active.TelemetryProcessorChainBuilder;            
                builder.Use((next) => new AppInsightsFilter(next));                                
                builder.Build();`

But the 'Process' method implemented in the class is never called, the class looks like:

`    class AppInsightsFilter : ITelemetryProcessor
    {
        private ITelemetryProcessor Next { get; set; }

        // Link processors to each other in a chain.
        public AppInsightsFilter(ITelemetryProcessor next)
        {
            this.Next = next;
        }

        public void Process(ITelemetry item)
        {
            // To filter out an item, just return
            if (!OKtoSend(item)) { return; }

            // Modify the item if required
            //ModifyItem(item);

            this.Next.Process(item);
        }

        private bool OKtoSend(ITelemetry item)
        {            
            // try something!
            return (DateTime.Now.Second <= 20); // only send for the first 20 seconds of a minute
        }
    }`

PS. We have tried setting config values SamplingPercentage (App Insights config) and maxTelemetryItemsPerSecond (host.json) as low as possible to reduce the telemetry data, but there is still too much telemetry.

_Copied from original issue: Azure/Azure-Functions#981_

luthus:
Should fix https://github.com/Azure/azure-functions-host/issues/3741

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
@@ -214,6 +214,7 @@ namespace Microsoft.Extensions.DependencyInjection
                     channel,
                     provider.GetServices<ITelemetryInitializer>(),
                     provider.GetServices<ITelemetryModule>(),
+                    provider.GetServices<ITelemetryProcessorFactory>(),
                     appIdProvider,
                     filterOptions,
                     roleInstanceProvider,
@@ -272,6 +273,7 @@ namespace Microsoft.Extensions.DependencyInjection
             ITelemetryChannel channel,
             IEnumerable<ITelemetryInitializer> telemetryInitializers,
             IEnumerable<ITelemetryModule> telemetryModules,
+            IEnumerable<ITelemetryProcessorFactory> telemetryProcessorFactories,
             IApplicationIdProvider applicationIdProvider,
             LoggerFilterOptions filterOptions,
             IRoleInstanceProvider roleInstanceProvider,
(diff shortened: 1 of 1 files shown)


Steg17 (approved):

Greybird:
Hello,

Any chance to have this fix merged ?

brettsam:
@luthus -- I know this is over a year old. Are you still available to push this through? If not, I can add a simple unit test here and get it in for you.

brettsam:
Pinging @cijothomas here as well for another set of eyes.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
```diff
@@ -352,6 +354,14 @@ internal static LoggerFilterOptions CreateFilterOptions(LoggerFilterOptions regi
                 configuration.TelemetryProcessorChainBuilder.UseSnapshotCollector(options.SnapshotConfiguration);
             }
 
+            if (telemetryProcessorFactories.Any())
+            {
+                foreach (ITelemetryProcessorFactory telemetryProcessorFactory in telemetryProcessorFactories)
+                {
+                    configuration.TelemetryProcessorChainBuilder.Use(telemetryProcessorFactory.Create);
```
cijothomas, line 361: https://github.com/microsoft/ApplicationInsights-dotnet/blob/main/NETCORE/src/Shared/Implementation/TelemetryConfigurationOptionsSetup.cs#L104

ApplicationInsights adds processors to default sink. Not sure if its applicable here.

liliankasem:
@brettsam @luthus @AnatoliB is this something we still want to get merged? Looks like just one more comment to address

RohitRanjanMS:
This approach won’t work as the ITelemeteryProcessorFactory is in Microsoft.ApplicationInsights.AspNetCore and this is marked as private in runtimeAssemblies. Will have to think of another approach to fix this.

Issue #2595, opened 2020-09-30
Expose current retry count in the context of the function


jeffhollan:
Per the new retry feature rolling out, one scenario that is desired is the ability to deadletter / capture something that is on its final retry.  For example, if I have an event hub triggered function and define a retry policy of 5, on the 5th retry if it fails I want to catch that failure and store it in a deadletter queue or something so I can go inspect it later.  However, today the context for the current retry count isn't surfaced directly, and using something like a local variable to increment may be tricky when multiple executions could be executing / retrying on the same host.

Proposal is that there is a new context passed into the `ExecutionContext` that includes information on the retry.

```javascript
try
{
}
catch()
{
   if(context.retries.count >= context.retries.max)
   { 
     // deadletter
   }
   throw exception;
}
```

// @pragnagopa @fabiocav @mathewc

jeffhollan:
And this would need to be piped into Java, Python, C#, PowerShell, and JavaScript context objects - not sure if additional work needed on the language libraries to support this

pragnagopa:
I agree `retrycount`  is useful to include in execution context.  Once this change is pulled into functions host --> Rpc layer needs to be updated to include this info. Language workers need to be updated to consume change in Rpc.

jeffhollan:
FWIW I think just the retry count would be good, retry count + defined maximum slightly better, but best case we pass in the entire `RetryContext` with info on the exception that was hit last

mathewc:
Yes, we might surface RetryContext on ExecutionContext. We'd just have to add MaxRetries to RetryContext - it's not there now. Not sure yet if we should surface that raw type, or have a new type with the members needed. E.g. RetryContext currently has an IFunctionInstance member on it that we might not want to expose here. Need to think about it more.

Another important thing - we need to be sure that users understand that there may actually be 2 levels of retries going on. In addition to the retries we're talking about (e.g. retry a single queue message dispatch N times), the queue trigger binding itself has a MaxDequeueCount which is the second level of retries. This is only an issue for triggers with retries already built in (Queue/ServiceBus). So applying this to your example above, you'd need to check both RetryCount as well as DequeueCount :)

jeffhollan:
Moving comment from @casper-79 here: https://github.com/Azure/azure-functions-java-library/issues/132#issuecomment-721812596

[...]

PR #2658, opened 2021-01-09
Expose current retry count in the context of the function (Issue #2595)

### LINKED ISSUE #2661: ExponentialBackoffRetryAttribute shares does not work properly as it shares state with other executions

alrod:
Related to #2595
Resolves #2661

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/ExecutionContext/ExecutionContext.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/ExecutionContext/ExecutionContext.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using Microsoft.Azure.WebJobs.Host;
 
 namespace Microsoft.Azure.WebJobs
 {
@@ -32,5 +33,10 @@ namespace Microsoft.Azure.WebJobs
         /// A host can set this via <see cref="CoreJobHostConfigurationExtensions.UseCore(JobHostConfiguration, string)"/>
         /// </remarks>
         public string FunctionAppDirectory { get; set; }
+
+        /// <summary>
+        /// Gets or sets value for retry context.
+        /// </summary>
+        public RetryContext RetryContext { get; set; }
     }
 }
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Threading;
+using Microsoft.Azure.WebJobs.Host.Executors;
 using Microsoft.Azure.WebJobs.Host.Protocols;
 using Microsoft.Extensions.DependencyInjection;
 
@@ -18,6 +19,7 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
         private readonly CancellationToken _functionCancellationToken;
         private readonly IServiceProvider _functionInvocationServices;
         private static Lazy<IServiceProvider> _emptyServiceProvider = new Lazy<IServiceProvider>(CreateEmptyServiceProvider);
+        private readonly RetryContext _retryContext;
 
         /// <summary>
         /// Creates a new instance.
@@ -53,6 +55,23 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
         }
 
         /// <summary>
+        /// Creates a new instance.
+        /// </summary>
+        /// <param name="functionInstanceId">The instance ID of the function being bound to.</param>
+        /// <param name="functionCancellationToken">The <see cref="CancellationToken"/> to use.</param>
+        /// <param name="functionInvocationServices">The user logger.</param>
+        /// <param name="functionDescriptor">Current function being executed. </param>
+        /// /// <param name="functionDescriptor">Current retry count. </param>
+        public FunctionBindingContext(IFunctionInstanceEx functionInstance, CancellationToken functionCancellationToken)
+        {
+            _functionInstanceId = functionInstance.Id;
+            _functionCancellationToken = functionCancellationToken;
+            _functionInvocationServices = functionInstance.InstanceServices ?? _emptyServiceProvider.Value;
+            _retryContext = functionInstance.RetryContext;
+            MethodName = functionInstance.FunctionDescriptor?.LogName;
+        }
+
+        /// <summary>
         /// Gets the instance ID of the function being bound to.
         /// </summary>
         public Guid FunctionInstanceId => _functionInstanceId;
@@ -71,6 +90,11 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
         /// The service provider for the current function invocation scope.
         /// </summary>
         public IServiceProvider InstanceServices { get; set; }
+
+        /// <summary>
+        /// Gets value for retry context.
+        /// </summary>
+        public RetryContext RetryContext => _retryContext;
 
         /// <summary>
         /// Creates an object instance with constructor arguments provided in the
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
@@ -92,7 +92,7 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
         public IServiceProvider InstanceServices { get; set; }
 
         /// <summary>
-        /// Gets value for retry context.
+        /// Gets the retry context if this invocation is being retried.
         /// </summary>
         public RetryContext RetryContext => _retryContext;
 
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
@@ -57,17 +57,17 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
         /// <summary>
         /// Creates a new instance.
         /// </summary>
-        /// <param name="functionInstanceId">The instance ID of the function being bound to.</param>
+        /// <param name="functionInstance">The function instance being bound to.</param>
         /// <param name="functionCancellationToken">The <see cref="CancellationToken"/> to use.</param>
-        /// <param name="functionInvocationServices">The user logger.</param>
-        /// <param name="functionDescriptor">Current function being executed. </param>
-        /// /// <param name="functionDescriptor">Current retry count. </param>
         public FunctionBindingContext(IFunctionInstanceEx functionInstance, CancellationToken functionCancellationToken)
         {
             _functionInstanceId = functionInstance.Id;
             _functionCancellationToken = functionCancellationToken;
             _functionInvocationServices = functionInstance.InstanceServices ?? _emptyServiceProvider.Value;
-            _retryContext = functionInstance.RetryContext;
+            if (functionInstance is FunctionInstance fi)
+            {
+                _retryContext = fi.RetryContext;
+            }
             MethodName = functionInstance.FunctionDescriptor?.LogName;
         }
 
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutorExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutorExtensions.cs
@@ -22,7 +22,8 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
             var attempt = 0;
             IDelayedException functionResult = null;
             ILogger logger = null;
-            RetryContext retryContext = new RetryContext();
+            RetryContext retryContext = null;
+            Exception exceptionFromLastExecution = null;
 
             while (true)
             {
@@ -32,16 +33,18 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                     logger = loggerFactory.CreateLogger(LogCategories.CreateFunctionCategory(functionInstance.FunctionDescriptor.LogName));
                 }
 
-                if (functionInstance is FunctionInstance instance)
+                // Set retry context only if retry strategy exists and this is not first attempt
+                if (functionInstance.FunctionDescriptor.RetryStrategy != null && attempt > 0 && functionInstance is FunctionInstance instance)
                 {
+                    retryContext = retryContext ?? new RetryContext();
                     retryContext.RetryCount = attempt;
                     retryContext.Instance = instance;
-                    retryContext.MaxRetryCount = (functionInstance.FunctionDescriptor.RetryStrategy == null) ? 0 : functionInstance.FunctionDescriptor.RetryStrategy.MaxRetryCount;
+                    retryContext.MaxRetryCount = functionInstance.FunctionDescriptor.RetryStrategy.MaxRetryCount;
+                    retryContext.Exception = exceptionFromLastExecution;
 
                     instance.RetryContext = retryContext;
                 }
                 functionResult = await executor.TryExecuteAsync(functionInstance, cancellationToken);
-                retryContext.Exception = functionResult?.Exception;
 
                 if (functionResult == null)
                 {
@@ -53,6 +56,8 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                     // retry is not configured
                     break;
                 }
+
+                exceptionFromLastExecution = functionResult?.Exception;
 
                 IRetryStrategy retryStrategy = functionInstance.FunctionDescriptor.RetryStrategy;
                 if (retryStrategy.MaxRetryCount != -1 && ++attempt > retryStrategy.MaxRetryCount)
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutorExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutorExtensions.cs
@@ -33,8 +33,8 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                     logger = loggerFactory.CreateLogger(LogCategories.CreateFunctionCategory(functionInstance.FunctionDescriptor.LogName));
                 }
 
-                // Set retry context only if retry strategy exists and this is not first attempt
-                if (functionInstance.FunctionDescriptor.RetryStrategy != null && attempt > 0 && functionInstance is FunctionInstance instance)
+                // Set retry context if retryContext is already created
+                if (retryContext != null && functionInstance is FunctionInstance instance)
                 {
                     retryContext.RetryCount = attempt;
                     retryContext.Instance = instance;
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutorExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutorExtensions.cs
@@ -22,7 +22,6 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
             var attempt = 0;
             IDelayedException functionResult = null;
             ILogger logger = null;
-            // Build retry context
             RetryContext retryContext = new RetryContext();
 
             while (true)
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutorExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutorExtensions.cs
@@ -22,6 +22,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
             var attempt = 0;
             IDelayedException functionResult = null;
             ILogger logger = null;
+            RetryContext retryContext = null;
 
             while (true)
             {
@@ -31,7 +32,25 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                     logger = loggerFactory.CreateLogger(LogCategories.CreateFunctionCategory(functionInstance.FunctionDescriptor.LogName));
                 }
 
-                functionResult = await executor.TryExecuteAsync(functionInstance, cancellationToken);
+                if (functionInstance is FunctionInstance instance)
+                {
+                    // Build retry context
+                    retryContext = new RetryContext
+                    {
+                        RetryCount = attempt,
+                        Exception = retryContext == null ? null : retryContext.Exception, // Pass exception from previos attempt
+                        Instance = functionInstance,
+                        RetryStrategy = functionInstance.FunctionDescriptor.RetryStrategy
+                    };
+
+                    instance.RetryContext = retryContext;
+                    functionResult = await executor.TryExecuteAsync(instance, cancellationToken);
+                    retryContext.Exception = functionResult?.Exception;
+                }
+                else
+                {
+                    functionResult = await executor.TryExecuteAsync(functionInstance, cancellationToken);
+                }
 
                 if (functionResult == null)
                 {
@@ -50,14 +69,6 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                     // retry count exceeded
                     break;
                 }
-
-                // Build retry context
-                var retryContext = new RetryContext
-                {
-                    RetryCount = attempt,
-                    Exception = functionResult.Exception,
-                    Instance = functionInstance
-                };
 
                 TimeSpan nextDelay = retryStrategy.GetNextDelay(retryContext);
                 logger.LogFunctionRetryAttempt(nextDelay, attempt, retryStrategy.MaxRetryCount);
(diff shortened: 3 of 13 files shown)


src/Microsoft.Azure.WebJobs.Host/Bindings/ExecutionContext/ExecutionContext.cs
```diff
@@ -32,5 +32,10 @@ public class ExecutionContext
         /// A host can set this via <see cref="CoreJobHostConfigurationExtensions.UseCore(JobHostConfiguration, string)"/>
         /// </remarks>
         public string FunctionAppDirectory { get; set; }
+
+        /// <summary>
+        /// Gets or sets value for retry count.
+        /// </summary>
```
pragnagopa, line 39: can we expose https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/RetryContext.cs instead of just `RetryCount` ? This will ensure out-of-proc language functions have access to retry count as well as the exception details for the failure

pragnagopa:
@alrod  - Please open a separate issue in host repo to track updating RPC layer to include RetryContext and link to this PR

src/Microsoft.Azure.WebJobs.Host/Bindings/ExecutionContext/ExecutionContext.cs
```diff
@@ -32,5 +32,10 @@ public class ExecutionContext
...
+        public int RetryCount { get; set; }
```
mathewc: Issue https://github.com/Azure/azure-webjobs-sdk/issues/2595 mentions that the use case is to be able to check if RetryCount = MaxRetries. This doesn't enable that, right? Was it decided that that wasn't needed? However, the logic for determining whether this is truly the last retry would also have to also take the message dequeue count into account in the case of SB/Queues, since there are two levels of retries going on. E.g. if we're at the max dequeue count AND the max retry, we know this is the very last attempt.

src/Microsoft.Azure.WebJobs.Host/Executors/IFunctionInstance.cs
```diff
@@ -22,6 +22,8 @@ public interface IFunctionInstance
         IFunctionInvoker Invoker { get; }
 
         FunctionDescriptor FunctionDescriptor { get; }
+
+        int RetryCount { get;  }
```
mathewc: This would be a breaking change to a public interface. There's an existing extensibility issue here as evidenced by the IFunctionInstanceEx interface which was added recently to add a single new property. One way we might solve this for now w/o breaking changes:

- add *internal* property RetryCount to FunctionInstance class. Alternatively to address Pragna's suggestion above this could be RetryContext itself
- add *internal* property FunctionInstance to FunctionBindingContext
- in the function executor extension where we're invoking (right before call to TryExecuteAsync), try/cast to FunctionInstance and set RetryCount (or RetryContext)
- in ExecutionContextBindingProvider, try/cast to FunctionInstance to get RetryCount/RetryContext and FunctionInstance.FunctionDescriptor.RetryStrategy.MaxRetries (from FunctionBindingContext) to flow through via ExecutionContext

This keeps it all internal - we can improve in the future w/o any breaking changes.

src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
```diff
@@ -52,6 +53,28 @@ public class FunctionBindingContext
...
 
+        /// <summary>
+        /// Creates a new instance.
+        /// </summary>
+        /// <param name="functionInstanceId">The instance ID of the function being bound to.</param>
+        /// <param name="functionCancellationToken">The <see cref="CancellationToken"/> to use.</param>
+        /// <param name="functionInvocationServices">The user logger.</param>
+        /// <param name="functionDescriptor">Current function being executed. </param>
+        /// /// <param name="functionDescriptor">Current retry count. </param>
+        public FunctionBindingContext(
```
mathewc: Let's just add a new ctor taking parameters IFunctionInstanceEx and CancellationToken. All the values needed come from the function instance. FunctionExecutor can then be simplified to call this overload.

src/Microsoft.Azure.WebJobs.Host/Bindings/ExecutionContext/ExecutionContext.cs
alrod: Good idea. Fixed.

src/Microsoft.Azure.WebJobs.Host/Bindings/ExecutionContext/ExecutionContext.cs
alrod: I did not read #2595 carefully enough. I added `IRetryStratagy` to `RetryPolicy` and exposed 'RetryPolicy' in 'ExecutionContext'.

> However, the logic for determining whether this is truly the last retry would also have to also take the message dequeue count into account in the case of SB/Queues, since there are two levels of retries going on. E.g. if we're at the max dequeue count AND the max retry, we know this is the very last attempt.

We need to point this out in our docs. By design the retry logic only execute the function code inside one instance and do not know about specific trigger retry mechanism(it's possible that another instance will retry execution for a storage/SB message).

@pragnagopa, is it ok add `IRetryStratagy` for RPC layer? or we need to use simple types?

src/Microsoft.Azure.WebJobs.Host/Executors/IFunctionInstance.cs
alrod: Fixed.

src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
alrod: Fixed.

alrod:
> @alrod - Please open a separate issue in host repo to track updating RPC layer to include RetryContext and link to this PR

https://github.com/Azure/azure-functions-host/issues/7048

src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
```diff
@@ -52,6 +54,23 @@ public class FunctionBindingContext
             MethodName = functionDescriptor?.LogName;
         }
 
+        /// <summary>
+        /// Creates a new instance.
+        /// </summary>
+        /// <param name="functionInstanceId">The instance ID of the function being bound to.</param>
+        /// <param name="functionCancellationToken">The <see cref="CancellationToken"/> to use.</param>
+        /// <param name="functionInvocationServices">The user logger.</param>
```
mathewc: These params don't match

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutorExtensions.cs
```diff
@@ -31,7 +32,25 @@ public static async Task<IDelayedException> TryExecuteAsync(this IFunctionExecut
...
                 }
 
-                functionResult = await executor.TryExecuteAsync(functionInstance, cancellationToken);
+                if (functionInstance is FunctionInstance instance)
+                {
+                    // Build retry context
+                    retryContext = new RetryContext
+                    {
+                        RetryCount = attempt,
+                        Exception = retryContext == null ? null : retryContext.Exception, // Pass exception from previos attempt
```
mathewc: nit: spelling "previos"

src/Microsoft.Azure.WebJobs.Host/Executors/IFunctionInstance.cs
```diff
@@ -27,5 +27,7 @@ public interface IFunctionInstance
     public interface IFunctionInstanceEx : IFunctionInstance
     {
         IServiceProvider InstanceServices { get; }
+
+        RetryContext RetryContext { get; }
```
mathewc: Again - this is a breaking change. You can't add new properties to public interfaces - it would break any existing implementations. I don't understand why you're adding this here.

src/Microsoft.Azure.WebJobs.Host/RetryContext.cs
```diff
@@ -21,6 +21,11 @@ public class RetryContext
         /// </summary>
         public Exception Exception { get; set; }
 
+        /// <summary>
+        /// Gets or sets max retry stratagy.
```
mathewc: nit: "Gets or sets the retry strategy"

src/Microsoft.Azure.WebJobs.Host/Bindings/ExecutionContext/ExecutionContext.cs
mathewc: It is a little strange to add IRetryStrategy to RetryContext, because that context was designed to pass into IRetryStrategy.GetNextDelay. We're now wanting to pass RetryContext in other contexts and want more info on it. Another option would be to add MaxRetries to RetryContext. That is also a bit of duplication from the standpoint of an IRetryStrategy implementor, but at least leaves us with a single context object with all the info needed for inspection by other consumers (e.g. ExecutionContext). Thoughts @pragnagopa? I don't think IRetryStrategy itself is needed for ExecutionContext, since that interface is for determining the next delay, something that we wouldn't want function code calling since calling it would advance the retry state.

alrod: anyway, I added MaxRetriesCount to ExecutionContext  instead of IRetryStrategy. I think it fits better.

src/Microsoft.Azure.WebJobs.Host/Executors/IFunctionInstance.cs
alrod: I though that IFunctionInstanceEx  is internal by some reason. Fixed.

src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
alrod: Fixed.

src/Microsoft.Azure.WebJobs.Host/RetryContext.cs
```diff
@@ -11,11 +13,18 @@ namespace Microsoft.Azure.WebJobs.Host
     /// </summary>
     public class RetryContext
     {
+        private Dictionary<string, object> _stateDictionary = new Dictionary<string, object>();
```
mathewc: Make this Lazy so it's only created when needed. E.g. for FixedDelay we'll never need it, so don't want the perf hit of unused object creation.

src/Microsoft.Azure.WebJobs.Host/RetryContext.cs
```diff
@@ -25,5 +34,10 @@ public class RetryContext
         /// Gets or sets the <see cref="IFunctionInstance"/> for the failed invocation.
         /// </summary>
         public IFunctionInstance Instance { get; set; }
+
+        /// <summary>
+        /// Gets the state dictionary.
+        /// </summary>
+        internal IDictionary<string, object> StateDictionary => _stateDictionary;
```
mathewc: Can we just name this "State"? We can also add a <remarks> doc section here to the effect "This property can be used by retry strategies to store arbitrary state for a retried invocation.".

mathewc: Can we just name this "State"? We can also add a <remarks> doc section here to the effect "This property can be used by retry strategies to store arbitrary state for a retried invocation.".

mathewc: Also this needs to be public - it's part of the programming model now for retry strategy authors. IRetryStrategy implementations used across invocations, so they need a place to store per invocation state.

src/Microsoft.Azure.WebJobs.Host/ExponentialBackoffRetryAttribute.cs
```diff
@@ -49,9 +48,21 @@ public ExponentialBackoffRetryAttribute(int maxRetryCount, string minimumInterva
...
+            IDelayStrategy delayStrategy;
+
+            if (context.StateDictionary.TryGetValue(nameof(RandomizedExponentialBackoffStrategy), out object stateObject))
+            {
+                delayStrategy = stateObject as IDelayStrategy;
+            }
+            else
+            {
+                delayStrategy = new RandomizedExponentialBackoffStrategy(_parsedMinimumInterval, _parsedMaximumInterval);
+                context.StateDictionary.Add(nameof(RandomizedExponentialBackoffStrategy), new RandomizedExponentialBackoffStrategy(_parsedMinimumInterval, _parsedMaximumInterval));
```
mathewc: for the dictionary key, can we use something like _msDelayStrategy? In general when we add fx properties to user accessible collecitons we use a MS prefix to avoid collisions.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutorExtensions.cs
```diff
@@ -31,7 +32,25 @@ public static async Task<IDelayedException> TryExecuteAsync(this IFunctionExecut
                     logger = loggerFactory.CreateLogger(LogCategories.CreateFunctionCategory(functionInstance.FunctionDescriptor.LogName));
                 }
 
-                functionResult = await executor.TryExecuteAsync(functionInstance, cancellationToken);
+                if (functionInstance is FunctionInstance instance)
+                {
+                    // Build retry context
+                    retryContext = new RetryContext
```
mathewc: Need to copy State as well here. If you don't have a test that is verifying that custom state is handled properly across retries, please add one. I wonder if to facilitate this, we shouldn't just make State a settable property rather than a Lazy.

src/Microsoft.Azure.WebJobs.Host/ExponentialBackoffRetryAttribute.cs
```diff
@@ -49,9 +48,21 @@ public ExponentialBackoffRetryAttribute(int maxRetryCount, string minimumInterva
 
         public override TimeSpan GetNextDelay(RetryContext context)
         {
+            IDelayStrategy delayStrategy;
+
+            if (context.StateDictionary.TryGetValue(nameof(RandomizedExponentialBackoffStrategy), out object stateObject))
```
mathewc: I just want to point out that this new State dictionary now enables our ExponentialBackoffRetryAttribute implementation as it existed. Note that if we weren't reusing the RandomizedBackoffRetryAttribute stateful class, this retry strategy could have been implemented without using State, since all required exponential backoff state can be determined/incremented solely based on the retry count which is maintained on the RetryContext.

So our addition of State is good since it adds flexibility for strategy implementors, but it'll be rare that people actually need to use it.

src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
```diff
@@ -72,6 +91,11 @@ public class FunctionBindingContext
         /// </summary>
         public IServiceProvider InstanceServices { get; set; }
 
+        /// <summary>
+        /// Gets value for retry context.
+        /// </summary>
+        public RetryContext RetryContext => _retryContext;
```
mathewc, line 97: nit: "Gets the retry context if this invocation is being retried."

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutorExtensions.cs
```diff
@@ -31,7 +32,25 @@ public static async Task<IDelayedException> TryExecuteAsync(this IFunctionExecut
...
+                    {
+                        RetryCount = attempt,
+                        Exception = retryContext == null ? null : retryContext.Exception, // Pass exception from previous attempt
+                        Instance = functionInstance,
+                        MaxRetryCount = (functionInstance.FunctionDescriptor.RetryStrategy == null) ? 0 : functionInstance.FunctionDescriptor.RetryStrategy.MaxRetryCount
+                    };
+
+                    instance.RetryContext = retryContext;
+                    functionResult = await executor.TryExecuteAsync(instance, cancellationToken);
```
mathewc: Looks like this code can be simplified by removing the else block below and executing these two lines outside of this if.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutorExtensions.cs
```diff
@@ -31,7 +33,16 @@ public static async Task<IDelayedException> TryExecuteAsync(this IFunctionExecut
                     logger = loggerFactory.CreateLogger(LogCategories.CreateFunctionCategory(functionInstance.FunctionDescriptor.LogName));
                 }
 
+                if (functionInstance is FunctionInstance instance)
+                {
+                    retryContext.RetryCount = attempt;
+                    retryContext.Instance = instance;
```
pragnagopa, line 37: nit: move this below condition ```if (functionInstance.FunctionDescriptor.RetryStrategy == null)``` and move ```retryContext.Exception = functionResult?.Exception;``` inside if block  , this will avoid initializing retry context if RetryStrategy is not specified.

src/Microsoft.Azure.WebJobs.Host/ExponentialBackoffRetryAttribute.cs
```diff
@@ -13,7 +13,9 @@ namespace Microsoft.Azure.WebJobs
     /// </summary>
     public class ExponentialBackoffRetryAttribute : RetryAttribute
     {
-        private IDelayStrategy _delayStrategy;
+        private TimeSpan _parsedMinimumInterval;
+        private TimeSpan _parsedMaximumInterval;
+        private static string _delayStrategyKeyName = "MS_DelayStratagy";
```
pragnagopa, line 18: nit: Move this to constants?

test/Microsoft.Azure.WebJobs.Host.UnitTests/Retry/RetryAttributeTests.cs
```diff
@@ -62,7 +62,6 @@ public void ExponentialBackOffDelay_Constructor_Throws()
         {
             Assert.Throws<ArgumentOutOfRangeException>(() => new ExponentialBackoffRetryAttribute(5, "01:020000000000:25", "00:00:10"));
             Assert.Throws<ArgumentOutOfRangeException>(() => new ExponentialBackoffRetryAttribute(5, "01:02:25", "100000000"));
-            Assert.Throws<ArgumentException>(() => new ExponentialBackoffRetryAttribute(5, "05:02:25", "00:00:10"));
```
pragnagopa: intentionally removed?

pragnagopa (commented):
Thanks for the PR. Added few comments. Since this is changing public surface, it would be good to prototype changes on the host and Rpc layer to ensure no other changes in SDK are needed.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Retry/RetryAttributeTests.cs
alrod: yes, the `ExponentialBackoffRetryAttribute` ctr does not throw the exception anymore.

src/Microsoft.Azure.WebJobs.Host/ExponentialBackoffRetryAttribute.cs
alrod: discussed with Mathew, you are ok to leave it here

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutorExtensions.cs
alrod: it does not work as we need to set the properties before the execution. we need the if to cast to `FunctionInstance` to access internal property `instance.RetryContext`.

alrod:
> Thanks for the PR. Added few comments. Since this is changing public surface, it would be good to prototype changes on the host and Rpc layer to ensure no other changes in SDK are needed.

I checked the change with host/node worker locally:

https://github.com/alrod/azure-webjobs-sdk-script/tree/alrod/rpcretry
https://github.com/Azure/azure-functions-nodejs-worker/tree/alrod/retryCount

test/Microsoft.Azure.WebJobs.Host.UnitTests/Retry/RetryAttributeTests.cs
pragnagopa: good to have this validation in place to avoid adding additional validation logic in the host. What do you think?

src/Microsoft.Azure.WebJobs.Host/RetryContext.cs
```diff
@@ -11,11 +13,18 @@ namespace Microsoft.Azure.WebJobs.Host
...
+        private Lazy<Dictionary<string, object>> _stateDictionary = new Lazy<Dictionary<string, object>>(() => new Dictionary<string, object>());
+
         /// <summary>
         /// Gets or sets the current retry count.
         /// </summary>
         public int RetryCount { get; set; }
 
+        /// <summary>
+        /// Gets the max retry count.
+        /// </summary>
```
pragnagopa, line 25: nit: Gets or sets

pragnagopa (approved):
Thanks for the prototyping changes in RPC layer!  Looking at https://github.com/alrod/azure-webjobs-sdk-script/tree/alrod/rpcretry , we should define "RetryContext" to ensure we can expose all the properties - including exception. We can discuss these changes in https://github.com/Azure/azure-functions-language-worker-protobuf/issues/52 . Rest looks OK.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Retry/RetryAttributeTests.cs
mathewc: You can retain the validation by refactoring the RandomizedExponentialBackoffStrategy ctor validation into a static Validate method on the RandomizedExponentialBackoffStrategy class that both that class ctor and ExponentialBackoffRetryAttribute ctor can call.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Retry/RetryAttributeTests.cs
```diff
@@ -81,5 +80,44 @@ public void ExponentialBackOffDelay_NextDelay_ReturnsExpected()
...
+            {
+                RetryCount = 1
+            };
+            var delay1 = retry.GetNextDelay(retryContext);
+            var delay2 = retry.GetNextDelay(retryContext);
+            Assert.True(delay2 > delay1);
+        }
+
+        [Fact]
+        public void ExponentialBackOffDelay_GetNextDelay_Throws()
```
mathewc: If you do the Validate refactor suggested above, this test can go away in favor of the existing ctor tests.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutorExtensions.cs
```diff
@@ -22,6 +22,8 @@ public static async Task<IDelayedException> TryExecuteAsync(this IFunctionExecut
             var attempt = 0;
             IDelayedException functionResult = null;
             ILogger logger = null;
+            // Build retry context
```
mathewc: nit: remove comment

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutorExtensions.cs
mathewc: Pragna does have a point though - we should only set this if the function actually uses a RetryStrategy, so we should expand the if clause here to also check RetryStrategy != null. Furthermore, can we also only create/set the context if attempt > 0? I.e. ExecutionContext.RetryContext is only set for failed invocations when we're doing retries? That allows us to avoid overhead of object creations for normal invocations.

We'd also move the creation of the retry context into this block as well then: retryContext = retryContext ?? new RetryContext();

pragnagopa: Yes, I agree. RetryContext should only be initialized if "RetryStrategy" is used by a function

mathewc (approved):

pragnagopa (approved):

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutorExtensions.cs
```diff
@@ -36,7 +36,6 @@ public static async Task<IDelayedException> TryExecuteAsync(this IFunctionExecut
                 // Set retry context only if retry strategy exists and this is not first attempt
                 if (functionInstance.FunctionDescriptor.RetryStrategy != null && attempt > 0 && functionInstance is FunctionInstance instance)
                 {
-                    retryContext = retryContext ?? new RetryContext();
```
mathewc: I see we can also simplify the above if clause (and comment) - we only need `if(retryContext != null && functionInstance is FunctionInstance instance)` since attempt will always be > 0 at that point, and we short circuit below if RetryStrategy is null.

mathewc (approved):

mathewc (approved):

Issue #2645, opened 2020-11-27
FunctionIndexingException thrown when same function method name is used in different classes


djjlewis:
Having a `[ServiceBusTrigger]` attribute `Message` parameter on one or methods with the same name, but in different classes, throws `Microsoft.Azure.WebJobs.Host.Indexers.FunctionIndexingException: Error indexing method  System.ArgumentException: An item with the same key has already been added.` in `FunctionIndexer.cs` line 93 when the webjob console app starts.

It appears this broke from version `3.0.22` as `3.0.19` still works OK.

#### Repro steps

1. Create a Webjob console application and reference version >= 3.0.22.

2. Create two classes `MessageOneHandler` and `MessageTwoHandler`.

3. Add methods to each class with the same name and containing a ServiceBusTrigger attribute parameter: `HandleMessage([ServiceBusTrigger(...)] Message message) {...} `

4. Run the console application

#### Expected behavior

No exception is thrown and both methods are invoked for the service bus topic / queue they are configured to use.

#### Actual behavior

Exception above is thrown:

`Microsoft.Azure.WebJobs.Host.Indexers.FunctionIndexingException: Error indexing method  System.ArgumentException: An item with the same key has already been added.` in `FunctionIndexer.cs` line 93:

```   at Microsoft.Azure.WebJobs.Host.RecoverableException.TryRecover(ILogger logger) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Exceptions\RecoverableException.cs:line 83
   at Microsoft.Azure.WebJobs.Host.Indexers.FunctionIndexer.IndexTypeAsync(Type type, IFunctionIndexCollector index, CancellationToken cancellationToken) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Indexers\FunctionIndexer.cs:line 98
   at Microsoft.Azure.WebJobs.Host.Indexers.FunctionIndexProvider.CreateAsync(CancellationToken cancellationToken) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Indexers\FunctionIndexProvider.cs:line 95
[...]

PR #2669, opened 2021-02-02
Fix name collision in indexer cache (#2645)

mathewc:
Fix for https://github.com/Azure/azure-webjobs-sdk/issues/2645. As noted in the issue, the bug was introduced in https://github.com/Azure/azure-webjobs-sdk/commit/76d44ca915b817ed3f3cc46e355aa6cb0f741b4b when a method name cache was introduced.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndex.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndex.cs
@@ -34,9 +34,17 @@ namespace Microsoft.Azure.WebJobs.Host.Indexers
             _functionsByMethod.Add(method, function);
 
             // For compat, accept either the short name ("Class.Name") or log name (just "Name")
-            _functionsByName.Add(descriptor.LogName, function);
-            if (descriptor.ShortName != descriptor.LogName)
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndex.cs
```diff
@@ -34,7 +34,12 @@ public void Add(IFunctionDefinition function, FunctionDescriptor descriptor, Met
             _functionsByMethod.Add(method, function);
 
             // For compat, accept either the short name ("Class.Name") or log name (just "Name")
-            _functionsByName.Add(descriptor.LogName, function);
+            if (!_functionsByName.ContainsKey(descriptor.LogName))
```
mathewc, line 37: This fix retains the previous logic of first one wins that existed before cache was added. See previous logic [here](https://github.com/Azure/azure-webjobs-sdk/commit/76d44ca915b817ed3f3cc46e355aa6cb0f741b4b#diff-1063e70146286d93a5d60e9e0c2cf6043c3fd84e07199fec91bc585f8b087eeaL54).

fabiocav (approved):

PR #2679, opened 2021-03-02
WEBSITE_CLOUD_ROLENAME overrides site name in Application Insights

### LINKED ISSUE #2635: AppInsights - Override function app properties in logs with Static Web Apps resource information
2020/12/10 updating to only overriding cloud rolename. The remainder will be in #2654 

https://github.com/Azure/static-web-apps/issues/204

@annaji-msft can you share more details on what properties to override and where to get the value from?

@brettsam FYI

kashimiz:
Resolves #2635

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsRoleEnvironmentTelmetryInitializer.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsRoleEnvironmentTelmetryInitializer.cs
@@ -19,7 +19,7 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
     {
         internal const string AzureWebsiteName = "WEBSITE_SITE_NAME";
         internal const string AzureWebsiteSlotName = "WEBSITE_SLOT_NAME";
-        internal const string AzureWebsiteCloudRolename = "WEBSITE_CLOUD_ROLENAME";
+        internal const string AzureWebsiteCloudRoleName = "WEBSITE_CLOUD_ROLENAME";
         private const string DefaultProductionSlotName = "production";
         private const string WebAppSuffix = ".azurewebsites.net";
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsRoleEnvironmentTelmetryInitializer.cs
```diff
@@ -35,12 +36,22 @@ public void Initialize(ITelemetry telemetry)
                 return;
             }
 
+            Lazy<string> websiteCloudRoleName = new Lazy<string>(() =>
+            {
+                return Environment.GetEnvironmentVariable(AzureWebsiteCloudRolename);
+            });
+
```
brettsam: I don't think we need this one wrapped in a `Lazy` because:
1. It's a much simpler lookup than `GetAzureWebsiteUniqueSlotName`.
2. It appears to always be required, so the Lazy will always be immediately evaluated anyway.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Loggers/WebJobsRoleEnvironmentTelemetryInitializerTests.cs
```diff
@@ -2,70 +2,83 @@
...
         {
-            Environment.SetEnvironmentVariable(WebJobsRoleEnvironmentTelemetryInitializer.AzureWebsiteName, websiteName);
-            Environment.SetEnvironmentVariable(WebJobsRoleEnvironmentTelemetryInitializer.AzureWebsiteSlotName, slotName);
-        }
+            var testCloudRoleName = "mycloudrolename";
 
-        public void Dispose()
-        {
-            SetEnvironmentVariables(null, null);
+            using (EnvVarHolder.Set(WebJobsRoleEnvironmentTelemetryInitializer.AzureWebsiteCloudRolename, testCloudRoleName))
```
brettsam: Can you have this test also set `AzureWebsiteName` and `AzureWebsiteSlotName` to something (like the test above). That way we can ensure that `AzureWebsiteCloudRolename` wins if all are specified.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsRoleEnvironmentTelmetryInitializer.cs
```diff
@@ -19,6 +19,7 @@ internal class WebJobsRoleEnvironmentTelemetryInitializer : ITelemetryInitialize
     {
         internal const string AzureWebsiteName = "WEBSITE_SITE_NAME";
         internal const string AzureWebsiteSlotName = "WEBSITE_SLOT_NAME";
+        internal const string AzureWebsiteCloudRolename = "WEBSITE_CLOUD_ROLENAME";
```
brettsam: nit: capitalize "Name".

brettsam (commented):
Nice! Couple minor comments...

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsRoleEnvironmentTelmetryInitializer.cs
kashimiz: Done

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsRoleEnvironmentTelmetryInitializer.cs
kashimiz: Done

test/Microsoft.Azure.WebJobs.Host.UnitTests/Loggers/WebJobsRoleEnvironmentTelemetryInitializerTests.cs
kashimiz: Done

brettsam (approved):

PR #2680, opened 2021-03-02
[pack] Adding Track2 Storage Providers for Blobs and Queues

karshinlin:
- New packages for Azure Blob and Queue Storage Providers.
- Adds necessary extension methods to implement the common StorageProvider
### Note: 
- This change brings in the dependency to Azure Storage SDKs v12 
- There's legacy logging that relies on v11 but these can probably be stripped out


src/Microsoft.Azure.WebJobs.Host.Storage/HostStorageProvider.cs
```diff
@@ -0,0 +1,70 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using Azure.Storage.Blobs;
+using Microsoft.Azure.WebJobs.StorageProvider.Blobs;
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.Azure.WebJobs
+{
+    public class HostStorageProvider
```
pakrym: Can we add Try** methods to the `BlobServiceClientProvider` and avoid an extra level of the hierarchy?

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/PublicSurfaceTests.cs
```diff
@@ -21,12 +21,22 @@ public void WebJobs_Host_Storage_VerifyPublicSurfaceArea()
 
             var expected = new[]
             {
+                "AzureBlobLeaseProvider",
```
karshinlin: Need sign off on these new public types @fabiocav

src/Microsoft.Azure.WebJobs.Host.Storage/AzureStorageProvider.cs
```diff
@@ -0,0 +1,94 @@
...
+        public virtual BlobContainerClient GetBlobContainerClient()
+        {
+            if (_storageOptions?.InternalSasBlobContainer != null)
+            {
+                return new BlobContainerClient(new Uri(_storageOptions.InternalSasBlobContainer));
+            }
+
+            if (!TryGetBlobServiceClientFromConnection(out BlobServiceClient blobServiceClient, ConnectionStringNames.Storage))
+            {
+                throw new InvalidOperationException("Could not create BlobServiceClient for the BlobLeaseDistributedLockManager");
```
karshinlin: nit: describe better for throw

src/Microsoft.Azure.WebJobs.Host.Storage/JobHostInternalStorageOptions.cs
```diff
@@ -1,9 +1,9 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
-using Microsoft.Extensions.Options;
-using Microsoft.Azure.Storage.Blob;
 using System;
+using Microsoft.Azure.Storage.Blob;
+using Microsoft.Extensions.Options;
```
fabiocav: nit: I don't believe these are required with the changes you've made.

src/Microsoft.Azure.WebJobs.Host.Storage/RuntimeStorageWebJobsBuilderExtensions.cs
```diff
@@ -44,6 +44,13 @@ public static IWebJobsBuilder AddAzureStorageCoreServices(this IWebJobsBuilder b
 
             return builder;
         }
+        
+        // Helper to add storage services for internal operations
+        public static void AddAzureStorageCoreServices(this IServiceCollection services)
+        {
+            var tempBuilder = new WebJobsBuilder(services);
```
fabiocav: We shouldn't be creating the builder here. If there isn't already an extension that performs the registration against the service collection, one should be introduced.

src/Microsoft.Azure.WebJobs.Shared.StorageProvider/StorageClientProvider.cs
```diff
@@ -0,0 +1,247 @@
...
+
+namespace Microsoft.Azure.WebJobs.Shared.StorageProvider
+{
+    /// <summary>
+    /// Abstraction to provide storage accounts from the connection names.
+    /// This gets the storage account name via the binding attribute's <see cref="IConnectionProvider.Connection"/>
+    /// property.
+    /// If the connection is not specified on the attribute, it uses a default account.
+    /// </summary>
+    public abstract class StorageClientProvider<TClient, TClientOptions> where TClientOptions : ClientOptions
```
fabiocav: As discussed offline, this type in the shared project should not be public.

src/Microsoft.Azure.WebJobs.Shared.StorageProvider/SkuUtility.cs
```diff
@@ -0,0 +1,33 @@
...
+using System;
+
+namespace Microsoft.Azure.WebJobs.Shared.StorageProvider
+{
+    internal static class SkuUtility
+    {
+        private const string AzureWebsiteSku = "WEBSITE_SKU";
+        private const string DynamicSku = "Dynamic";
+
+        private static readonly Lazy<bool> s_isDynamicSku = new Lazy<bool>(() =>
```
fabiocav: remove `s_` prefix throughout.

src/Microsoft.Azure.WebJobs.Shared.StorageProvider/StorageClientProvider.cs
```diff
@@ -0,0 +1,247 @@
...
+        /// <returns>retrieval success</returns>
+        protected virtual bool TryGetServiceUri(IConfiguration configuration, out Uri serviceUri)
+        {
+            try
+            {
+                var serviceUriConfig = string.Format(CultureInfo.InvariantCulture, "{0}ServiceUri", ServiceUriSubDomain);
+
+                if (configuration.GetValue<string>("accountName") != null)
+                {
+                    var accountName = configuration.GetValue<string>("accountName");
```
fabiocav: Avoid duplicating the call to `GetValue` in both cases. You can use an inline declaration here.

src/Microsoft.Azure.WebJobs.Shared.StorageProvider/StorageClientProvider.cs
```diff
@@ -0,0 +1,230 @@
...
+        {
+            // If connection string is present, it will be honored first
+            if (!IsConnectionStringPresent(configuration) && TryGetServiceUri(configuration, out Uri serviceUri))
+            {
+                var constructor = typeof(TClient).GetConstructor(new Type[] { typeof(Uri), typeof(TokenCredential), typeof(TClientOptions) });
+                return (TClient)constructor.Invoke(new object[] { serviceUri, tokenCredential, options });
+            }
+
+            return (TClient)_componentFactory.CreateClient(typeof(TClient), configuration, tokenCredential, options);
+        }
```
fabiocav: Let's avoid all of the reflection code here making assumptions about the constructor signature. Let's just make this an abstract method and have the derived types implement the instantiation. The other option would be to introduce a factory here, but the abstract method approach should suffice.

src/Microsoft.Azure.WebJobs.Shared.StorageProvider/StorageClientProvider.cs
```diff
@@ -0,0 +1,230 @@
...
+            try
+            {
+                var configuration = new ConfigurationBuilder().AddInMemoryCollection(new[]
+                {
+                    new KeyValuePair<string, string>("connectionString", connectionString)
+                }).Build();
+
+                // AzureComponentFactory assumes IConfigurationSection with a value is a connection string
+                var configSection = configuration.GetSection("connectionString");
+                client = CreateClient(configSection, null, null);
```
fabiocav: This is a fair amount of work just to pass the connection string. Can we have an overload of `CreateClient` that directly takes the connection string instead so we can avoid all of this?

src/Microsoft.Azure.WebJobs.Shared.StorageProvider/StorageClientProvider.cs
```diff
@@ -0,0 +1,230 @@
...
+        }
+
+        /// <summary>
+        /// Creates client options from the given configuration
+        /// </summary>
+        /// <param name="configuration">Registered <see cref="IConfiguration"/></param>
+        /// <returns>Client options</returns>
+        protected virtual TClientOptions CreateClientOptions(IConfiguration configuration)
+        {
+            var clientOptions = (TClientOptions)_componentFactory.CreateClientOptions(typeof(TClientOptions), null, configuration);
```
fabiocav: Is there a reason we're using the component factory here instead of using the standard options patterns?

src/Microsoft.Azure.WebJobs.StorageProvider.Blobs/AzureBlobBuilderExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Microsoft.Azure.WebJobs.StorageProvider.Blobs
+{
+    /// <summary>
+    /// Extension methods for Storage Blobs integration.
+    /// </summary>
+    public static class AzureBlobBuilderExtensions
+    {
+        /// <summary>
+        /// Adds the Storage Blobs extension to the provided <see cref="IServiceCollection"/>.
```
fabiocav: Is this description accurate? This adds the host services. Extensions, in the context of WebJobs, are something different.

src/Microsoft.Azure.WebJobs.StorageProvider.Blobs/BlobServiceClientProvider.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Microsoft.Azure.WebJobs.StorageProvider.Blobs
+{
+    /// <summary>
+    /// Provider to create BlobServiceClient objects with the registered <see cref="IConfiguration"/>
+    /// </summary>
+    public class BlobServiceClientProvider : StorageClientProvider<BlobServiceClient, BlobClientOptions>
+    {
+        /// <summary>
+        /// BlobStorageClientProvider
```
fabiocav: nit: Better description here.

src/Microsoft.Azure.WebJobs.StorageProvider.Blobs/BlobServiceClientProvider.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// </summary>
+        /// <param name="configuration">Registered <see cref="IConfiguration"/></param>
```
fabiocav: nit: better description for the parameters as well. Unclear what `Registered` would mean if actually using this constructor.

build/common.props
```diff
@@ -1,10 +1,11 @@
 <Project>
   <PropertyGroup>
     <!-- Packages can have independent versions and only increment when released -->
-    <Version>3.0.29$(VersionSuffix)</Version>
+    <Version>3.1.0$(VersionSuffix)</Version>
     <ExtensionsStorageVersion>4.0.5$(VersionSuffix)</ExtensionsStorageVersion>
-    <HostStorageVersion>4.0.2$(VersionSuffix)</HostStorageVersion>
+    <HostStorageVersion>4.1.0$(VersionSuffix)</HostStorageVersion>
     <LoggingVersion>4.0.2$(VersionSuffix)</LoggingVersion>
+    <StorageProviderVersion>1.0.0$(VersionSuffix)</StorageProviderVersion>
```
fabiocav: These need to be flagged as beta if we're expecting them to depend on beta packages.

src/Microsoft.Azure.WebJobs.StorageProvider.Queues/AzureQueueBuilderExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace Microsoft.Azure.WebJobs.StorageProvider.Queues
+{
+    /// <summary>
+    /// Extension methods for Storage Queues integration.
+    /// </summary>
+    public static class AzureQueueBuilderExtensions
+    {
+        /// <summary>
+        /// Adds the Storage Queues extension to the provided <see cref="IServiceCollection"/>.
```
fabiocav: Same comment about the `extension` term here.

src/Microsoft.Azure.WebJobs.StorageProvider.Queues/QueueServiceClientProvider.cs
```diff
@@ -0,0 +1,30 @@
...
+using Azure.Storage.Queues;
+using Microsoft.Azure.WebJobs.Shared.StorageProvider;
+using Microsoft.Extensions.Azure;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.StorageProvider.Queues
+{
+    /// <summary>
+    /// Provider to create QueueServiceClient objects with the registered <see cref="IConfiguration"/>
```
fabiocav: Update the description as `registered` here doesn't apply.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/InternalStorageTests.cs
```diff
@@ -13,14 +13,15 @@
...
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.Azure.WebJobs.Host.EndToEndTests
 {
     public class InternalStorageTests
     {
         // End-2-end test that we can run a JobHost from purely a SAS connection string. 
         [Fact]
-        public async Task Test()
+        public async Task Test1()
```
fabiocav: Better/more descriptive name?

fabiocav (commented):
Initial round of feedback.

Looking at the tests, looks like there's a fair amount of new surface being added that don't seem to have test coverage. Are you planning on bringing additional tests in?

fabiocav:
@mathewc also expressed concerns about the added public API surface for some of the extension methods. Can you review those and make sure they're all still needed?

src/Microsoft.Azure.WebJobs.Shared.StorageProvider/StorageClientProvider.cs
karshinlin: Don't think this is possible. Base classes can't be less accessible than derived classes: https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs0060?f1url=%3FappId%3Droslyn%26k%3Dk(CS0060)

src/Microsoft.Azure.WebJobs.Host.Storage/JobHostInternalStorageOptions.cs
karshinlin: Microsoft.Azure.Storage.Blob needed for CloudBlobContainer (existing implementation - no change in this pr)
Microsoft.Extensions.Options needed for IOptions (no change in this pr). 

I just changed the ordering to put System first.

src/Microsoft.Azure.WebJobs.Host.Storage/RuntimeStorageWebJobsBuilderExtensions.cs
karshinlin: This is just a helper that performs registration against a service collection and repurposes the existing extension method. I can copy over the registrations from the other method to remove the temp builder.

src/Microsoft.Azure.WebJobs.Shared.StorageProvider/StorageClientProvider.cs
fabiocav: Yes :) You can't have a derived type with broader visibility. 

To make this clearer; my comment was that we can't have this type be public if it's being pulled into distinct projects/defined in a shared project. This is what we discussed offline. Your options would be to either introduce a proper project for the abstraction, or add it to the existing project (which I believe that's what we had discussed) and reference that from the others

karshinlin: Created WebJobs.StorageProvider.Common to hold this common StorageProvider. This avoids bringing v12 storage into any existing packages.

src/Microsoft.Azure.WebJobs.Host.Storage/RuntimeStorageWebJobsBuilderExtensions.cs
karshinlin: Moved this overload with IServiceCollection to StorageServiceCollectionExtensions.cs (already existed, same namespace). I call this new extension method in RuntimeStorageWebJobsBuilderExtensions.cs here.

src/Microsoft.Azure.WebJobs.Shared.StorageProvider/SkuUtility.cs
karshinlin: Sku logic no longer in webjobs.

src/Microsoft.Azure.WebJobs.Shared.StorageProvider/StorageClientProvider.cs
karshinlin: This logic in componentFactory is already shipped and handles all the client creation. It's probably best to unify client creation in the componentFactory rather than creating it ourselves (unless we have a different configuration like the serviceUri specification)

src/Microsoft.Azure.WebJobs.Host.Storage/JobHostInternalStorageOptions.cs
```diff
@@ -40,5 +40,7 @@ public DistributedLockManagerContainerProvider(IOptions<JobHostInternalStorageOp
     public class JobHostInternalStorageOptions
     {
         public string InternalSasBlobContainer { get; set; }
+
+        public string InternalContainerName { get; set; }
```
mathewc: This new public member isn't used anywhere in the WebJobs SDK. This public contract implies that if this is set, we'll use this container across the board rather than the default "azure-webjobs-hosts", which isn't the case. E.g. if this were set, it should result in InternalContainer above for the lock manager using this option.

Even before your changes, these blob container options aren't actually honored as the code summary intends. It's only the lock manager that is using InternalSasBlobContainer. If you're adding InternalContainerName, I think it should be honored above as well for symmetry.

mathewc: Actually I see in the code below the comment "$$$ get rid of LegacyConfig" where this type is used. So this was intended to be temporary all along. Rather than expanding this, perhaps we just remove this and in Functions Host you just use the default container name? Not sure we need to make this configurable.

karshinlin: That would be the simplest path if we don't need a configurable container. When I tested this override container, only the Lock Manager uses it, FunctionsSyncManager, secrets, etc. all use the default container. 

If we did want to support it, then we wouldn't be able to incorporate it as part of DistributedLockManagerContainerProvider since we need access to the StorageAccountOptions (internal object). We _could_ implement as part of the Create() method for the Lock Manager.

src/Microsoft.Azure.WebJobs.Host/Extensions/IConfigurationExtensions.cs
```diff
@@ -93,5 +93,37 @@ public static bool IsSettingEnabled(this IConfiguration configuration, string se
...
+            return section;
+        }
+
+        /// <summary>
+        /// Looks for a connection string by first checking the ConfigurationStrings section, and then the root.
+        /// </summary>
+        /// <param name="configuration">The configuration.</param>
+        /// <param name="connectionName">The connection string key.</param>
+        /// <returns></returns>
+        public static IConfigurationSection GetConnectionStringOrSettingSection(this IConfiguration configuration, string connectionName)
```
mathewc: Doesn't need to be public right, since it appears it's only used by the method above?

src/Microsoft.Azure.WebJobs.Host/Extensions/IConfigurationExtensions.cs
```diff
@@ -93,5 +93,37 @@ public static bool IsSettingEnabled(this IConfiguration configuration, string se
             
             return false;
         }
+
+        public static IConfigurationSection GetWebJobsConnectionStringSection(this IConfiguration configuration, string connectionStringName)
```
mathewc: Does this new public method have to live in this assembly, or will it work if it lives in the Microsoft.Azure.WebJobs.StorageProvider.Common? Trying to scope public surface changes to the new stopgap assemblies

karshinlin: I think it can live in the StorageProvider.Common assembly. I originally put it in WebJobs.Host since it already had similar extension methods.

karshinlin: Actually the GetWebJobsConnectionStringSection method is used in WebJobs.Script.WebHost so if we moved it into the StorageProvider.Common package, we would need to add that package directly I think. Since this is the only needed method, I can make the others private/internal, but this logic may be more suited to be in WebJobs.Host. Let me know what you think.

src/Microsoft.Azure.WebJobs.Host/Extensions/IConfigurationExtensions.cs
karshinlin: right. I'll make it private

src/Microsoft.Azure.WebJobs.Host.Storage/JobHostInternalStorageOptions.cs
karshinlin: Talked with Fabio. InternalContainerName would be added behavior - I'm going to just take it out, since the new concrete lock implementation in the host already supports the sas container override.

src/Microsoft.Azure.WebJobs.Shared.StorageProvider/StorageClientProvider.cs
fabiocav: I don't think that addresses the initial question. This is for creation of the options class, not the client. Is there a reason this type is not just taking a dependency on `IOptions<TClientOptions>` instead?

src/Microsoft.Azure.WebJobs.StorageProvider.Common/StorageClientProvider.cs
```diff
@@ -0,0 +1,222 @@
...
+
+            // $$$ Where does validation happen?
+            IConfigurationSection connectionSection = _configuration.GetWebJobsConnectionStringSection(name);
+            if (!connectionSection.Exists())
+            {
+                // Not found
+                throw new InvalidOperationException($"Storage account connection string '{IConfigurationExtensions.GetPrefixedConnectionStringName(name)}' does not exist. Make sure that it is a defined App Setting.");
+            }
+
+            _logForwarder.Start();
```
fabiocav: Will this type handle multiple calls to `Start` gracefully?

fabiocav (commented):
Small questions at this point. The follow up on options and the `Start` call on the forwarder

src/Microsoft.Azure.WebJobs.StorageProvider.Common/StorageClientProvider.cs
karshinlin: It should, but moving this to the constructor to lessen the calls to it.

src/Microsoft.Azure.WebJobs.Shared.StorageProvider/StorageClientProvider.cs
karshinlin: BlobClientOptions and QueueClientOptions need parameterless constructors to use with the Options pattern. Not possible here.

fabiocav (approved):

build/common.props
```diff
@@ -3,7 +3,7 @@
     <!-- Packages can have independent versions and only increment when released -->
     <Version>3.0.29$(VersionSuffix)</Version>
     <ExtensionsStorageVersion>4.0.5$(VersionSuffix)</ExtensionsStorageVersion>
-    <HostStorageVersion>4.0.2$(VersionSuffix)</HostStorageVersion>
+    <HostStorageVersion>4.1.0$(VersionSuffix)</HostStorageVersion>
```
karshinlin: @fabiocav @mathewc given the reduced changes, should this still be 4.1.0?

mathewc: Yes, 4.0.3 is probably better - following our regular patch versioning.

mathewc (approved):
Approved - just change version to 4.0.3.

PR #2693, opened 2021-04-08
Time calls to flush, close and complete to accurately measure I/O time

gohar94:
The existing timer logs seemingly undercount the time taken to write to Blob storage, likely because of ongoing buffering and some pending operations that complete after flush/close which are not being accounted for in the timer.
This PR times flush/close/complete methods also to more accurately measure the entire time spent in I/O.


goiri (approved):

gohar94:
@pragnagopa please review whenever you get a chance. Thanks a lot.

liliankasem:
@pragnagopa is this something we still need, can you take another look?

PR #2701, opened 2021-04-15
create inner OptionsFactory via DI

### LINKED ISSUE #2522: WebJobsOptionsFactory implementation ignores the "validations" parameter
Because the implementation of the `WebJobsOptionsFactory` ignores the validation, it does not perform the intended validation.

#### Repro steps

```csharp
{
    public Startup()
    {
        var config = new ConfigurationBuilder()
            .AddEnvironmentVariables();

        Configuration = config.Build();
    }

    public IConfiguration Configuration { get; }

    {
        builder.Services.AddOptions<SampleOptions>()
               .Bind(Configuration.GetSection("Sample"))
               .ValidateDataAnnotations();
    }
}

public class SampleOptions
{
    [Required]
    public string Value { get; set; }
}
```

#### Expected behavior

I expect the validation to run.
`Sample:Value` key if the application settings are not set and `OptionsValidationException` is thrown.

#### Actual behavior

No validation is performed and the property values are not checked.

#### Known workarounds

```csharp
{
    {
        // Replace default implementation
        builder.Services.Replace(ServiceDescriptor.Transient(typeof(IOptionsFactory<>), typeof(OptionsFactory<>)));
    }
}
```

#### Related information 

https://github.com/Azure/azure-webjobs-sdk/blob/f386e54e5000ddddf255f529689ab2a9c3ee78cd/src/Microsoft.Azure.WebJobs.Host/Hosting/OptionsFormatter/WebJobsOptionsFactory.cs#L28

brettsam:
fixes #2522 

In later versions of .NET Core, there is an additional parameter for `validations` accepted by the `OptionsFactory<>` constructor. These weren't being picked up because we were creating the factory directly rather than letting DI handle it.

Confirmed by updating the Functions host locally with this change and validations are now called.

### DIFF
--- a/build/common.props
+++ b/build/common.props
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <!-- Packages can have independent versions and only increment when released -->
-    <Version>3.0.28$(VersionSuffix)</Version>
+    <Version>3.0.29$(VersionSuffix)</Version>
     <ExtensionsStorageVersion>4.0.5$(VersionSuffix)</ExtensionsStorageVersion>
     <HostStorageVersion>4.0.2$(VersionSuffix)</HostStorageVersion>
     <LoggingVersion>4.0.2$(VersionSuffix)</LoggingVersion>
--- a/src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsServiceCollectionExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsServiceCollectionExtensions.cs
@@ -8,7 +8,6 @@ using Microsoft.Azure.WebJobs.Host.Config;
 using Microsoft.Azure.WebJobs.Host.Configuration;
 using Microsoft.Azure.WebJobs.Host.Dispatch;
 using Microsoft.Azure.WebJobs.Host.Executors;
-using Microsoft.Azure.WebJobs.Host.Hosting;
 using Microsoft.Azure.WebJobs.Host.Indexers;
 using Microsoft.Azure.WebJobs.Host.Listeners;
 using Microsoft.Azure.WebJobs.Host.Loggers;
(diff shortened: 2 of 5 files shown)


src/Microsoft.Azure.WebJobs/WebJobs.csproj
```diff
@@ -6,7 +6,7 @@
     <Description>This library simplifies the task of adding background processing to your Microsoft Azure Web Sites. The SDK uses Microsoft Azure Storage, triggering a function in your program when items are added to Queues and Blobs. A dashboard provides rich monitoring and diagnostics for the programs that you write by using the SDK. For more information, please visit http://go.microsoft.com/fwlink/?LinkID=320971</Description>
     <AssemblyName>Microsoft.Azure.WebJobs</AssemblyName>
     <RootNamespace>Microsoft.Azure.WebJobs</RootNamespace>
-    <DocumentationFile>$(OutputPath)\$(AssemblyName).xml</DocumentationFile>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
```
brettsam, line 9: This build failed locally for me again (we'd seen this in other repos), so following what we do elsewhere.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsServiceCollectionExtensions.cs
```diff
@@ -110,6 +109,7 @@ public static IWebJobsBuilder AddWebJobs(this IServiceCollection services, Actio
             services.AddSingleton(typeof(IWebJobsExtensionConfiguration<>), typeof(WebJobsExtensionConfiguration<>));
 
             // Options logging
+            services.AddTransient(typeof(OptionsFactory<>));
```
brettsam, line 112: Let DI create this instead of us.

build/common.props
```diff
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <!-- Packages can have independent versions and only increment when released -->
-    <Version>3.0.28$(VersionSuffix)</Version>
+    <Version>3.0.29$(VersionSuffix)</Version>
```
brettsam, line 4: bumping this now that 3.0.28 is already pulled into the host.

fabiocav (approved):

PR #2707, opened 2021-04-28
Adding IDistributedlockManager implementation for Kubernetes

divyagandhisethi:
Fixes https://github.com/microsoft/k4apps/issues/234 

We need to have a non-storage based lease management systems for Timer triggers when running in Kubernetes hosted environments.


divyagandhisethi:
We'll add test in the next iteration. This is a draft PR for preliminary review.

fabiocav:
@divyagandhisethi after a discussion on this, we believe this would be better suited in the host. Could you please move this implementation there?

PR #2721, opened 2021-05-26
Handle exceptions when getting binding properties

### LINKED ISSUE #2698: BindingDataProvider.GetProperties should gracefully handle exceptions
The code [here](https://github.com/Azure/azure-webjobs-sdk/blob/3f4ec78be9f43bb041937425ced00b341883aa42/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingDataProvider.cs#L70) should handle exceptions when accessing a property value and if an exception was thrown skip adding it to binding data.

This came up recently in the context of the new track 2 ServiceBus extension, where the target trigger type ServiceBusReceivedMessage Body property can throw in some cases.

ankitkumarr:
Fixes https://github.com/Azure/azure-webjobs-sdk/issues/2698

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingDataProvider.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingDataProvider.cs
@@ -64,13 +64,7 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
             }
             else
             {
-                Dictionary<string, object> bindingData = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
-                foreach (var propertyHelper in _propertyHelpers)
-                {
-                    object propertyValue = propertyHelper.GetValue(value);
-                    bindingData.Add(propertyHelper.Name, propertyValue);
-                }
-                return bindingData;
+                return GetBindingDataFromPropertyHelpers(value);
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.WebJobs.Host/Bindings/BindingDataProvider.cs
```diff
@@ -126,5 +120,24 @@ public static BindingDataProvider FromTemplate(string template, bool ignoreCase
...
+            foreach (var propertyHelper in _propertyHelpers)
+            {
+                try
+                {
+                    object propertyValue = propertyHelper.GetValue(value);
+                    bindingData.Add(propertyHelper.Name, propertyValue);
+                }
+                catch (Exception)
+                {
+                    // Skip if accessing the property throws
```
fabiocav, line 136: Should we be logging here?

fabiocav (approved):
Approving, but with a small comment on something that may improve our ability to diagnose issues in this area.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingDataProvider.cs
ankitkumarr: Yes I did attempt to do that, but I didn't find an available logger. This class is instantiated via `BindingDataProvider.FromType` which is called by a couple of bindings and their static helpers with no logger. I think passing a logger all the way to this type would be invasive and may break public contracts.

I could log to console directly, but it may be better to consider passing a proper logger. Do you know of a convenient way to do that? If not, it may be a follow-up item to restructure and flow down a logger.

mathewc: I'm fine with not logging for now. For the scenario in which @joshlove-msft logged this, it's not clear to me whether every invocation would throw. If so, we wouldn't want to be logging an exception on every invocation.

mathewc (approved):

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingDataProvider.cs
ankitkumarr: That makes sense. If we did log, I'd guess it'd be a debug level logging. But I merge this for now as it sounds like we should be OK without it.

Issue #2088, opened 2019-01-17
WebJobs Shutdown signal not working


mathewc:
WebJobsShutdownWatcher is designed to monitor a sentinel file that the Kudu continuous webjobs infrastructure uses to signal that the WebJob is shutting down. In v2 this shutdown signal would cause RunAndBlock to complete, code [here](https://github.com/Azure/azure-webjobs-sdk/blob/5407e75fb875f9afb79e06bd45d3109f9bda4d94/src/Microsoft.Azure.WebJobs.Host/JobHost.cs#L193) stopping the WebJob exe and allowing any cleanup code to run.

In v3, we're not honoring this - causing a shutdown signal that is picked up by the WebJobsShutdownWatcher does not cause the `await host.RunAsync()` to complete.

We shouldn't be ignoring the shutdown signal - it's part of graceful shutdown. As it is now the WebJob will continue running until Kudu kills the process.

divinci:
At last some information regarding why my shutdown code is not working :)

Thanks @mathewc do you have information for a workaround in the mean time?

divinci:
This doesn't seem to work.

`            var host = builder.Build();

            var cancellationToken = new WebJobsShutdownWatcher().Token;
            cancellationToken.Register(async () => await host.StopAsync());

            using (host)
            {
                await host.RunAsync(cancellationToken);
            }`

divinci:
> This doesn't seem to work.
> 
> ```
> var host = builder.Build();
> var cancellationToken = new WebJobsShutdownWatcher().Token;
> cancellationToken.Register(async () => await host.StopAsync());
> 
> using (host)
> {
>     await host.RunAsync(cancellationToken);
> }
> ```

After some debugging, this code allows for graceful shutdown in my local environment when I introduce the environment variable `WEBJOBS_SHUTDOWN_FILE` and manually create it.  But when deployed, something means it will not work :/

divinci:
Any news on this issue @fabiocav @mathewc?

aherrick:
Just to be clear, there isn't any workaround to stopping a continuous web job via the azure portal? (not Kudu process kill)

divinci:
> Just to be clear, there isn't any workaround to stopping a continuous web job via the azure portal? (not Kudu process kill)

I can't find one.

alhardy:
Having the same issue

JSCProjects:
Facing same issue here.

[...]

PR #2730, opened 2021-06-08
Ensure graceful host shutdown when shutdown file is created (#2088)

mathewc:
Addresses https://github.com/Azure/azure-webjobs-sdk/issues/2088

As the referenced issue indicates, currently when a shutdown is triggered, listeners are stopped, but the host itself isn't stopped, nor are hosted services, and IHost.RunAsync doesn't complete. With my changes everything shuts down gracefully, RunAsync completes allowing the WebJobs console app to run any cleanup code before exiting.


suwatch:
:shipit:

brettsam (approved):

PR #2740, opened 2021-06-24
A few Dynamic Concurrency fixes/updates

mathewc:
In doing some OOP E2E testing in the Functions Host I discovered a couple bugs.


WebJobs.sln
```diff
@@ -53,8 +53,6 @@ Project("{6EC3EE1D-3C4E-46DD-8F32-0CC8E7565705}") = "FSharpFunctions", "test\Tes
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Benchmarks", "test\Benchmarks\Benchmarks.csproj", "{78F8086D-8313-477A-B24F-E475A690880A}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "TestChildProcess", "TestChildProcess\TestChildProcess.csproj", "{CEBC078A-9DD2-4558-829B-DBAE6BEA4264}"
```
mathewc, line 56: I had a temporarily skipped E2E test for process monitoring that was spinning up test processes. The test ran fine locally and also on CI when I RDP'ed in and ran from VS. But for some reason in actual CI runs, the test would fail. The child processes were getting killed prematurely by someone. So I've made some refactorings to make Process fully mockable, since the goal of the test was to test behaviors when processes were killed (e.g. language worker dies).

src/Microsoft.Azure.WebJobs.Host/Scale/DefaultConcurrencyThrottleManager.cs
```diff
@@ -66,7 +66,7 @@ public ConcurrencyThrottleAggregateStatus GetStatus()
 
         private void UpdateThrottleState()
         {
-            IEnumerable<ConcurrencyThrottleStatus> throttleResults = _throttleProviders.Select(p => p.GetStatus(_logger));
+            IEnumerable<ConcurrencyThrottleStatus> throttleResults = _throttleProviders.Select(p => p.GetStatus(_logger)).ToArray();
```
mathewc, line 69: E2E testing showed that there was a subtle bug here due to the fact that this enumerable wasn't cached. Because of this, I could get inconsistent/flaky results below.

src/Microsoft.Azure.WebJobs.Host/Scale/DefaultProcessMetricsProvider.cs
```diff
@@ -1,40 +0,0 @@
...
-
-using System;
-using System.Diagnostics;
-
-namespace Microsoft.Azure.WebJobs.Host.Scale
-{
-    /// <summary>
-    /// Default implementation, that just delegates to the underlying Process.
-    /// </summary>
-    internal class DefaultProcessMetricsProvider : IProcessMetricsProvider
```
mathewc, line 15: This type/interface existed to make process stats mockable. I'm removing this and going deeper by making Process itself mockable, which was needed for some other tests I'm adding here around Process.HasExited. These are all internal types.

src/Microsoft.Azure.WebJobs.Host/Scale/ProcessMonitor.cs
```diff
@@ -91,13 +89,20 @@ public virtual ProcessStats GetStats()
 
         private void OnTimer(object state)
         {
-            if (_disposed || _process.HasExited)
```
mathewc, line 94: This was the point of the bug I discovered. When running in the Functions Host, if I for example killed the OOP Python worker, sometimes HasExited would actually throw because the process had been fully cleaned up. So I'm adding error handling for that case (see ProcessWrapper.HasExited).

test/Microsoft.Azure.WebJobs.Host.UnitTests/Scale/DefaultHostProcessMonitorTests.cs
```diff
@@ -327,26 +340,27 @@ public void GetStatus_MultiProcess_HighMemory_ReturnsOverloaded()
             Assert.Equal("[HostMonitor] Host memory threshold exceeded (8500 >= 8000)", log.FormattedMessage);
         }
 
-        [Fact(Skip = "Failing on CI. Need to investigate.")]
```
mathewc, line 330: This was the failing test that I wanted to enable. For whatever reason the 3 child processes I'd spawn below would start up and run for a short period of time, but would then die for some reason. Not reproable when I actually RDP'ed, or ran locally. The mocking I'm enabling in this PR is superior anyways.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Scale/ProcessMonitorTests.cs
```diff
@@ -51,6 +52,62 @@ public ProcessMonitorTests()
...
+        }
+
+        [Fact]
+        public void Process_PrivateMemoryBytes_ReturnsExpectedResult()
+        {
+            var process = Process.GetCurrentProcess();
+            var monitor = new ProcessMonitor(process);
+
+            Assert.Equal(monitor.Process.PrivateMemoryBytes, process.PrivateMemorySize64);
+            Assert.Equal(monitor.Process.PrivateMemoryBytes, process.PrivateMemorySize64);
```
paulbatum: Whats the point of doing this twice?

test/Microsoft.Azure.WebJobs.Host.UnitTests/Scale/ProcessMonitorTests.cs
```diff
@@ -51,6 +52,62 @@ public ProcessMonitorTests()
...
+        }
+
+        [Fact]
+        public void Process_Id_ReturnsExpectedResult()
+        {
+            var process = Process.GetCurrentProcess();
+            var monitor = new ProcessMonitor(process);
+
+            Assert.Equal(monitor.Process.Id, process.Id);
```
paulbatum, line 87: Is this just testing the property accessor? It feels a little unnecessary.

src/Microsoft.Azure.WebJobs.Host/Scale/DefaultHostProcessMonitor.cs
```diff
@@ -66,7 +69,9 @@ public void UnregisterChildProcess(Process process)
             ProcessMonitor monitor = null;
             lock (_syncLock)
             {
-                monitor = _childProcessMonitors.SingleOrDefault(p => p.Process == process);
+                // try to locate the monitor for the specified process
+                // and remove it if found
+                monitor = _childProcessMonitors.SingleOrDefault(p => p.Process.Inner == process);
```
paulbatum: Having to access Inner in product code kind of breaks the abstraction. I get this interface was introduced for testing purposes, but once you incur the cost of introducing the interface, I'd like to try to stick to it wherever possible. Could we do this lookup by ID instead?

src/Microsoft.Azure.WebJobs.Host/Scale/ProcessMonitor.cs
```diff
@@ -191,5 +196,59 @@ public void Dispose()
...
+                {
+                    try
+                    {
+                        return _process.HasExited;
+                    }
+                    catch
+                    {
+                        // When the process has been Closed/Disposed
+                        // Process members like HasExited will throw
+                        return true;
```
paulbatum, line 236: We can't check if its disposed, instead of catching the exception?

mathewc: The IDisposable interface doesn't have an "IsDisposed".

src/Microsoft.Azure.WebJobs.Host/Scale/DefaultHostProcessMonitor.cs
mathewc: Yes I realize that and I struggled here. The issue is that when the process is closed/disposed it's property accessors throw InvalidOperation exceptions. So I can't go by ID unfortunately.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Scale/ProcessMonitorTests.cs
mathewc: Yeah, it's testing the wrapper abstraction ensuring that all the wrappers return the expected values. I included this test for completness.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Scale/ProcessMonitorTests.cs
mathewc: Removed the extra line here and in the other tests.

src/Microsoft.Azure.WebJobs.Host/Scale/DefaultHostProcessMonitor.cs
mathewc: I've resolved this issue by removing the Inner property and instead overriding Equals, allowing a wrapper to compare (and be equal to) the process instance it wraps.

paulbatum (approved):

Karmaibringkaosanddrama (approved):

PR #2743, opened 2021-07-01
[pack] Support shared listener pattern in FunctionExecutor concurrency tracking

mathewc:
We've discovered that for extensions that use a single shared listener for all functions of that type (e.g. BlobTrigger/Durable), the Function ID tracking used for concurrency management wasn't working properly, preventing concurrency from increasing. Normally functions map 1:1 to their listeners.

The ID used by the listener when calling ConcurrencyManager.GetStatus must be the same ID used in FunctionExecutor when calling ConcurrencyManager.FunctionStarted/Completed. If they're not, the function invocations aren't tracked and CM will never vote to increase concurrency for that function. This is because CM only ever increases concurrency if the current concurrency level is being fully utilized (logic [here](https://github.com/Azure/azure-webjobs-sdk/blob/206e194e03d82a101a6b5dffb7b4c3904730801d/src/Microsoft.Azure.WebJobs.Host/Scale/ConcurrencyStatus.cs#L186)).

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Scale/ConcurrencyManagerService.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Scale/ConcurrencyManagerService.cs
@@ -127,10 +127,11 @@ namespace Microsoft.Azure.WebJobs.Host.Scale
         {
             var snapshot = _concurrencyManager.GetSnapshot();
 
-            // only persist snapshots for functions that are in our current index
-            // this ensures we prune any stale entries from a previously applied snapshot
+            // Only persist snapshots for functions that are in our current index. This ensures
+            // we prune any stale entries from a previously applied snapshot.
+            // Note that functions using a shared listener are treated specially here.
             var functionIndex = await _functionIndexProvider.GetAsync(CancellationToken.None);
-            var functions = functionIndex.ReadAll().ToLookup(p => p.Descriptor.Id, StringComparer.OrdinalIgnoreCase);
+            var functions = functionIndex.ReadAll().ToLookup(p => p.Descriptor.SharedListenerId ?? p.Descriptor.Id, StringComparer.OrdinalIgnoreCase);
             snapshot.FunctionSnapshots = snapshot.FunctionSnapshots.Where(p => functions.Contains(p.Key)).ToDictionary(p => p.Key, p => p.Value);
 
             if (!snapshot.Equals(_lastSnapshot))
--- a/src/Microsoft.Azure.WebJobs.Host/SharedListenerAttribute.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/SharedListenerAttribute.cs
@@ -1,0 +1,25 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// Attribute applied to a <see cref="ITriggerBinding"/> implementation when the binding uses
+    /// a shared listener for all functions.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class)]
+    public class SharedListenerAttribute : Attribute
+    {
+        public SharedListenerAttribute(string sharedListenerId)
+        {
+            SharedListenerId = sharedListenerId;
+        }
+
+        /// <summary>
+        /// Gets the shared ID used for all functions.
+        /// </summary>
+        public string SharedListenerId { get; }
+    }
+}
--- a/test/Microsoft.Azure.WebJobs.Host.UnitTests/Executors/TriggeredFunctionExecutorTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.UnitTests/Executors/TriggeredFunctionExecutorTests.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
-using System.Linq;
 using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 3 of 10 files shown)


test/Microsoft.Azure.WebJobs.Host.UnitTests/Executors/TriggeredFunctionExecutorTests.cs
```diff
@@ -74,15 +76,76 @@ public async Task TryExecuteAsync_WithInvokeHandler_InvokesHandler()
...
+                DynamicConcurrencyEnabled = dynamicConcurrencyEnabled
+            };
+            var functionExecutor = GetTestFunctionExecutor(concurrencyOptions: concurrencyOptions);
+            var triggerExecutor = new TriggeredFunctionExecutor<int>(functionDescriptor, functionExecutor, instaceFactoryMock.Object, NullLoggerFactory.Instance);
+
+            string functionId = functionInstance.FunctionDescriptor.Id;
+            
+            if (sharedListenerFunctionId != null)
+            {
+                triggerData.TriggerDetails.Add(LogConstants.SharedListenerFunctionIdKey, sharedListenerFunctionId);
```
mathewc: For extensions like BlobTrigger/Durable they'll flow their listener Function ID along with the rest of the trigger details. I'll be updating the Track 2 BlobTrigger executor to do this.

cgillum: Durable triggers don't currently use `TriggerDetails`, so I guess we'll need to add that. The key in this dictionary always needs to be `LogConstants.SharedListenerFunctionIdKey`, right?  Seems a little weird to have this constant defined in a class called `LogConstants`...

mathewc: Yes, you might decide you want to add additional details in there as well in the future. Other triggers add useful information which flows to AppInsights (e.g. QueueTrigger [here](https://github.com/Azure/azure-webjobs-sdk/blob/3f4ec78be9f43bb041937425ced00b341883aa42/src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Listeners/QueueTriggerExecutor.cs#L39)). Regarding this constants class - I put it there because we're already using this for other data in TriggerDetails, e.g. [here](https://github.com/Azure/azure-webjobs-sdk/blob/524decf320ca9cf8a35f10b675af7b4a36ecc1c5/src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/Constants/LogConstants.cs#L178). @brettsam - is this an OK place for this key?

mathewc: Changed the model. Due to other issues I discovered, I've taken another approach. The shared ID is no longer flowed via trigger details, it's specified declaratively via a new SharedListenerAttribute applied to the ITriggerBinding implementation for the extension.

src/Microsoft.Azure.WebJobs.Host/SharedListenerAttribute.cs
```diff
@@ -0,0 +1,25 @@
...
+using System;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// Attribute applied to a <see cref="ITriggerBinding"/> implementation when the binding uses
+    /// a shared listener for all functions.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Class)]
+    public class SharedListenerAttribute : Attribute
```
mathewc, line 13: This attribute must be applied to the ITriggerBinding implementation of the extension using a shared listener. E.g. for BlobTrigger, it'll get applied to BlobTriggerBinding [here](https://github.com/Azure/azure-sdk-for-net/blob/4162f6fa2445b2127468b9cfd080f01c9da88eba/sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Blobs/src/Triggers/BlobTriggerBinding.cs#L28).

src/Microsoft.Azure.WebJobs.Host/Scale/ConcurrencyManagerService.cs
```diff
@@ -127,10 +127,11 @@ internal async Task WriteSnapshotAsync()
...
             var snapshot = _concurrencyManager.GetSnapshot();
 
-            // only persist snapshots for functions that are in our current index
-            // this ensures we prune any stale entries from a previously applied snapshot
+            // Only persist snapshots for functions that are in our current index. This ensures
+            // we prune any stale entries from a previously applied snapshot.
+            // Note that functions using a shared listener are treated specially here.
             var functionIndex = await _functionIndexProvider.GetAsync(CancellationToken.None);
-            var functions = functionIndex.ReadAll().ToLookup(p => p.Descriptor.Id, StringComparer.OrdinalIgnoreCase);
+            var functions = functionIndex.ReadAll().ToLookup(p => p.Descriptor.SharedListenerId ?? p.Descriptor.Id, StringComparer.OrdinalIgnoreCase);
```
mathewc, line 134: This was the other issue I discovered in testing DC with BlobTrigger end to end. The issue was with this logic that prunes snapshots for IDs that don't exist in the function index. The DC snapshots use the shared listener ID, but the function index previously didn't contain any information on the shared listener ID. That meant we'd never persist snapshots for these functions. Now, for functions using a shared listener, we use the shared ID and everything lines up.

src/Microsoft.Azure.WebJobs.Host/SharedListenerAttribute.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        public SharedListenerAttribute(string sharedListenerId)
+        {
+            SharedListenerId = sharedListenerId;
```
cgillum, line 17: Is `null` a valid value for this ID? If so, what's the expected behavior?

mathewc: If the value is null, it won't be used - i.e. the trigger won't be treated as having a shared listener. I avoided throwing from the constructor here because in general that's not a good practice.

cgillum (approved):
LGTM!

PR #2745, opened 2021-07-13
Some Dynamic Concurrency updates

mathewc:
Changes in this PR:

* Changing the API for callers to determine the available invocation count. The previous AvailableInvocationCount property was too simplistic to handle all listener scenarios. While it worked well for single listener loop that always ensured they dispatched the invocations before calling GetStatus again, it doesn't work well in scenarios where there might be multiple concurrent listener loops for the same function. In that case it would be possible for 2 or more threads to call GetStatus and receive the same AvailableInvocationCount. If both those threads then started that number of invocations, 2x the actual allowed number would be started. The API is changed to a method that expects the caller to pass in the # of outstanding work items it has fetched/started. In a multi-threaded scenario, it's up to the caller to keep a count of the total number of items it has started, and ensure that count is synchronized across the listener loops to ensure OutstandingInvocations is never greater than CurrentConcurrency. E.g. that might be done by having each listener loop treat the GetStatus/FetchItems step as a critical section that only one loop at a time is executing, and increment their own outstandingInvocations count within that critical section.
* Exposing ThrottleStatus to provide more visibility into concurrency status. E.g. if zero is returned for # available invocations and the reason is because throttles are enabled, you can inspect the throttle details.
* Removing the NextStatusDelay property from ConcurrencyStatus for now. It was based on AvailableInvocationCount which was removed. I think it's fine for now to allow listeners to decide their delay when no invocations are available.
* Fixing a bug where it would be possible for a persisted snapshot to be applied that has a concurrency level greater than the configured MaximumFunctionConcurrency. That could happen for example if the customer changed their config after snapshots were persisted.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Scale/ConcurrencyStatus.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Scale/ConcurrencyStatus.cs
@@ -12,7 +12,6 @@ namespace Microsoft.Azure.WebJobs.Host.Scale
     /// </summary>
     public class ConcurrencyStatus
     {
-        internal const int NextStatusDelayDefaultSeconds = 1;
         internal const int DefaultFailedAdjustmentQuietWindowSeconds = 30;
         internal const int AdjustmentRunWindowSeconds = 10;
         internal const int DefaultMinAdjustmentFrequencySeconds = 5;
@@ -58,58 +57,30 @@ namespace Microsoft.Azure.WebJobs.Host.Scale
         public string FunctionId { get; }
 
         /// <summary>
-        /// Gets the current number of new invocations of this function the host can process.
-        /// When throttling is enabled, this may return 0 meaning no new invocations should be
-        /// started.
-        /// </summary>
-        public int AvailableInvocationCount
-        {
-            get
-            {
-                if (_concurrencyManager.ThrottleEnabled || OutstandingInvocations > CurrentConcurrency)
-                {
-                    // we can't take any work right now
-                    return 0;
-                }
-                else
-                {
-                    // no throttles are enabled, so we can take work up to the current concurrency level
-                    return CurrentConcurrency - OutstandingInvocations;
-                }
-            }
-        }
-
-        /// <summary>
-        /// Gets the current concurrency level for this function. This adjusts
+        /// Gets the current maximum allowed concurrency level for this function. This adjusts
         /// dynamically over time.
         /// </summary>
+        /// <remarks>
+        /// The number of outstanding invocations should always be less than or equal
+        /// to this number. When concurrency is dynamically adjusted down, it might be
+        /// possible for the number of outstanding invocations to exceed this number
+        /// for a short period of time, but in general the above holds.
+        /// </remarks>
         public int CurrentConcurrency { get; internal set; }
 
         /// <summary>
-        /// Gets the current number of in progress invocations of this function.
+        /// Gets the current number of actively executing invocations of this function.
         /// </summary>
         public int OutstandingInvocations { get; internal set; }
 
         /// <summary>
-        /// Gets the recommended delay for the next time <see cref="ConcurrencyManager.GetStatus(string)"/> should be called
-        /// for this function.
+        /// Gets the current throttle status.
         /// </summary>
-        public TimeSpan NextStatusDelay
+        public ConcurrencyThrottleAggregateStatus ThrottleStatus
         {
             get
             {
-                if (AvailableInvocationCount == 0)
-                {
-                    // currently hardcoded, but can be made dynamic/configurable in the future
-                    // if the host is currently throttling or otherwise can't take any more work,
-                    // caller should delay the next status call, to give time for the situation
-                    // to change
-                    return TimeSpan.FromSeconds(NextStatusDelayDefaultSeconds);
-                }
-                else
-                {
-                    return TimeSpan.Zero;
-                }
+                return _concurrencyManager.ThrottleStatus;
             }
         }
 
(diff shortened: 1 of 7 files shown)


src/Microsoft.Azure.WebJobs.Host/Scale/ConcurrencyStatus.cs
```diff
@@ -143,6 +115,47 @@ public TimeSpan NextStatusDelay
         /// </summary>
         internal double TotalInvocationTimeSinceLastAdjustmentMs { get; set; }
 
+        /// <summary>
+        /// Gets the current number of new invocations of this function the host can process.
+        /// When throttling is enabled, this may return 0 meaning no new invocations should be
+        /// started.
+        /// </summary>
+        /// <param name="outstandingInvocations">The number of outstanding invocations the caller
+        /// is tracking. Must be greater than zero. Note that this number may be greater than
```
cgillum: > Must be greater than zero.

Did you mean "must be greater than *or equal to* zero"? The implementation only checks for less than zero. Also, one of your tests explicitly passes zero to this method.

src/Microsoft.Azure.WebJobs.Host/Scale/ConcurrencyStatus.cs
```diff
@@ -143,6 +115,47 @@ public TimeSpan NextStatusDelay
...
+        {
+            if (outstandingInvocations < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(outstandingInvocations));
+            }
+
+            // the number of outstanding invocations as tracked by the caller may be greater than
+            // the number we're tracking, since work items for new invocations may have been started
+            // by the caller, but may not have actually started executing
+            outstandingInvocations = Math.Max(outstandingInvocations, OutstandingInvocations);
```
cgillum: It seems odd that there are two independent variables for tracking outstanding invocations, and they both have the exact same name. I think we need to change the names to more clearly reflect what we're actually tracking or else people might get confused.  For example, it seems to me that one of these represents "in-flight" executions while the other represents "pending or not completed" or perhaps "reserved" executions. Is that accurate?

src/Microsoft.Azure.WebJobs.Host/Scale/ConcurrencyStatus.cs
```diff
@@ -143,6 +115,47 @@ public TimeSpan NextStatusDelay
...
+
+            if (_concurrencyManager.ThrottleStatus.State == ThrottleState.Enabled || outstandingInvocations >= CurrentConcurrency)
+            {
+                // we can't take any work right now
+                return 0;
+            }
+            else
+            {
+                // no throttles are enabled, so we can take work up to the current concurrency level
+                return CurrentConcurrency - outstandingInvocations;
```
cgillum: Another point of confusion: how is "outstandingInvocations" different from "CurrentConcurrency"? These sound like synonyms, leaving me confused about what either of them are actually referring to. Should "CurrentConcurrency" actually be "Current**Max**Concurrency"?

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/DynamicConcurrencyEndToEndTests.cs
```diff
@@ -666,12 +667,13 @@ public void OnTimer(object state)
...
                                 var concurrencyStatus = _concurrencyManager.GetStatus(_descriptor.Id);
-                                currentBatchSize = Math.Min(concurrencyStatus.AvailableInvocationCount, 32);
+                                int availableInvocationCount = concurrencyStatus.GetAvailableInvocationCount(_outstandingInvocationCount);
+                                currentBatchSize = Math.Min(availableInvocationCount, 32);
                                 if (currentBatchSize == 0)
                                 {
                                     // if we're not healthy or we're at our limit, we'll wait
                                     // a bit before checking again
-                                    _timer.Change((int)concurrencyStatus.NextStatusDelay.TotalMilliseconds, Timeout.Infinite);
+                                    _timer.Change(1000, Timeout.Infinite);
```
cgillum, line 676: Just double-checking that this change was intentional?

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/DynamicConcurrencyEndToEndTests.cs
```diff
@@ -702,6 +704,15 @@ public void OnTimer(object state)
...
                         }
                     }
+
+                    private async Task ExecuteFunctionAsync(TestEvent evt)
+                    {
+                        Interlocked.Increment(ref _outstandingInvocationCount);
+
+                        await _executor.TryExecuteAsync(new TriggeredFunctionData { TriggerValue = evt }, CancellationToken.None);
+
+                        Interlocked.Decrement(ref _outstandingInvocationCount);
```
cgillum: Should this go in a `finally` block in case `TryExecuteAsync` throws for some unexpected reason?

cgillum (approved):
No blockers detected, but a few things to consider.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/DynamicConcurrencyEndToEndTests.cs
mathewc: Yes - I've removed ConcurrencyStatus.NextStatusDelay. Listeners can decide themselves the delay to use before checking again when no invocations are currently available.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/DynamicConcurrencyEndToEndTests.cs
mathewc: Yes - this is test code so I was loose, but you're right.

src/Microsoft.Azure.WebJobs.Host/Scale/ConcurrencyStatus.cs
mathewc: Yes, good catch

src/Microsoft.Azure.WebJobs.Host/Scale/ConcurrencyStatus.cs
mathewc: Will rename the method parameter to "pendingInvocations" to make the difference clearer. OutstandingInvocations as tracked by CM is the number of outstanding invocations that have started executing but haven't completed yet. PendingInvocations as tracked by the caller represents the number of work items that are "in-flight" that may or may not have started executing yet.

src/Microsoft.Azure.WebJobs.Host/Scale/ConcurrencyStatus.cs
mathewc: Yes, CurrentConcurrency is the maximum concurrency that the function is allowed to execute up to. OutstandingInvocations is the current number of invocations that have started executing but haven't completed yet, so you're right that it represents the real time actual concurrency level.

While it could have been named CurrentMaxConcurrency, I think the existing name is fine and I'll keep it at this point. The "Current" portion of the name "CurrentConcurrency" refers to the fact that the value dynamically adjusts over time. I'll update the help text for these properties to make this clearer.

Issue #2729, opened 2021-06-08
Consider supporting wildcard parameter name in DataBindingProvider


JoshLove-msft:
Currently, [DataBindingProvider ](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Bindings/Data/DataBindingProvider.cs#L24) matches on the parameter name specified in the bindingDataContract dictionary. There are times when this is not necessary as there would only ever be a single instance of the type that corresponds to the parameter. For example, we allow binding to the [ServiceBusMessageAction](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Triggers/ServiceBusTriggerBindingStrategy.cs#L60) type in the new Service Bus extensions. It shoudn't really matter here what the user uses as the parameter name. It would be great if an element could be added to the Dictionary with an "*" for the key to indicate that any name is acceptable. The side effect of this change would be users could define multiple parameters binding to the same instance, this could either be allowed or cause an exception depending on what is idiomatic for the framework.

It is possible to get the desired behavior without this, but it is more cumbersome, as I don't believe you would be able to use a BindingStrategy and instead would have to implement an IBinding yourself.

JoshLove-msft:
Would be happy to contribute this if this it is something that would be accepted.
 
/cc @fabiocav @paulbatum

JoshLove-msft:
A customer ran into this while trying out the new ServiceBus trigger - https://stackoverflow.com/questions/68350196/how-to-manually-handle-message-completion-in-c-sharp-azure-function-with-azure-s

mathewc:
Thanks for logging the issue @JoshLove-msft. Yes the fact that data contract bindings for non-trigger parameters must match the static contract name is something that confuses users. From the beginning of the WebJobs SDK this is how it's behaved.

[...]

PR #2756, opened 2021-08-23
Improving indexing exceptions for potential failed contract bindings.

mathewc:
Improvement to address https://github.com/Azure/azure-webjobs-sdk/issues/2729. Basically, if a parameter binding fails and we have binding contract information, we expand the binding error to include the following:

_Cannot bind parameter 'receiver' to type MessageReceiver. Make sure the parameter Type is supported by the binding. **The binding supports the following parameter names for type MessageReceiver: (MessageReceiver). If you're trying to bind to one of those values, rename your parameter to match the contract name (case insensitive).** If you're using binding extensions (e.g. Azure Storage, ServiceBus, Timers, etc.) make sure you've called the registration method for the extension(s) in your startup code (e.g. builder.AddAzureStorage(), builder.AddServiceBus(), builder.AddTimers(), etc.)._


brettsam (approved):
Nice!

PR #2775, opened 2021-10-11
[pack] Adding FunctionActivityStatusProvider to monitor retries or active invocations.

alrod:
Implement a new Functions Host drain status API that DataRole will poll to determine when it's safe to reclaim a draining instance.

Internal work [item](https://dev.azure.com/azfunctions/Planning/_workitems/edit/827/) and [design](https://microsoft.sharepoint.com/teams/Antares/_layouts/OneNote.aspx?id=%2fteams%2fAntares%2fShared%20Documents%2fAntares%20Feature%20Crew&wd=target%28Azure%20Functions.one%7cBD6BD264-4C3E-463B-8B54-35FCCBE73FA1%2fFunctions%20Retry%5c%2fDrain%20Issue%7cBA4C3DBF-F9CB-4FD2-B495-6232391B0020%2f%29)

The change also addresses Azure/azure-functions-host#7188 

Related function host PR Azure/azure-functions-host/pull/7760

### DIFF
--- a/build/common.props
+++ b/build/common.props
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <!-- Packages can have independent versions and only increment when released -->
-    <Version>3.0.31$(VersionSuffix)</Version>
+    <Version>3.0.32$(VersionSuffix)</Version>
     <HostStorageVersion>5.0.0-beta.1$(VersionSuffix)</HostStorageVersion>
     <LoggingVersion>4.0.2$(VersionSuffix)</LoggingVersion>
     
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/DelegatingFunctionExecutor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/DelegatingFunctionExecutor.cs
@@ -6,10 +6,10 @@ using System.Threading.Tasks;
 
 namespace Microsoft.Azure.WebJobs.Host.Executors
 {
-    abstract internal class DelegatingFunctionExecutor : IRetryNotifier, IFunctionExecutor
+    internal abstract class DelegatingFunctionExecutor : IRetryNotifier, IFunctionExecutor
     {
-        readonly private IFunctionExecutor _innerExecutor;
-        readonly private IRetryNotifier _retryNotifier;
+        private readonly IFunctionExecutor _innerExecutor;
+        private readonly IRetryNotifier _retryNotifier;
 
         public DelegatingFunctionExecutor(IFunctionExecutor innerExecutor)
         {
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/DelegatingFunctionExecutor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/DelegatingFunctionExecutor.cs
@@ -1,0 +1,39 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Host.Executors
+{
+    abstract internal class DelegatingFunctionExecutor : IRetryNotifier, IFunctionExecutor
+    {
+        readonly protected IFunctionExecutor _innerExecutor;
+
+        public DelegatingFunctionExecutor(IFunctionExecutor innerExecutor)
+        {
+            _innerExecutor = innerExecutor;
+        }
+
+        public virtual Task<IDelayedException> TryExecuteAsync(IFunctionInstance instance, CancellationToken cancellationToken)
+        {
+            return Task.FromResult<IDelayedException>(null);
+        }
+
+        public void RetryPending()
+        {
+            if (_innerExecutor is IRetryNotifier retryNotifier)
+            {
+                retryNotifier.RetryPending();
+            }
+        }
+
+        public void RetryComplete()
+        {
+            if (_innerExecutor is IRetryNotifier retryNotifier)
+            {
+                retryNotifier.RetryComplete();
+            }
+        }
+    }
+}
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/DelegatingFunctionExecutor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/DelegatingFunctionExecutor.cs
@@ -8,32 +8,28 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
 {
     abstract internal class DelegatingFunctionExecutor : IRetryNotifier, IFunctionExecutor
     {
-        readonly protected IFunctionExecutor _innerExecutor;
+        readonly private IFunctionExecutor _innerExecutor;
+        readonly private IRetryNotifier _retryNotifier;
 
         public DelegatingFunctionExecutor(IFunctionExecutor innerExecutor)
         {
             _innerExecutor = innerExecutor;
+            _retryNotifier = innerExecutor as IRetryNotifier;
         }
 
         public virtual Task<IDelayedException> TryExecuteAsync(IFunctionInstance instance, CancellationToken cancellationToken)
         {
-            return Task.FromResult<IDelayedException>(null);
+            return _innerExecutor.TryExecuteAsync(instance, cancellationToken);
         }
 
-        public void RetryPending()
+        public virtual void RetryPending()
         {
-            if (_innerExecutor is IRetryNotifier retryNotifier)
-            {
-                retryNotifier.RetryPending();
-            }
+            _retryNotifier?.RetryPending();
         }
 
-        public void RetryComplete()
+        public virtual void RetryComplete()
         {
-            if (_innerExecutor is IRetryNotifier retryNotifier)
-            {
-                retryNotifier.RetryComplete();
-            }
+            _retryNotifier?.RetryComplete();
         }
     }
 }
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionActivityStatus.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionActivityStatus.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Host.Executors
+{
+    /// <summary>
+    /// Represents activity status the of job host
+    /// </summary>
+    public class FunctionActivityStatus
+    {
+        /// <summary>
+        /// Gets or sets number of outstanding invocations
+        /// </summary>
+        public int OutstandingInvocations { get; set; }
+
+        /// <summary>
+        /// Gets or sets number of outstanding retries
+        /// </summary>
+        public int OutstandingRetries { get; set; }
+    }
+}
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
@@ -21,7 +21,7 @@ using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.WebJobs.Host.Executors
 {
-    internal class FunctionExecutor : IFunctionExecutor
+    internal class FunctionExecutor : IFunctionExecutor, IFunctionActivityStatusProvider, IRetryNotifier
     {
         private readonly IFunctionInstanceLogger _functionInstanceLogger;
         private readonly IWebJobsExceptionHandler _exceptionHandler;
@@ -31,6 +31,8 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
         private readonly IEnumerable<IFunctionFilter> _globalFunctionFilters;
         private readonly IDrainModeManager _drainModeManager;
         private readonly ConcurrencyManager _concurrencyManager;
+        private int _outstandingInvocations;
+        private int _outstandingRetries;
 
         private readonly Dictionary<string, object> _inputBindingScope = new Dictionary<string, object>
         {
@@ -85,9 +87,10 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
             FunctionInstanceLogEntry instanceLogEntry = null;
             Stopwatch sw = Stopwatch.StartNew();
             string concurrencyFunctionId = functionInstanceEx.FunctionDescriptor.SharedListenerId ?? functionInstanceEx.FunctionDescriptor.Id;
-
+            
             try
             {
+                Interlocked.Increment(ref _outstandingInvocations);
                 if (_concurrencyManager.Enabled)
                 {
                     _concurrencyManager.FunctionStarted(concurrencyFunctionId);
@@ -147,10 +150,31 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                     _concurrencyManager.FunctionCompleted(concurrencyFunctionId, sw.Elapsed);
                 }
 
+                Interlocked.Decrement(ref _outstandingInvocations);
+
                 ((IDisposable)functionInstanceEx)?.Dispose();
             }
 
             return exceptionInfo != null ? new ExceptionDispatchInfoDelayedException(exceptionInfo) : null;
+        }
+
+        public FunctionActivityStatus GetStatus()
+        {
+            return new FunctionActivityStatus()
+            {
+                OutstandingInvocations = _outstandingInvocations,
+                OutstandingRetries = _outstandingRetries
+            };
+        }
+
+        public void RetryPending()
+        {
+            Interlocked.Increment(ref _outstandingRetries);
+        }
+
+        public void RetryComplete()
+        {
+            Interlocked.Decrement(ref _outstandingRetries);
         }
 
         private FunctionInstanceLogEntry CreateInstanceLogEntry(FunctionCompletedMessage functionStartedMessage)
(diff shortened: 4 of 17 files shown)


src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutorExtensions.cs
```diff
@@ -17,62 +17,86 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
     /// </summary>
     internal static class FunctionExecutorExtensions
     {
+        public static int _outstandingRetries;
+
+        public static int GetOutstandingRetries()
```
alrod: @mathewc, `IRetryNotifier` does not work as we have different implementations for `IFunctionExecutor `

pragnagopa:
Thanks for the PR!  Will the status exposed here eventually can be accessed in host status api - https://github.com/Azure/azure-functions-host/issues/7188 ?

alrod:
> Thanks for the PR! Will the status exposed here eventually can be accessed in host status api - [Azure/azure-functions-host#7188](https://github.com/Azure/azure-functions-host/issues/7188) ?

yes, Azure/azure-functions-host/pull/7760

build/common.props
```diff
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <!-- Packages can have independent versions and only increment when released -->
-    <Version>3.0.31$(VersionSuffix)</Version>
+    <Version>3.0.32$(VersionSuffix)</Version>
```
mathewc, line 4: Need to double check, but I don't think we've released 3.0.31 so you may not need to bump this. Kar is doing some work and he did also mistakenly bump this for his PR so on the myget feed there are some prerelease 3.0.32 versions, but I think this can stay at 31.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionActivityStatus.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Host.Executors
+{
+    public class FunctionActivityStatus
```
mathewc, line 9: add public xml comments for all new public surface area, here and for other types

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutorExtensions.cs
mathewc: You're talking about the Abort/Heartbeat/Shutdown executor nesting? Yes, each of these wrappers would have to implement IRetryNotifier, and simply delegate to their inner executor, as they do for TryExecuteAsync.

We don't want to use a static for this - we want each JobHost instance to track its own invocation and other state.

src/Microsoft.Azure.WebJobs.Host/Executors/IFunctionActivityStatusProvider.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Host.Executors
+{
+    /// <summary>
+    /// The interface is used to get current avtivity status of job host
```
mathewc: Fix spelling. Also, I'd reword this to something like "Interface providing access to job function activity information for the job host.".

src/Microsoft.Azure.WebJobs.Host/Executors/RetryFunctionExecutor.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.WebJobs.Host.Executors
+{
+    abstract internal class RetryFunctionExecutor : IRetryNotifier
```
mathewc: Let's generalize this further, naming it DelegatingFunctionExecutor, and it implements both IRetryNotifier and IFunctionExecutor. It'd expose a protected virtual method for TryExecuteAsync that subclasses can override to do their specific logic, then they'd call base to invoke the inner executor. The three wrapper executors we have (Shutdown/Heartbeat/Abort) would just derive from this then.

mathewc: Also, have you verified that we actually need to implement IRetryNotifier on these? I.e. are you seeing that the retry executor helper function was receiving IFunctionExecutor instances that were wrappers?

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -21,7 +21,7 @@
 
 namespace Microsoft.Azure.WebJobs.Host.Executors
 {
-    internal class FunctionExecutor : IFunctionExecutor
+    internal class FunctionExecutor : IFunctionExecutor, IFunctionActivityStatusProvider, IRetryNotifier
```
mathewc, line 24: Will need to add tests to FunctionExecutor tests for the new GetStatus behavior. E.g. a test that invokes a few functions verifies counts before/after they complete, etc. These unit tests would also call RetryPending/RetryCompleted directly and verify GetStatus reflects.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Executors/TriggeredFunctionExecutorTests.cs
```diff
@@ -165,7 +167,29 @@ public async Task TryExecuteWithRetries_Test(bool invocationThrows)
...
+            var functionActivityStatusProvider = functionExecutor as IFunctionActivityStatusProvider;
+
+            ManualResetEvent resetEvent = new ManualResetEvent(false);
+            _ = Task.Run(async () =>
+            {
+                if (invocationThrows)
+                {                    
+                    await TestHelpers.Await(() =>
+                    {
+                        var status = functionActivityStatusProvider.GetStatus();
```
mathewc: This test logic is tricky and hard to understand (at least for me :)) Thinking we can simplify. If we have the FunctionExecutor unit tests I mentioned above, I think we'll have good coverage if we then just add some new tests to FunctionExecutorExtensionsTests to verify that the retry notifier methods are called at the right times. I.e. we've then tested the new activity tracking at both layers via unit tests.

mathewc (changes requested):

src/Microsoft.Azure.WebJobs.Host/Executors/RetryFunctionExecutor.cs
alrod: `TryExecuteAsync`  method from `DelegatingFunctionExecutor` will be `public virtual` as it's interface implementation. 

Yes, I saw that `ShutdownExecutor `with internal `FunctionExecutor ` was received in FunctionExecutorExtension when run a nodejs test in function host repo.

src/Microsoft.Azure.WebJobs.Host/Executors/DelegatingFunctionExecutor.cs
```diff
@@ -0,0 +1,39 @@
...
+    abstract internal class DelegatingFunctionExecutor : IRetryNotifier, IFunctionExecutor
+    {
+        readonly protected IFunctionExecutor _innerExecutor;
+
+        public DelegatingFunctionExecutor(IFunctionExecutor innerExecutor)
+        {
+            _innerExecutor = innerExecutor;
+        }
+
+        public virtual Task<IDelayedException> TryExecuteAsync(IFunctionInstance instance, CancellationToken cancellationToken)
```
mathewc, line 20: I think the base implementation here should call TryExecuteAsync on the inner executor. Then derived classes can override to provide their before/after custom logic and call base.TryExecuteAsync. Will that work? IFunctionExecutor doesn't need to be a protected member then - derived classes don't need to access it.

src/Microsoft.Azure.WebJobs.Host/Executors/DelegatingFunctionExecutor.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            return Task.FromResult<IDelayedException>(null);
+        }
+
+        public void RetryPending()
```
mathewc: for completeness, these retry methods should also be virtual I think

src/Microsoft.Azure.WebJobs.Host/Executors/DelegatingFunctionExecutor.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            if (_innerExecutor is IRetryNotifier retryNotifier)
```
mathewc: nit: if you initialize n IRetryNotifier field on construction, we can avoid casting each time and this becomes simply _retryNotifier?.RetryPending()

src/Microsoft.Azure.WebJobs.Host/Executors/HeartbeatFunctionExecutor.cs
```diff
@@ -7,21 +7,19 @@
 
 namespace Microsoft.Azure.WebJobs.Host.Executors
 {
-    internal class HeartbeatFunctionExecutor : IFunctionExecutor
+    internal class HeartbeatFunctionExecutor : DelegatingFunctionExecutor, IFunctionExecutor
```
mathewc: don't need IFunctionExecutor anymore

src/Microsoft.Azure.WebJobs.Host/Executors/ShutdownFunctionExecutor.cs
```diff
@@ -6,18 +6,16 @@
 
 namespace Microsoft.Azure.WebJobs.Host.Executors
 {
-    internal class ShutdownFunctionExecutor : IFunctionExecutor
+    internal class ShutdownFunctionExecutor : DelegatingFunctionExecutor, IFunctionExecutor
```
mathewc: don't need IFunctionExecutor anymore

test/Microsoft.Azure.WebJobs.Host.UnitTests/Executors/FunctionExecutorExtensionsTests.cs
```diff
@@ -47,6 +48,36 @@ public async Task TryExecuteWithRetries_ExitRetryLoop()
...
+            });
+
+            functionExecutor.As<IRetryNotifier>().Setup(x => x.RetryComplete()).Callback(() =>
+            {
+                retryCompletedCallsCount++;
+            });
+         
+            var result = await functionExecutor.Object.TryExecuteAsync(instanceFactory, NullLoggerFactory.Instance, CancellationToken.None);
+
+            Assert.Equal(retryPendingCallsCount, 1);
```
mathewc, line 77: The function seems to always throw, and retries are configured for up to 5 retries, so I don't understand why this is only 1? Also, if this test will run to 5 retries, recommend having GetFunctionDescriptor take params for maxRetryCount/retryDelay so the test can be made to run faster with short delay.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Executors/FunctionExecutorTests.cs
```diff
@@ -418,6 +418,58 @@ public void OnFunctionTimeout_DoesNotCancel_IfDebugging()
...
+            Assert.Equal(status.OutstandingRetries, 0);
+
+            ManualResetEvent resetEvent = new ManualResetEvent(false);
+            _ = Task.Run(async () =>
+            {
+                // validate that we have 2 active invocation and 2 retries
+                await TestHelpers.Await(() =>
+                {
+                    var status = (executor as IFunctionActivityStatusProvider).GetStatus();
+                    return status.OutstandingInvocations == 2 && status.OutstandingRetries == 2;
```
mathewc, line 449: The timing of this works because the test helper polls every 50ms, and the function invocation delays 1s, which guarantees we'll get this condition?

test/Microsoft.Azure.WebJobs.Host.UnitTests/Executors/FunctionExecutorTests.cs
```diff
@@ -418,6 +418,58 @@ public void OnFunctionTimeout_DoesNotCancel_IfDebugging()
...
+            Action action = new Action(async () =>
+            {
+                (executor as IRetryNotifier).RetryPending();
+                await executor.TryExecuteAsync(functionInstance, CancellationToken.None);
+                (executor as IRetryNotifier).RetryComplete();
+            });
+            tasks.Add(Task.Run(action));
+            tasks.Add(Task.Run(action));
+
+            Task.WaitAll(tasks.ToArray(), 5000);
```
mathewc: nit: don't need the List<Task> - you can just Task.WaitAll(Task.Run(action), Task.Run(action))

test/Microsoft.Azure.WebJobs.Host.UnitTests/Executors/FunctionExecutorTests.cs
```diff
@@ -418,6 +418,58 @@ public void OnFunctionTimeout_DoesNotCancel_IfDebugging()
...
+            bool result = resetEvent.WaitOne(5000);
+            Assert.True(result);
+
+            Assert.Equal(status.OutstandingInvocations, 0);
```
mathewc, line 471: can you put another GetStatus call here for safety to make it clear we're testing final/udpated values?

test/Microsoft.Azure.WebJobs.Host.UnitTests/Executors/FunctionExecutorTests.cs
```diff
@@ -418,6 +418,58 @@ public void OnFunctionTimeout_DoesNotCancel_IfDebugging()
...
+            {
+                (executor as IRetryNotifier).RetryPending();
+                await executor.TryExecuteAsync(functionInstance, CancellationToken.None);
+                (executor as IRetryNotifier).RetryComplete();
+            });
+            tasks.Add(Task.Run(action));
+            tasks.Add(Task.Run(action));
+
+            Task.WaitAll(tasks.ToArray(), 5000);
+            bool result = resetEvent.WaitOne(5000);
```
mathewc: Why do you need to use a reset event here? It seems all you need is a simple boolean flag that gets set when your condition gets reached. We're waiting for the function invocations to complete already, so we don't need to wait on another event it seems.

build/common.props
mathewc: Funny - since I made this comment, Kar did end up releasing 3.0.31, so now we DO need to bump this :)

test/Microsoft.Azure.WebJobs.Host.UnitTests/Executors/FunctionExecutorExtensionsTests.cs
alrod: because we do not count retries but we count executions with retries. In retry loop `RetryPending `will be called only once.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Executors/FunctionExecutorTests.cs
alrod: correct

test/Microsoft.Azure.WebJobs.Host.UnitTests/Executors/FunctionExecutorTests.cs
alrod: even the bool flag is not needed - can wait for 3 task to complete.

src/Microsoft.Azure.WebJobs.Host/Executors/DelegatingFunctionExecutor.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Host.Executors
+{
+    abstract internal class DelegatingFunctionExecutor : IRetryNotifier, IFunctionExecutor
```
mathewc: nit: fix modifier ordering to follow conventions we use in all our codebases. E.g. this should be "internal abstract class". The properties below should be "private readonly". See https://stackoverflow.com/questions/191929/is-there-a-convention-to-the-order-of-modifiers-in-c/33237262. Not sure why our stylecop rules don't catch this - maybe they need updated.

src/Microsoft.Azure.WebJobs.Host/Executors/IFunctionActivityStatusProvider.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Microsoft.Azure.WebJobs.Host.Executors
+{
+    /// <summary>
+    /// Interface providing access to job function activity information for the job host.
+    /// </summary>
+    public interface IFunctionActivityStatusProvider
+    {
+        /// <summary>
+        /// Returns activity status of the job host
```
mathewc: Gets the current function activity status of the job host.

src/Microsoft.Azure.WebJobs.Host/Executors/IFunctionActivityStatusProvider.cs
```diff
@@ -0,0 +1,17 @@
...
+        /// </summary>
+        /// <returns></returns>
```
mathewc, line 14: The current cref=FunctionActivityStatus for the job host.

mathewc (approved):
Couple final comments otherwise looks good

PR #2785, opened 2021-11-18
Adding SupportsRetryAttribute

alrod:


src/Microsoft.Azure.WebJobs.Host/Bindings/SupportsRetryAttribute.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs.Host
```
mathewc: Recommend moving this to the root folder, peer to RetryAttribute (not in Bindings folder)

src/Microsoft.Azure.WebJobs.Host/Bindings/SupportsRetryAttribute.cs
```diff
@@ -0,0 +1,15 @@
...
+{
+    /// <summary>
+    /// Attribute applied to a <see cref="ITriggerBinding"/> implementation when the binding supports retry.
```
mathewc: nit: "... implementation to declare that the trigger supports cref=RetryAttribute."

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
```diff
@@ -182,6 +182,12 @@ internal async Task IndexMethodAsyncCore(MethodInfo method, IFunctionIndexCollec
                     {
                         throw new InvalidOperationException("More than one trigger per function is not allowed.");
                     }
+
+                    if ((method.GetCustomAttribute<RetryAttribute>() != null || method.DeclaringType.GetCustomAttribute<RetryAttribute>() != null) &&
```
mathewc: We're already pulling the RetryAttribute in the FromMethod call that CreateFunctiondescriptor makes. So rather than do it twice, recommend determining supportsRetry here and passing that down for the check later.

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
```diff
@@ -182,6 +182,12 @@ internal async Task IndexMethodAsyncCore(MethodInfo method, IFunctionIndexCollec
...
+                        triggerBinding.GetType().GetCustomAttribute<SupportsRetryAttribute>() == null)
+                    {
+                        _logger?.LogWarning($"Retries are not allowed for {triggerBinding.GetType().Name}");
```
mathewc: nit: the trigger binding type is internal, so shouldn't be logged to the user. Instead just say something like "Retries are not supported by the trigger binding for {functionName}."

mathewc (approved):

PR #2790, opened 2021-11-23
Add telemetry for retry enabled functions

alrod:
Adding telemetry that can help to identify how many customers are using retries and on what triggers

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
@@ -460,7 +460,12 @@ namespace Microsoft.Azure.WebJobs.Host.Indexers
         {
             if (descr.RetryStrategy != null)
             {
-                logger?.LogDebug($"Retry strategy is defined for function '{descr?.ShortName}' with trigger binding type '{triggerBinding?.GetType().Name}'. Strategy = '{JsonConvert.SerializeObject(descr?.RetryStrategy)}'");
+                string functionName = descr?.ShortName;
+                string triggerBindingName = triggerBinding?.GetType().Name;
+                string strategy = JsonConvert.SerializeObject(descr?.RetryStrategy);
+
+                logger?.LogDebug("Retry strategy is defined for function '{FunctionName}' with trigger binding type 'TriggerBindingName'. Strategy = '{Strategy}'", 
+                    functionName, triggerBindingName, strategy);
 
                 //Temporary comment: https://github.com/Azure/azure-webjobs-sdk/issues/2788
                 //if (triggerBinding != null && triggerBinding.GetType().GetCustomAttribute<SupportsRetryAttribute>() == null)
--- a/src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
@@ -460,11 +460,11 @@ namespace Microsoft.Azure.WebJobs.Host.Indexers
         {
             if (descr.RetryStrategy != null)
             {
-                string functionName = descr?.ShortName;
+                string functionName = descr.ShortName;
                 string triggerBindingName = triggerBinding?.GetType().Name;
-                string strategy = JsonConvert.SerializeObject(descr?.RetryStrategy);
+                string strategy = JsonConvert.SerializeObject(descr.RetryStrategy);
 
-                logger?.LogDebug("Retry strategy is defined for function '{FunctionName}' with trigger binding type 'TriggerBindingName'. Strategy = '{Strategy}'", 
+                logger?.LogDebug("Retry strategy is defined for function '{FunctionName}' with trigger binding type '{TriggerBindingName}'. Strategy = '{Strategy}'", 
                     functionName, triggerBindingName, strategy);
 
                 //Temporary comment: https://github.com/Azure/azure-webjobs-sdk/issues/2788
--- a/src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
@@ -21,6 +21,7 @@ using Microsoft.Azure.WebJobs.Logging;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
 
 namespace Microsoft.Azure.WebJobs.Host.Indexers
 {
@@ -429,9 +430,8 @@ namespace Microsoft.Azure.WebJobs.Host.Indexers
             ITriggerBinding triggerBinding, IReadOnlyDictionary<string, IBinding> nonTriggerBindings, string sharedListenerId)
         {
             var descr = FromMethod(method, _configuration, this._activator, _nameResolver, _defaultTimeout, _defaultRetryStrategy);
-
-            // Temporary comment: https://github.com/Azure/azure-webjobs-sdk/issues/2788
-            // CheckRetrySupport(triggerBinding, descr, _logger);
+            
+            CheckRetrySupport(triggerBinding, descr, _logger);
 
             List<ParameterDescriptor> parameters = new List<ParameterDescriptor>();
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
```diff
@@ -458,10 +458,17 @@ internal static string GetSharedListenerIdOrNull(ITriggerBinding triggerBinding)
 
         internal static void CheckRetrySupport(ITriggerBinding triggerBinding, FunctionDescriptor descr, ILogger logger)
         {
-            if (descr.RetryStrategy != null && triggerBinding != null && triggerBinding.GetType().GetCustomAttribute<SupportsRetryAttribute>() == null)
+            if (descr.RetryStrategy != null)
             {
-                logger?.LogWarning($"Retries are not supported by the trigger binding for { descr.ShortName}.");
-                descr.RetryStrategy = null; // Need to fix tests
+                logger?.LogDebug($"Retry strategy is defined for function '{descr.ShortName}' with trigger binding type '{triggerBinding.GetType().Name}' + " +
+                    $"Strategy = '{JsonConvert.SerializeObject(descr.RetryStrategy)}'");
```
NickCraver: Hey there! Adding a perf note since I was in the area looking at some allocations here. On this, we should add a guard to not log if we're not debug logging enabled (e.g. `if (logger?.IsEnalbed(LogLevel.Debug)` wrapper), and prefer a single formattable string (rather than 2 concatenated together). This is because we pay the price for the string formatting and JSON serialization _even if we're not logging_ as is (guard helps there), but additionally in .NET 6 we can get some advantages with overloads we'll add to `LogDebug()` that don't format the string unless it's enabled either.

Normally we wouldn't need the guard because of the second thing, but in this case we have JSON serialization in play so we don't want to do that heading in to the maybe-formatter later either.

alrod: We always log to FunctionsLogs(kusto logs) whatever log level is, so the guard would be redundant

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
```diff
@@ -458,10 +458,16 @@ internal static string GetSharedListenerIdOrNull(ITriggerBinding triggerBinding)
 
         internal static void CheckRetrySupport(ITriggerBinding triggerBinding, FunctionDescriptor descr, ILogger logger)
         {
-            if (descr.RetryStrategy != null && triggerBinding != null && triggerBinding.GetType().GetCustomAttribute<SupportsRetryAttribute>() == null)
+            if (descr.RetryStrategy != null)
             {
-                logger?.LogWarning($"Retries are not supported by the trigger binding for { descr.ShortName}.");
-                descr.RetryStrategy = null; // Need to fix tests
+                logger?.LogDebug($"Retry strategy is defined for function '{descr?.ShortName}' with trigger binding type '{triggerBinding?.GetType().Name}'. Strategy = '{JsonConvert.SerializeObject(descr?.RetryStrategy)}'");
```
pragnagopa: nit: Use log params instead of string interpolation - https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loggerextensions.loginformation?view=dotnet-plat-ext-3.1

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Indexers/FunctionIndexerTests.cs
```diff
@@ -348,8 +352,11 @@ public void CheckRetrySupport_LeavesRetry()
 
             // Act & Assert
             FunctionIndexer.CheckRetrySupport(new SupportRetryTestTriggerBinding(), desc, logger);
-            Assert.NotNull(desc.RetryStrategy);
-            Assert.DoesNotContain(loggerProvider.GetAllLogMessages(), x => x.FormattedMessage.Contains("Retries are not supported by the trigger binding for "));
+            Assert.Contains(loggerProvider.GetAllLogMessages(), x => x.FormattedMessage.Contains("Retry strategy is defined for function ") && x.Level == LogLevel.Debug);
+
+            //Temporary comment: https://github.com/Azure/azure-webjobs-sdk/issues/2788
+            //Assert.NotNull(desc.RetryStrategy);
```
pragnagopa, line 358: delete commented code?

pragnagopa (commented):
Added minor comments

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/Indexers/FunctionIndexerTests.cs
alrod: no, this code will be uncommented once https://github.com/Azure/azure-webjobs-sdk/issues/2788 is resolved

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
```diff
@@ -458,10 +458,21 @@ internal static string GetSharedListenerIdOrNull(ITriggerBinding triggerBinding)
...
-            if (descr.RetryStrategy != null && triggerBinding != null && triggerBinding.GetType().GetCustomAttribute<SupportsRetryAttribute>() == null)
+            if (descr.RetryStrategy != null)
             {
-                logger?.LogWarning($"Retries are not supported by the trigger binding for { descr.ShortName}.");
-                descr.RetryStrategy = null; // Need to fix tests
+                string functionName = descr?.ShortName;
+                string triggerBindingName = triggerBinding?.GetType().Name;
+                string strategy = JsonConvert.SerializeObject(descr?.RetryStrategy);
+
+                logger?.LogDebug("Retry strategy is defined for function '{FunctionName}' with trigger binding type 'TriggerBindingName'. Strategy = '{Strategy}'",
```
NickCraver: Thanks for going structured here! Needs just a touch of formatting love for the middle arg:
```suggestion
                logger?.LogDebug("Retry strategy is defined for function '{FunctionName}' with trigger binding type '{TriggerBindingName}'. Strategy = '{Strategy}'", 
```

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
```diff
@@ -458,10 +458,21 @@ internal static string GetSharedListenerIdOrNull(ITriggerBinding triggerBinding)
...
         internal static void CheckRetrySupport(ITriggerBinding triggerBinding, FunctionDescriptor descr, ILogger logger)
         {
-            if (descr.RetryStrategy != null && triggerBinding != null && triggerBinding.GetType().GetCustomAttribute<SupportsRetryAttribute>() == null)
+            if (descr.RetryStrategy != null)
             {
-                logger?.LogWarning($"Retries are not supported by the trigger binding for { descr.ShortName}.");
-                descr.RetryStrategy = null; // Need to fix tests
+                string functionName = descr?.ShortName;
+                string triggerBindingName = triggerBinding?.GetType().Name;
+                string strategy = JsonConvert.SerializeObject(descr?.RetryStrategy);
```
NickCraver: Minor: this is assured not to be null, so can simplify to:
```suggestion
                string strategy = JsonConvert.SerializeObject(descr.RetryStrategy);
```

NickCraver (commented):
I can't speak to comments (new to the codebase!) but spotted a quick-fix issue in there, thanks for the ping and structured logging tweaks helping keep allocations down <3

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
mathewc: If the goal of this new log is to just measure retry usage, it seems the LogFunctionRetryAttempt logging we already do on the execution path is sufficient for that? I.e. let's just query that existing log rather than add a new one.

alrod: we also want to log:
1. Retry is specified but there were no retry executions.
2. Number of retries specified by trigger type.

NickCraver (approved):
I can't speak on logging itself (need more experience there), but on perf :+1: - thanks for the tweaks!

src/Microsoft.Azure.WebJobs.Host/Indexers/FunctionIndexer.cs
mathewc: My issue with putting this log in the SDK indexing like this is that it establishes a pattern we don't want at this layer. Note that there are lots of features like Singleton, Timeout, etc. that users can use for functions, and we don't log the status of each one on startup.

This telemetry is only needed for Azure Functions, and the established pattern for this is to log metric events (logged to FunctionsMetrics table). E.g. you can see an example [here](https://github.com/Azure/azure-functions-host/blob/ebe1aafa9bf01ed56202b21a1a47a1897d2cebf9/src/WebJobs.Script.WebHost/Diagnostics/FunctionInstanceLogger.cs#L78) where we log binding type usage metrics. We should do something similar for retry metrics. This will also simplify querying of these metrics.

pragnagopa: Tagging @divyagandhisethi as FYI as the logging approach agreed here needs to be leveraged in building detectors

alrod:
closing the PR in favor of the comment https://github.com/Azure/azure-webjobs-sdk/pull/2790#discussion_r762177958

Issue #755, opened 2016-06-23
Unhandled Exception: Microsoft.WindowsAzure.Storage.StorageException: The client could not finish the operation within 


prohith14:
my continuous web job with queue is able to complete multiple messages in when run in local, but is able to process only one message when deployed in azure from visual studio. if the queue receives multiple messages it fails with above error
could any one please help me
below are the logs
[sampleResults.txt](https://github.com/Azure/azure-webjobs-sdk/files/329436/sampleResults.txt)

mathewc:
What version of the SDK are you using, and what version of Azure Storage SDK?

prohith14:
Many thanks for reply Mathew
below are the details
WebJob SDK version 1.1.2
Storage version 7.0.0

kpfaulkner:
Having the same issue here. Same vresions of Webjob SDK and Storage lib. I consistently get this error, usually around 7 minutes into the execution of the webjob. Clarification, I consistently get this error for a particular queue message (one which generates a larger/longer to process load). 

Have looked at various timeout app setting options (also various attributes such as "Timeout") but with no luck.

Any advice on how to stop this timeout?

mathewc:
Do you have these same issues if you downgrade your storage version to 6.2.0? Note that the Storage SDK version that the SDK targets (and is tested against) is 4.3 (a really old version). In the 1.1.2 release, we removed the version cap to allow people to move to later versions, however that is at your own discretion. We have run our full test passes successfully against 6.2.0, but there are known breaking changes in 7.0.0 on some codepaths.

Please reopen this if you are still experiencing issues on version 6.2.0.

prohith14:
Hi Mathew
I have changed the version and deployed it was working fine for 4 to 5 queue messages but was getting error if submitted more. below is the error
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at LPTWebJob.Functions.ProcessQueueMessage(String message, TextWriter log)

kpfaulkner:
I also tried downgrading to 6.2.0...  no luck then 4.3.0  still no luck. Same exception as before :/

prohith14:
Hey Ken, do you have3 any idea how to re-open this issue? If so please let me know so that we can re-open this am not sure if Mathew gets to know abt our issue

[...]

PR #2794, opened 2021-11-30
Discussion: WebJobs SDK Throughput Performance Improvements & Opportunities

NickCraver:
This is a draft PR **for discussion and not meant to be merged**, think "issue with some code". I figured this was the easiest way to talk about areas we can improve throughput performance in the least spammy way possible.

## Proposals Summary
These diffs are *not* proposed PRs. They are pain points or opportunities identified in profiling and experimentation where I think we can improve performance. It's likely some are viable and some are not due to constraints I'm unaware of (so let's talk!). The hope is that some of these findings help in making us faster and spur new discussions on the remaining. 

#### Goals
- Optimize what we have, with equivalent functionality - not remove any functionality.
- Don't break any use cases we have today
- Make any improvements as small single-thing PRs, not one big batch

A lot of these I file under "papercuts". Except 1 very recent regression in the `4.x` branch (which @fabiocav and @brettsam are looking at already), there is no large single item affecting performance that I can see. However, there are many small items we can potentially improve that add up. I'm numbering these in lists below so that they're easy to refer to and talk about in discussions in a quick way.

It's worth breaking these down and testing each we consider viable, though each may be a small gain in itself. In aggregate I'm seeing 24-38% gains (depending on scenario) with our typical 2 and 4 core VMs for a hello world scenario. We need to test this in App Services yet (private stamp issues...) and the gains will be proportionately smaller when the function itself is more expensive. [See #7908 for performance data](https://github.com/Azure/azure-functions-host/pull/7908).

These proposals are split across 3 repos and need discussion PRs for each (I'll update this as I create PRs)
- ([#7908](https://github.com/Azure/azure-functions-host/pull/7908)) https://github.com/Azure/azure-functions-host
- (**This PR**) https://github.com/Azure/azure-webjobs-sdk
- ([#755](https://github.com/Azure/azure-webjobs-sdk-extensions/pull/755)) https://github.com/Azure/azure-webjobs-sdk-extensions

#### Optimizations: WebJobs SDK 

- [ ] 1. `net6.0` as an additional build here yields several areas we can save or improve on, but I'm not sure of what the added cost here is for another TFM aside from double testing and such.
- [x] 2. (#2803) `Environment.CurrentDirectory` was 4-5% of all allocations since [in Windows this allocates every call](https://github.com/dotnet/runtime/blob/6cc686f5800feffb3f04bf48487cdfbd7b48a32d/src/libraries/System.Private.CoreLib/src/System/Environment.Windows.cs#L13-L39). I hope we can safely cache this.
    - Notes: In calls thus far, we couldn't think of a legitimate reason `Environment.CurrentDirectory` could change, if anything, this may be a guard against in-proc user code.
- [ ] 3. `ExecutionContextValueProvider.GetValueAsync()` creates a context per-call, but doesn't change per instance - we can re-use the created instance here.
- [x] 4. (PR: [azure-functions-host #8032](https://github.com/Azure/azure-functions-host/pull/8032), Issue: [azure-functions-host #8015](https://github.com/Azure/azure-functions-host/issues/8015)) `LoggerTraceWriter` can save allocations when not logging (this may have no production impact).
    - Notes: @paulbatum and @mathewc felt this was odd and it shouldn't be present in 4.x, so need to dig in and see if the `LoggerTraceWriter` created via `TraceWriterBinding` is intentionally still active - it origins in `builder.Services.AddSingleton<IBindingProvider, TraceWriterBindingProvider>();`.
- [ ] 5. (#2802) `IFunctionInstanceLogger` is all async adding async/await overhead, but the interface is internal and nothing is actually asynchronous. This could either be synchrnous (as in PR) or we can go `ValueTask` if it needs to remain asynchronous.
- [ ] (TBD PR & #2800) 6. `FunctionExecutor` can save a few allocations on `Stopwatch`, async overhead from number 5 above, and a few informed dictionary constructors.
- [ ] 7. `FunctionInvoker` calls `await Task.Yield()` _in case_ the function isn't async, this throws it on the task pool protecting us from a hung continuation in the synchronous case, but it also means we pay start/stop overhead on simpler cases that can be completed inline. 
    - Notes: Need follow-up with @brettsam to see if we can tweak how we're handling the root problem [described here](https://github.com/Azure/azure-webjobs-sdk/pull/799/files#r77062029). Perhaps we could kick the timeout off separately, rather than the code we want to complete quicker, etc.
- [x] 8. (**Punted**) `FastTableLoggerProvider` is populating `FunctionInstanceLogEntry.LogOutput` but this seems to only be used in testing - it's commented in the PR but we could put this capture/allocation behind options or something so that it's not a cost we pay in production.
     - Notes: This is another @mathewc and @paulbatum thought was odd in review - it's possible this shouldn't be running at all. I think this was caught in profiling 3.x before the move to 4.x and isn't a production issue today. Still, if it's active anywhere, we can tweak and improve this.
- [x] 9. (#2801) `JobHost` has a common/rare async path as well - when the host is initialized, we can elide the async/await state machine and overhead between `CallAsync` and `CallAsyncCore`.
- [ ] 10. `FunctionInstanceLogEntry` maintains a `Stopwatch` allocation/ref where we can use the start/stop ticks to get the same duration data.
- [ ] 11. `FunctionResultAggregator` has a bounded capacity on the input, which creates contention in the form of `_buffer.SendAsync()` stalls in load tests - if we could increase or remove the buffer it'd help the load test scenario without a great deal of impact given the flush interval, though this won't affect standard user throughput much if at all.
- [ ] 12. `LogCategories` has a double allocation on `CreateFunctionUserCategory` we can avoid.
- [x] 13. (#2800) `LoggerExtensions` has dictionary resize allocations we can avoid with some constructor overload additions.
- [ ] 14. `TaskSeriesTimer` if often started with a `Task.Delay()` - these can take the `CancellationToken` to avoid firing when there's no need to and shave some work off our queue.
- [ ] 15. `TaskSeriesTimer.StopAsyncCore` is always allocating a `TaskCompletionSource` and the call weight with those registrations, but the common case is a successful completion of the main task, so we can avoid that work if we've already completed. Additionally, we can use `myTask.WaitAsync(cancellationToken)` in .NET 6 her for more efficiency.
- [ ] 16. `TaskSeriesTimer.RunAsync` creates a `TaskCompletionSource` every loop though the cancellation is the same, so we can move this out of the loop - additionally we can use `.WaitAsync()` in .NET 6 here as well.
- [x] 17. (#2800) `TriggeredFunctionBinding` can size the initial dictionary to prevent a resize.

#### Other areas of interest
- Logging: as Mathew has already pointed out in earlier research, logging is where _a lot_ of our cost is. We're allocating strings in `SystemLogger` more than anywhere due to formatting before handing it off to ETW, Linux files, etc. However, this isn't true on Linux where we do further string manipulation in `LinuxAppServiceEventGenerator`.
    - Due to which `IEventGenerator`, there's a huge allocation source difference between platforms to be aware of when profiling given how many of our strings come from this path.
    - Related to the Functions Host, if we can find an alternative to Dictionary and logging scopes we are creating now, we may get a decent performance boost from it, but compat is a large concern.
- `LogFunctionTraceEvent` which allocates many more.
- Logging (dates): For simplicity/reuse we're using the constant `"O"` for our `DateTime.ToString()` calls - this would be more efficient but not as centralized as `"foo {timeVar:O}"` in formattable strings to eliminate the intermediary date string.
- Logging (theoretical): I want to raise the question if we could have structured logging (passing formattable strings and payloads down) and not allocating the strings until the point of output, e.g. to the file stream in the Linux case.
   - Notes: this is similar to saving we potentially have in the Functions Host - a small project in itself but one with likely significant burst gains. I'll at least benchmark this out but after this round of proposals since it's non-trivial.
- `IValueProvider.GetValueAsync()` is async and Task-creating when it needn't be likely the vast majority of the time - only actual async sources would need to be. We could save here with `ValueTask` but likely can't change anything given it's a public API.
   - Notes: this is public, so we're trapped here. I think it's worth benchmarking to have the numbers, but I highly doubt this would meet any threshold for being worth changing.

#### Diff Notes
- Some of the `?` additions are to fix `net6.0` build findings
- The `FunctionDescriptor.GetFunctionLogger()` is to ensure compat - ideally, we'd move this to a constructor. It's a bit odd because the Functions Host has another descriptor and I'm not sure how these should, if at all, align.
- Some `netcoreapp2.1` -> `netcoreapp3.1` upgrades to alleviate build warnings

I talked through most of the above with various people (thanks @mathewc @paulbatum and @fabiocav especially), and the next step is to start on the smaller and more straightforward PRs before moving to ones that require any discussion. If anyone has feedback of comments on any of the above proposals (presented here as code too), please add! I'm missing immense amounts of context and assume that I may have missed some edge cases which is why this discussion PR exists :)

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
@@ -79,13 +79,13 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
         public async Task<IDelayedException> TryExecuteAsync(IFunctionInstance functionInstance, CancellationToken cancellationToken)
         {
             var functionInstanceEx = (IFunctionInstanceEx)functionInstance;
-            var logger = _loggerFactory.CreateLogger(LogCategories.CreateFunctionCategory(functionInstanceEx.FunctionDescriptor.LogName));
+            var logger = functionInstanceEx.FunctionDescriptor.GetFunctionLogger(_loggerFactory);
             var functionStartedMessage = CreateStartedMessage(functionInstanceEx);
             var parameterHelper = new ParameterHelper(functionInstanceEx);
             ExceptionDispatchInfo exceptionInfo = null;
             string functionStartedMessageId = null;
             FunctionInstanceLogEntry instanceLogEntry = null;
-            Stopwatch sw = Stopwatch.StartNew();
+            var funcStart = Stopwatch.GetTimestamp();
             string concurrencyFunctionId = functionInstanceEx.FunctionDescriptor.SharedListenerId ?? functionInstanceEx.FunctionDescriptor.Id;
             
             try
@@ -121,8 +121,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
 
                     // If function started was logged, don't cancel calls to log function completed.
                     bool loggedStartedEvent = functionStartedMessageId != null;
-                    var logCompletedCancellationToken = loggedStartedEvent ? CancellationToken.None : cancellationToken;
-                    await _functionInstanceLogger.LogFunctionCompletedAsync(functionStartedMessage, logCompletedCancellationToken);
+                    _functionInstanceLogger.LogFunctionCompleted(functionStartedMessage);
 
                     if (instanceLogEntry != null)
                     {
@@ -133,7 +132,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
 
                     if (loggedStartedEvent)
                     {
-                        await _functionInstanceLogger.DeleteLogFunctionStartedAsync(functionStartedMessageId, cancellationToken);
+                        _functionInstanceLogger.DeleteLogFunctionStarted(functionStartedMessageId);
                     }
                 }
 
@@ -144,10 +143,11 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
             }
             finally
             {
-                sw.Stop();
+                var funcEnd = Stopwatch.GetTimestamp();
+                var elapsed = new TimeSpan((funcEnd - funcStart) * Stopwatch.Frequency);
                 if (_concurrencyManager.Enabled)
                 {
-                    _concurrencyManager.FunctionCompleted(concurrencyFunctionId, sw.Elapsed);
+                    _concurrencyManager.FunctionCompleted(concurrencyFunctionId, elapsed);
                 }
 
                 Interlocked.Decrement(ref _outstandingInvocations);
@@ -188,7 +188,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                 TriggerReason = functionStartedMessage.ReasonDetails,
                 StartTime = functionStartedMessage.StartTime.DateTime,
                 Properties = new Dictionary<string, object>(),
-                LiveTimer = Stopwatch.StartNew()
+                StartTimeTicks = Stopwatch.GetTimestamp()
             };
             Debug.Assert(logEntry.IsStart);
 
@@ -197,9 +197,8 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
 
         private void CompleteInstanceLogEntry(FunctionInstanceLogEntry instanceLogEntry, IDictionary<string, string> arguments, ExceptionDispatchInfo exceptionInfo)
         {
-            instanceLogEntry.LiveTimer.Stop();
             instanceLogEntry.EndTime = DateTime.UtcNow;
-            instanceLogEntry.Duration = instanceLogEntry.LiveTimer.Elapsed;
+            instanceLogEntry.Duration = Utility.GetDuration(instanceLogEntry.StartTimeTicks);
             instanceLogEntry.Arguments = arguments;
             if (exceptionInfo != null)
             {
@@ -264,7 +263,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
         {
             var outputDefinition = await _functionOutputLogger.CreateAsync(instance, cancellationToken);
             var outputLog = outputDefinition.CreateOutput();
-            var updateOutputLogTimer = StartOutputTimer(outputLog.UpdateCommand, _exceptionHandler);
+            var updateOutputLogTimer = StartOutputTimer(outputLog.UpdateCommand, _exceptionHandler, cancellationToken);
 
             try
             {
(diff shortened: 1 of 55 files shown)


src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -143,8 +143,7 @@ public async Task<IDelayedException> TryExecuteAsync(IFunctionInstance functionI
             }
             finally
             {
-                var funcEnd = Stopwatch.GetTimestamp();
```
mathewc: Are you sync'ed to current dev branch? The code [here](https://github.com/Azure/azure-webjobs-sdk/blob/bba8d6a69f77f3fea870082efcd540a138a788b4/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs#L147) isn't the same as in your PR?

mathewc: Also, there are several places in the execution pipeline where we're calling StopWatch.StartNew and/or creating per invocation StopWatches. Is there really a perf issue with using SW? If so, then I think we'd want to discuss a consistent replacement pattern to use consistently throughout. We use SW alot both in the SDK as well as in Functions Host.

NickCraver: I updated to latest (was a bit behind) - yep the Stopwatches add up overall enough to make an impact on RPS. I agree we should look at replacements overall  - wanted to put this forth as the location more than the fix.

A common pattern has been to include `ValueStopwatch` as an internal copy from the framework itself (e.g. https://github.com/Azure/azure-sdk-for-net/blob/4162f6fa2445b2127468b9cfd080f01c9da88eba/sdk/core/Azure.Core/src/Shared/ValueStopwatch.cs) - this gives us a consistent usage throughout, but making it internal would means `InternalsVisibleTo`. The pattern we have elsewhere is `[Obsolete("Not ready for public consumption.")]` - could go that way for consistency too.

Thoughts on including `ValueStopwatch` and doing one of the above for a consistent pattern? I could go through and replace all references that way.

NickCraver:
@paulbatum @mathewc Sidenote to this PR: I did poke at where TraceWriter is coming from - though deprecated, it's still in our samples: https://github.com/Azure/azure-functions-host/blob/dev/sample/CSharp/HttpTrigger/run.csx#L5 - I'll make a PR over on the host to replace it with `ILogger`.

PR #2802, opened 2021-12-06
Performance: de-async IFunctionInstanceLogger

NickCraver:
Part of #2794. In the logging path we pay overhead in async/await quite a bit around logging a function via various `IFunctionInstanceLogger` implementations, but since it's internal and none of them are actually async, we can eliminate the overhead there.


NickCraver:
@mathewc I'd love to get Crank going here, but probably need to pair on how you were doing hello world previously to get it going!

kshyju (approved):
nit: Can also remove the `using System.Threading.Tasks;` from all these files.

NickCraver:
@kshyju Good call - absolutely! Pushed up some tidy :)

kshyju (approved):

PR #2803, opened 2021-12-06
Performance: eliminate Environment.CurrentDirectory hot path

NickCraver:
Part of #2794. `Environment.CurrentDirectory` does some work and allocates every single call, though it never changes (TODO: Discuss if that's somehow an invalid assumption - @mathewc and @paulbatum didn't _think_ it could change on a quick sync here, but we need to be sure w.r.t. specialization). The exact call structure can be seen here: https://github.com/dotnet/runtime/blob/b201a16e1a642f9532c8ea4e42d23af8f4484a36/src/libraries/System.Private.CoreLib/src/System/Environment.Windows.cs#L13-L39 - it accounts for 4-5% of our string allocations on top of the calculation cost per call.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/ExecutionContext/ExecutionContextBindingProvider.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/ExecutionContext/ExecutionContextBindingProvider.cs
@@ -88,6 +88,9 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
             {
                 private readonly ValueBindingContext _context;
                 private readonly IOptions<ExecutionContextOptions> _options;
+                // This is ~4.5% of our allocations vs. the 4.x baseline and 3.3.x lib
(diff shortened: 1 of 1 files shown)


paulbatum:
I suspect this doesn't work correctly with specialization. My guess is that the current directory changes during specialization, and so we'd need to update the static at that point.

NickCraver:
If that the case, I wonder if then we should add this to an options object like `AppServiceOptions` in the host somehow (or maybe just a prop on the host?) Anywhere we can store it quickly that updates when needed is probably fair game - if it changes a property on `JobHost` _sounds_ correct in my limited expertise here - thoughts?

NickCraver:
@paulbatum I took another peek at this - while I do believe moving it higher is the right approach (higher than original PR) if we need to cache...looking at this _we don't even need to call it_ the majority of the time. We throw away the value in the hot path. I changed up the approach here to not actually invoke it unless needed, this should eliminate the race/reload question while optimizing the vast majority case :)

paulbatum (approved):
This approach makes sense to me. @mathewc can you take a quick look?

NickCraver:
@mathewc Could you take a peek at this one please? It's a surprisingly common allocation source (at least to me!) that we can get a quick win on.

brettsam (approved):

PR #2811, opened 2021-12-20
Performance: Eliminate OfType iterator allocations in FunctionExecutor

NickCraver:
This eliminates the `OfType` iterator (4 of them per function invocation) which we can the same way in almost as little code with pattern matching now.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
@@ -611,28 +611,36 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
         {
             // If the job method is an instance method and the job class implements
             // the filter interface, this filter runs first before all other filters
-            TFilter instanceFilter = instance as TFilter;
-            if (instanceFilter != null)
+            if (instance is TFilter instanceFilter)
             {
                 yield return instanceFilter;
             }
 
             // Add any global filters
-            foreach (var filter in globalFunctionFilters.OfType<TFilter>())
+            foreach (var filter in globalFunctionFilters)
             {
-                yield return filter;
+                if (filter is TFilter tFilter)
+                {
+                    yield return tFilter;
+                }
             }
 
             // Next, any class level filters are added
-            foreach (var filter in functionDescriptor.ClassLevelFilters.OfType<TFilter>())
+            foreach (var filter in functionDescriptor.ClassLevelFilters)
             {
-                yield return filter;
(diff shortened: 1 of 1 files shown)


src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -611,28 +611,36 @@ private static ITaskSeriesTimer StartParameterLogTimer(IRecurrentCommand updateC
...
-            TFilter instanceFilter = instance as TFilter;
-            if (instanceFilter != null)
+            if (instance is TFilter instanceFilter)
             {
                 yield return instanceFilter;
             }
 
             // Add any global filters
-            foreach (var filter in globalFunctionFilters.OfType<TFilter>())
+            foreach (var filter in globalFunctionFilters)
```
mathewc, line 620: What is the gain here? We're still iterating once through the filters as we were before, right?

NickCraver: Missed this! Yes we are iterating but not double iterating - we're not allocating the intermediate OfTypeIterator through LINQ per call ([which has its own `foreach`](https://github.com/dotnet/runtime/blob/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/src/libraries/System.Linq/src/System/Linq/Cast.cs#L23-L29)). Here's a sample comparison benchmark of the 2 approaches:
```cs
[ShortRunJob]
[MemoryDiagnoser]
public class IteratorBenchmarks
{
	private static List<object> things = new() {
		"foo",
		"bar",
		new object(),
		new AggregateException(),
		new object(),
		"foo",
	};
	
	[Benchmark]
	public void Direct()
	{
		foreach (var obj in things)
		{
			if (obj is string stringObj)
			{
				var i = stringObj;
			}
		}
	}

	[Benchmark]
	public void OfType()
	{
		foreach (var stringObj in things.OfType<string>())
		{
			var i = stringObj;
		}
	}
}
```
And the comparison:
| Method |       Mean |    Error |    StdDev |  Gen 0 | Allocated |
|------- |-----------:|---------:|----------:|-------:|----------:|
| Direct |   6.922 ns | 1.326 ns | 0.0727 ns |      - |         - |
| OfType | 105.029 ns | 4.618 ns | 0.2531 ns | 0.0057 |      96 B |

NickCraver:
Sorry here - clicked assigned rather than review request, my fault!

brettsam (approved):
Very cool -- did not realize any of this!

mathewc (approved):

PR #2814, opened 2021-12-29
Fix InstanceServices is null on FunctionBindingContext

swoog:
I have used FunctionBindingContext with dependency injection. 
But I cannot access to the IServiceProvider scoped to the Function instance. The property InstanceServices is null. I think this is a mistake. I do not see a set of this property. 

I have test to use _functionInvocationServices directly in my code and the behavior seems to be good.


gorillapower:
Will this  be merged in the next release? Is there a work around for now? I need to access some injected services from my Durable Entity static Run() method.

swoog:
Actual workaround I used : 
```csharp
var fieldInfo = context.FunctionContext.GetType().GetField("_functionInvocationServices", BindingFlags.Instance | BindingFlags.NonPublic);

if (fieldInfo == null)
{
    throw new ArgumentException("Not found _functionInvocationServices in the Azure Function SDK on FunctionContext (Hack not working)");
}

var serviceProvider = fieldInfo.GetValue(context.FunctionContext) as IServiceProvider;
```

eOkadas:
This was exactly what we needed. Thanks @swoog

src/Microsoft.Azure.WebJobs.Host/Bindings/FunctionBindingContext.cs
```diff
@@ -89,7 +89,7 @@ public FunctionBindingContext(IFunctionInstanceEx functionInstance, Cancellation
         /// <summary>
         /// The service provider for the current function invocation scope.
         /// </summary>
-        public IServiceProvider InstanceServices { get; set; }
+        public IServiceProvider InstanceServices => this._functionInvocationServices;
```
mathewc, line 92: Hmmm, looks like this issue has existed since this property was originally added in https://github.com/Azure/azure-webjobs-sdk/commit/f7da27d133b9ce0aadf9918770f6ed45748f3c8a.

Please add a test to the existing FunctionBindingContext test suite that tests this property and potentially others as well.

mathewc:
Closing due to no activity.

PR #2817, opened 2022-01-04
Allow to restart task series timer

bhugot:
A listener can be stopped and start again but if there is a TaskSeriesTimer under the listener.
This pr allow to start and stop and start again a listener when we want (really usefull for testing purpose).

### DIFF
--- a/test/Microsoft.Azure.WebJobs.Host.UnitTests/Timers/TaskSeriesTimerTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Host.UnitTests/Timers/TaskSeriesTimerTests.cs
@@ -303,7 +303,7 @@ namespace Microsoft.Azure.WebJobs.Host.UnitTests.Timers
         }
 
         [Fact]
-        public void StopAsync_TriggersCommandCancellationToken()
+        public async Task StopAsync_TriggersCommandCancellationToken()
         {
             // Arrange
             using (EventWaitHandle executeStarted = new ManualResetEvent(initialState: false))
@@ -331,7 +331,7 @@ namespace Microsoft.Azure.WebJobs.Host.UnitTests.Timers
 
                     // Assert
                     Assert.NotNull(task);
-                    task.GetAwaiter().GetResult();
+                    await task;
                     Assert.True(executeFinished.WaitOne(1000)); // Guard
                     Assert.True(cancellationTokenSignalled);
                 }
@@ -339,7 +339,7 @@ namespace Microsoft.Azure.WebJobs.Host.UnitTests.Timers
         }
 
         [Fact]
-        public void StopAsync_WaitsForExecuteToFinishToCompleteTask()
+        public async Task StopAsync_WaitsForExecuteToFinishToCompleteTask()
         {
             // Arrange
             bool executeFinished = false;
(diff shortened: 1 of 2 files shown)


test/Microsoft.Azure.WebJobs.Host.UnitTests/Timers/TaskSeriesTimerTests.cs
```diff
@@ -289,6 +289,19 @@ public void Start_IfStarted_Throws()
...
         }
 
+        [Fact]
+        public void Start_IfStopped_DoesNotThrow()
+        {
+            // Arrange
+            ITaskSeriesCommand command = CreateStubCommand(TimeSpan.Zero);
+            ITaskSeriesTimer product = CreateProductUnderTest(command);
+            product.Start();
+            product.StopAsync(CancellationToken.None).GetAwaiter().GetResult();
```
NickCraver: Instead of using a blocking pattern here, please make the test `async Task` instead and await the stop :)

bhugot: Done and cleaned all the tests file

NickCraver:
Code is looking good (thanks!) but builds are unhappy for unrelated reasons - I'll try and find someone with access tomorrow!

bhugot:
@NickCraver some news about the build? I tried to fix it on another PR but without success

NickCraver:
@bhugot builds seem happy but I don't have permissions to trigger them. @liliankasem @brettsam could you please poke this build please? (or can we tweak perms on AppVeyor?)

liliankasem:
> @bhugot builds seem happy but I don't have permissions to trigger them. @liliankasem @brettsam could you please poke this build please? (or can we tweak perms on AppVeyor?)

I also don't have permissions, but maybe @fabiocav or @mathewc can also help :)

Edit: permissions acquired, but build seems to still fail because of versioning? Which is odd

brettsam (commented):
To be honest, it makes me nervous to change the behavior of this class (it's been around long before I was on the project). We'd have to think through any potential race conditions that this could open up (e.g. quick start/stop/start and state not being correct)

We've gotten around this elsewhere by creating a new timer whenever we needed one -- is that possible in your scenario?

bhugot:
@brettsam it's not possible because it's created by listeners. I don't see how we could have race condition more than now on this.

brettsam:
Can you share an example? As far as I know Listeners are not meant to be re-started after they've been stopped, which is why this type also cannot be restarted.

bhugot:
Sure for example  [BlobListerner](https://github.com/Azure/azure-webjobs-sdk/blob/3f4ec78be9f43bb041937425ced00b341883aa42/src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/BlobListener.cs). To me either all listener can't be restart either their component should be able to be restarted.

liliankasem:
@brettsam @mathewc Do we still have concerns about changing this behaviour?

mathewc:
Yes I agree with Brett's point that listeners in general aren't supposed to be restarted. I get that BlobListener seems to be a counter example to that, but the general point holds. Is this change meant to address an issue or something - what's the motivation?

Also, the PR as it stands now still has some issues. For example, once stopped _cancellationTokenSource is cancelled and will remain in a cancelled state, meaning after a subsequent Start, the actual target operation won't run because the token is cancelled. Similar questions for the semantics around _initialWait which is a task passed on on construction that acts as an initial delay. For a restarted timer, that task will likely already be completed so no delay. The class was really designed to not be stopped/restarted.

So I'd say unless there is a strong reason to change this behavior, we shouldn't.

liliankasem:
Closing unless there is a strong case proposed for this change

Issue #2318, opened 2019-08-27
Azure Functions host - Lease lock http 409


hsubramaniam:
#### Question

I have a Azure function triggered by Azure Service Bus topic. Occasionally I see a lot of http 409 errors in Application Insights when the functions host tries to acquire a lock i believe and persist that lock in Azure blob storage.

url - https://<functionstoragename>.blob.core.windows.net:443/azure-webjobs-hosts/locks/functionappname/host?comp=lease

Maxconcurrentcalls is set at 16

What is causing this and is there a documentation on how this locking works? There are the other suggestions to control the web app scaling by setting its limit to 1. Not clear as to how this works. Any documentation to the locking would be helpful.

brettsam:
Are you seeing these in the Dependencies in app insights? 

The 409s for lease ownership are expected and not actually errors. When a host starts, it attempts to acquire a lock -- if another host has it, it gets a 409 back. App Insights treats anything as 400+ as an error, so it flags these. If you can share exactly where you're seeing it, we can see if there's any way we can minimize it.

OskarNet:
Any news on this?  Annoying to see thousands of dependency errors in App Insights

zmarty:
@brettsam, is it normal to see about 10K AcquireBlobLease per minute in one app service plan with 7-8 apps? Is it one lease per .. what? Per function?

vigouredelaruse:
i'm also trying to drive down the associated costs of high blob lease tranctions associated with function apps - unfortunately this suggests 1 app per storage account https://docs.microsoft.com/en-us/azure/azure-functions/storage-considerations 

i think this is less than ideal and i'm looking for a more efficient solution because of this graph
![image](https://user-images.githubusercontent.com/31052867/109530558-d7386080-7a84-11eb-9e97-57f46997c5b5.png)

the sharp drop in traffic (every 10 minutes HTTP GET /) occurred when i changed AZUREWEBJOBSSTORAGE to point to a different storage account 
![image](https://user-images.githubusercontent.com/31052867/109530900-3f874200-7a85-11eb-8fe1-ea6efb6eeefe.png)

with this kind of blob lease traffic the functions hub is essentially functioning as a denial of service attack against the associated storage account. can this be 'better'?

joaocc:
Hi
In our case we are running the function as container in AKS, already with one storage per job.
The lock seems to arise from 2 instances of the function container trying to lock a file whose name is identical, due to truncation of the pod instance.
```
[...]

PR #2819, opened 2022-01-12
Check if lease is available before trying to acquire

liliankasem:
Addressing concerns in this issue #2318 where there is a lot of noise around lease acquisition when using more than 1 instance.

This PR makes it so that we check if we can acquire a lease before trying to, this should reduce the log noise for 409 errors when non-primary instances continue to try and acquire a lease.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host.Storage/Singleton/BlobLeaseDistributedLockManager.cs
+++ b/src/Microsoft.Azure.WebJobs.Host.Storage/Singleton/BlobLeaseDistributedLockManager.cs
@@ -167,8 +167,11 @@ namespace Microsoft.Azure.WebJobs.Host
             bool blobDoesNotExist = false;
             try
             {
-                // Check if a lease is available before trying to acquire. The blob may
-                // not yet exist; if it doesn't we handle the 404, create it, and retry below
+                // Check if a lease is available before trying to acquire. The blob may not
+                // yet exist; if it doesn't we handle the 404, create it, and retry below.
+                // The reason we're checking to see if the lease is available before trying
+                // to acquire is to avoid the flood of 409 errors that Application Insights
+                // picks up when a lease cannot be acquired due to conflict; see issue #2318.
                 var blobProperties = await ReadLeaseBlobMetadata(blobClient, cancellationToken);
 
                 switch (blobProperties?.LeaseState)
--- a/src/Microsoft.Azure.WebJobs.Host.Storage/Singleton/BlobLeaseDistributedLockManager.cs
+++ b/src/Microsoft.Azure.WebJobs.Host.Storage/Singleton/BlobLeaseDistributedLockManager.cs
@@ -152,12 +152,6 @@ namespace Microsoft.Azure.WebJobs.Host
             return blobClient.GetBlobLeaseClient(proposedLeaseId);
         }
 
-        // // Allows the extension method to be mocked for testing
-        protected virtual BlobProperties GetBlobProperties(BlobClient blobClient)
-        {
-            return blobClient.GetProperties();
-        }
-
         // Allows the extension method to be mocked for testing
         protected virtual BlobContainerClient GetParentBlobContainerClient(BlobClient blobClient)
         {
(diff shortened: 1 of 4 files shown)


src/Microsoft.Azure.WebJobs.Host.Storage/Singleton/BlobLeaseDistributedLockManager.cs
```diff
@@ -167,10 +173,15 @@ protected virtual BlobContainerClient GetParentBlobContainerClient(BlobClient bl
...
             {
-                // Optimistically try to acquire the lease. The blob may not yet
-                // exist. If it doesn't we handle the 404, create it, and retry below
-                var leaseResponse = await GetBlobLeaseClient(blobClient, proposedLeaseId).AcquireAsync(leasePeriod, cancellationToken: cancellationToken);
-                return leaseResponse.Value.LeaseId;
+                // Check if a lease is available before trying to acquire.
+                // If it doesn't we handle the 404, create it, and retry below
+                var blobProperties = GetBlobProperties(blobClient);
+
+                if (blobProperties?.LeaseState != LeaseState.Leased)
```
liliankasem: Should probably consider other states and if they're safe to attempt to acquire on, maybe we should only acquire when the state is 'Available' or 'Expired'.

Unlocked lease states are: Expired, Broken, Available. Locked states: Leased, Breaking

https://docs.microsoft.com/en-us/rest/api/storageservices/lease-blob#remarks

NickCraver:
How many 409s do we get as a percentage? Asking because from a performance standpoint sometimes the cheaper path is to recognize the 409 error and then run a further check as to why and decide if to ignore. That can be pretty appealing if the percentage is low, rather than an additional call on every hit.

liliankasem:
> How many 409s do we get as a percentage? Asking because from a performance standpoint sometimes the cheaper path is to recognize the 409 error and then run a further check as to why and decide if to ignore. That can be pretty appealing if the percentage is low, rather than an additional call on every hit.

I'm not sure how to get the percentage, but the request to acquire is every 5 seconds per worker, and the request to renew from the primary host is every 12 seconds. The issue I'm trying to fix is that those 409s are coming up in application insights and we get a bunch of issues/CRIs about them. I think essentially people don't want to see them at all. So if we recognize the 409, we still have the same problem where the 409s are showing up for customers. So we figured just best to avoid the 409 all together? @brettsam for thoughts on this too!

NickCraver:
Ahh got it I think - they get logged _before_ we see the result here as part of that pipe? If so makes total sense.

liliankasem:
> Ahh got it I think - they get logged _before_ we see the result here as part of that pipe? If so makes total sense.

Yup! My understanding is that it's a trace log that's automatically picked up when you have AI enabled, maybe through the exception table? 

Something like:

- Request PUT https://storageaccount.blob.core.windows.net/azure-webjobs-hosts/locks/funcapp/host?comp=lease x-ms-lease-action:acquire x-ms-lease-duration:15
- Error response 409 There is already a lease present.

brettsam:
@liliankasem's explanation is correct. The log happens from App Insights before we can do anything (they use a Diagnostic Listener to log all outgoing requests).

App Insights flags 400+ as errors, which is ultimately the issue -- 404/409/etc are completely fine in REST, but customers don't like seeing those as errors in their logs.

We've gotten lots of pings about stuff like this over the years and while it's less performant, it does make things nicer for logs. I did something similar with another API years ago here: https://github.com/Azure/azure-webjobs-sdk/pull/1946

src/Microsoft.Azure.WebJobs.Host.Storage/Singleton/BlobLeaseDistributedLockManager.cs
```diff
@@ -167,10 +167,21 @@ protected virtual BlobContainerClient GetParentBlobContainerClient(BlobClient bl
...
             {
-                // Optimistically try to acquire the lease. The blob may not yet
-                // exist. If it doesn't we handle the 404, create it, and retry below
-                var leaseResponse = await GetBlobLeaseClient(blobClient, proposedLeaseId).AcquireAsync(leasePeriod, cancellationToken: cancellationToken);
-                return leaseResponse.Value.LeaseId;
+                // Check if a lease is available before trying to acquire. The blob may
+                // not yet exist; if it doesn't we handle the 404, create it, and retry below
+                var blobProperties = await ReadLeaseBlobMetadata(blobClient, cancellationToken);
+
+                switch (blobProperties?.LeaseState)
```
brettsam, line 177: Can you add a comment here explaining why we're doing it this way?

liliankasem: Can do!

mindraptor:
> How many 409s do we get as a percentage? Asking because from a performance standpoint sometimes the cheaper path is to recognize the 409 error and then run a further check as to why and decide if to ignore. That can be pretty appealing if the percentage is low, rather than an additional call on every hit.

FYI, With 2 instances I see 1430 messages ("Error response [542b3fbc-b0c7-4ca5-82bd-c7f81e2725ab] 409 There is already a lease present....") in the last hour in our case. These are logged as trace warnings.

mindraptor:
Which loglevel category can I use in the hosts.json to exclude warnings from this class for the time being?
The application insights logs mostly contain these 409 warnings which is a waste of data now.

brettsam (approved):

Issue #2830, opened 2022-02-08
BindingDataProvider.FromType Fails to Resolve Shadowed Properties


jsquire:
When a trigger binds to a type that shadows a base class property, the call to `BindingTypeProvider.FromType` fails with:

**Error Message:**
```text
Microsoft.Azure.WebJobs.Host.Indexers.FunctionIndexingException : Error indexing method '[[ METHOD ]]'
----> System.InvalidOperationException : Multiple properties named '[[ PROPERTY ]]' found in type '[[ TYPE ]]'.
```

**Stack Trace:**
```text
--InvalidOperationException
   at Microsoft.Azure.WebJobs.Host.Bindings.BindingDataProvider.FromType(Type type) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Bindings\BindingDataProvider.cs:line 103
   at Microsoft.Azure.WebJobs.Host.Triggers.CustomTriggerArgumentBinding`2..ctor(ITriggerBindingStrategy`2 bindingStrategy, IConverterManager converterManager, FuncAsyncConverter converter, Type userType) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Triggers\TriggerArgumentBinding\CustomTriggerArgumentBinding.cs:line 37
   at Microsoft.Azure.WebJobs.Host.Bindings.BindingFactoryHelpers.GetDirectTriggerBindingWorker[TMessage,TTriggerValue](ITriggerBindingStrategy`2 bindingStrategy, IConverterManager converterManager, Type userType) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Bindings\BindingFactoryHelpers.cs:line 42
   at Microsoft.Azure.WebJobs.Host.Bindings.BindingFactoryHelpers.GetDirectTriggerBinding[TMessage,TTriggerValue](Type exactType, ITriggerBindingStrategy`2 bindingStrategy, IConverterManager converterManager) in C:\projects\azure-webjobs-sdk-rqm4t\src\Microsoft.Azure.WebJobs.Host\Bindings\BindingFactoryHelpers.cs:line 30
   at Microsoft.Azure.WebJobs.Host.Bindings.BindingFactoryHelpers.GetTriggerArgumentElement
```

[...]

PR #2831, opened 2022-02-14
[BindingDataProvider] Allow Property Shadowing

jsquire:
# Summary

The focus of these changes is to allow the type being bound to contain properties that shadow members from the class hierarchy when declared directly.  Properties declared on the type will be preferred over those declared elsewhere in the hierarchy.   In the event that properties are shadowed in the hierarchy but not declared directly on the type being bound, it is not clear which property definition should be used, so the operation will fail.

Please see #2830 for additional context and discussion.

# References and Resources

- [BindingDataProvider.FromType Fails to Resolve Shadowed Properties (#2830)](https://github.com/Azure/azure-webjobs-sdk/issues/2830)

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingDataProvider.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingDataProvider.cs
@@ -89,44 +89,25 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
             }
 
             // The properties on user-defined types are valid binding data.
-            IReadOnlyList<PropertyHelper> bindingDataProperties = PropertyHelper.GetProperties(type);
+            IReadOnlyList<PropertyHelper> bindingDataProperties = PropertyHelper.GetVisibleProperties(type);
             if (bindingDataProperties.Count == 0)
             {
                 return null;
             }
 
             Dictionary<string, Type> contract = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);
-            Dictionary<string, Type> declaringTypes = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);
-
             foreach (PropertyHelper property in bindingDataProperties)
             {
-                if (!contract.ContainsKey(property.Name))
+                // Because only visible properties are being returned from the type hierarchy, there should
+                // be no duplication due to shadowing and this condition should never be met.  This check
+                // exists only as guard to ensure a meaningful error message should something unexpected occur.
+                if (contract.ContainsKey(property.Name))
                 {
-                    contract.Add(property.Name, property.Property.PropertyType);
-                    declaringTypes.Add(property.Name, property.Property.DeclaringType);
+                    throw new InvalidOperationException(
+                        string.Format(CultureInfo.InvariantCulture,
+                        "Multiple visible properties named '{0}' found for type '{1}'.", property.Name, type.Name));
                 }
-                else
-                {
-                    // If the mapped property is declared on the type being bound, then the property was
-                    // shadowed.  Consider the direct property to have priority and replace the existing member.
-                    if (property.Property.DeclaringType == type)
-                    {
-                        contract[property.Name] = property.Property.PropertyType;
-                        declaringTypes[property.Name] = property.Property.DeclaringType;
-                    }
-                    else if (declaringTypes[property.Name] != type)
-                    {
-                        // If the mapped property was not directly declared on the type being bound and neither
-                        // is the current property, then the property was shadowed somewhere in the base hierarchy.
-                        // Because it's not clear which property we should use, consider this an invalid scenario.
-                        throw new InvalidOperationException(
-                            string.Format(CultureInfo.InvariantCulture,
-                            "Multiple properties named '{0}' found in type '{1}'.", property.Name, type.Name));
-                    }
-
-                    // If the mapped property was directly declared on the type being bound and the current property
-                    // was not, take no action.  Keep the direct property and silently ignore the shadowed version.
-                }
+                contract.Add(property.Name, property.Property.PropertyType);
             }
 
             return new BindingDataProvider(type, contract, bindingDataProperties);
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingDataProvider.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingDataProvider.cs
@@ -96,15 +96,37 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
             }
 
             Dictionary<string, Type> contract = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);
+            Dictionary<string, Type> declaringTypes = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);
+
             foreach (PropertyHelper property in bindingDataProperties)
             {
-                if (contract.ContainsKey(property.Name))
+                if (!contract.ContainsKey(property.Name))
                 {
-                    throw new InvalidOperationException(
-                        string.Format(CultureInfo.InvariantCulture,
-                        "Multiple properties named '{0}' found in type '{1}'.", property.Name, type.Name));
+                    contract.Add(property.Name, property.Property.PropertyType);
+                    declaringTypes.Add(property.Name, property.Property.DeclaringType);
                 }
-                contract.Add(property.Name, property.Property.PropertyType);
+                else
+                {
+                    // If the mapped property is declared on the type being bound, then the property was
+                    // shadowed.  Consider the direct property to have priority and replace the existing member.
+                    if (property.Property.DeclaringType == type)
+                    {
+                        contract[property.Name] = property.Property.PropertyType;
+                        declaringTypes[property.Name] = property.Property.DeclaringType;
+                    }
+                    else if (declaringTypes[property.Name] != type)
+                    {
+                        // If the mapped property was not directly declared on the type being bound and neither
+                        // is the current property, then the property was shadowed somewhere in the base hierarchy.
+                        // Because it's not clear which property we should use, consider this an invalid scenario.
+                        throw new InvalidOperationException(
+                            string.Format(CultureInfo.InvariantCulture,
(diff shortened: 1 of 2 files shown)


jsquire:
//fyi:  @fabiocav, @joshlove-msft, @mathewc

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingDataProvider.cs
```diff
@@ -96,15 +96,37 @@ public static BindingDataProvider FromType(Type type)
             }
 
             Dictionary<string, Type> contract = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);
+            Dictionary<string, Type> declaringTypes = new Dictionary<string, Type>(StringComparer.OrdinalIgnoreCase);
```
jsquire: We need to track the declaring type for inspection.  I considered changing the `contract`  set to `Dictionary<string, PropertyInfo>` and then using a `ToDictionary` call to transform for the return, but I believe using an additional dictionary will be more efficient.

jsquire:
CI failures do not appear to be related to these changes; is this expected?

JoshLove-msft (approved):

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingDataProvider.cs
```diff
@@ -96,15 +96,37 @@ public static BindingDataProvider FromType(Type type)
...
+                    contract.Add(property.Name, property.Property.PropertyType);
+                    declaringTypes.Add(property.Name, property.Property.DeclaringType);
+                }
+                else
                 {
-                    throw new InvalidOperationException(
-                        string.Format(CultureInfo.InvariantCulture,
-                        "Multiple properties named '{0}' found in type '{1}'.", property.Name, type.Name));
+                    // If the mapped property is declared on the type being bound, then the property was
+                    // shadowed.  Consider the direct property to have priority and replace the existing member.
```
mathewc: You can have shadow properties all up the hierarchy though, not just the most derived, right? So it doesn't seem this solves the general case. Add a test case with 4 levels with shadowing at a couple levels to verify.

It seems what we want is to simply use the most derived property for the contract. The problem is the reflection APIs return separate properties for shadow properties that change the return Type. We could order the properties here based on place in the hierarchy (I don't think we can assume any ordering from Type.GetRuntimeProperties()). Then the caller can just implement first one wins.

JoshLove-msft: I think we want to use the property as long as it is directly on the type being bound to. If there is another shadowed property on a more derived type, I think we still want things to work, so we may need to relax the validation so that we only throw if there are multiple properties with the same name AND we didn't find the property on the bound type.

JoshLove-msft: Since these properties would be using "new", I don't think it is correct in general to use the most derived property.

jsquire: > You can have shadow properties all up the hierarchy though, not just the most derived, right? So it doesn't seem this solves the general case. 

That's correct; it's an explicit non-goal of this implementation to support shadowed properties deeper in the hierarchy.  _(see: #2830)_  

The rationale for this is that we don't have a strong heuristic to apply in the general case to capture the caller's intent.   For this specific case, we're being explicitly asked to bind to `type`, which makes it reasonable to assume that its direct properties are preferred and that it was most likely the type that was serialized.   

Once we start going further down, can we reasonably assume that the intent is to bind to the most direct ancestor?  Maybe.  It would support the statement that `type` was most likely what was serialized.   Maybe not, though.  It could be a case where someone serialized an ancestor deeper in the hierarchy which cannot be deserialized into the target.   Though, that statement could also be made for the direct `type` shadow as well, I believe it's less likely in the scenarios that we know about today.

We could start explicitly walking the hierarchy and accepting the property on the most direct ancestor, but do we think this is a mainstream enough case to take the performance hit?  Should we put the engineering investment in to try and include an optimized path for the non-shadowed case and fall back to the more robust inspection if we do detect shadowing?  I don't know that this scenario is common enough for either; I'm not sure that I would advocate for a more robust approach until we can prove that we need it.  Worst case, we can always revisit and extend.

> Add a test case with 4 levels with shadowing at a couple levels to verify.

With respect to tests, `FromType_DuplicatePropertiesDetected_Throws` on line 57 verifies that shadowing in the hierarchy is not supported, following the existing behavior of throwing an `InvalidOperationException`.

JoshLove-msft: Deleted my comments as I wasn't thinking about this correctly and @jsquire summarized the considerations.

mathewc: For a trigger binding of type T, only T and derived types can be bound, not ancestors. So for a BindingDataProvider for T, it will only ever be asked to resolve properties of T or derived types (see [here](https://github.com/Azure/azure-webjobs-sdk/blob/d8ac3619d29461f068b657051df20474d39dd200/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingDataProvider.cs#L51)).

The intent of the binding contract is to provide the value for each contract property from the instance being bound. I.e. what value would you get if you invoke each contract property by name on the instance being bound to. Those would be  the most derived property definitions.

As far performance, this isn't a concern since property helpers are created once then cached. I think something like I try in https://github.com/Azure/azure-webjobs-sdk/tree/bindingdata-fix would work.

jsquire: Yup; that's pretty similar to the approach that I was thinking of for "walk the hierarchy".   If we're not concerned with the performance impact, then we could bypass my thoughts on an optimized path and a fallback path and just pay for the additional allocations and reflection for all cases, which simplifies things like you've prototyped.

I'm not 100% clear on the safety, and it may just be my ignorance, but I'm assuming someone chose to not support shadowing for a reason.   The case that I'm thinking of is something like:

```csharp
public class Top : SingleBase
{
    public string Name { get; set; }
}

public class SingleBase : BaseBase
{
    public int Count { get; set; }
}

public class BaseBase
{
    public long Count { get; set; }
}
```

Let's say that, for some reason, the service received a serialized instance of `BaseBase` and returns it to the trigger with the following:

```json
{
    "Count": 9223372036854775804
}
```

 The trigger attempts to bind to `Top` so when things get bound, we end up either:

1) Throwing when attempting to bind 

1)  With an instance that has an incorrect value due to the down-casting:

```csharp
instance of Top
{
    public string Name = <null>,
    public int Count = -4
}
```

To be transparent, I do not have enough depth of knowledge to know whether this is a legit concern or whether there are additional safeguards in there.   In the first case where we throw, assuming that gets surfaced correctly, then it seems like we're good.   In the second case or if that exception gets swallowed, that seems like we'd be doing the wrong thing.

From your perspective, do we need to be concerned there, or do we think that building a map of the visible properties of `type` by walking the hierarchy is safe?

mathewc: Yes you're right. Initially I was thinking that "new" properties completely replaced the base properties they hide, but further experimentation shows that they are still there and usable. E.g. for an instance of Top t, I can set both `t.Count = int.MaxValue` and `((BaseBase)t).Count = long.MaxValue`.

Likely there was a reason why this hard exclusion was put in place, but I'm not aware of the history. In the code there is actually a [PropertyHelper.GetVisibleProperties](https://github.com/Azure/azure-webjobs-sdk/blob/3f4ec78be9f43bb041937425ced00b341883aa42/src/Microsoft.Azure.WebJobs.Host/PropertyHelper.cs#L140) method which we're not using for binding providers.

However, the IBindingDataProvider is never used for producing/deserializing an instance. It's only used for reading binding data members defined in the contract. The instance passed to IBindingDataProvider.GetBindingData is produced by the trigger itself. For the purposes of binding data access, it seems safe to bind to only the most derived property. @brettsam any thoughts on this?

mathewc: I think we can simply switch to using the GetVisibleProperties method. There's low risk for breaking changes here, since we don't currently support types with duplicate/hidden properties, and for types w/o such properties GetVisibleProperties returns the same results as GetProperties.

I prototyped this in https://github.com/Azure/azure-webjobs-sdk/tree/bindingdata-fix.

jsquire: Nice... I overlooked `GetVisibleProperties`.     So, to sum up the discussion and next steps:

- We're comfortable supporting full-hierarchy shadowing, though we don't understand the history of why it was explicitly disallowed.

- The infrastructure to do so already exists; we would only need to swap from `PropertyHelper::GetProperties` to `PropertyHelper::GetVisibleProperties`.

- We'll need to remove the existing test that validates shadowing is not supported.

- We'll want to add some additional test cases to validate shadowing at the top level and lower in the hierarchy.


@mathewc:  Does that match what you're thinking?

mathewc: Yes, that's what I'm thinking

jsquire: Updated per discussion.

mathewc (approved):

brettsam (approved):

mathewc:
I had to create a new branch PR to get this to build on CI (we have issues with forks). PR https://github.com/Azure/azure-webjobs-sdk/pull/2836 was merged. Thanks for your contribution :)

jsquire:
@mathewc:  Thank you for helping to push this through.   Do we know approximately when it will be included in a release?

mathewc:
These changes will be part of the 3.0.33 version of [Microsoft.Azure.WebJobs](https://www.nuget.org/packages/Microsoft.Azure.WebJobs/). Might be a few weeks. We have some schedule constraints on these releases - we first have to release a new Azure Functions version taking a dependency on the new package from our myget feed, then once that is rolled out fully to prod, we can update nuget.

PR #2832, opened 2022-02-15
Add ExtensionInfo to the DI system

TsuyoshiUshio:
Enable to fetch ExtensionInfo from DI from `WebJobsExtensionOptionsConfiguration<TOptions>` 

This instance is used from Azure Functions Host for a new API for `host.json`  payload with default value.


TsuyoshiUshio:
Hi @mathewc 
One thing I want to discuss is these lines.
https://github.com/Azure/azure-webjobs-sdk/blob/12e195b86ad3a9221ca2c242e102281639c6c2eb/src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsExtensionBuilderExtensions.cs#L74-L75

I wanted to use the same instance for it. I tried several ways. The 2nd method success for testing, however, according to the result of the testing (ConfigureAndBindOptions_BindsToConfiguration test, not hydrated the appsettings), it is lambda expression, so that, it might depend which interface is used first if the hydration happens.  So that I talked with @pragnagopa  and decided to keep both implementation factory that create an instance.  If you have any suggestions, please let me know. 

_test fails_

```
builder.Services.AddSingleton<IWebJobsExtensionOptionsConfiguration<TOptions>>((IServiceProvider p) => new WebJobsExtensionOptionsConfiguration<TOptions>(p.GetService<IConfiguration>(), builder.ExtensionInfo, configure));

builder.Services.AddSingleton<IWebJobsExtensionOptionsConfiguration<TOptions>>((IServiceProvider p) => new WebJobsExtensionOptionsConfiguration<TOptions>(p.GetService<IConfiguration>(), builder.ExtensionInfo, configure));
            builder.Services.AddSingleton<IConfigureOptions<TOptions>>((IServiceProvider p) => {
                var s = p.GetService<IWebJobsExtensionOptionsConfiguration<TOptions>>();
                return s;    
            });
```

_works_ 

```
builder.Services.AddSingleton<IConfigureOptions<TOptions>>((IServiceProvider p) => new WebJobsExtensionOptionsConfiguration<TOptions>(p.GetService<IConfiguration>(), builder.ExtensionInfo, configure));

builder.Services.AddSingleton<IWebJobsExtensionOptionsConfiguration<TOptions>>((IServiceProvider p) => (IWebJobsExtensionOptionsConfiguration<TOptions>)p.GetService<IConfigureOptions<TOptions>>());

```

TsuyoshiUshio:
Hi @chiangvincent 
This is the PR for the new API.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsExtensionBuilderExtensions.cs
```diff
@@ -69,9 +71,9 @@ public static class WebJobsExtensionBuilderExtensions
                 throw new ArgumentNullException(nameof(configure));
             }
 
-            builder.Services.AddSingleton<IConfigureOptions<TOptions>>(p =>
-                new WebJobsExtensionOptionsConfiguration<TOptions>(p.GetService<IConfiguration>(), builder.ExtensionInfo.ConfigurationSectionName, configure));
-
+            builder.Services.AddSingleton<IConfigureOptions<TOptions>>((IServiceProvider p) => new WebJobsExtensionOptionsConfiguration<TOptions>(p.GetService<IConfiguration>(), builder.ExtensionInfo, configure));
```
mathewc: This will add multiple different instances to the container which isn't what we want. We want to register the same single instance and have that same instance be returned regardless of which interface it is requested through. There are already examples of how to do this, e.g. [here](https://github.com/Azure/azure-functions-host/blob/d49c3b94d89d0cec419b49ef22999fe6ae57d9c3/src/WebJobs.Script.WebHost/WebHostServiceCollectionExtensions.cs#L159). In this example, WebJobsScriptHostService is registered as WebJobsScriptHostService, IHostedService, and later below as IScriptHostManager.

With the IWebJobsExtensionOptionsConfiguration interface non-generic, we'll add each of these instances via `services.TryAddEnumerable<IWebJobsExtensionOptionsConfiguration>(p => p.GetRequiredService<IConfigureOptions<TOptions>())` which is how you'll be able to query for all of them from host in a single call in a non-generic way.

src/Microsoft.Azure.WebJobs.Host/Config/IWebJobsExtensionOptionsConfiguration.cs
```diff
@@ -0,0 +1,16 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Options;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.WebJobs.Host.Config
+{
+    public interface IWebJobsExtensionOptionsConfiguration<TOptions> : IConfigureOptions<TOptions> where TOptions : class
```
mathewc: This interface shouldn't be generic - remove the <TOptions> from it

src/Microsoft.Azure.WebJobs.Host/Extensions/ExtensionInfo.cs
```diff
@@ -11,7 +11,7 @@
 
 namespace Microsoft.Azure.WebJobs
 {
-    public class ExtensionInfo
+    public class ExtensionInfo : IExtensionInfo
```
mathewc: We don't need the IExtensionInfo interface - ExtensionInfo is already a public type. Why can't you use that type in the interface above?

src/Microsoft.Azure.WebJobs.Host/Config/IWebJobsExtensionOptionsConfiguration.cs
TsuyoshiUshio: Hi @mathewc  
I think we need it. The following code is usage of the Azure Functions Host side. If we have multiple extensions registered, how can we identify it? If you know the better way, please let me know. 

https://github.com/Azure/azure-functions-host/blob/178dc626799e747be0b9575ebe2d380b868eca35/src/WebJobs.Script.WebHost/HostJsonConfigProvider.cs#L77-L81

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsExtensionBuilderExtensions.cs
TsuyoshiUshio: Hi @mathewc  
I tried, TryAdd with non-generic interface strategy, however, it throws exception. TryAddEnumerable only allows ServiceDescriptor.   I tested two ways. 

```csharp
builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IWebJobsExtensionOptionsConfiguration>(p =>
            
      (IWebJobsExtensionOptionsConfiguration)p.GetRequiredService<IConfigureOptions<TOptions>>()
 ));
```
It causes 

```
System.ArgumentException : Implementation type cannot be 'Microsoft.Azure.WebJobs.Host.Config.IWebJobsExtensionOptionsConfiguration' because it is indistinguishable from other services registered for 'Microsoft.Azure.WebJobs.Host.Config.IWebJobsExtensionOptionsConfiguration'. (Parameter 'descriptor')
```
I found one issue on the dotnet repo. https://github.com/dotnet/runtime/issues/56105 I use the workaround as well, however, it doesn't work. Am I missing something?

src/Microsoft.Azure.WebJobs.Host/Config/IWebJobsExtensionOptionsConfiguration.cs
TsuyoshiUshio: Done

src/Microsoft.Azure.WebJobs.Host/Extensions/ExtensionInfo.cs
TsuyoshiUshio: Testability. The other solution is introduce a class that implements `IExtensionConfigProvider` on the TestClass. If you prefer the approach I will do it.

src/Microsoft.Azure.WebJobs.Host/Config/IWebJobsExtensionOptionsConfiguration.cs
TsuyoshiUshio: Done

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsExtensionBuilderExtensions.cs
TsuyoshiUshio: Hi @mathewc 
I tried the approach. TryAddEnumerable strategy works well. However, TryAddEnumerable + `p.GetRequiredService<IConfigureOptions<TOptions>()` approach doesn't work. 

## The `p.GetRequiredService` approach doesn't work
Using this approach, it works with testing on the WebJobs side, however, when I work from Azure Functions Host side, I see the following error. It looks confusing with the other configuration. It can be possible that my DI configuration on Azure Functions is wrong. 

```csharp
          builder.Services.AddSingleton<IConfigureOptions<TOptions>>(p => new WebJobsExtensionOptionsConfiguration<TOptions>(p.GetService<IConfiguration>(), builder.ExtensionInfo, configure));
            builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IWebJobsExtensionOptionsConfiguration, WebJobsExtensionOptionsConfiguration<TOptions>>(p => (WebJobsExtensionOptionsConfiguration<TOptions>)p.GetRequiredService<IConfigureOptions<TOptions>>()));
```

_error_

```
System.InvalidCastException: 'Unable to cast object of type 
'Microsoft.Extensions.Options.ConfigureNamedOptions`1[Microsoft.Azure.WebJobs.EventHubs.EventHubOptions]' to type 
'Microsoft.Azure.WebJobs.Host.Config.WebJobsExtensionOptionsConfiguration`1[Microsoft.Azure.WebJobs.EventHubs.EventHubOptions]'.'
```

## current approach works
However, I have duplicate instance. That the part I don't like. 

```csharp
builder.Services.AddSingleton<IConfigureOptions<TOptions>>(p => new WebJobsExtensionOptionsConfiguration<TOptions>(p.GetService<IConfiguration>(), builder.ExtensionInfo, configure));
builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IWebJobsExtensionOptionsConfiguration, WebJobsExtensionOptionsConfiguration<TOptions>>(p => new WebJobsExtensionOptionsConfiguration<TOptions>(p.GetService<IConfiguration>(), builder.ExtensionInfo, configure)));
```

src/Microsoft.Azure.WebJobs.Host/Extensions/ExtensionInfo.cs
mathewc: We don't want to add new unnecessary public surface area just for our testability. To solve your test issue, just use one of the ExtensionInfo factory methods to create an instance.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsExtensionBuilderExtensions.cs
```diff
@@ -69,8 +72,8 @@ public static class WebJobsExtensionBuilderExtensions
                 throw new ArgumentNullException(nameof(configure));
             }
 
-            builder.Services.AddSingleton<IConfigureOptions<TOptions>>(p =>
-                new WebJobsExtensionOptionsConfiguration<TOptions>(p.GetService<IConfiguration>(), builder.ExtensionInfo.ConfigurationSectionName, configure));
+            builder.Services.AddSingleton<IConfigureOptions<TOptions>>(p => new WebJobsExtensionOptionsConfiguration<TOptions>(p.GetService<IConfiguration>(), builder.ExtensionInfo, configure));
+            builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IWebJobsExtensionOptionsConfiguration, WebJobsExtensionOptionsConfiguration<TOptions>>(p => new WebJobsExtensionOptionsConfiguration<TOptions>(p.GetService<IConfiguration>(), builder.ExtensionInfo, configure)));
```
mathewc: We'll need to find a way to not return duplicate instances - there has to be a way. We can't have it be the case that we have 2 different instances floating around with different behaviors.

mathewc: I wonder if a better solution which also sidesteps the instance issue is to instead register instances of a type ExtensionOptionsInfo which can be constructed using TOptions and builder.ExtensionInfo so we don't have to delay create. This way we don't have to modify the existing configure type with the new interface.

Having it be a class rather than an interface will allow us to add additional info to it if we need to in the future (interfaces make that hard).

src/Microsoft.Azure.WebJobs.Host/Extensions/ExtensionInfo.cs
TsuyoshiUshio: Sure. will do.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsExtensionBuilderExtensions.cs
TsuyoshiUshio: Done

src/Microsoft.Azure.WebJobs.Host/Extensions/ExtensionInfo.cs
TsuyoshiUshio: Removed

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsExtensionBuilderExtensions.cs
TsuyoshiUshio: I introduce the change.  I can't find the interface with just add the instance. However, make it clear difference between two.

TsuyoshiUshio:
Tested with this change on Azure Functions Host. 
https://github.com/Azure/azure-functions-host/pull/8148/commits/5d588ac5d8ea2819146b99c54a9e55bcc00958be

src/Microsoft.Azure.WebJobs.Host/Extensions/IExtensionInfo.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.WebJobs
+{
+    public interface IExtensionInfo
```
mathewc: why is this still here? you said you were removing it

src/Microsoft.Azure.WebJobs.Host/Config/WebJobsExtensionOptionsConfiguration.cs
```diff
@@ -9,20 +9,24 @@ namespace Microsoft.Azure.WebJobs.Host.Config
 {
     internal class WebJobsExtensionOptionsConfiguration<TOptions> : IConfigureOptions<TOptions> where TOptions : class
     {
-        private readonly string _extensionName;
         private readonly IConfiguration _configuration;
         private readonly Action<IConfiguration, string, TOptions> _configure;
+        private readonly ExtensionInfo _extensionInfo;
 
-        public WebJobsExtensionOptionsConfiguration(IConfiguration configuration, string extensionName, Action<IConfiguration, string, TOptions> configure)
+        public WebJobsExtensionOptionsConfiguration(IConfiguration configuration, ExtensionInfo extensionInfo, Action<IConfiguration, string, TOptions> configure)
```
mathewc: Changes to this file are no longer needed now, are they?

src/Microsoft.Azure.WebJobs.Host/Config/ExtensionOptionsInfo.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs.Host.Config
+{
+    public class ExtensionOptionsInfo<TOptions> : IExtensionOptionsInfo
```
mathewc: As in my initial suggestion for this type, make it non-generic, and remove the interface. We can just register instances under service type ExtensionOptionsInfo, and the type is then easily extensible in the future.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Configuration/WebJobsExtensionOptionsConfigurationTests.cs
```diff
@@ -19,7 +21,7 @@ public void Configuration_IsBound()
                 { "azurewebjobs:extensions:testext:testvalue2", "test2" },
             };
 
-            TestOptions options = ConfigureOptions<TestOptions>(config, "testext");
+            TestOptions options = ConfigureOptions<TestOptions, TestExtensionConfigProvider>(config);
```
mathewc: I don't see any new tests for the new functionality? E.g. I'd expect a test verifying that when ConfigureOptions is called for an extension, the new ExtensionOptionsInfo gets registered

src/Microsoft.Azure.WebJobs.Host/Config/ExtensionOptionsInfo.cs
TsuyoshiUshio: Hi @mathewc  
I understood your suggestion, however, I was not sure how we can register multiple `ExtensionOptionInfo` instances on the DI? For example.  The current registration using TryAddEnumerable to register multiple `ExtensionOptionsInfo<T>` instances with `IExtensionOptionsInfo`.  However, we can't register something like this `ServiceDescriptor.Singleton<ExtensionOptionInfo, ExtensionOptionInfo>(p => new ExtensionOptionInfo(builder.ExtensionInfo));`. Is there any other way to register multiple instances of the same class?
```
builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IExtensionOptionsInfo, ExtensionOptionsInfo<TOptions>>(p => new ExtensionOptionsInfo<TOptions>(builder.ExtensionInfo)));
```

mathewc: I see - yes, it seems we need an interface then. @fabiocav what are your thoughts on this? As part of the options configuration pipeline, we're registering instances of this type to enable us to later query all of them to discover all extensions/option Types registered. Is it strange to register "info" types like this? Usually types registered are services that provide behavior.

I guess another option would be for us to register a singleton IExtensionOptionsProvider (or some such name), and that service is responsible for returning the ExtensionOptionInfos.

src/Microsoft.Azure.WebJobs.Host/Extensions/IExtensionInfo.cs
TsuyoshiUshio: Thanks. Removed.

src/Microsoft.Azure.WebJobs.Host/Config/WebJobsExtensionOptionsConfiguration.cs
TsuyoshiUshio: Good point. Rollback the change.

src/Microsoft.Azure.WebJobs.Host/Config/ExtensionOptionsInfo.cs
TsuyoshiUshio: I talked with @fabiocav and change to the current design.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Configuration/WebJobsExtensionOptionsConfigurationTests.cs
TsuyoshiUshio: Added Functions Testing instead.

TsuyoshiUshio:
Hi @fabiocav and @mathewc 

Could you have a look? I introduced a change that I discussed with @fabiocav . 
To avoid reflection on the Azure Functions Side, We add Options class as `IOptionsFormatter`. I tested with Azure Functions Host  side. I updated the PR as well. https://github.com/Azure/azure-functions-host/pull/8148

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsExtensionBuilderExtensions.cs
```diff
@@ -38,7 +42,7 @@ public static class WebJobsExtensionBuilderExtensions
         /// <param name="builder">The <see cref="IWebJobsExtensionBuilder"/> to configure.</param>
         /// <param name="configure">The <see cref="Action{IConfigurationSection, TOptions}"/> that will be invoked when configuring the options.</param>
         /// <returns>The configured <see cref="IWebJobsExtensionBuilder"</returns>
-        public static IWebJobsExtensionBuilder ConfigureOptions<TOptions>(this IWebJobsExtensionBuilder builder, Action<IConfigurationSection, TOptions> configure) where TOptions : class
+        public static IWebJobsExtensionBuilder ConfigureOptions<TOptions>(this IWebJobsExtensionBuilder builder, Action<IConfigurationSection, TOptions> configure) where TOptions : class, new()
```
mathewc: I think the addition of new() might be potentially breaking here? I think all options classes already have to have default constructors (i.e. if they didn't they already wouldn't work) but technically this would be a breaking change I think to the methods, so if not necessary I'd say we omit this.

src/Microsoft.Azure.WebJobs.Host/Config/ExtensionOptionsInfo.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.WebJobs.Hosting;
+using System;
+
+namespace Microsoft.Azure.WebJobs.Host.Config
+{
+    public class ExtensionOptionsInfo : IExtensionOptionsInfo
```
mathewc: Because the interface is public, the implementation type ExtensionOptionsInfo can remain internal, right? To reduce public surface area.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsExtensionBuilderExtensions.cs
```diff
@@ -72,6 +76,11 @@ public static class WebJobsExtensionBuilderExtensions
             builder.Services.AddSingleton<IConfigureOptions<TOptions>>(p =>
                 new WebJobsExtensionOptionsConfiguration<TOptions>(p.GetService<IConfiguration>(), builder.ExtensionInfo.ConfigurationSectionName, configure));
 
+            if (typeof(IOptionsFormatter).IsAssignableFrom(typeof(TOptions)))
+            {
+                builder.Services.AddSingleton<IExtensionOptionsInfo>(p => new ExtensionOptionsInfo(builder.ExtensionInfo, p.GetService<IOptions<TOptions>>().Value as IOptionsFormatter));
```
mathewc: It's possible ConfigureOptions to be called multiple times for the same options type, with different configure actions. In that case, above we correctly register multiple IConfigureOptions instances so at build time they'll be applied in order.

For our new IExtensionOptionsInfo instances however, they won't vary based on the action. Can you test what happens here - perhaps update your unit test so that for one of your test extensions, you call ConfigureOptions directly multiple times, with different actions that set options properties. I'm assuming we'll end up with multiple duplicate IExtensionOptionsInfos registered which likely doesn't cause any problems, however it is strange. I wonder if there's a way we can avoid that.

fabiocav: This will be equivalent of registering enumerables. You'll have a collection of them that you can get by taking a dependency on `IEnumerable<IExtensionOptionsInfo>`

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsExtensionBuilderExtensions.cs
fabiocav: Yes. This would need to be evaluated against the consumers as it is a constraint addition and would be breaking if violated.

fabiocav: @TsuyoshiUshio can you try updating this to use `IOptionsMonitor` as opposed to `IOptions` here?

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsExtensionBuilderExtensions.cs
mathewc: Right - in both cases, you'll end up with a collection of separate instances, each of which is effectively the same. Likely doesn't happen that often, and not a huge concern since consumer can just pick one, I just didn't know if there was a simple pattern we could employ here to only register ONE for a given TOptions/builder pair.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsExtensionBuilderExtensions.cs
TsuyoshiUshio: Done!

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsExtensionBuilderExtensions.cs
TsuyoshiUshio: Update as we talked.

src/Microsoft.Azure.WebJobs.Host/Config/ExtensionOptionsInfo.cs
TsuyoshiUshio: Good point. Done.

mathewc (approved):

Issue #2660, opened 2021-01-19
Storage Extension Feature Request: Log the details about the Storage account bindings for Queues and Blobs 


FinVamp1:
#### Repro steps

Customers use the Queue and Blob Storage account bindings very frequently.
If these don't work then we need to ask the customer for the details of those storage accounts to see if there is an issue on that side. 
Adding this information when the Storage extension starts to listen on the Queue or Blob container would allow us to be able to 
a) Build Detectors that access this information if the customer does raise a support request.
b) Look at automated solutions that will check connectivity to the Storage account and allow the customer to resolve transient issues on their own. 

#### Expected behavior

1.  A log entry is written when we attach to the binding. 

#### Actual behavior

1.  We log information about the queue name or the blob container but not the FQDN of the Storage account. 

#### Known workarounds

No workarounds exist.

sidkri:
Option 1: Expand on logic in azure-functions-host\src\WebJobs.Script.WebHost\Management to include event source information
Option 2: Log these details at startup and have an analytics job that keeps a copy of startup events in a separate table.
Option 3: Log these details on each trigger event

fabiocav:
@sidkri do we need to discuss this work item, or is this something you're expecting to handle?

sidkri:
@fabiocav I'm handling this one but will be reviewing the solution with you before submitting a PR.

mathewc:
Is the ask only for triggers, or for input/output bindings as well? What about binding types other than Queues (e.g. CosmosDB, etc.)? For triggers, I can see us logging some account/queue details on listener startup.

fabiocav:
We also need to make sure any changes are ported to the track 2 SDK extensions.

sidkri:
At present the focus is for both triggers and input/output bindings (any place a customer configures a connection with another Azure service).  Based on frequency of issues, the initial focus is Storage, Service Bus and Event Hubs.

fabiocav:
@sidkri should we move this to the next sprint?

sidkri:
> @sidkri should we move this to the next sprint?

Yes, I have updated it just now to the next sprint.

fabiocav:
Moving. this back to triaged. Please assign to an appropriate sprint if needed.

alrod:
This is done for both track1 and track2

PR #2856, opened 2022-04-12
Log Details of Storage Blob/Queue Listeners 

hossam-nasr:
Log the details of a storage blob/queue listener on start/stop operations. Addresses this WI: https://dev.azure.com/FunctionsExtensions/Functions%20Extensions/_workitems/edit/36 and https://github.com/Azure/azure-webjobs-sdk/issues/2660

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/BlobListener.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/BlobListener.cs
@@ -23,6 +23,7 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
         // for mock test purposes only
         internal BlobListener(ISharedListener sharedListener)
         {
+            _sharedListener = sharedListener;
         }
 
         public BlobListener(ISharedListener sharedListener, CloudBlobContainer container, ILoggerFactory loggerFactory)
--- a/src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/BlobListener.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/BlobListener.cs
@@ -35,23 +35,15 @@ namespace Microsoft.Azure.WebJobs.Host.Blobs.Listeners
 
         public Task StartAsync(CancellationToken cancellationToken)
         {
-            try
-            {
-                ThrowIfDisposed();
+            ThrowIfDisposed();
 
-                if (_started)
-                {
-                    throw new InvalidOperationException("The listener has already been started.");
-                }
-
-                _logger.LogDebug($"Storage blob listener started ({_details})");
-                return StartAsyncCore(cancellationToken);
-            }
-            catch (Exception ex)
+            if (_started)
             {
-                _logger.LogError(ex, $"Storage blob listener exception during starting ({_details})");
-                throw;
+                throw new InvalidOperationException("The listener has already been started.");
             }
+
+            _logger.LogDebug($"Storage blob listener started ({_details})");
+            return StartAsyncCore(cancellationToken);
         }
 
         private async Task StartAsyncCore(CancellationToken cancellationToken)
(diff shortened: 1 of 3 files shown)


src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/BlobListener.cs
```diff
@@ -6,31 +6,52 @@
...
+        // for mock test purposes only
+        internal BlobListener(ISharedListener sharedListener)
         {
             _sharedListener = sharedListener;
         }
 
+        public BlobListener(ISharedListener sharedListener, CloudBlobContainer container, ILoggerFactory loggerFactory)
+        {
+            _sharedListener = sharedListener;
+            _details = new Lazy<string>(() => $"blob container={container.Name}, storage account name={container.ServiceClient.GetAccountName()}");
```
alrod: why do we use `Lazy` here?

alrod:
Original task was only about including the source details(blob - account, container, path, queue - account, queue name):
[Issue 34](https://dev.azure.com/FunctionsExtensions/Functions%20Extensions/_workitems/edit/34): Log the event source that functions trigger off of

Let's remove the logic around stopping listener and catching events. Please [see](https://github.com/Azure/azure-sdk-for-net/pull/25568/files).

src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/BlobListener.cs
alrod: also would be useful to log full[ blob name pattern](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-storage-blob-trigger?tabs=in-process%2Cextensionv5&pivots=programming-language-csharp#blob-name-patterns) from `IBlobPathSource`

src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Listeners/QueueListener.cs
```diff
@@ -141,17 +143,29 @@ public Task StartAsync(CancellationToken cancellationToken)
         {
             ThrowIfDisposed();
             _timer.Start();
+            _logger.LogDebug($"Storage queue listener started: {_details.Value}");
```
alrod: We use internal queue for blob triggers - after detecting a new blob we create a message in internal queue and then execute the function based on internal queue. Do we use the same "QueueListener" for the purpose? Will we log the details for internal queue?

src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/BlobListener.cs
hossam-nasr: I saw the `Lazy` pattern in other logging PRs (for example: https://github.com/Azure/azure-sdk-for-net/pull/25566/files). I'm not sure what it accomplished there either, but I just copied that example

hossam-nasr: Removed `Lazy` and added logging the full blob name pattern

src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Listeners/QueueListener.cs
hossam-nasr: It looks like yes, we do :/ I could add a flag to the constructor of `QueueListener`, e.g. `shouldLogDetails`, which is `false` for this scenario and `true` for all others. Let me know what you think

src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/BlobListener.cs
```diff
@@ -6,21 +6,33 @@
...
+        private string _details;
 
-        public BlobListener(ISharedListener sharedListener)
+        // for mock test purposes only
+        internal BlobListener(ISharedListener sharedListener)
         {
             _sharedListener = sharedListener;
         }
 
+        public BlobListener(ISharedListener sharedListener, CloudBlobContainer container, ILoggerFactory loggerFactory, IBlobPathSource blobPathSource)
```
alrod: Lets move `ILoggerFactory` to the last place in the ctr signature

src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Listeners/QueueListener.cs
alrod: I think it's ok, it can be also be useful info for us when we starting/stopping listening the internal queue

alrod (approved):
with a minor comment

src/Microsoft.Azure.WebJobs.Extensions.Storage/Queues/Listeners/QueueListener.cs
hossam-nasr: Sounds good!

src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/BlobListener.cs
hossam-nasr: Done!

PR #2858, opened 2022-04-15
do not format ObjectDisposedExceptions

brettsam:
One part of improving https://github.com/Azure/azure-functions-host/issues/7873.

Our `HostDisposedException` grabs the full stack trace -- but the formatting done here overwrites that. This skips the special formatting if it's an `ObjectDisposedException`, which I'm updating `HostDisposedException` to derive from.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Protocols/FunctionFailure.cs
+++ b/src/Microsoft.Azure.WebJobs.Protocols/FunctionFailure.cs
@@ -33,7 +33,7 @@ namespace Microsoft.Azure.WebJobs.Host.Protocols
             {
                 Exception = ex,
                 ExceptionType = ex.GetType().FullName,
-                ExceptionDetails = ex.ToDetails()
+                ExceptionDetails = ex is ObjectDisposedException ? ex.ToString() : ex.ToDetails() // Do not format ObjectDisposedExceptions to aid in debugging
             };
         }
     }


src/Microsoft.Azure.WebJobs.Protocols/FunctionFailure.cs
```diff
@@ -33,7 +33,7 @@ internal static FunctionFailure FromException(Exception ex)
             {
                 Exception = ex,
                 ExceptionType = ex.GetType().FullName,
-                ExceptionDetails = ex.ToDetails()
+                ExceptionDetails = ex is ObjectDisposedException ? ex.ToString() : ex.ToDetails() // Do not format ObjectDisposedExceptions to aid in debugging
```
kshyju, line 36: Should we do this inside the `ToDetails`? I wonder folks will miss doing this conditional check in future when they have to use this exception type in some new code they are writing.

kshyju, line 36: Should we do this inside the `ToDetails`? I wonder folks will miss doing this conditional check in future when they have to use this exception type in some new code they are writing.

kshyju (approved):

fabiocav (approved):

src/Microsoft.Azure.WebJobs.Protocols/FunctionFailure.cs
brettsam: Good point, but I didn't want `ToDetails` to get even more fancy and this is meant to just handle one very specific case with the `HostDisposedException` stack trace getting overwritten.

PR #2859, opened 2022-04-19
Adding more storage hanging timeouts handling.

alrod:


mathewc:
I'd like to get @brettsam's thoughts/review on this as well since he did the original work/investigations for this issue and has the most context.

Also, it seems more likely to me that adding timeout handling to the log scan ListBlobsAsync call [here](https://github.com/Azure/azure-webjobs-sdk/blob/3d31a1cdc59721d42833a25aa299ebcaa5f74f44/src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/BlobLogListener.cs#L169) is more in line with the previous fix, rather than the 2 places you added.

src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/StorageAnalyticsLogParser.cs
```diff
@@ -80,7 +85,12 @@ public StorageAnalyticsLogParser(ILogger<BlobListener> logger)
         {
             List<StorageAnalyticsLogEntry> entries = new List<StorageAnalyticsLogEntry>();
 
-            using (TextReader tr = new StreamReader(await blob.OpenReadAsync(cancellationToken)))
+            OperationContext operationContext = new OperationContext { ClientRequestID = Guid.NewGuid().ToString() };
+            using (TextReader tr = new StreamReader(await TimeoutHandler.ExecuteWithTimeout("ParseLogAsync", operationContext.ClientRequestID, _exceptionHandler,
+                    _logger, cancellationToken, () =>
+                    {
+                        return blob.OpenReadAsync(accessCondition: null, options: null, operationContext: operationContext, cancellationToken);
```
mathewc, line 92: Not clear to me that the OpenReadAsync API could possibly be affected by the original issue addressed by https://github.com/Azure/azure-storage-net/pull/985 - @brettsam?

alrod:
> Also, it seems more likely to me that adding timeout handling to the log scan ListBlobsAsync call [here](https://github.com/Azure/azure-webjobs-sdk/blob/3d31a1cdc59721d42833a25aa299ebcaa5f74f44/src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/BlobLogListener.cs#L169) is more in line with the previous fix, rather than the 2 places you added.

the ListBlobsAsync call is already [doing](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/CloudBlobClientExtensions.cs#L34) timeout handling inside

src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/StorageAnalyticsLogParser.cs
brettsam: The hangs were almost all a result of the SDK internally parsing XML responses from the storage apis like ListBlobs (if i recall). You can see the stack in the linked issue above.

This operation directly on a blob doesn't look to be in the same family as those -- but it seems like we need to start gathering more details here to figure out why this is stopping.

brettsam:
Have we been able to get any dumps of these scenarios yet? The other possibility is that there's some unhandled exception being thrown that's killing the polling -- although I would that we would have identified that by now for all the years this code has been running.

mathewc:
> Have we been able to get any dumps of these scenarios yet? The other possibility is that there's some unhandled exception being thrown that's killing the polling -- although I would that we would have identified that by now for all the years this code has been running.

Another thing we should be doing for cases when they come up is to check the "Slow Request Stack Tracing" AppLens detector which gives stack traces from dumps automatically taken during platform proactive AutoHeal events https://azure.github.io/AppService/2017/08/17/Introducing-Proactive-Auto-Heal.html. A few recent CRIs I've investigated that involved hangs/deadlocks this feature provided stack traces pointing to the actual issue. Perhaps we should wait on this until we get another case where we get some actual stack details?

alrod:
Ok, lets wait for the another occurrence of the issue and try to figure out the stack trace with "Slow Request Stack Tracing" AppLens detector or a dump. Closing the PR for now.

PR #2866, opened 2022-05-13
Add hook for telemetry processor from host

madelinegordon:
Part of changes for host issue [#4296](https://github.com/Azure/azure-functions-host/issues/4296)

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsLoggingBuilderExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsLoggingBuilderExtensions.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using Microsoft.ApplicationInsights.Extensibility;
 using Microsoft.Azure.WebJobs.Logging.ApplicationInsights;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging.Configuration;
@@ -43,11 +44,22 @@ namespace Microsoft.Extensions.Logging
         /// Registers Application Insights and <see cref="ApplicationInsightsLoggerProvider"/> with an <see cref="ILoggingBuilder"/>.
         /// </summary>        
         public static ILoggingBuilder AddApplicationInsightsWebJobs(
+             this ILoggingBuilder builder,
+             Action<ApplicationInsightsLoggerOptions> configure)
+        {
+            return builder.AddApplicationInsightsWebJobs(configure, _ => { });
+        }
+
+        /// <summary>
+        /// Registers Application Insights and <see cref="ApplicationInsightsLoggerProvider"/> with an <see cref="ILoggingBuilder"/>.
+        /// </summary>  
+        public static ILoggingBuilder AddApplicationInsightsWebJobs(
             this ILoggingBuilder builder,
-            Action<ApplicationInsightsLoggerOptions> configure)
+            Action<ApplicationInsightsLoggerOptions> configureOptions,
+            Action<TelemetryConfiguration> additionalConfiguration)
         {
             builder.AddConfiguration();
-            builder.Services.AddApplicationInsights(configure);
+            builder.Services.AddApplicationInsights(configureOptions, additionalConfiguration);
 
             builder.Services.PostConfigure<LoggerFilterOptions>(o =>
             {
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsLoggingBuilderExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsLoggingBuilderExtensions.cs
@@ -45,9 +45,9 @@ namespace Microsoft.Extensions.Logging
         /// </summary>        
         public static ILoggingBuilder AddApplicationInsightsWebJobs(
              this ILoggingBuilder builder,
-             Action<ApplicationInsightsLoggerOptions> configure)
+             Action<ApplicationInsightsLoggerOptions> loggerOptionsConfiguration)
         {
-            return builder.AddApplicationInsightsWebJobs(configure, null);
+            return builder.AddApplicationInsightsWebJobs(loggerOptionsConfiguration, null);
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
```diff
@@ -41,7 +48,8 @@ public static IServiceCollection AddApplicationInsights(this IServiceCollection
             return services;
         }
 
-        public static IServiceCollection AddApplicationInsights(this IServiceCollection services)
```
kshyju, line 44: Are we removing an existing `public` method? Wouldn't that be a breaking change?

madelinegordon: Adding an overload - this method signature still exists line 34

kshyju: Yea, I missed that. Thanks!

kshyju:
Don't we need a package version update as well?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsLoggingBuilderExtensions.cs
```diff
@@ -42,12 +43,23 @@ public static class ApplicationInsightsLoggingBuilderExtensions
...
+        }
+
+        /// <summary>
+        /// Registers Application Insights and <see cref="ApplicationInsightsLoggerProvider"/> with an <see cref="ILoggingBuilder"/>.
+        /// </summary>  
         public static ILoggingBuilder AddApplicationInsightsWebJobs(
             this ILoggingBuilder builder,
-            Action<ApplicationInsightsLoggerOptions> configure)
+            Action<ApplicationInsightsLoggerOptions> configureOptions,
+            Action<TelemetryConfiguration> additionalConfiguration)
```
liliankasem: Can we call this telemetryConfiguration to match the type? additional config sounds very vague

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsLoggingBuilderExtensions.cs
```diff
@@ -42,12 +43,23 @@ public static class ApplicationInsightsLoggingBuilderExtensions
         /// <summary>
         /// Registers Application Insights and <see cref="ApplicationInsightsLoggerProvider"/> with an <see cref="ILoggingBuilder"/>.
         /// </summary>        
+        public static ILoggingBuilder AddApplicationInsightsWebJobs(
+             this ILoggingBuilder builder,
+             Action<ApplicationInsightsLoggerOptions> configure)
```
liliankasem: Also feel like we can name is to be more descriptive i.e. loggerOptions or loggerConfiguration and apply the naming across the application insights methods

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
```diff
@@ -41,7 +48,8 @@ public static IServiceCollection AddApplicationInsights(this IServiceCollection
...
+        internal static IServiceCollection AddApplicationInsights(this IServiceCollection services, Action<ApplicationInsightsLoggerOptions> configureOptions,
```
liliankasem: nit: break out the new params into a new line for readability

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
```diff
@@ -41,7 +48,8 @@ public static IServiceCollection AddApplicationInsights(this IServiceCollection
...
+            Action<TelemetryConfiguration> additionalConfiguration)
```
liliankasem: just wondering: why are these actions?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
```diff
@@ -31,6 +31,13 @@ namespace Microsoft.Extensions.DependencyInjection
...
     internal static class ApplicationInsightsServiceCollectionExtensions
     {
+
+        public static IServiceCollection AddApplicationInsights(this IServiceCollection services)
+        {
+            return services.AddApplicationInsights(_ => { }, _ => { });
+        }
+
+
         public static IServiceCollection AddApplicationInsights(this IServiceCollection services, Action<ApplicationInsightsLoggerOptions> configure)
```
liliankasem: I wonder if you could keep the base method for `AddApplicationInsights` the same and just create a new method like this one that takes in the telemetry configuration where you call `AddApplicationInsights(configure)` and after that you can call `additionalConfiguration?.Invoke(configuration);` - it looks like that `configuration` value is returned so you can use it?

> side note: same comments in this class about clearing up the naming i.e. ApplicationInsightsLoggerOptions -> loggerOptions or loggerConfiguration and additionalConfiguration -> telemetryConfiguration

madelinegordon: additionalConfiguration?.Invoke(configuration); is called inside SetupTelemetryConfiguration, which operates on the TelemetryConfiguration object that's instantiated in line 186 inside of a services.AddSingleton\<TelemetryConfiguration\> call; that config object isn't returned by AddApplicationInsights.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
madelinegordon: My understanding is that they're actions so that the webjobs sdk doesn't need to know about the specific classes implementing TelemetryConfiguration and ApplicationInsightsLoggerOptions in the host.

brettsam (approved):

PR #2887, opened 2022-06-30
Providing public GetMemoryLimitBytes

alrod:
Making public TotalAvailableMemoryBytes to access from function host

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/AppServicesHostingUtility.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/AppServicesHostingUtility.cs
@@ -1,0 +1,73 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    public static class AppServicesHostingUtility
+    {
+        private const int BytesPerGB = 1024 * 1024 * 1024;
+
+        public static long GetMemoryLimitBytes(string sku = null, int? numCores = null)
+        {
+            sku ??= Utility.GetWebsiteSku();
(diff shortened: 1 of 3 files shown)


src/Microsoft.Azure.WebJobs.Host/Scale/IHostProcessMonitor.cs
```diff
@@ -12,6 +12,11 @@ namespace Microsoft.Azure.WebJobs.Host.Scale
     /// </summary>
     public interface IHostProcessMonitor
     {
+        /// <summary>
+        /// Total available memory in bytes.
+        /// </summary>
+        long TotalAvailableMemoryBytes { get; }
```
mathewc: This doesn't belong on this interface. Furthermore, it would be a breaking change to a public interface. If the goal is to expose the sku/core memory limit calculation, then the simple way to do it would be to expose a public helper method that does the calculation.

However, the reason all of this is internal currently is because testing revealed issues and it wasn't working as expected. So before we make any of it public, I'd like to see tests showing that behaviors based on dynamic memory calculations actually work and improve target scenarios.

alrod: I thought we can change public interfaces in webjobs sdk if we are releasing webjobs nuget after function runtime release -  [example](https://github.com/Azure/azure-webjobs-sdk/pull/2745/files) (changing public class). Do we never modify public classes/interfaces defined in webjob sdk or we can do it some circumstances?
Anyway a helper method looks better approach, what class should I use to add the method?

I did prod tests when gradually added language workers which consume RAM - after exceeding the RAM limit the function app became unhealthy - some of the workers became unhealthy or function host restarted abruptly.

mathewc (approved):

PR #2906, opened 2022-08-25
Bump app insights dependencies

### LINKED ISSUE #2900: Update dependency on Microsoft.ApplicationInsights
A new [version](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fwww.nuget.org%2Fpackages%2FMicrosoft.ApplicationInsights%2F2.21.0&data=05%7C01%7CLove.Josh%40microsoft.com%7Cee33cb46b71c4d831e7808da770a8908%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637953184376476224%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=UV6kova49ZDTzhCJDIF3Yv4jFwBQwgm1DifDaK07Qmc%3D&reserved=0) of the App Insights library has been released that adds additional support for the new Service Bus library. In order for Service Bus extension users to benefit from these changes, we would need to update the dependency version used in the WebJobs SDK, which would in turn pull up the version used by the Functions host.

ilmax:
Fixes #2900

I've updated all the application insights dependencies to 2.21 and the SnapshotCollector to the latest stable 1.4.3, anything else that needs to be done here @JoshLove-msft ?

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/WebJobs.Logging.ApplicationInsights.csproj
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/WebJobs.Logging.ApplicationInsights.csproj
@@ -24,16 +24,16 @@
   </ItemGroup>
   
   <ItemGroup>
-  <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.17.0" />
-  <PackageReference Include="Microsoft.ApplicationInsights.PerfCounterCollector" Version="2.17.0" />
-  <PackageReference Include="Microsoft.ApplicationInsights.SnapshotCollector" Version="1.3.7.5" />
-  <PackageReference Include="Microsoft.ApplicationInsights.WindowsServer" Version="2.17.0" />
-  <PackageReference Include="Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel" Version="2.17.0" />
-  <PackageReference Include="StyleCop.Analyzers" Version="1.1.0-beta004">
+    <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.21.0" />
+    <PackageReference Include="Microsoft.ApplicationInsights.PerfCounterCollector" Version="2.21.0" />
+    <PackageReference Include="Microsoft.ApplicationInsights.SnapshotCollector" Version="1.4.3" />
+    <PackageReference Include="Microsoft.ApplicationInsights.WindowsServer" Version="2.21.0" />
+    <PackageReference Include="Microsoft.ApplicationInsights.WindowsServer.TelemetryChannel" Version="2.21.0" />
+    <PackageReference Include="StyleCop.Analyzers" Version="1.1.0-beta004">
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
-    <PackageReference Include="Microsoft.ApplicationInsights" Version="2.17.0" />
-    <PackageReference Include="Microsoft.ApplicationInsights.DependencyCollector" Version="2.17.0" />
+    <PackageReference Include="Microsoft.ApplicationInsights" Version="2.21.0" />
+    <PackageReference Include="Microsoft.ApplicationInsights.DependencyCollector" Version="2.21.0" />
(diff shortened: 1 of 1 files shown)


ilmax:
@mathewc By any chance, can you have a look at this one?

mathewc:
Looks good to me - @brettsam see any issues here? I notice that even the Functions Host hasn't pulled in this new version yet - [it's still on 2.20](https://github.com/Azure/azure-functions-host/blob/438b1d4daf431dccefbe09d76f2f96fb8acdd694/src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj#L66).

brettsam:
I swear there was a reason we had to abandon updates to this package -- maybe a dependency could cause v3 to break if we took it due to some major bump?

Let me try to figure it out...

brettsam:
That doesn't seem to be it -- we may have just started doing these changes directly in azure-functions-host for convenience. That meant we didn't have to do a release, etc, just to bump to a newer version.

@ilmax -- I assume the end-goal for these versions to be in Functions production?

brettsam:
n/m -- appears to be a WebJobs requirement based on the linked issue. Right now, customers should be able to directly reference the required App Insights nuget packages in their WebJobs app; nothing should be preventing that...

Unfortunately, our release flow for these packages is slow due to the dependency in Functions production. So once this is in, the flow needs to be:

1. We should probably follow-up and sever the project dependency on WebJobs.Host so we don't need a full release of everything just for this. Instead, we can have it depend on the last deployed nuget...
2. Get an official build and move it all the way to host in v3.x and dev.
3. Wait until this makes its way to production.
4. Then publish this package on nuget.

brettsam:
The build fails here b/c of some test dependencies. Let me pull this and fix it up.

ilmax:
> n/m -- appears to be a WebJobs requirement based on the linked issue. Right now, customers should be able to directly reference the required App Insights nuget packages in their WebJobs app; nothing should be preventing that...
> 
So we just need to reference the right version of AppInsights to fix the issue? I'd like still to update AppInsight package here to do the right thing by default, since the old ServiceBus library is deprecated and the new one isn't supported properly by the referenced version of the AppInsights library

brettsam:
Any direct reference in your project will override what we reference here and should get you running with the latest; this shouldn't be blocking anyone. It's not required that we update (although it's a nicer experience).

That being said, I've created a new PR to update b/c it was easier to manage from testing, etc. I'm going to close this one in favor of that: https://github.com/Azure/azure-webjobs-sdk/pull/2916

JoshLove-msft:
> Any direct reference in your project will override what we reference here and should get you running with the latest; this shouldn't be blocking anyone. It's not required that we update (although it's a nicer experience).

Is this true for the Isolated worker? I was under the impression that users couldn't influence the dependencies used there.

ilmax:
> Is this true for the Isolated worker? 

I updated my webjobs and referenced the 2.21 version of application insights, but still the issue is not resolved, within application insights I still don't have end to end tracing of service bus messages.

@brettsam Any updates on #2916? This is causing real issues to customers. Any ETA on when the PR one can be merged an a new version of the WebJob SDK be released?

brettsam:
@ilmax -- to be clear, you are running a WebJobs app, not a Functions app?

You may need to explain what you're trying to do. WebJobs and Functions don't do much with SB tracing so if updating to 2.21.0 didn't work, we may need to follow up with SB and App Insights teams.

Next release of Functions will include 2.21.0 -- https://github.com/Azure/azure-functions-host/pull/8838

ilmax:
>@ilmax -- to be clear, you are running a WebJobs app, not a Functions app?

Correct.

In order to get it working I had to add to every project an additional reference (Microsoft.ApplicationInsights.DependencyCollector v 2.21.0) that's actually where the code for dealing with dependency is.

So it can be worked around, it just takes times, it would be much easier to reference the correct AI version in the sdk.

brettsam:
Thanks for confirming you were able to get it working.

The commit is in, but it takes a while (a month maybe) before it can be released to nuget.org as it needs to be released globally in Functions production first.

PR #2907, opened 2022-08-26
Target base scale

### LINKED ISSUE #2909: Target base scale in runtime driven scale
Intension
In order to GA target base scale we need to add the feature into runtime driven scale, the feature was already implemented in `ScaleController`. This work intersects with SC redesign but should be released before as SC redesign is bigger item and it will take longer. SC redesign will use the change in the PR for target base scale(TBS).

Changes/Design
New interfaces:
`ITargetScaler` - Interface defining an Azure Functions scaler that makes scale decisions based on current event source metrics and function concurrency. Needs to be implemented in an extension.
`ITargetScalerManager` - Manager for registering and accessing `ITargetScaler` instances for `JobHost` instance. Implemented in WebJobs SDK.
`ITargetScalerProvider` - Provider interface for returning an `ITargetScaler` instance. Needs to be implemented in an extension.

Flow:
* An extension implements `ITargetScaler`
* `HostListenerFactory` in webjobs registers `ITargetScaler` using `ITargetScalerManager.Register`
* SC calls FunctionHost's api: `admin/host/scale/status`
* FunctionHost gets all the registered scalers `ITargetScalers` by `ITargetScalerManager.GetTargetScalers` 
* FunctionHost figures out `ITargetScalers` and `IScaleMonitors` to process. If an extension implements both and TBS is enabled only `ITargetScalers` is used.
* FunctionHost calls `ITargetScaler.GetScaleResultAsync` and `IScaleMonitor.GetScaleStatus` and return response to SC

The flow is similar to regular(incremental) runtime scale.

alrod:
Resolves: https://github.com/Azure/azure-webjobs-sdk/issues/2909


build/common.props
```diff
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <!-- Packages can have independent versions and only increment when released -->
-    <Version>3.0.34$(VersionSuffix)</Version>
+    <Version>3.0.35$(VersionSuffix)</Version>
```
mathewc: We haven't released official version 3.0.34 so we're still in the same minor version train and shouldn't need to update

src/Microsoft.Azure.WebJobs.Host/Scale/ITargetScaleMonitor.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Host.Scale
+{
+    public interface ITargetScaleMonitor : IScaleMonitor
+    {
+        Task<int> GetScaleVoteAsync(ScaleStatusContext context);
```
mathewc, line 9: Rather than add a new interface, we should be able to just extend the existing ScaleStatus response returned from the existing IScaleMonitor.GetScaleStatus method. E.g. we could add a new TargetWorkerCount property (defaults to -1). Then for an extension supporting TBS, it can return this along with the ScaleVote it already returns.

src/Microsoft.Azure.WebJobs.Host/Scale/DynamicTargetValueProvider.cs
```diff
@@ -0,0 +1,103 @@
...
+
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Host.Scale
+{
+    internal class DynamicTargetValueProvider : IDynamicTargetValueProvider
```
mathewc, line 12: Extensions using ConcurrencyManager already have access to the persisted concurrency value via their calls to ConcurrencyManager.GetStatus which they make as part of dynamic scaling. There's a hosted service running [ConcurrencyManagerService](https://github.com/Azure/azure-webjobs-sdk/blob/62bbc94767672b17600f8329a5d36fee364a7c72/src/Microsoft.Azure.WebJobs.Host/Scale/ConcurrencyManagerService.cs#L65) responsible for reading the status and applying it on startup, as well as persisting updates on a timer. So I'm wondering if rather than adding a new service to read this value, we can't instead just use the value provided by [ConcurrencyStatus.CurrentConcurrency](https://github.com/Azure/azure-webjobs-sdk/blob/1f7f383c2f48237a76f9b3112c79e4a3800ec4d9/src/Microsoft.Azure.WebJobs.Host/Scale/ConcurrencyStatus.cs#L71)?

src/Microsoft.Azure.WebJobs.Host/Scale/ITargetScaleMonitor.cs
alrod: Adding a public member to a public class is a "potential" breaking change in some cases. I tested the new member with FunctionHost referencing new WebJobs SDK + old extension referencing old WebJobs SDK and it works as expected. 
Do you think it's safe to add a new member to the public class here?

src/Microsoft.Azure.WebJobs.Host/Scale/DynamicTargetValueProvider.cs
alrod: @chiangvincent, as @mathewc said above we can read the dynamic concurrency value is currently used by a function host so there is no need to read from the blob. Do you think we can go this path?

Scale controller has the logic to cache the concurrency value, also the concurrency value can be expired and we return fallback in this case.

chiangvincent: Interesting, this should be doable in scale controller, as long as there are no permissions issues for the blob in storage when using the ConcurrencyManagerService.

alrod: I do not think it would be doable in SC as `ConcurrencyManagerService `have access to executions and dynamic concurrency engine directly to return the dynamic concurrency value. `ConcurrencyManagerService `reads the blob once on startup to set up initial concurrency and then primary host instance stores current dynamic value to the blob.

alrod: Question is if it's ok to use `ConcurrencyManager.GetStatus` value for target base scale. The value is calculated dynamically based on executions and different metrics.

chiangvincent: @pragnagopa @TsuyoshiUshio @seilorjunior as fyi

seilorjunior:
@pragnagopa @seilorjunior @TsuyoshiUshio

src/Microsoft.Azure.WebJobs.Host/Scale/DynamicTargetValueProvider.cs
chiangvincent: The current logic in scale controller has an additional layer of logic on retrieval of this value. For example, logic to check if the value has stabilized and if we should use it at all. As long as we have stabilization logic built on top of the value retrieved from `ConcurrencyManager.GetStatus` it should be ok to use.

alrod: it looks like the dynamic currency its own logic around stabilization:
https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Scale/ConcurrencyStatus.cs#L16

do we really need another layer of the stabilization?
@mathewc, do you think dynamic concurrency value can be unstable  some time and we need do additional stabilization for the target base scale?

mathewc: The concurrency value for each function is adjusted over time. Initially it starts at 1 and over time gets adjusted based on extensions calling into ConcurrencyManager.GetStatus. Periodically that value is written to storage by ConcurrencyManagerService [here](https://github.com/Azure/azure-webjobs-sdk/blob/62bbc94767672b17600f8329a5d36fee364a7c72/src/Microsoft.Azure.WebJobs.Host/Scale/ConcurrencyManagerService.cs#L111). So what do you mean exactly by "unstable" - it's dynamic, so yes will tend to stabilize over time.

My main point was that we don't need a new service that extensions use to read this value from storage, since extensions already have access to the source of that value already - ConcurrencyStatus.CurrentConcurrency. An extension when asked to determine it's scale status (vote/TargetWorkerCount) should just use the most current information it has.

src/Microsoft.Azure.WebJobs.Host/Scale/ITargetScaleMonitor.cs
mathewc: Yes it's fine to add a new property to a public type - these types were designed with this in mind to make them extensible, exactly for reasons such as this.

src/Microsoft.Azure.WebJobs.Host/Scale/DynamicTargetValueProvider.cs
alrod: Regarding main point it makes sense - we do not need to read the value from storage as we can get current value using `ConcurrencyManager.GetStatus`. Agree.

"Unstable" means that the dynamic value can fluctuate a lot and we want to ignore peak values to avoid overscale/underscale. @chiangvincent correct me if I am wrong. Question is do we want this to be implemented as in SC controller or we want to rely only on current dynamic value.

pragnagopa: I recommend going with what is built into SDK - current dynamic value. We can have thresholds in SC specifically to keep checks on overscale / underscale

alrod:
closing in favor https://github.com/Azure/azure-webjobs-sdk/pull/2918

PR #2916, opened 2022-09-16
updating ApplicationInsights to 2.21; decoupling ApplicationInsights package from Host

brettsam:
Bumping App Insights, but also changing the project to allow for it to be released independently of a WebJobs release.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/WebJobs.Logging.ApplicationInsights.csproj
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/WebJobs.Logging.ApplicationInsights.csproj
@@ -2,6 +2,7 @@
   <Import Project="..\Microsoft.Azure.WebJobs.Shared\WebJobs.Shared.projitems" Label="Shared" />
   <Import Project="..\..\build\common.props" />
   <PropertyGroup>
+    <Version>3.0.34$(VersionSuffix)</Version>
     <TargetFramework>netstandard2.0</TargetFramework>
     <PackageId>Microsoft.Azure.WebJobs.Logging.ApplicationInsights</PackageId>
     <Description>This package contains the runtime assemblies for Microsoft.Azure.WebJobs.Logging.ApplicationInsights. For more information, please visit http://go.microsoft.com/fwlink/?LinkID=320971</Description>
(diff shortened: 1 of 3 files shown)


brettsam:
Rather than do this here -- I've just updated it directly in Functions at https://github.com/Azure/azure-functions-host/pull/8838

brettsam:
This appears to be confusing enough people that I'll re-open and submit it here. But if you stumble across this issue before the new nuget is published -- you can directly add references to the latest AppInsights packages in your WebJobs project to workaround this.

RohitRanjanMS (approved):

fowl2:
> This appears to be confusing enough people that I'll re-open and submit it here. But if you stumble across this issue before the new nuget is published -- you can directly add references to the latest AppInsights packages in your WebJobs project to workaround this.

It would be really great if the release notes (and docs) made clear the best way of handling app insights dependencies!

czb182:
I see this release in 3.0.34, but its not on nuget yet

ilmax:
Hey @brettsam any update on this? This is fixing a **critical** functionality, it's has been merged 3 months ago and still no package update available. Why is this one taking so long?

brettsam:
You should be able to directly reference the newest App Insights in your project. I thought this change was a (major) convenience but didn't think it was blocking anything critical.

That being said -- I'll work on getting it released this week.

ilmax:
Hey @brettsam thanks for the answer.

This is not blocking because there's a workaround but I think it's critical because the current state is broken by default. On top of that people need to spend a lot of time to find the workaround.

The underlying issue has been fixed months ago, so I was assuming a new release to be available soon after.

I am personally waiting for this to undo the workarounds.

Thank you

brettsam:
@ilmax -- I went to do this yesterday and realized that this package is still not fully released to Functions production (due to the holidays, which set things back a while). This is a requirement before we push these to nuget -- if we push to nuget too early, it can lead to runtime issues if a customer deploys a newer version than what's in prod.

The good news is there's only one release stage left so this should be good to go to nuget sometime next week.

ilmax:
@brettsam thank you for the info! 
Looking forward to it next week then.

brettsam:
Sorry for the additional delay. It's now published: https://www.nuget.org/packages/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/3.0.34

PR #2919, opened 2022-09-27
Roranjan/tranmission status

### LINKED ISSUE #2649: Detect if Application Insights is not able to connect to the Telemetry and Live Metrics Endpoints.
Customers run their Functions in a number of environments.
In some of these environments like Isolated and\or Functions Premium it is possible that they may have configured the Virtual Networks in such a way as to block access to the endpoints.
E.G DNS, Route Tables, NSG's etc.

https://docs.microsoft.com/en-us/azure/azure-monitor/app/ip-addresses 

The customer may also choose to block ingestion from the public internet.

https://docs.microsoft.com/en-us/azure/azure-monitor/platform/private-link-security#configure-application-insights

#### Repro steps

1. Configure a New Function App and enable Application Insights. 

2. Turn off Public Network Access ingestion on the Application Insights resource. 

image

#### Expected behavior

That the Functions runtime would let you know when it is unable to send data to Application Insights. 

#### Actual behavior

No monitor data is available and there's no way to know why.

#### Known workarounds

No workarounds exist.

RohitRanjanMS:
Fixes [this](https://github.com/Azure/azure-webjobs-sdk/issues/2649) issue.

Track ApplicationInsights ingestion service response by subscribing to transmission event. It will only log in case of non-sucess response or if the debug trace feature flag (EnableDebugTracing) is enabled, this will log all the request.

e.g. 
```sh
Diagnostic source 'Microsoft.Azure.Functions.Host.ApplicationInsights' emitted event 'TransmissionStatus': {"statusCode":400,"statusDescription":"Invalid instrumentation key","id":"vkZu1wA1u3c=","items":[{"type":"AppTraces","count":3},{"type":"AppRequests","count":1},{"type":"AppMetrics","count":12}],"responseTimeInMs":65,"timeout":"00:01:40"}
```

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Constants/LoggingConstants.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Constants/LoggingConstants.cs
@@ -8,5 +8,7 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
         public const string ZeroIpAddress = "0.0.0.0";
         public const string Unknown = "[Unknown]";
         public const string ClientIpKey = "ClientIp";
+        public const string HostDiagnosticSourcePrefix = "Microsoft.Azure.Functions.Host.";
+        public const string HostDiagnosticSourceDebugEventNamePrefix = "debug-";
     }
 }
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
@@ -80,6 +80,7 @@ namespace Microsoft.Extensions.DependencyInjection
             services.AddSingleton<ITelemetryInitializer, MetricSdkVersionTelemetryInitializer>();
             services.AddSingleton<QuickPulseInitializationScheduler>();
             services.AddSingleton<QuickPulseTelemetryModule>();
+            services.AddSingleton<ITelemetryModule, TransmissionStatusTelemetryModule>();
 
             services.AddSingleton<ITelemetryModule>(provider =>
             {
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TransmissionStatusTelemetryModule.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TransmissionStatusTelemetryModule.cs
@@ -47,82 +47,63 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
             source.Dispose();
         }
 
-        public void Handler(object sender, TransmissionStatusEventArgs args)
+        internal void Handler(object sender, TransmissionStatusEventArgs args)
         {
-            // Do not block the main thread
-            Task.Run(() =>
+            if (args == null)
             {
-                if (sender != null && args != null)
+                source.Write("TransmissionStatus", "Unable to parse transmission status response, args is null");
+                return;
+            }
+
+            // Always log if the response is non-success or if the feature flag is enabled
+            if (args.Response?.StatusCode != 200 || source.IsEnabled(debugTracingEventName))
+            {
+                // Do not block the main thread
+                Task.Run(() =>
                 {
-                    // Always log if the response is non-success
-                    if (args.Response.StatusCode != 200)
-                    {
-                        var transmission = sender as Transmission;
-                        string id = null;
-                        if (transmission != null)
-                        {                        
-                            id = transmission.Id;
-                        }
-
-                        var log = new
-                        {
-                            statusCode = args.Response?.StatusCode,
-                            description = args.Response?.StatusDescription,
-                            id
-                        };
-                        source.Write("TransmissionStatus", JsonConvert.SerializeObject(log, options));
-                    }
-
-                    // Log everything if the debug trace feature flag is enabled
-                    if (source.IsEnabled(debugTracingEventName))
-                    {
-                        source.Write(debugTracingEventName, FormattedLog(sender, args));
-                    }
-                }
-            });
+                    source.Write("TransmissionStatus", FormattedLog(sender, args));
+                });
+            }
         }
         internal string FormattedLog(object sender, TransmissionStatusEventArgs args)
         {
             var transmission = sender as Transmission;
-            if (transmission != null)
+            var items = transmission?.TelemetryItems.GroupBy(n => n.GetEnvelopeName())
+                        .Select(n => new
+                        {
+                            type = n.Key,
+                            count = n.Count()
+                        });
+
+            BackendResponse backendResponse = null;
+            if (!string.IsNullOrWhiteSpace(args?.Response?.Content))
             {
-                var items = transmission.TelemetryItems.GroupBy(n => n.GetEnvelopeName())
-                            .Select(n => new
-                            {
-                                type = n.Key,
-                                count = n.Count()
-                            });
-
-                BackendResponse backendResponse = null;
-                if (!string.IsNullOrWhiteSpace(args.Response?.Content))
-                {
-                    backendResponse = JsonConvert.DeserializeObject<BackendResponse>(args.Response.Content, options);
-                }
-                string topErrorMessage = null;
-                int? topStatusCode = null;
-                if (backendResponse?.Errors != null && backendResponse.Errors.Count() > 0)
-                {
-                    topErrorMessage = backendResponse.Errors[0].Message;
-                    topStatusCode = backendResponse?.Errors[0].StatusCode;
-                }
-
-                var log = new
-                {
-                    items = items,
-                    statusCode = args.Response?.StatusCode,
-                    statusDescription = args.Response?.StatusDescription,
-                    retryAfterHeader = args.Response?.RetryAfterHeader,
-                    id = transmission.Id,
-                    timeout = transmission.Timeout,
-                    responseTimeInMs = args.ResponseDurationInMs,
-                    received = backendResponse?.ItemsReceived,
-                    accepted = backendResponse?.ItemsAccepted,
-                    errorMessage = topErrorMessage,
-                    errorCode = topStatusCode,
-                };
-                return JsonConvert.SerializeObject(log, options);
+                backendResponse = JsonConvert.DeserializeObject<BackendResponse>(args.Response.Content, options);
             }
-            return "Unable to parse transmission status response";
+
+            string topErrorMessage = null;
+            int? topStatusCode = null;
+            if (backendResponse?.Errors != null && backendResponse.Errors.Length > 0)
+            {
+                topErrorMessage = backendResponse.Errors[0].Message;
+                topStatusCode = backendResponse?.Errors[0].StatusCode;
+            }
+
+            var log = new
+            {                
+                statusCode = args?.Response?.StatusCode,
+                statusDescription = args?.Response?.StatusDescription,
+                id = transmission?.Id,
+                items = items,
+                received = backendResponse?.ItemsReceived,
+                accepted = backendResponse?.ItemsAccepted,
+                errorMessage = topErrorMessage,
+                errorCode = topStatusCode,
+                responseTimeInMs = args?.ResponseDurationInMs,
+                retryAfterHeader = args?.Response?.RetryAfterHeader,
+                timeout = transmission?.Timeout,
+            };
+            return JsonConvert.SerializeObject(log, options);
         }
     }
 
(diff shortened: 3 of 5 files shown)


src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TransmissionStatusTelemetryModule.cs
```diff
@@ -0,0 +1,149 @@
...
+
+        /// <summary>
+        /// Disposes the object.
+        /// </summary>
+        public void Dispose()
+        {
+            source.Dispose();
+        }
+
+        public void Handler(object sender, TransmissionStatusEventArgs args)
```
kshyju: Should this be `public` ?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TransmissionStatusTelemetryModule.cs
```diff
@@ -0,0 +1,149 @@
...
+        {
+            // Do not block the main thread
+            Task.Run(() =>
+            {
+                if (sender != null && args != null)
+                {
+                    // Always log if the response is non-success
+                    if (args.Response.StatusCode != 200)
```
kshyju: Should we do `> 299`  so that all success responses are covered?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TransmissionStatusTelemetryModule.cs
```diff
@@ -0,0 +1,149 @@
...
+    {
+        [JsonProperty("itemsReceived")]
+        public int ItemsReceived { get; set; }
+
+        [JsonProperty("itemsAccepted")]
+        public int ItemsAccepted { get; set; }
+
+        [JsonProperty("errors")]
+        public Error[] Errors { get; set; }
+        public class Error
```
kshyju: Can this be `internal` ?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TransmissionStatusTelemetryModule.cs
```diff
@@ -0,0 +1,149 @@
...
+                            });
+
+                BackendResponse backendResponse = null;
+                if (!string.IsNullOrWhiteSpace(args.Response?.Content))
+                {
+                    backendResponse = JsonConvert.DeserializeObject<BackendResponse>(args.Response.Content, options);
+                }
+                string topErrorMessage = null;
+                int? topStatusCode = null;
+                if (backendResponse?.Errors != null && backendResponse.Errors.Count() > 0)
```
kshyju: nit: `Errors.Any()`

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TransmissionStatusTelemetryModule.cs
```diff
@@ -0,0 +1,149 @@
...
+                    timeout = transmission.Timeout,
+                    responseTimeInMs = args.ResponseDurationInMs,
+                    received = backendResponse?.ItemsReceived,
+                    accepted = backendResponse?.ItemsAccepted,
+                    errorMessage = topErrorMessage,
+                    errorCode = topStatusCode,
+                };
+                return JsonConvert.SerializeObject(log, options);
+            }
+            return "Unable to parse transmission status response";
```
kshyju: nit: line break before return

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TransmissionStatusTelemetryModule.cs
```diff
@@ -0,0 +1,149 @@
...
+        {
+            source.Dispose();
+        }
+
+        public void Handler(object sender, TransmissionStatusEventArgs args)
+        {
+            // Do not block the main thread
+            Task.Run(() =>
+            {
+                if (sender != null && args != null)
```
kshyju: nit: invert if to return early. This will reduce nesting of ifs and improve readability.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TransmissionStatusTelemetryModule.cs
```diff
@@ -0,0 +1,149 @@
...
+        JsonSerializerSettings options = new JsonSerializerSettings() { Error = (sender, error) => error.ErrorContext.Handled = true, NullValueHandling = NullValueHandling.Ignore };
+
+
+        /// <summary>
+        /// Initializes the telemetry module.
+        /// </summary>
+        /// <param name="configuration">Telemetry configuration to use for initialization.</param>
+        public void Initialize(TelemetryConfiguration configuration)
+        {
+            // Prevent the telemetry module from being initialized multiple times.
```
kshyju, line 33: Question out of curiosity: Would the Initialize gets called more than once? On what scenario?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TransmissionStatusTelemetryModule.cs
kshyju: Also, should we move this sanity check and return before `Task.Run` ?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
```diff
@@ -80,6 +80,7 @@ public static IServiceCollection AddApplicationInsights(this IServiceCollection
             services.AddSingleton<ITelemetryInitializer, MetricSdkVersionTelemetryInitializer>();
             services.AddSingleton<QuickPulseInitializationScheduler>();
             services.AddSingleton<QuickPulseTelemetryModule>();
+            services.AddSingleton<ITelemetryModule, TransmissionStatusTelemetryModule>();
```
kshyju, line 83: This dependency is used in the cold start code path. Is that correct?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TransmissionStatusTelemetryModule.cs
```diff
@@ -0,0 +1,149 @@
...
+                    if (source.IsEnabled(debugTracingEventName))
+                    {
+                        source.Write(debugTracingEventName, FormattedLog(sender, args));
+                    }
+                }
+            });
+        }
+        internal string FormattedLog(object sender, TransmissionStatusEventArgs args)
+        {
+            var transmission = sender as Transmission;
```
kshyju, line 70: nit: Pattern matching and return early to reduce one level of nesting.

````
if (!(sender is Transmission transmission))
{
    return "Unable to parse transmission status response";
}
````

File: src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TransmissionStatusTelemetryModule.cs (hunk 2 above)
Comment (RohitRanjanMS) [reply src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TransmissionStatusTelemetryModule.cs]: We should log everything except 200. This will help in investigating missing logs scenario. AI ingestion endpoint returns 206 for partial content.

File: src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TransmissionStatusTelemetryModule.cs (hunk 4 above)
Comment (RohitRanjanMS) [reply src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TransmissionStatusTelemetryModule.cs]: Length would be more appropriate.

File: src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs (hunk 8 above)
Comment (RohitRanjanMS) [reply src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs:83]: yes

File: src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TransmissionStatusTelemetryModule.cs (hunk 9 above)
Comment (RohitRanjanMS) [reply src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TransmissionStatusTelemetryModule.cs:70]: optimized code.

File: src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TransmissionStatusTelemetryModule.cs (hunk 7 above)
Comment (RohitRanjanMS) [reply src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TransmissionStatusTelemetryModule.cs:33]: It shouldn't in the ideal scenario.

File: src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TransmissionStatusTelemetryModule.cs [hunk 10]
```diff
@@ -0,0 +1,130 @@
...
+        /// </summary>
+        /// <param name="configuration">Telemetry configuration to use for initialization.</param>
+        public void Initialize(TelemetryConfiguration configuration)
+        {
+            // Prevent the telemetry module from being initialized multiple times.
+            if (isInitialized)
+            {
+                return;
+            }
+            (configuration.TelemetryChannel as ServerTelemetryChannel).TransmissionStatusEvent += Handler;
```
brettsam, line 38: It's possible that this isn't a ServerTelemetryChannel (in tests, etc).

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TransmissionStatusTelemetryModule.cs
```diff
@@ -0,0 +1,130 @@
...
+            {
+                source.Write("TransmissionStatus", "Unable to parse transmission status response, args is null");
+                return;
+            }
+
+            // Always log if the response is non-success or if the feature flag is enabled
+            if (args.Response?.StatusCode != 200 || source.IsEnabled(debugTracingEventName))
+            {
+                // Do not block the main thread
+                Task.Run(() =>
```
brettsam, line 62: I'm not sure doing this in a Task is necessary. The .NET framework and ASP.NET write directly to the source all over the place. Plus it should be rare.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TransmissionStatusTelemetryModule.cs
```diff
@@ -0,0 +1,130 @@
...
+    /// Transmission Status Telemetry Module 
+    /// </summary>
+    internal class TransmissionStatusTelemetryModule : ITelemetryModule, IDisposable
+    {
+        private bool isInitialized = false;
+        DiagnosticListener source = new DiagnosticListener(LoggingConstants.HostDiagnosticSourcePrefix + "ApplicationInsights");
+        string debugTracingEventName = LoggingConstants.HostDiagnosticSourceDebugEventNamePrefix + "TransmissionStatus";
+        JsonSerializerSettings options = new JsonSerializerSettings() { Error = (sender, error) => error.ErrorContext.Handled = true, NullValueHandling = NullValueHandling.Ignore };
+
+
```
brettsam, line 26: nit: be sure to fix spacing in this document.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Constants/LoggingConstants.cs
```diff
@@ -8,5 +8,7 @@ internal static class LoggingConstants
         public const string ZeroIpAddress = "0.0.0.0";
         public const string Unknown = "[Unknown]";
         public const string ClientIpKey = "ClientIp";
+        public const string HostDiagnosticSourcePrefix = "Microsoft.Azure.Functions.Host.";
```
brettsam, line 11: Since this is taking advantage of a host feature... it seems to me like this code should be in the host. It'd be a lot simpler to test e2e as well as you wouldn't need to make an update, then consume the package, etc.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TransmissionStatusTelemetryModule.cs
RohitRanjanMS: Added a check

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Constants/LoggingConstants.cs
RohitRanjanMS: I will close this PR and move the logic to Host

RohitRanjanMS:
Moving logic to Host. PR->https://github.com/Azure/azure-functions-host/pull/8828

PR #2923, opened 2022-10-04
Create reference object for SDK-type binding support

liliankasem:
Related issue: https://github.com/Azure/azure-functions-dotnet-worker/issues/1035

Implementation of a reference object that we can use to hydrate SDK-type bindings in out-of-proc function workers. Usage of this new type can be seen in these PRs:

- Proto: https://github.com/Azure/azure-functions-language-worker-protobuf/pull/83
- Host: https://github.com/Azure/azure-functions-host/pull/8815
- Blob Extension: https://github.com/Azure/azure-sdk-for-net/pull/31757
- Dotnet Worker: https://github.com/Azure/azure-functions-dotnet-worker/pull/1107
- Dotnet Worker Blob Extension: https://github.com/Azure/azure-functions-dotnet-worker/pull/1108


src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// <summary> Initializes a new instance of the <see cref="ParameterBindingData"/> class. </summary>
+        /// <param name="source"> Identifies the extension this event is coming from </param>
+        /// <param name="jsonSerializableData"> An object containing any required information to hydrate an SDK-type object in the out-of-process worker </param>
+        /// <param name="dataSerializationType"> The type to use when serializing the data.
+        /// If not specified, <see cref="object.GetType()"/> will be used on <paramref name="jsonSerializableData"/>.</param>
+        public ParameterBindingData(string source, object jsonSerializableData, Type dataSerializationType = default)
+        {
+            if (jsonSerializableData is BinaryData)
+            {
+                throw new InvalidOperationException("This constructor does not support BinaryData. Use the constructor that takes a BinaryData instance.");
```
jviau: nit: `ArgumentException` seems more fitting.

src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
```diff
@@ -0,0 +1,67 @@
...
+using System;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// A reference type for supporting SDK-type bindings in out-of-process Function workers
+    /// This type has built-in serialization using System.Text.Json.
+    /// </summary>
+    [JsonConverter(typeof(ParameterBindingData))]
```
jviau: I might be missing something, but I don't think this is how `JsonConverterAttribute` is used. Additionally, how is this type expected to be serialized? I think it can only be written to JSON, not read from it as is. It will need some work to know which ctor to use for deserialization.

src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        /// <summary>
+        /// The schema version of the ParameterBindingData
+        /// </summary>
+        public string Version { get; set; }
+
+        /// <summary>
+        /// Gets or sets the extension source of the event i.e CosmosDB, BlobStorage
+        /// </summary>
+        public string Source { get; set; }
```
jviau: Given there are ctors with this argument, are the `set` necessary for `Source`, `ContentType`, and `Content`? I would extend this to `Version`. I like to start with immutability, and only introduce mutability if the scenario requires it.

jviau (commented):
Not sure if there is a json serialization issue? I haven't used STJ much.

src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
```diff
@@ -0,0 +1,67 @@
...
+        public ParameterBindingData(string source, object jsonSerializableData, Type dataSerializationType = default)
+        {
+            if (jsonSerializableData is BinaryData)
+            {
+                throw new InvalidOperationException("This constructor does not support BinaryData. Use the constructor that takes a BinaryData instance.");
+            }
+
+            Version = "1.0";
+            Source = source;
+            ContentType = "application/json";
```
fabiocav: It would be good to make it very clear in the constructor docs that `application/json` will be used when this constructor is used.

fabiocav (commented):
Very minor comments.

src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
liliankasem: I honestly just copied the CloudEvent format and changed it to work with the parameter binding data schema we are designing: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Messaging/CloudEvent.cs

They also include a [deserialization](https://github.com/Azure/azure-sdk-for-net/blob/6ea1458cc053a2061e8846fedb6d03c0df61212f/sdk/core/Azure.Core/src/Messaging/CloudEvent.cs#L198) method  and we could do something similar?

kshyju: Also the value we are passing in the constructor is incorrect. A valid value should be the type of an actual converter implementation (A class which derives from [JsonConverter ](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.serialization.jsonconverter-1?view=net-6.0)and override the `Read` and `Write` methods as needed). 

I think this attribute decoration can be removed. If we want to **_customize_** the serialization/desreialization, we will have to write a converter implementation and use that.

src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
kshyju: since it is a public type, consider adding the exceptions this method/constructor throws in the method summary(IDEs usually shows that in the tooltip). [Example ](https://github.com/Azure/azure-functions-dotnet-worker/blob/main/extensions/Worker.Extensions.Abstractions/src/BindingPropertyNameAttribute.cs#L18)

src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// </summary>
+    [JsonConverter(typeof(ParameterBindingData))]
+    public class ParameterBindingData
+    {
+        /// <summary> Initializes a new instance of the <see cref="ParameterBindingData"/> class. </summary>
+        /// <param name="source"> Identifies the extension this event is coming from </param>
+        /// <param name="jsonSerializableData"> An object containing any required information to hydrate an SDK-type object in the out-of-process worker </param>
+        /// <param name="dataSerializationType"> The type to use when serializing the data.
+        /// If not specified, <see cref="object.GetType()"/> will be used on <paramref name="jsonSerializableData"/>.</param>
+        public ParameterBindingData(string source, object jsonSerializableData, Type dataSerializationType = default)
```
kshyju: Should it throw if `source` is null ?

src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
```diff
@@ -0,0 +1,67 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// A reference type for supporting SDK-type bindings in out-of-process Function workers
+    /// This type has built-in serialization using System.Text.Json.
```
kshyju: nit: Not sure we should keep the serializer information here.

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/TriggerAdapterBindingProvider.cs
```diff
@@ -36,6 +35,7 @@ IConverterManager converterManager
         // Listed in precedence for providing via DefaultType.
         // Precedence is more important than how we produce the default type (a direct conversion vs. a converter)
         private static readonly Type[] _defaultTypes = new Type[] {
```
kshyju, line 37: I see `_defaultBatchTypes` has similar information. Should that be also updated?

src/Microsoft.Azure.WebJobs/WebJobs.csproj
```diff
@@ -25,6 +25,7 @@
     </PackageReference>
     <PackageReference Include="System.ComponentModel.Annotations" Version="4.4.0" />
     <PackageReference Include="System.Diagnostics.TraceSource" Version="4.3.0" />
+    <PackageReference Include="System.Memory.Data" Version="1.0.1" />
```
kshyju, line 28: Does this PR changes uses any type from this package?  I couldn't spot any. Remove if not used.

kshyju (approved):
Do we also need a version bump to publish a new package with this change?

src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
jviau: Yeah if you want to support serialization and deserialization, this type looks simple enough to do it without needing a custom converter. I think what you will need is:

1. Add `[JsonConstructor]` to this constructor: `public ParameterBindingData(string source, BinaryData content, string dataContentType)`
2. Update the above constructor's parameters to match the properties they map to: `public ParameterBindingData(string source, BinaryData content, string contentType)`

src/Microsoft.Azure.WebJobs/WebJobs.csproj
liliankasem: Yup, BinaryData. WebJobs is on the version of .net standard that does not have BinaryData built-in

liliankasem:
> Do we also need a version bump to publish a new package with this change?

Yup, will need to version bump

src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/TriggerAdapterBindingProvider.cs
liliankasem: I don't think we're supporting batch types for the new ParameterBindingData yet, and we should update `_defaultBatchTypes` when we work on that feature

surgupta-msft (approved):

src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
```diff
@@ -0,0 +1,87 @@
...
+            ContentType = "application/json";
+            Content = new BinaryData(jsonSerializableData, type: dataSerializationType ?? jsonSerializableData?.GetType());
+        }
+
+        /// <summary> Initializes a new instance of the <see cref="ParameterBindingData"/> class using binary event data.</summary>
+        /// <param name="source"> Identifies the extension this event is coming from </param>
+        /// <param name="content"> Binary data containing any required information to hydrate an SDK-type object in the out-of-process worker </param>
+        /// <param name="contentType"> Content type of the payload. A content type different from "application/json" should be specified if payload is not JSON. </param>
+        /// <exception cref="ArgumentNullException">Throws if source, content or contentType is null.</exception>
+        public ParameterBindingData(string source, BinaryData content, string contentType)
```
jviau: Given there are 2 public constructors, I think you may need `[JsonConstructor]` on this one to support deserialization.

build/common.props
```diff
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <!-- Packages can have independent versions and only increment when released -->
-    <Version>3.0.34$(VersionSuffix)</Version>
+    <Version>3.0.35$(VersionSuffix)</Version>
```
jviau, line 4: Should this be a minor version rev as it is a new feature being added?

liliankasem: I remember someone saying it was patch but I could be wrong and it should be minor 🤔

@mathewc and @fabiocav for validation

src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
liliankasem: Woops, forgot to add this - thanks for catching

build/common.props
liliankasem: Confirmed with Fabio that minor bump works

fabiocav (approved):

src/Analyzer/WebJobs.VSAnalyzer.Core/packages.config
```diff
@@ -69,7 +69,7 @@
   <package id="System.Reflection.Primitives" version="4.3.0" targetFramework="net461" />
   <package id="System.Resources.ResourceManager" version="4.3.0" targetFramework="net461" />
   <package id="System.Runtime" version="4.3.0" targetFramework="net461" />
-  <package id="System.Runtime.CompilerServices.Unsafe" version="4.4.0" targetFramework="net461" />
+  <package id="System.Runtime.CompilerServices.Unsafe" version="5.0.0" targetFramework="net461" />
```
fabiocav: Discussing offline. But the major bump here is concerning. We need to evaluate the serialization needs for the new type to make a decision on how to proceed.

liliankasem: Deserialization is not a use case we need to support, reverting those changes

src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
```diff
@@ -0,0 +1,87 @@
...
+            {
+                throw new ArgumentNullException(nameof(contentType));
+            }
+
+            if (content is null)
+            {
+                throw new ArgumentNullException(nameof(content));
+            }
+
+            Version = "1.0";
```
mathewc: Looks like this is hardcoded to 1, and the actual producer/consumer of this type can't control the version? Is this the version of this type, or the version of the contents? How is this version intended to be used?

jviau (approved):

src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
```diff
@@ -0,0 +1,87 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// A reference type for supporting SDK-type bindings in out-of-process Function workers
```
mathewc: nit: Id prefer if this doc wasn't functions/scenario specific, since the SDK is general and not for Functions only.

How about something like "Represents data a parameter binding provides that can be used by a function to perform the bind operation to the target itself."

src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
```diff
@@ -0,0 +1,87 @@
...
+    /// </summary>
+    public class ParameterBindingData
+    {
+        /// <summary> Initializes a new instance of the <see cref="ParameterBindingData"/> class with the Content-Type set to application/json</summary>
```
mathewc: nit: looks like you're adding an unnecessary space at the beginning of each of these lines here and below

src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
```diff
@@ -0,0 +1,87 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// A reference type for supporting SDK-type bindings in out-of-process Function workers
+    /// </summary>
+    public class ParameterBindingData
```
mathewc: Thinking we should mark this type "sealed" because we're relying on strict type equality and don't really support subclasses, right? Or do we want to allow for extensions to provide their own subclasses, e.g. ServiceBusParameterBindingData?

src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
```diff
@@ -0,0 +1,87 @@
...
+    {
+        /// <summary> Initializes a new instance of the <see cref="ParameterBindingData"/> class with the Content-Type set to application/json</summary>
+        /// <param name="source"> Identifies the extension this event is coming from </param>
```
mathewc: nit: "Identifies the extension providing the binding data."

src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// <param name="jsonSerializableData"> An object containing any required information to hydrate an SDK-type object in the out-of-process worker </param>
+        /// <param name="dataSerializationType"> The type to use when serializing the data.
+        /// If not specified, <see cref="object.GetType()"/> will be used on <paramref name="jsonSerializableData"/>.</param>
+        /// <exception cref="ArgumentNullException">Throws when source is null.</exception>
+        /// <exception cref="ArgumentException">Throws when jsonSerializableData is of type BinaryData.</exception>
+        public ParameterBindingData(string source, object jsonSerializableData, Type dataSerializationType = default)
+        {
+            if (jsonSerializableData is BinaryData)
```
mathewc: Wondering if we need this ctor at all, since in the end it's just creating a BinaryData. Why not just have callers create that themselves? Otherwise, we're in the business of doing validation like this, and also responsible for surfacing BinaryData ctor parameters. For example, your ctor doesn't allow them to specify JsonSerializerOptions.

src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
```diff
@@ -0,0 +1,87 @@
...
+                throw new ArgumentNullException(nameof(source));
+            }
+
+            Version = "1.0";
+            Source = source;
+            ContentType = "application/json";
+            Content = new BinaryData(jsonSerializableData, type: dataSerializationType ?? jsonSerializableData?.GetType());
+        }
+
+        /// <summary> Initializes a new instance of the <see cref="ParameterBindingData"/> class using binary event data.</summary>
```
mathewc: nit: it's not really "event data". Perhaps just simplify this to "Initializes a new instance of the ParameterBindingData class using the specified BindingData and content type."

src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// <param name="source"> Identifies the extension this event is coming from </param>
+        /// <param name="content"> Binary data containing any required information to hydrate an SDK-type object in the out-of-process worker </param>
```
mathewc: nit: "BinaryData containing the binding data."

src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// </summary>
+        public string Source { get; }
+
+        /// <summary>
+        /// Gets the content type of the content data
+        /// </summary>
+        public string ContentType { get; }
+
+        /// <summary>
+        /// Gets the event content as <see cref="BinaryData"/>. Using BinaryData, one can deserialize
```
mathewc: nit: "The binding data content."

test/Microsoft.Azure.WebJobs.Host.UnitTests/JobHostMetadataProviderTests.cs
```diff
@@ -174,19 +174,19 @@ public void DefaultTypeForOpenTypeTrigger()
...
+                        .Build();
             IJobHostMetadataProvider metadataProvider = host.CreateMetadataProvider();
 
             var attr = new Test9Attribute(null);
             var type = metadataProvider.GetDefaultType(attr, FileAccess.Write, null);
 
-            // The trigger handles Open type, which means it will first pull byte[]. 
-            Assert.Equal(typeof(byte[]), type);
+            // The trigger handles Open type, which means it will first pull ParameterBindingData.
+            Assert.Equal(typeof(ParameterBindingData), type);
```
mathewc, line 189: Wondering if this indicates a potential breaking change? Does ParameterBindingData need to come first in the precedence list? Wondering what potential impact this change will have.

liliankasem: Yeah I am worried about this too, and maybe this is something you can help me understand better.

Without ParameterBindingData being at the top of the list, we never end up with the ParameterBindingData converter which ends up breaking the entire SDK-binding scenario. I did test out the other extensions that don't support ParameterBindingData and things were working as expected, so I haven't noticed anything breaking yet - but that doesn't mean that nothing will break 🥲

Do you happen to know why this was the original order of things? And bytes has to be first, do you have suggestions on how we can get the correct converter in the ParameterBindingData scenario?

src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
liliankasem: The version is the schema version, so this type. Ideally consumers are checking for this for backwards compatibility. So whenever we make breaking changes to this type, we can bump the version. But I'm open to better suggestions for versioning!

src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
liliankasem: I don't think there is any intent for subclasses, I can't think of a scenario right now where extension would need to build their own since the content is pretty flexible

src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
liliankasem: Yup, that's a fair point. I guess if they wanted to do it themselves with JsonSerializerOptions they could always use the other ctor. This was just to make it easier but I'm not attached to it so can remove

src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
liliankasem: Had a sync with Fabio about this, and agree that it should be the version of the content and not the schema. Updating this so that version is provided by the provider

build/common.props
liliankasem: Reverting to patch as suggested by Mathew - we're not changing SDK behaviors and just providing new capabilities for extensions to leverage so patch fits the case fine.

src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
```diff
@@ -0,0 +1,47 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs
+{
+    /// <summary>
+    /// Represents data a parameter binding provides that can be used by a function to perform the bind operation to the target itself.
+    /// </summary>
+    sealed public class ParameterBindingData
```
mathewc: This should be ordered "public sealed". It might be generating a warning?

mathewc (approved):

src/Microsoft.Azure.WebJobs/ParameterBindingData.cs
liliankasem: done!

PR #2931, opened 2022-10-25
Update parameter binding default type

liliankasem:
We should only be trying to bind to ParameterBindingData if it is the requestedType, otherwise fallback to default behaviour

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/TriggerAdapterBindingProvider.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/TriggerAdapterBindingProvider.cs
@@ -69,7 +69,7 @@ namespace Microsoft.Azure.WebJobs.Host.Bindings
 
                 // We should only be requesting ParameterBindingData if it is supported by the
                 // extension we're binding against. In theory, this should never happen.
-                throw new ArgumentException($"Converter for {requestedType} not found.");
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.WebJobs.Host/Bindings/BindingProviders/TriggerAdapterBindingProvider.cs
```diff
@@ -45,20 +44,34 @@ internal class TriggerAdapterBindingProvider<TAttribute, TTriggerValue> :
...
+            if (requestedType == typeof(ParameterBindingData) || requestedType == typeof(ParameterBindingData[]))
+            {
+                if (_converterManager.HasConverter<TAttribute>(typeof(TTriggerValue), requestedType))
+                {
+                    return requestedType;
+                }
+
+                // We should only be requesting ParameterBindingData if it is supported by the
+                // extension we're binding against. In theory, this should never happen.
+                throw new ArgumentException($"Converter for {requestedType} not found.");
```
liliankasem: I don't think this is the right exception to use, so open to suggestions here

fabiocav: For this, an `InvalidOperationException` perhaps?

fabiocav (approved):

mathewc (approved):

PR #2932, opened 2022-11-05
Adding FunctionId to ScaleMonitorDescriptor

alrod:


src/Microsoft.Azure.WebJobs.Host/Scale/ScaleMonitorDescriptor.cs
```diff
@@ -13,6 +13,11 @@ public ScaleMonitorDescriptor(string id)
             Id = id;
         }
 
+        public ScaleMonitorDescriptor(string id, string functionId) : this(id)
```
mathewc, line 19: Since our goal going forward is for people to use this new constructor, I recommend marking the other one [Obsolete] to start people migrating

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleMonitorDescriptor.cs
```diff
@@ -22,5 +27,10 @@ public ScaleMonitorDescriptor(string id)
         /// be of the form "{FunctionId}-QueueTrigger-{QueueName}".
         /// </remarks>
         public string Id { get; }
+
+        /// <summary>
+        /// Gets the ID of the function associated with this scaler.
```
mathewc: "... associated with this monitor"

mathewc (approved):

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleMonitorDescriptor.cs
JoshLove-msft: FYI this is potentially a breaking change for consumers of this library. If they have WarnAsError configured as part of their build, it will break them.

PR #2937, opened 2022-12-10
Exposing common scaling code to consume by SC v3

### LINKED ISSUE #2938: Exposing common scaling code to consume by SC v3
For SC v3(redesign) we can reuse RDS code from function host such as:
1. Decision making.
`ScaleManager`
2. Metric saving. 
`ScaleMonitorService` to periodically save metrics. 
`InMemoryScaleMetricsRepository` - SC saves metrics in memory. 
3. Dynamic concurrency value reading.
`BlobStorageConcurrencyStatusRepository` - making public.

The plan is that SC v3 and function host will be referencing webjob sdk use the classes/interfaces in the PR.

alrod:
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/2938

Function host change:
https://github.com/Azure/azure-functions-host/pull/8977

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Scale/IScaleMetricsRepository.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Scale/IScaleMetricsRepository.cs
@@ -1,0 +1,28 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Host.Scale
+{
+    /// <summary>
+    /// Interface defining methods for reading/writing scale metrics to a persistent store.
+    /// </summary>
+    public interface IScaleMetricsRepository
+    {
+        /// <summary>
+        /// Persist the metrics for each monitor.
+        /// </summary>
+        /// <param name="monitorMetrics">The collection of metrics for each monitor.</param>
+        /// <returns>A task.</returns>
+        Task WriteMetricsAsync(IDictionary<IScaleMonitor, ScaleMetrics> monitorMetrics);
+
+        /// <summary>
+        /// Read the metrics.
+        /// </summary>
+        /// <param name="monitors">The current collection of monitors.</param>
+        /// <returns>Map of metrics per monitor.</returns>
+        Task<IDictionary<IScaleMonitor, IList<ScaleMetrics>>> ReadMetricsAsync(IEnumerable<IScaleMonitor> monitors);
+    }
+}
--- a/src/Microsoft.Azure.WebJobs.Host/Scale/ScaleManager.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Scale/ScaleManager.cs
@@ -1,0 +1,306 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Host.Scale
+{
+    /// <summary>
+    /// Class to get scaling votes for all the triggers or for an individual trigger.
+    /// </summary>
+    internal class ScaleManager : IScaleManager
+    {
+        private readonly IScaleMonitorManager _monitorManager;
+        private readonly ITargetScalerManager _targetScalerManager;
+        private readonly IScaleMetricsRepository _metricsRepository;
+        private readonly IConcurrencyStatusRepository _concurrencyStatusRepository;
+        private readonly ILogger _logger;
+        private readonly HashSet<string> _targetScalersInError;
+        private IOptions<ScaleOptions> _scaleOptions;
+
+        public ScaleManager(
+            IScaleMonitorManager monitorManager,
+            ITargetScalerManager targetScalerManager,
+            IScaleMetricsRepository metricsRepository,
+            IConcurrencyStatusRepository concurrencyStatusRepository,
+            IOptions<ScaleOptions> scaleConfiguration,
+            ILoggerFactory loggerFactory)
+        {
+            _monitorManager = monitorManager;
+            _targetScalerManager = targetScalerManager;
+            _metricsRepository = metricsRepository;
+            _concurrencyStatusRepository = concurrencyStatusRepository;
+            _logger = loggerFactory?.CreateLogger<ScaleManager>();
+            _targetScalersInError = new HashSet<string>();
+            _scaleOptions = scaleConfiguration;
+        }
+
+        // for mock testing only
+        internal ScaleManager()
+        {
+        }
+
+
+        /// <summary>
+        /// Gets overall scale status for all the triggers.
+        /// </summary>
+        /// <param name="context">The scale status context</param>
+        /// <returns>A task that returns <see cref="ScaleStatus"/> for all ther triggers.</returns>
+        public async Task<ScaleStatus> GetScaleStatusAsync(ScaleStatusContext context)
+        {
+            GetScalersToSample(out List<IScaleMonitor> scaleMonitorsToProcess, out List<ITargetScaler> targetScalersToProcess);
+
+            var scaleMonitorStatuses = await GetScaleMonitorsResultAsync(context, scaleMonitorsToProcess);
+            var targetScalerStatuses = await GetTargetScalersResultAsync(context, targetScalersToProcess);
+
+            return new ScaleStatus
+            {
+                Vote = GetAggregateScaleVote(scaleMonitorStatuses.Values.Select(x => x.Vote).Union(targetScalerStatuses.Select(x => x.Value.Vote)), context, _logger),
+                TargetWorkerCount = targetScalerStatuses.Any() ? targetScalerStatuses.Max(x => x.Value.TargetWorkerCount) : null
+            };
+        }
+
+        /// <summary>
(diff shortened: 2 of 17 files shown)


src/Microsoft.Azure.WebJobs.Host/Scale/FunctionsScaleManager.cs
```diff
@@ -0,0 +1,295 @@
...
+            _logger = loggerFactory.CreateLogger<FunctionsScaleManager>();
+            _targetScalersInError = new HashSet<string>();
+        }
+
+        // for mock testing only
+        internal FunctionsScaleManager()
+        {
+        }
+
+        public virtual bool IsTargetBaseScalingEnabled => false;
```
chiangvincent: nit: targetbasedscaling

src/Microsoft.Azure.WebJobs.Host/Scale/InMemoryScaleMetricsRepository.cs
```diff
@@ -0,0 +1,91 @@
...
+                        result[monitor] = _monitorMetrics.ContainsKey(monitor) ? _monitorMetrics[monitor] : result[monitor];
+                    }
+                }
+                return Task.FromResult(result);
+            }
+        }
+
+        public Task WriteMetricsAsync(IDictionary<IScaleMonitor, ScaleMetrics> monitorMetrics)
+        {
+            lock (_monitorMetrics)
```
chiangvincent: Only writes to the same IScaleMonitor should be locked right? (Instead of locking the entire metrics cache) We should allow concurrent writes to different keys in the dictionary since they are separate triggers -- perhaps a `ConcurrentDictionary` is a better data structure for this use case.

src/Microsoft.Azure.WebJobs.Host/Scale/InMemoryScaleMetricsRepository.cs
```diff
@@ -0,0 +1,91 @@
...
+            {
+                foreach (IScaleMonitor scaleMonitor in monitorMetrics.Keys)
+                {
+                    if (_monitorMetrics.TryGetValue(scaleMonitor, out IList<ScaleMetrics> existingMetrics))
+                    {
+                        existingMetrics.Add(monitorMetrics[scaleMonitor]);
```
chiangvincent: We should also enforce a max number of metrics per key (10 samples in current sc, for example) to ensure we don't overflow memory

src/Microsoft.Azure.WebJobs.Host/Scale/IScalerFactory.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Azure.WebJobs.Host.Scale
+{
+    public interface IScalerFactory
```
mathewc: Not sure this interface needs to be in the SDK. I don't see it used from either the SDK or the Host PR you linked to, so likely it's a SC specific thing and can just live there.

src/Microsoft.Azure.WebJobs.Host/Scale/InMemoryScaleMetricsRepository.cs
```diff
@@ -0,0 +1,91 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+
+namespace Microsoft.Azure.WebJobs.Host.Scale
+{
+    public class InMemoryScaleMetricsRepository : IScaleMetricsRepository
```
mathewc: This is also a ScaleController specific implementation that would only be used by ScaleController so doesn't need to be in the SDK? I.e. it's not an implementation that needs to be shared.

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleManager.cs
```diff
@@ -0,0 +1,295 @@
...
+using Microsoft.Extensions.Logging;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Host.Scale
+{
+    public class ScaleManager
```
mathewc: For any new public types we expose, we'll need to have good xml doc on classes/methods

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleStatus.cs
```diff
@@ -12,5 +12,21 @@ public class ScaleStatus
         /// Gets or sets the current scale decision.
         /// </summary>
         public ScaleVote Vote { get; set; }
+
+        /// <summary>
+        /// Gets or sets the currebt target worker count.
```
mathewc: spelling

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleStatus.cs
```diff
@@ -12,5 +12,21 @@ public class ScaleStatus
...
+        /// </summary>
+        public int? TargetWorkerCount { get; set; }
+
+        /// <summary>
+        /// Gets the current scale decision as int.
+        /// </summary>
+        public int VoteAsInt
```
mathewc: This int value makes for not a clean Type interface. We should just use ScaleVote in the SDK, and if SC needs to do a conversion/interpretation, it should do it itself in that layer.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Scale/FunctionsScaleMonitorServiceTests.cs
```diff
@@ -0,0 +1,197 @@
...
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs.Host.Scale;
+using Microsoft.Azure.WebJobs.Host.TestCommon;
+using Microsoft.Azure.WebJobs.Host.UnitTests.Scale;
+using Microsoft.Extensions.Logging;
+
+using Moq;
+using Xunit;
+
+namespace Microsoft.Azure.WebJobs.Script.Tests.Scale
```
mathewc: any tests we move over should have their namespaces fixed for SDK (i.e. no "Script")

test/Microsoft.Azure.WebJobs.Host.UnitTests/Scale/FunctionsScaleMonitorServiceTests.cs
```diff
@@ -0,0 +1,197 @@
...
+{
+    public class FunctionsScaleMonitorServiceTests
```
mathewc: Type name was changed, so tests should reflect that. Check for all the other tests as well.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Scale/FunctionsScaleMonitorServiceTests.cs
```diff
@@ -0,0 +1,197 @@
...
+    {
+        private readonly ScaleMonitorService _monitor;
+        private readonly TestMetricsRepository _metricsRepository;
+        private readonly TestLoggerProvider _loggerProvider;
+        private List<IScaleMonitor> _monitors;
+        private List<ITargetScaler> _scalers;
+
+        public FunctionsScaleMonitorServiceTests()
+        {
+            //_isPrimaryHost = true;
```
mathewc: remove commented out code

src/Microsoft.Azure.WebJobs.Host/Scale/IScaleMetricsRepository.cs
```diff
@@ -0,0 +1,28 @@
...
+
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Host.Scale
+{
+    /// <summary>
+    /// Interface defining methods for reading/writing scale metrics to a persistent store.
+    /// </summary>
+    public interface IScaleMetricsRepository
```
mathewc, line 12: In the ScaleController we won’t be persisting metrics to storage. Is the idea that SC would use in-memory, as a replacement for current scale controller metrics capturing?

src/Microsoft.Azure.WebJobs.Host/Scale/InMemoryScaleMetricsRepository.cs
alrod: `InMemoryScaleMatricsRepository` is moving to `ScaleController.Common`

src/Microsoft.Azure.WebJobs.Host/Scale/IScalerFactory.cs
alrod: All the extensions will implement the interface to use it from ScaleController. WebJobs SDK is common part for all the extensions.

src/Microsoft.Azure.WebJobs.Host/Scale/InMemoryScaleMetricsRepository.cs
alrod: Agreed. `InMemoryScaleMetricsRepository` is moved to ScaleController repository.

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleStatus.cs
alrod: Ok, this can be moved to SC.

src/Microsoft.Azure.WebJobs.Host/Scale/IScaleMetricsRepository.cs
alrod: yes, SC will be using in-memory implementation.

alrod:
closing in favor of #2954

Issue #2847, opened 2022-03-30
Use new reliable way of flushing Application Insights logs


marnilss:
As discussed here (https://github.com/Microsoft/ApplicationInsights-dotnet/issues/407) the old synchronous `telemetryClient.Flush()` was not reliable since it wasn't really synchronous, but instead relied upon some variation of `Thread.Sleep(n)` afterwards.

This is implemented with a sleep of 2 seconds in Dispose() in ApplicationInsightsLoggerProvider.cs (https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerProvider.cs)

But when functions exceed the function timeout value, and the host shuts down, the following trace-messages are never seen in Application Insights with the current sleep-setting.

> "A function timeout has occurred. Host is shutting down."
> "Timeout value of 00:00:30 exceeded by function 'AddressImportOnlyFunction_LogVersion' (Id: 'a9ac9cd6-f299-4d72-9ecf-d74f0b032299'). Initiating cancellation."
> 

neither is the Exception:

> Microsoft.Azure.WebJobs.Host.FunctionTimeoutException

Solution: he new asynchronous FlushAsync() method should be awaited instead, and then the sleep can be removed.

#### Repro steps

1. Create new Function App project with application insights logging enabled
2. Create a new function which calls `Thread.Sleep(TimeSpan.FromMinutes(5));`
3. Set `"functionTimeout": "00:00:30"` in host.json
4. Run and observe that the log messages detailed above is never seen in Application Insights database.

#### Expected behavior

Both trace- and exceptions-messages surrounding timeout should be visible in Application Insights.

#### Actual behavior

None of the trace- and exceptions-messages surrounding timeout are visible in Application Insights.

#### Known workarounds

An ugly workaround is to add the following function to the project, but which also proves that FlushAsync solves the problem:
```
    public class LogFlusher
    {
        private static bool _isSetup = false;

        private readonly TelemetryClient _telemetryClient;
#pragma warning disable CS0618 // Type or member is obsolete - BUT Function App is still relying on it
        private readonly IApplicationLifetime _applicationLifetime;

        public LogFlusher(TelemetryClient telemetryClient, IApplicationLifetime applicationLifetime)
#pragma warning restore CS0618 // Type or member is obsolete
        {
            _telemetryClient = telemetryClient;
            _applicationLifetime = applicationLifetime;
[...]

PR #2939, opened 2022-12-19
Update Application Insights flushing method to be FlushAsync

michaelpeng36:
Addresses issue #2847 and adds a `DiagnosticListener` so consumers of the SDK can subscribe to events where flushing fails upon disposal of the `ApplicationInsightsLoggerProvider`.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerProvider.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerProvider.cs
@@ -3,7 +3,9 @@
 
 using System;
 using System.Diagnostics;
+using System.Linq;
 using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.ApplicationInsights;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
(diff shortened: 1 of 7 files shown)


src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerProvider.cs
```diff
@@ -34,15 +38,17 @@ public void Dispose()
                 {
                     try
                     {
-                        _client.Flush();
-
-                        // This sleep isn't ideal, but the flush is async so it's the best we have right now. This is
-                        // being tracked at https://github.com/Microsoft/ApplicationInsights-dotnet/issues/407
-                        Thread.Sleep(2000);
+                        // Flushing logs upon disposal should not be canceled
+                        var task = _client.FlushAsync(CancellationToken.None);
```
RohitRanjanMS: FlushAsync returns false when transfer of telemetry data to server has failed with non-retriable http status. We should log if Flush Async returns false.

RohitRanjanMS: There's no timeout and this can take longer. We should cancel flush if it takes longer (~5 seconds).

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerProvider.cs
```diff
@@ -34,20 +50,42 @@ public void Dispose()
                 {
                     try
                     {
-                        _client.Flush();
+                        // Default timeout of 5 seconds
+                        var timeout = new TimeSpan(hours: 0, minutes: 0, seconds: 5);
+                        _cancellationTokenSource = _cancellationTokenSource ?? new CancellationTokenSource();
```
NickCraver: The created `_cancellationTokenSource` should be disposed here as well :)

RohitRanjanMS (approved):

Issue #2824, opened 2022-01-18
Performance Area: Application Insights


NickCraver:
In profiling production vs. local one of the big differences in the common setup is Application Insights being on or not. For most customers the advice is to have it on, so let's assume it is. The problem is impact of it being on - in Hello World this is approximately a 40% decrease in throughput and a near doubling of latency. In EUAP the difference for a single instance running hello world with Application Insights (1,100-1,200 requests/sec) to without Application Insights (1,650-1,750 requests/sec) is a significant jump. This is confirmed locally even at higher core counts.

Where the runtime cost lies in AI is in 2 areas:
1. Our feeding data into it.
2. The serialization and sending data inside it.

#### Piping data to Application Insights: Current
The first area we spend a lot of time in compute and allocations is in getting the data into AI. This happens primarily through `ILogger.BeginScope(<dictionary>)`. The aggregate allocations in this are:
1. The dictionary going into the scope (a few are cached/re-used, but most are not - and can't be).
   - This is primarily the allocation of a dictionary itself, though we can save a few resizes (covered in #2800)
2. The receiving side of [`ApplicationInsightsLogger.BeginScope()`](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs#L445-L455) with [`DictionaryLoggerScope`](https://github.com/Azure/azure-webjobs-sdk/blob/8e26083eab043b5c0794365be6466f71e9fe7b65/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/DictionaryLoggerScope.cs)
   - This [creates a copy of the `Dictionary<string, object>` and wraps it in another `ReadOnlyDictionary`](https://github.com/Azure/azure-webjobs-sdk/blob/8e26083eab043b5c0794365be6466f71e9fe7b65/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/DictionaryLoggerScope.cs#L38-L60).
   - It also creates an `AsyncLocal<T>` storage pattern shift as we move past the optimized "up to 3 items" threshold for its storage pattern.
[...]

PR #2946, opened 2023-01-09
Performance Improvements - Logging

RohitRanjanMS:
This PR addresses the issue [#2842](https://github.com/Azure/azure-webjobs-sdk/issues/2824). 
Enabling Application Insights has significant performance and throughput overhead. Function supports adding context to log/metric telemetry using BeginScope(). Each call to BeginScope() creates another level of nesting, so that logs/metrics produced within the scope can be tagged with information about where they came from. While this is done extensively in host, customers can create scopes as well (.Net In-Proc) . Nick did an [investigation](https://github.com/Azure/azure-webjobs-sdk/issues/2824) to highlight the performance overhead of using scopes. This is primarily due to multiple Dictionaries that get allocated to support scope. One at the time of passing the scope attributes, another when these attributes are received and wrapped in ReadOnlyDictionary. There’s another set of allocation when the current scope is built where-in it recursively iterate through the nested scopes.

Improvements in this PR
1.	`ApplicationInsightsLogger.cs`
a.	Removed `BeginScope` from the `Log()` method. All the `known properties` are passed as param and added to the telemetry property. This reduces the level of nesting and also reduces the cost of building `scopeInfo`.
b.     Reduced the numer of times `GetMergedStateDictionaryOrNull` is called.
c.	Removed Linq from `ApplyScopeProperties()`. 
d.     Saved some boxing/unboxing and made type casting more efficient.

2.	`LogLevelEnumHelper .cs`
Enum helper to convert `enum` into `string`. This uses a cached string array. No allocation and very efficient compared to `ToString()`. 

3.	`WebJobsTelemetryInitializer.cs`
a.     Restricted the call to `GetMergedStateDictionaryOrNull()` for only `DependencyTelemetry`. 
b.     Category, LogLevel, EventId and EventName are set in `ApplicationInsightsLogger.Log()` method.

4.	`DictionaryLoggerScope.cs`
a.	`Push()` - If state is a dictionary, then use it otherwise build a new dictionary.
b.	`GetMergedStateDictionaryOrNull()` – Return null if `Current` is `null`. 
c.      Cached the `scopeInfo` and return from cache if exist. The cache is tied to the current scope.
e.	`_itemCount` – to avoid dictionary resizing.

**Benchmarking**:
Memory Allocation:
- Dictionary<string, object> allocation is down by ~99.75%. 
- Resize is almost negligible.
 
image

- Req/sec and avg response time is down by ~8%
Screenshot 2023-03-10 112101

image

Execution Time
image

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
@@ -138,14 +138,14 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
                     case MetricCountKey when entry.Value is int count:
                         telemetry.Count = count;
                         break;
-                    case MetricMinKey when entry.Value is double min:
-                        telemetry.Min = min;
+                    case MetricMinKey:
+                        telemetry.Min = Convert.ToDouble(entry.Value);
                         break;
-                    case MetricMaxKey when entry.Value is double max:
-                        telemetry.Max = max;
+                    case MetricMaxKey:
+                        telemetry.Max = Convert.ToDouble(entry.Value);
                         break;
-                    case MetricStandardDeviationKey when entry.Value is double dev:
-                        telemetry.StandardDeviation = dev;
+                    case MetricStandardDeviationKey:
+                        telemetry.StandardDeviation = Convert.ToDouble(entry.Value);
                         break;
                     default:
                         // Otherwise, it's a custom property.
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
@@ -16,7 +16,6 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.Azure.WebJobs.Logging.ApplicationInsights.Extensions;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
-using static Microsoft.ApplicationInsights.MetricDimensionNames;
 
 namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
 {
@@ -76,9 +75,6 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
             // Initialize stateValues so the rest of the methods don't have to worry about null values.
             stateValues = stateValues ?? new Dictionary<string, object>();
 
-            // Add some well-known
-            // properties to the scope dictionary so the TelemetryInitializer can add them
-            // for all telemetry.
             if (_isUserFunction && eventId.Id == LogConstants.MetricEventId)
             {
                 // Log a metric from user logs only
@@ -110,7 +106,9 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
         {
             MetricTelemetry telemetry = new MetricTelemetry();
             // Always apply scope first to allow state to override.
-            ApplyScopeProperties(telemetry.Properties, telemetry);
+            ApplyScopeProperties(telemetry);
+
+            // Add known properties like category, logLevel and event
             ApplyKnownProperties(telemetry.Properties, logLevel, eventId);
 
             foreach (var entry in values)
@@ -167,23 +165,23 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
         }
 
         // Applies scope properties; filters most system properties, which are used internally
-        private static void ApplyScopeProperties(IDictionary<string, string> properties, ITelemetry telemetry)
+        private static void ApplyScopeProperties(ITelemetry telemetry)
         {
             var scopeProperties = DictionaryLoggerScope.GetMergedStateDictionaryOrNull();            
             if (scopeProperties != null)
             {
                 foreach (var scopeProperty in scopeProperties)
                 {
-                    if (!SystemScopeKeys.Contains(scopeProperty.Key))
+                    if (!SystemScopeKeys.Contains(scopeProperty.Key, StringComparer.Ordinal))
                     {
-                        ApplyProperty(properties, scopeProperty.Key, scopeProperty.Value, true);
+                        ApplyProperty(telemetry.Context.Properties, scopeProperty.Key, scopeProperty.Value, true);
                     }
                 }
 
                 string invocationId = scopeProperties.GetValueOrDefault<string>(ScopeKeys.FunctionInvocationId);
                 if (invocationId != null)
                 {
-                    properties[LogConstants.InvocationIdKey] = invocationId;
+                    telemetry.Context.Properties[LogConstants.InvocationIdKey] = invocationId;
                 }
             }
             telemetry.Context.Operation.Name = scopeProperties.GetValueOrDefault<string>(ScopeKeys.FunctionName);
@@ -256,7 +254,7 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
         // Makes sure these are done in the correct order. If there are duplicate keys, the last State property wins.
         private static void ApplyScopeAndStateProperties(IDictionary<string, string> properties, IEnumerable<KeyValuePair<string, object>> state, ITelemetry telemetry)
         {
-            ApplyScopeProperties(properties, telemetry);
+            ApplyScopeProperties(telemetry);
             ApplyProperties(properties, state, true);
         }
 
@@ -276,10 +274,10 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
                 }
             }
         }
-        internal static void ApplyProperty(IDictionary<string, string> properties, string key, object value, bool applyPrefix = false)
+        internal static void ApplyProperty(IDictionary<string, string> properties, string key, object objectValue, bool applyPrefix = false)
         {
             // do not apply null values
-            if (value == null)
+            if (objectValue == null)
             {
                 return;
             }
@@ -287,21 +285,21 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
             string stringValue = null;
 
             // Format dates
-            if (value is string st)
+            if (objectValue is string value)
             {
-                stringValue = st;
+                stringValue = value;
             }
-            if (value is DateTime date)
+            if (objectValue is DateTime date)
             {
                 stringValue = date.ToUniversalTime().ToString(DateTimeFormatString);
             }
-            else if (value is DateTimeOffset dateOffset)
+            else if (objectValue is DateTimeOffset dateOffset)
             {
                 stringValue = dateOffset.UtcDateTime.ToString(DateTimeFormatString);
             }
             else
             {
-                stringValue = value.ToString();
+                stringValue = objectValue.ToString();
             }
 
             string prefixedKey = applyPrefix ? _prefixedProperyNames.GetOrAdd(key, k =>
@@ -345,13 +343,13 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
                         // We won't use the format string here
                         break;
                     default:                        
-                        if (value.Value is int i) 
+                        if (value.Value is int intValue) 
                         {
-                            metrics.Add(value.Key, Convert.ToDouble(i));
+                            metrics.Add(value.Key, Convert.ToDouble(intValue));
                         }
-                        else if (value.Value is double d)
+                        else if (value.Value is double doubleValue)
                         {
-                            metrics.Add(value.Key, d);
+                            metrics.Add(value.Key, doubleValue);
                         }
                         else if (value.Value is TimeSpan timeSpan)
                         {
@@ -363,7 +361,7 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
                 }
             }
 
-            IDictionary<string, string> properties = new Dictionary<string, string>(4);
+            IDictionary<string, string> properties = new Dictionary<string, string>(2);
             ApplyKnownProperties(properties, logLevel, eventId);
 
             foreach (KeyValuePair<string, double> metric in metrics)
@@ -455,7 +453,6 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
                             break;
                     }
                 }
-
                 
                 currentActivity.AddTag(LogConstants.CategoryNameKey, _categoryName);
                 currentActivity.AddTag(LogConstants.LogLevelKey, LogLevelEnumHelper.ToStringOptimized(logLevel));
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
@@ -122,11 +122,11 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
                 // next entry if found.
                 switch (entry.Key)
                 {
-                    case LogConstants.NameKey:
-                            telemetry.Name = entry.Value.ToString();
+                    case LogConstants.NameKey when entry.Value is string name:
+                        telemetry.Name = name;
                         continue;
-                    case LogConstants.MetricValueKey:
-                        telemetry.Sum = (double)entry.Value;
+                    case LogConstants.MetricValueKey when entry.Value is double sum:
+                        telemetry.Sum = sum;
                         continue;
                     default:
                         break;
@@ -135,17 +135,17 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
                 // Now check for case-insensitive matches
                 switch (entry.Key.ToLowerInvariant())
                 {
-                    case MetricCountKey:
-                        telemetry.Count = Convert.ToInt32(entry.Value);
+                    case MetricCountKey when entry.Value is int count:
+                        telemetry.Count = count;
                         break;
-                    case MetricMinKey:
-                        telemetry.Min = Convert.ToDouble(entry.Value);
+                    case MetricMinKey when entry.Value is double min:
+                        telemetry.Min = min;
                         break;
-                    case MetricMaxKey:
-                        telemetry.Max = Convert.ToDouble(entry.Value);
+                    case MetricMaxKey when entry.Value is double max:
+                        telemetry.Max = max;
                         break;
-                    case MetricStandardDeviationKey:
-                        telemetry.StandardDeviation = Convert.ToDouble(entry.Value);
+                    case MetricStandardDeviationKey when entry.Value is double dev:
+                        telemetry.StandardDeviation = dev;
                         break;
                     default:
                         // Otherwise, it's a custom property.
@@ -171,8 +171,7 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
                     }
                 }
 
-                string invocationId = scopeProperties.GetValueOrDefault<string>(ScopeKeys.FunctionInvocationId);
-                if (invocationId != null)
+                if (scopeProperties.GetValueOrDefault<string>(ScopeKeys.FunctionInvocationId) is string invocationId)
                 {
                     telemetry.Context.Properties[LogConstants.InvocationIdKey] = invocationId;
                 }               
@@ -325,8 +324,8 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
             {
                 switch (value.Key)
                 {
-                    case LogConstants.NameKey:
-                        functionName = value.Value.ToString();
+                    case LogConstants.NameKey when value.Value is string name:
+                        functionName = name;
                         break;
                     case LogConstants.TimestampKey:
                     case LogConstants.OriginalFormatKey:
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
@@ -9,11 +9,13 @@ using System.Globalization;
 using System.Linq;
 using System.Text;
 using Microsoft.ApplicationInsights;
+using Microsoft.ApplicationInsights.Channel;
 using Microsoft.ApplicationInsights.DataContracts;
 using Microsoft.ApplicationInsights.Extensibility;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
+using static Microsoft.ApplicationInsights.MetricDimensionNames;
 
 namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
 {
@@ -33,7 +35,7 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
         internal const string MetricMinKey = "min";
         internal const string MetricMaxKey = "max";
         internal const string MetricStandardDeviationKey = "standarddeviation";
-
+        
         private static readonly HashSet<string> SystemScopeKeys = new HashSet<string>
             {
                 LogConstants.CategoryNameKey,
@@ -73,50 +75,42 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
             // Initialize stateValues so the rest of the methods don't have to worry about null values.
             stateValues = stateValues ?? new Dictionary<string, object>();
 
-            // Add some well-known properties to the scope dictionary so the TelemetryInitializer can add them
+            // Add some well-known
+            // properties to the scope dictionary so the TelemetryInitializer can add them
             // for all telemetry.
-            using (BeginScope(new Dictionary<string, object>
+            if (_isUserFunction && eventId.Id == LogConstants.MetricEventId)
             {
-                [LogConstants.CategoryNameKey] = _categoryName,
-                [LogConstants.LogLevelKey] = (LogLevel?)logLevel,
-                [LogConstants.EventIdKey] = eventId.Id,
-                [LogConstants.EventNameKey] = eventId.Name,
-            }))
+                // Log a metric from user logs only
+                LogMetric(stateValues, logLevel, eventId);
+            }
+            else if (_categoryName == LogCategories.Results)
             {
-                if (_isUserFunction && eventId.Id == LogConstants.MetricEventId)
-                {
-                    // Log a metric from user logs only
-                    LogMetric(stateValues);
-                }
-                else if (_categoryName == LogCategories.Results)
-                {
-                    // Log a function result
-                    LogFunctionResult(stateValues, logLevel, exception);
-                }
-                else if (_categoryName == LogCategories.Aggregator)
-                {
-                    // Log an aggregate record
-                    LogFunctionResultAggregate(stateValues);
-                }
-                else if (exception != null)
-                {
-                    // Log an exception
-                    LogException(logLevel, stateValues, exception, formattedMessage);
-                }
-                else
-                {
-                    // Otherwise, log a trace
-                    LogTrace(logLevel, stateValues, formattedMessage);
-                }
+                // Log a function result
+                LogFunctionResult(stateValues, logLevel, exception, eventId);
+            }
+            else if (_categoryName == LogCategories.Aggregator)
+            {
+                // Log an aggregate record
+                LogFunctionResultAggregate(stateValues, logLevel, eventId);
+            }
+            else if (exception != null)
+            {
+                // Log an exception
+                LogException(logLevel, stateValues, exception, formattedMessage, eventId);
+            }
+            else
+            {
+                // Otherwise, log a trace
+                LogTrace(logLevel, stateValues, formattedMessage, eventId);
             }
         }
 
-        private void LogMetric(IEnumerable<KeyValuePair<string, object>> values)
+        private void LogMetric(IEnumerable<KeyValuePair<string, object>> values, LogLevel logLevel, EventId eventId)
         {
             MetricTelemetry telemetry = new MetricTelemetry();
-
             // Always apply scope first to allow state to override.
-            ApplyScopeProperties(telemetry.Properties);
+            ApplyScopeProperties(telemetry.Properties, telemetry);
+            ApplyKnownProperties(telemetry.Properties, logLevel, eventId);
 
             foreach (var entry in values)
             {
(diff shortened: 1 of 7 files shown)


src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/LogLevelEnumHelper.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        private static readonly string[] LogLevelEnumStrings;
+
+        static LogLevelEnumHelper()
+        {
+            LogLevelEnumStrings = GetEnumOrderedStrings();
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static string ToStringOptimized(LogLevel enumValue)
```
jviau: Consider making this an extension method so we can call `logLevel.ToStringOptimized()`.

Also, can we use a switch statement?

``` CSharp
public static string ToStringOptimized(this LogLevel logLevel)
{
    return logLevel switch =>
    {
        LogLevel.Trace => "Trace",
        LogLevel.Debug => "Debug",
        // ... etc
       _ => logLevel.ToString(), // or throw here?
    }
}
```

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -33,7 +35,7 @@ internal class ApplicationInsightsLogger : ILogger
         internal const string MetricMinKey = "min";
         internal const string MetricMaxKey = "max";
         internal const string MetricStandardDeviationKey = "standarddeviation";
-
+
```
jviau: nit: remove extra whitespace

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -165,17 +165,29 @@ private void LogMetric(IEnumerable<KeyValuePair<string, object>> values)
         }
 
         // Applies scope properties; filters most system properties, which are used internally
-        private static void ApplyScopeProperties(IDictionary<string, string> properties)
+        private static void ApplyScopeProperties(ITelemetry telemetry)
         {
-            var scopeProperties = DictionaryLoggerScope.GetMergedStateDictionaryOrNull();
+            var scopeProperties = DictionaryLoggerScope.GetMergedStateDictionaryOrNull();
```
jviau: nit: remove extra whitespace
```suggestion
            var scopeProperties = DictionaryLoggerScope.GetMergedStateDictionaryOrNull();
```

jviau: I am curious why we do not use `ISupportExternalScope`? Would it be possible for us to switch to it, or is that work for another time?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -236,35 +252,54 @@ private void LogTrace(LogLevel logLevel, IEnumerable<KeyValuePair<string, object
...
-        internal static void ApplyProperty(IDictionary<string, string> properties, string key, object value, bool applyPrefix = false)
+        internal void ApplyKnownProperties(IDictionary<string, string> properties, LogLevel logLevel, EventId eventId)
+        {
+            properties[LogConstants.CategoryNameKey] = _categoryName;
+            properties[LogConstants.LogLevelKey] = LogLevelEnumHelper.ToStringOptimized(logLevel);
+            if (eventId != null)
+            {
+                if (eventId.Id != 0)
+                {
+                    properties[LogConstants.EventIdKey] = Convert.ToString(eventId.Id);
```
jviau: Can you explain why `Convert.ToString(int)` over `int.ToString()`? Or rather `int.ToString(CultureInfo.InvariantCulture)`, which can be made into an extension metho `ToStringInviariant()`.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -236,35 +252,54 @@ private void LogTrace(LogLevel logLevel, IEnumerable<KeyValuePair<string, object
...
+            ApplyScopeProperties(telemetry);
             ApplyProperties(properties, state, true);
         }
 
-        internal static void ApplyProperty(IDictionary<string, string> properties, string key, object value, bool applyPrefix = false)
+        internal void ApplyKnownProperties(IDictionary<string, string> properties, LogLevel logLevel, EventId eventId)
+        {
+            properties[LogConstants.CategoryNameKey] = _categoryName;
+            properties[LogConstants.LogLevelKey] = LogLevelEnumHelper.ToStringOptimized(logLevel);
+            if (eventId != null)
```
jviau: If you are interested, pattern matching can be used to reduce the nesting here

```csharp
if (eventId { Id: != 0 })
{
}

if (eventId { Name: not null })
{
}
```

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -236,35 +252,54 @@ private void LogTrace(LogLevel logLevel, IEnumerable<KeyValuePair<string, object
...
             string stringValue = null;
 
             // Format dates
-            Type propertyType = value.GetType();
-            if (propertyType == typeof(DateTime))
+            if (objectValue is string value)
+            {
+                stringValue = value;
+            }
+            if (objectValue is DateTime date)
```
jviau: Note sure if you intentionally didn't use `else if` here.
```suggestion
            else if (objectValue is DateTime date)
```

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -307,40 +342,46 @@ private void LogFunctionResultAggregate(IEnumerable<KeyValuePair<string, object>
                         // Timestamp is created automatically
                         // We won't use the format string here
                         break;
-                    default:
-                        if (value.Value is TimeSpan)
+                    default:
```
jviau: nit: whitespace
```suggestion
                    default:
```

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/DictionaryLoggerScope.cs
```diff
@@ -17,7 +18,10 @@ private DictionaryLoggerScope(IReadOnlyDictionary<string, object> state, Diction
             State = state;
             Parent = parent;
         }
-
+        // Cache merged dictionary. Invalidate cache on push/dispose 
+        private IDictionary<string, object> _currentScope;
```
jviau: nit: place fields before the ctor.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
RohitRanjanMS: I did consider IExternalScopeProvider and we use this in all the other loggers as well. In AI logger, scope is built more than once, and I am able to cache it in the `DictionaryLoggerScope` to save allocation. Additionally, we may later remove AI logger in favor of OTel.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
RohitRanjanMS: There's not much gain here, just my preferred way to convert to string. I will add InvariantCulture.

image

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
RohitRanjanMS: I missed it.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
RohitRanjanMS: eventId is a struct, null check for eventId is redundant.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -130,7 +123,14 @@ private void LogMetric(IEnumerable<KeyValuePair<string, object>> values)
...
                     case LogConstants.NameKey:
-                        telemetry.Name = entry.Value.ToString();
+                        if (entry.Value is string name)
+                        {
+                            telemetry.Name = name;
+                        }
+                        else
+                        {
+                            telemetry.Name = entry.Value.ToString();
+                        }
```
NickCraver: Can unconditionally call .ToString() here to be a bit cleaner - it'll just get inlined as a direct ref return anyhow: https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/String.cs#L541

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
NickCraver: We should consider changing this though - we throw away anything from `GetMergedStateDictionaryOrNull` that's not `!SystemScopeKeys.Contains` or `ScopeKeys.FunctionName` below. For example, a minimal allocation reduction (as well as dictionary size optimization) would be to pass a static predicate to `GetMergedStateDictionaryOrNull` limiting the return and only needing to exclude FunctionName in the foreach below. Happy to brainstorm much more efficient ways to do this!

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
NickCraver: I'm confused in reviewing the resolved conversations here - are there changes not pushed up perhaps?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
NickCraver: FWIW a switch statement would be much more concise here with the `else if` change in play (I can't see this reviewing latest on GitHub though)

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -308,39 +343,45 @@ private void LogFunctionResultAggregate(IEnumerable<KeyValuePair<string, object>
...
                 }
             }
 
+            IDictionary<string, string> properties = new Dictionary<string, string>(2);
+            ApplyKnownProperties(properties, logLevel, eventId);
+
             foreach (KeyValuePair<string, double> metric in metrics)
             {
-                _telemetryClient.TrackMetric($"{functionName} {metric.Key}", metric.Value);
+                _telemetryClient.TrackMetric($"{functionName} {metric.Key}", metric.Value, properties);
```
NickCraver, line 359: Is `metric.Key` a finite set of values? Allocations per-concat here would be something we can eliminate if so

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/LogLevelEnumHelper.cs
NickCraver: Agreed - there's no way these change - recommend simplifying to like the core framework has: https://github.com/dotnet/runtime/blob/9b76c28567640e4cbe0d20e18b765b8f1a47473f/src/libraries/Microsoft.Extensions.Logging.Console/src/JsonConsoleFormatter.cs#L92 This is both faster and has zero startup time cost.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
```diff
@@ -49,47 +49,61 @@ public void Initialize(ITelemetry telemetry)
...
-                string eventName = scopeProps.GetValueOrDefault<string>(LogConstants.EventNameKey);
-                if (eventName != null)
-                {
-                    telemetryProps[LogConstants.EventNameKey] = eventName;
+                    if (!telemetryContext.Properties.ContainsKey(LogConstants.LogLevelKey))
+                    {
+                        LogLevel? logLevel = scopeProps.GetValueOrDefault<LogLevel?>(LogConstants.LogLevelKey);
+                        if (logLevel != null)
+                        {
+                            telemetryContext.Properties[LogConstants.LogLevelKey] = logLevel.Value.ToString();
```
NickCraver: Can use the switched .ToString method here to avoid an allocation

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
```diff
@@ -49,47 +49,61 @@ public void Initialize(ITelemetry telemetry)
...
-                    telemetryProps[LogConstants.EventIdKey] = eventId.Value.ToString();
-                }
+                    // Apply Category, LogLevel event details to all telemetry
+                    if (!telemetryContext.Properties.ContainsKey(LogConstants.CategoryNameKey))
+                    {
+                        string category = scopeProps.GetValueOrDefault<string>(LogConstants.CategoryNameKey);
+                        if (category != null)
+                        {
+                            telemetryContext.Properties[LogConstants.CategoryNameKey] = category;
+                        }
```
NickCraver: FWIW can simplify these with your new format, e.g. 
```cs
if (scopeProps.GetValueOrDefault<string>(LogConstants.CategoryNameKey) is string category)
```

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
```diff
@@ -49,47 +49,61 @@ public void Initialize(ITelemetry telemetry)
...
+                            telemetryContext.Properties[LogConstants.LogLevelKey] = logLevel.Value.ToString();
+                        }
+                    }
+
+                    if (!telemetryContext.Properties.ContainsKey(LogConstants.EventIdKey))
+                    {
+                        int? eventId = scopeProps.GetValueOrDefault<int?>(LogConstants.EventIdKey);
+                        if (eventId != null && eventId.HasValue && eventId.Value != 0)
+                        {
+                            telemetryContext.Properties[LogConstants.EventIdKey] = eventId.Value.ToString();
```
NickCraver: Should this be culture aware as well?

NickCraver (commented):
Changes are looking good - My main concern is around the `GetMergedStateDictionaryOrNull` shifts - have we done memory profiling on that tree in particular? (I see the total profiling - nice! - I'm only asking about the remainder) I worry with the new approach we may be adding dictionary allocations overall. I think we can scope this in some places but that's messy too - IMO we need another way to go about doing this all together, but I'm not sure where else it's still used looking only at the diff scope of this PR.

Overall: I think there's probably a round of changes here I'm not seeing in GitHub's current version given resolved comments, happy to take a peek at latest if helpful - nice work!

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
RohitRanjanMS: I removed the `State` from the `DictionaryLoggerScope `.  I am building the aggregate scope every time a new scope is pushed. `GetMergedStateDictionaryOrNull ` simply returns the current scope.

image

image

RohitRanjanMS: I will push the code after testing.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
RohitRanjanMS: yeah, but functionName varies as well.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
RohitRanjanMS: Done

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/LogLevelEnumHelper.cs
RohitRanjanMS: Done

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
RohitRanjanMS: Done.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
RohitRanjanMS: Done.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
RohitRanjanMS: Done.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/DictionaryLoggerScope.cs
```diff
@@ -12,16 +12,16 @@ internal class DictionaryLoggerScope
     {
         private static AsyncLocal<DictionaryLoggerScope> _value = new AsyncLocal<DictionaryLoggerScope>();
 
-        private DictionaryLoggerScope(IReadOnlyDictionary<string, object> state, DictionaryLoggerScope parent)
```
brettsam, line 15: Have you looked into getting rid of this type entirely? I wrote this years ago before ExternalScopes were supported and have changed every other instance of logging to use that rather than this. The only holdout is how we access the scope in the TelemetryInitializer, but I never did enough research to figure out how to get around that. 

I'd be curious how much improvement we get by using the built-in ScopeProvider, which I assume is optimized for performance...

I'd be interested to see performance reports if this was used (even if it is fairly hacked together for now).

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
brettsam: Agree with investigating external scopes -- I just commented elsewhere (I didn't see this comment chain) -- but this DictionaryLoggerScope code was written before IExternalScopeProvider stuff existed (it came in .NET Core 2.1; I added this stuff in 2.0 if I recall). What happens if we switch to that and abandon this thing? I know we access it in the initializer, but we may be able to hack something together to test it at least.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/DictionaryLoggerScope.cs
RohitRanjanMS: I did try `IExternalScopeProvider` but it didn't work.
1. As you mentioned, we need the scope in the `TelemetryInitializer` for Dependencies. All telemetry items flow through the `ApplicationInsightsLogger` except dependencies. `TelemetryInitializer` is the only way I can think of to enrich dependencies with additional information.
2. IExternalScopeProvider calls an internal `BeginScope` instead of the [one](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs#:~:text=public%20IDisposable-,BeginScope,-%3CTState%3E(TState%20state)) provided by us. We need this for `StartTelemetryIfFunctionInvocation`. `ApplicationInsightsLogger` is the only logger that uses `BeginScope`.

I did some benchmarking against IExternalScopeProvider and it's very efficient and works like magic. Unfortunately, couldn't use it here. 
The proposed changes are efficient as well, we are saving almost 99.74% allocation even with all the complexities.

jviau: @RohitRanjanMS can you share your code with the `IExternalScopeProvider` implementation?

RohitRanjanMS: sure, I don't have a commit, but I can make a quick change and push to a branch.

NickCraver (commented):
Current code looks good from an efficiency standpoint - I'm only reviewing in text so limited here and profiles are very much appreciated!

My only suggestion here would be: do we want to beef up the testing slightly specifically on scopes/expected values? There's enough of a change here I'm not sure existing tests cover all the cases that can happen with overrides, what if the value set in an override is null, etc - just from a 1:1 behavior with the old for nested properties.

RohitRanjanMS:
> Current code looks good from an efficiency standpoint - I'm only reviewing in text so limited here and profiles are very much appreciated!
> 
> My only suggestion here would be: do we want to beef up the testing slightly specifically on scopes/expected values? There's enough of a change here I'm not sure existing tests cover all the cases that can happen with overrides, what if the value set in an override is null, etc - just from a 1:1 behavior with the old for nested properties.

I will have a look at test cases and add more if we don't have enough coverage.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/DictionaryLoggerScope.cs
RohitRanjanMS: I created a branch that uses [`IExternalScopeProvider`](https://github.com/Azure/azure-webjobs-sdk/compare/roranjan/IExternalScopeProvider?expand=1). I don't see any issue with `BeginScope()` now and the only thing that's stopping us from using this is [`WebJobsTelemetryInitializer`](https://github.com/Azure/azure-webjobs-sdk/blob/roranjan/IExternalScopeProvider/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs#:~:text=IDictionary%3Cstring%2C%20object%3E%20scopeProps%20%3D%20new%20Dictionary%3Cstring%2C%20object%3E()%3B//DictionaryLoggerScope.GetMergedStateDictionaryOrNull()%3B).

RohitRanjanMS: I did some [benchmarking ](https://github.com/RohitRanjanMS/LoggerFactoryScopeProviderBenchmarking/tree/master/LoggerFactoryScopeProviderBenchmarking) using `LoggerFactoryScopeProvider`. While the performance is almost same, memory allocation is significantly less.

image

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
```diff
@@ -165,17 +158,28 @@ private void LogMetric(IEnumerable<KeyValuePair<string, object>> values)
         }
 
         // Applies scope properties; filters most system properties, which are used internally
-        private static void ApplyScopeProperties(IDictionary<string, string> properties)
+        private static void ApplyScopeProperties(ITelemetry telemetry)
         {
-            var scopeProperties = DictionaryLoggerScope.GetMergedStateDictionaryOrNull();
+            var scopeProperties = DictionaryLoggerScope.GetMergedStateDictionaryOrNull();
```
jviau: nit: remove extra whitespace

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/DictionaryLoggerScope.cs
```diff
@@ -12,16 +12,16 @@ internal class DictionaryLoggerScope
...
+        // Cache merged dictionary.  
+        internal IReadOnlyDictionary<string, object> CurrentScope { get; private set; }
 
         internal DictionaryLoggerScope Parent { get; private set; }
 
+        private DictionaryLoggerScope(IReadOnlyDictionary<string, object> currentScope, DictionaryLoggerScope parent)
+        {
+            CurrentScope = currentScope;
+            Parent = parent;
+        }
```
jviau, line 24: nit: add new line after `}`

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/DictionaryLoggerScope.cs
jviau: Thank you for the branch. I understand your issue now of needing access to the log scope - or rather to a telemetry property bag for multiple forms of telemetry (traces and logs here). I believe this is a concept solved by OTel Baggage, but we don't have access to it here. I don't really like the idea of rolling our own scope and having this discrepancy in how this _one_ `ILogger` works - but addressing this is not that straight forward.

As for the allocation difference you see - yes, your sample branch is allocating an extra dictionary. You can remove that perf hit by avoiding the dictionary allocation and attaching scope data directly to where it needs to go within the `ForEachScope` delegate. Additionally, you want to use the `state` parameter to pass in any variables you access within the `ForEachScope` delegate to avoid function closures. With those 2 changes you should see significantly less allocations.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/LogLevelExtension.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Globalization;
+
+namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights.Extensions
+{
+    internal static class LogLevelExtension
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal static string ToStringOptimized(this LogLevel logLevel)
+        {
+            switch (logLevel)
```
jviau, line 17: style nit: consider switch expression.

``` CSharp
return logLevel switch
{
    LogLevel.Trace => "Trace",
    ...
    _ => logLevel.ToString(CultureInfo.InvariantCulture),
};
```

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
```diff
@@ -49,53 +50,61 @@ public void Initialize(ITelemetry telemetry)
...
-                telemetryContext.Operation.Name = scopeProps.GetValueOrDefault<string>(ScopeKeys.FunctionName);
+                // Apply our special scope properties
+                IReadOnlyDictionary<string, object> scopeProps = DictionaryLoggerScope.GetMergedStateDictionaryOrNull();
 
-                // Apply Category and LogLevel to all telemetry
-                string category = scopeProps.GetValueOrDefault<string>(LogConstants.CategoryNameKey);
-                if (category != null)
+                // this could be telemetry tracked in scope of function call - then we should apply the logger scope
+                // or RequestTelemetry tracked by the WebJobs SDK or AppInsight SDK - then we should apply Activity.Tags
+                if (scopeProps != null && scopeProps.Count > 0)
```
jviau: small suggestion: 
```suggestion
                if (scopeProps?.Count > 0)
```

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/DictionaryLoggerScope.cs
```diff
@@ -37,54 +37,62 @@ public static DictionaryLoggerScope Current
...
+                BuildCurrentScope(currentState);
+            }
+            else if (state is IEnumerable<KeyValuePair<string, object>> stateEnum)
             {
-                // Convert this to a dictionary as we have scenarios where we cannot have duplicates. In this
-                // case, if there are dupes, the later entry wins.
+                IDictionary<string, object> stateValues;
+                // Convert this to a dictionary as we have scenarios where we cannot have duplicates.
+                // In this case, if there are dupes, the later entry wins.
                 stateValues = new Dictionary<string, object>();
```
jviau, line 50: If you want even more performance out of this scope implementation, I recommend looking at [LoggerFactoryScopeProvider](https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Logging/src/LoggerFactoryScopeProvider.cs) and basically following the `IExternalScopeProvider` pattern. It does a great job at avoiding extra allocations. For example here you are allocating an extra dictionary. And then `BuildCurrentScope` is going to allocate _another_ dictionary yet. `IExternalScopeProvider` avoids these allocations by not collapsing/processing the scope upfront, but rather letting consumers iterate through the scope values and they must perform type checking and consuming each scope level appropriately - and they also need to decide how to handle duplicates.

Not a blocker for this PR, but I thought I would mention this as the primary focus of this is performance.

jviau (approved):
Small nits and some perf suggestions.

My primary concern is the way this makes our one `ApplicationInsightsLogger` unique. I worry that we have two ways log scopes are done, and it is not clear if you need to use `ILogger.BeginScope` or this other `DictionaryLoggerScope`. I get that we are trying to bring log scope properties over to distributed tracing (AI dependencies here). Did we explore if `Activity.Tags` would work here? Or is there some other prescribed pattern for having a general "telemetry scope" instead of a log scope?

I am approving this as the PR goal is performance and the `DictionaryLoggerScope` deviation (which is my primary concern) predates this PR.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/DictionaryLoggerScope.cs
RohitRanjanMS: The dictionary allocation in the `Push` method is only if the state is `IEnumerable`. If it's already a dictionary, then there is no allocation. In all our code path we use a dictionary for state. `IEnumerable` is just to make sure we don't break if any customer is using it for scope.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLogger.cs
RohitRanjanMS: Fixed.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/DictionaryLoggerScope.cs
RohitRanjanMS: Done.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/LogLevelExtension.cs
RohitRanjanMS: This feature is not available in C# 7.3

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
RohitRanjanMS: Done.

PR #2953, opened 2023-02-06
Scale API refinements

mathewc:
Proposed changes to active scale API work


src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsHostBuilderExtensions.cs
```diff
@@ -137,16 +139,32 @@ public static IHostBuilder ConfigureWebJobs(this IHostBuilder builder, Action<Ho
...
 
         /// <summary>
-        /// Applies scale configuration to the specified <see cref="IHostBuilder"/>.
+        /// Configures the specified <see cref="IHostBuilder"/> as a scale manager host.
         /// </summary>
         /// <param name="builder">The <see cref="IHostBuilder"/> to configure.</param>
+        /// <param name="configureScaleOptions">Configuration action for <see cref="ScaleOptions"/>.</param>
         /// <returns>The <see cref="IHostBuilder"/>.</returns>
-        public static IHostBuilder AddScale(this IHostBuilder builder)
+        public static IHostBuilder ConfigureWebJobsScale<TConcurrencyStatusRepository, TScaleMetricsRepository>(this IHostBuilder builder, Action<ScaleOptions> configureScaleOptions, string hostId)
```
mathewc: Required repository impls. provided as type params. HostID also specified, and using existing fixed ID provider

src/Microsoft.Azure.WebJobs.Host/Scale/IScaleManager.cs
```diff
@@ -8,20 +8,16 @@
...
+        /// <param name="context">The <see cref="ScaleStatusContext"/>.</param>
+        /// <returns>The current <see cref="ScaleStatus"/>.</returns>
         Task<ScaleStatus> GetScaleStatusAsync(ScaleStatusContext context);
-
-        /// <summary>
-        /// Gets <see cref="ScaleStatus"> for each trigger.
-        /// </summary>
-        /// <param name="context"></param>
-        /// <returns></returns>
-        Task<IDictionary<string, ScaleStatus>> GetScaleStatusesAsync(ScaleStatusContext context);
```
mathewc, line 25: We don't need a new method for this - thinking we can just include per function status as part of the existing result type

src/Microsoft.Azure.WebJobs.Host/Scale/IScalerProvider.cs
```diff
@@ -0,0 +1,27 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.WebJobs.Host.Scale
+{
+    /// <summary>
+    /// Interface for providing scalers.
+    /// </summary>
+    public interface IScalerProvider
```
mathewc, line 13: This ends up being required if we want to allow extensions to use DI properly w/o having to new instances up manually, which breaks DI flow. TBD on exact name

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleOptions.cs
```diff
@@ -78,20 +80,27 @@ public string Format()
...
-        /// Gets or sets the function to checks if target scaler is supported for specific <see cref="ITargetScaler"/>
+        /// Gets or sets a function to check if a particular <see cref="ITargetScaler"/> is enabled.
         /// </summary>
-        public Func<ITargetScaler, bool> IsTargetBasedScalingEnabledForTriggerFunc { get; set; }
+        public Func<ITargetScaler, bool> CheckTargetScalerEnabled { get; set; }
+
+        /// <summary>
+        /// Gets or sets the function metadata to configure the host with.
+        /// </summary>
+        public JArray FunctionMetadata { get; set; }
```
mathewc: Stuck this here for now, though it's not really ideal. Considering better ways to pass this through

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
```diff
@@ -1,191 +1,155 @@
...
+                options.CheckTargetScalerEnabled = ts =>
                 {
-                    hostBuilder.AddScaleForCosmosDbTrigger(jtoken.ToString());
-                }
-            }
+                    return true;
+                };
+                options.MetricsPurgeEnabled = false;
+                options.FunctionMetadata = functionMetadata;
+            }, hostId)
```
mathewc: With these changes, you can see how much simpler host config is

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
```diff
@@ -276,60 +302,56 @@ public Task<TargetScalerResult> GetScaleResultAsync(TargetScalerContext context)
...
-            {
-                _functionName = functionName;
-            }
-
-            public IScaleMonitor GetMonitor()
+            builder.ConfigureServices((context, services) =>
             {
-                return new ScaleMonitorImpl(_functionName, _functionName);
-            }
+                services.AddSingleton<IScalerProvider, TestExtensionAScalerProvider>();
```
mathewc, line 309: Extensions simply register a single provider

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
```diff
@@ -230,12 +194,74 @@ public Task WriteMetricsAsync(IDictionary<IScaleMonitor, ScaleMetrics> monitorMe
...
         }
+    }
+
+    internal abstract class TestExtensionScalerProvider : IScalerProvider
+    {
+        private IOptions<ScaleOptions> _scaleOptions;
+        private List<IScaleMonitor> _scaleMonitors;
+        private List<ITargetScaler> _targetScalers;
+
+        public TestExtensionScalerProvider(IConfiguration config, IOptions<ScaleOptions> scaleOptions)
```
mathewc: Extension providers can take their parameters from DI - we're not newing one up explicitly anywhere

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleOptions.cs
```diff
@@ -78,20 +80,27 @@ public string Format()
...
         /// </summary>
-        public bool IsTargetBasedScalingEnabled { get; set; }
+        public bool IsTargetScalingEnabled { get; set; }
 
         /// <summary>
-        /// Gets or sets the function to checks if target scaler is supported for specific <see cref="ITargetScaler"/>
+        /// Gets or sets a function to check if a particular <see cref="ITargetScaler"/> is enabled.
         /// </summary>
-        public Func<ITargetScaler, bool> IsTargetBasedScalingEnabledForTriggerFunc { get; set; }
+        public Func<ITargetScaler, bool> CheckTargetScalerEnabled { get; set; }
```
mathewc: Because we only have a temporary need for this while we're doing staged rollout, we shouldn't add a specific first class API for it. I wonder instead if we don't just need a general feature flags API that we can use for this and other things? I.e. also push the host feature flags interface into SDK in some form as well. This pattern is going to continue in the future, and we don't want to litter interfaces with shortlived config APIs like this.

PR #2954, opened 2023-02-07
Scaler apis

### LINKED ISSUE #2938: Exposing common scaling code to consume by SC v3

alrod:
Resolves https://github.com/Azure/azure-webjobs-sdk/issues/2938
FunctionHost PR: https://github.com/Azure/azure-functions-host/pull/9131
SB: https://github.com/Azure/azure-sdk-for-net/pull/34557
EH: https://github.com/Azure/azure-sdk-for-net/pull/35344
CosmosDb: https://github.com/Azure/azure-webjobs-sdk-extensions/pull/841
Queue/Blob:https://github.com/Azure/azure-sdk-for-net/pull/35358


test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
```diff
@@ -0,0 +1,357 @@
...
+                // configure scale options
+                options.IsTargetScalingEnabled = tbsEnabled;
+                options.CheckTargetScalerEnabled = ts =>
+                {
+                    return true;
+                };
+                options.MetricsPurgeEnabled = false;
+                options.FunctionMetadata = functionMetadata;
+            }, hostId)
+            .ConfigureTestExtensionAScale()
```
alrod: I am thinking about [dynamic assembly loading](https://learn.microsoft.com/en-us/dotnet/core/tutorials/creating-app-with-plugin-support), in this case we do not know about exact name of the method.
we can add a new interface:
```
    interface IConfigureExtensionScale
    {
        void ConfigureExtensionScale(IHostBuilder builder);
    }
```

this approach is similar to [IWebJobsStartup](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Hosting/IWebJobsStartup.cs) and [IWebJobsStartup2](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Hosting/IWebJobsStartup2.cs).

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
```diff
@@ -0,0 +1,357 @@
...
+            .ConfigureWebJobsScale<TestConcurrencyStatusRepository, TestScaleMetricsRepository>(options =>
+            {
+                // configure scale options
+                options.IsTargetScalingEnabled = tbsEnabled;
+                options.CheckTargetScalerEnabled = ts =>
+                {
+                    return true;
+                };
+                options.MetricsPurgeEnabled = false;
+                options.FunctionMetadata = functionMetadata;
```
alrod: We also want to pass token credentials per trigger from SC to extension to initiate azure services clients for the scalers.

We can override CreateTokenCredential in [AzureComponentFactory](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/extensions/Microsoft.Extensions.Azure/src/AzureComponentFactory.cs) and pass the custom factory from SC (internal implementation is [AzureComponentFactoryImpl](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/extensions/Microsoft.Extensions.Azure/src/Internal/AzureComponentFactoryImpl.cs)) to create azure clients.

To pass various parameters we can add `IDictionary<object, object> Properties`; (something like we have for [HostBuilderContext](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.hosting.hostbuildercontext.properties?view=dotnet-plat-ext-7.0#microsoft-extensions-hosting-hostbuildercontext-properties)). We can not specify `AzureComponentFactory` or `TokenCredential` in web jobs sdk contract as it requires a new reference in webjobs sdk.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsHostBuilderExtensions.cs
```diff
@@ -135,6 +138,38 @@ public static IHostBuilder ConfigureWebJobs(this IHostBuilder builder, Action<Ho
             return builder;
         }
 
+        /// <summary>
+        /// Configures the specified <see cref="IHostBuilder"/> as a scale manager host.
+        /// </summary>
+        /// <param name="builder">The <see cref="IHostBuilder"/> to configure.</param>
+        /// <param name="configureScaleOptions">Configuration action for <see cref="ScaleOptions"/>.</param>
+        /// <returns>The <see cref="IHostBuilder"/>.</returns>
+        public static IHostBuilder ConfigureWebJobsScale<TConcurrencyStatusRepository, TScaleMetricsRepository>(this IHostBuilder builder, Action<ScaleOptions> configureScaleOptions, string hostId)
```
alrod, line 147: Also need `IAzureBlobStorageProvider` to initialize `BlobStorageConcurrencyStatusRepository`.
Both SC and function runtime need to use `BlobStorageConcurrencyStatusRepository` but `IAzureBlobStorageProvider `should be different.

alrod: It looks like we need `IAzureBlobStorageProvider`

src/Microsoft.Azure.WebJobs.Host.Storage/BlobStorageConcurrencyStatusRepository.cs
```diff
@@ -17,7 +17,7 @@
 
 namespace Microsoft.Azure.WebJobs.Host
 {
-    internal class BlobStorageConcurrencyStatusRepository : IConcurrencyStatusRepository
+    public class BlobStorageConcurrencyStatusRepository : IConcurrencyStatusRepository
```
mathewc: Rather than making this implementation public, scale hosts should register this implementation using an appropriate host builder method. Currently, AddAzureStorageCoreServices is called to register this implementation. If we can't use that method, we could introduce a more targeted one, the goal being to keep this type internal.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsHostBuilderExtensions.cs
mathewc: The AddAzureStorageCoreServices method I mentioned above also registers IAzureBlobStorageProvider

mathewc: Open question on whether we should base these new host builder extensions on IWebJobsBuilder. Need to check to see what extension methods exist on IWebJobsBuilder and whether those make sense for scaler hosts

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleOptions.cs
```diff
@@ -0,0 +1,106 @@
...
+
+        /// <summary>
+        /// Gets or sets a value indicating whether target base scaling is enabled at the host level.
+        /// </summary>
+        public bool IsTargetScalingEnabled { get; set; }
+
+        /// <summary>
+        /// Gets or sets a function to check if a particular <see cref="ITargetScaler"/> is enabled.
+        /// </summary>
+        public Func<ITargetScaler, bool> CheckTargetScalerEnabled { get; set; }
```
mathewc: This is a temporary hack that will be removed. This is for supporting feature flags as the functions host does now. We'll need a first class interface for feature flags that can be used, that various hosts can implement.

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleOptions.cs
```diff
@@ -0,0 +1,106 @@
...
+
+        /// <summary>
+        /// Gets or sets the function metadata to configure the host with.
+        /// </summary>
+        public JArray FunctionMetadata { get; set; }
```
mathewc: This is a temporary hack - we'll need to introduce an interface for providing function metadata to the host

src/Microsoft.Azure.WebJobs.Host.Storage/StorageServiceCollectionExtensions.cs
```diff
@@ -31,6 +31,15 @@ public static void AddAzureStorageCoreServices(this IServiceCollection services)
...
             services.AddAzureClientsCore();
 
+            services.AddAzureStorageScaleServices(null);
+        }
+
+        public static void AddAzureStorageScaleServices(this IServiceCollection services, Action<JobHostInternalStorageOptions> configure)
+        {
+            if (configure != null)
+            {
+                services.Configure(configure);
```
mathewc: Why do we need to configure this SAS at all in the scaler scenarios? Scaler hosts won't even be using the infrastructure storage account, right?

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleMonitorService.cs
```diff
@@ -0,0 +1,149 @@
...
+        public void Dispose()
+        {
+            Dispose(true);
+        }
+
+        private async Task TakeMetricsSamplesAsync()
+        {
+            try
+            {
+                // TODO: Need to fix this cast to ScaleManager
```
alrod: [Mathew]: We need to find a way to remove this cast ideally

src/Microsoft.Azure.WebJobs.Host/Scale/TriggerMetadata.cs
```diff
@@ -0,0 +1,41 @@
...
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Azure.WebJobs.Host.Scale
+{
+    /// <summary>
+    /// Represents trigger metadata.
+    /// </summary>
+    public class TriggerMetadata
+    {
+        public TriggerMetadata(string functionName, JObject metadata)
```
alrod: [Mathew]: I revamped the trigger metadata / provider APIs to make them cleaner. We may need to iterate more.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
```diff
@@ -0,0 +1,415 @@
...
+        private const string Function3Name = "Function3";
+
+        [Theory]
+        [InlineData(false)]
+        [InlineData(true)]
+        public async Task ScaleManager_GetScaleStatusAsync_ReturnsExpected(bool tbsEnabled)
+        {
+            var triggerMetadata = new List<TriggerMetadata>()
+            {
+                new TriggerMetadata(Function1Name, new JObject { { "type", "testExtensionATrigger" } }),
```
alrod: [Mathew]: I'd like to discuss more why this `Properties` bag is needed for triggers, but for now assuming it is, I retain it in the API.

[Alexey]: This is needed to pass `AzureComponentFactory` for example (maybe some other properties in the future). We can not define `AzureComponentFactory` in `TriggerMetadata` as it would bring another nuget package reference and we do no want that.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
```diff
@@ -0,0 +1,415 @@
...
+            {
+                // Adding host.json here
+                config.AddJsonStream(new MemoryStream(Encoding.UTF8.GetBytes(hostJson)));
+
+                // Adding app settings
+                config.AddInMemoryCollection(new Dictionary<string, string>
+                {
+                    { "app_setting1", "value1" },
+                    { "app_setting2", "value2"},
+                    { "FeatureManagement:Microsoft.Azure.WebJobs.Host.EndToEndTests", "true" }
```
alrod: [Mathew]: Here's how a feature flag can be added. We can also control the root "FeatureManagement" name - we could have it be "WebJobsFeatures" or something.

src/Microsoft.Azure.WebJobs.Host/Scale/TriggerMetadataProvider.cs
```diff
@@ -0,0 +1,24 @@
...
+    internal class TriggerMetadataProvider : ITriggerMetadataProvider
+    {
+        private readonly IEnumerable<TriggerMetadata> _triggerMetadata;
+
+        public TriggerMetadataProvider(IEnumerable<TriggerMetadata> triggerMetadata)
+        {
+            _triggerMetadata = triggerMetadata;
+        }
+
+        public IEnumerable<TriggerMetadata> GetTriggerMetadata()
```
alrod: Do we want to filer by trigger type? Extension author needs to filter extensions' trigger through all trigger data.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
alrod: It must be the same as in function host. In function host we are using "[FunctionsHostingConfig](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/ScriptConstants.cs#L224)". But it can be changes to whatever name we like.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsHostBuilderExtensions.cs
```diff
@@ -135,6 +140,38 @@ public static IHostBuilder ConfigureWebJobs(this IHostBuilder builder, Action<Ho
...
+                if (configureScaleOptions != null)
+                {
+                    services.Configure(configureScaleOptions);
+                }
+
+                services.AddFeatureManagement();
+                services.TryAddSingleton<IScaleMonitorManager, ScaleMonitorManager>();
+                services.TryAddSingleton<ITargetScalerManager, TargetScalerManager>();
+                services.TryAddSingleton<IScaleManager, ScaleManager>();
+                services.TryAddSingleton<IHostedService, ScaleMonitorService>();
```
alrod: We need to add `ScaleMonitorService` in function host. As implementation is internal we need to call `ConfigureWebJobsScale` in function host or add a new method.

https://github.com/Azure/azure-functions-host/blob/alrod/sc-redesign2/src/WebJobs.Script/ScriptHostBuilderExtensions.cs#L314

src/Microsoft.Azure.WebJobs.Host/Constants.cs
```diff
@@ -15,5 +15,6 @@ internal static class Constants
         public const string AzureWebsiteInstanceId = "WEBSITE_INSTANCE_ID";
         public const string AzureWebsiteContainerName = "CONTAINER_NAME";
         public const string DateTimeFormatString = "yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'fffK";
+        public const string HostingConfigSectionName = "HostingConfig";
```
mathewc, line 18: What is this here for - I don't see it being used.

mathewc:
Adding @brettsam for general WebJobs SDK review, as well as specific review for the DI and host builder changes, since he has a lot of experience in that area

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsHostBuilderExtensions.cs
```diff
@@ -135,6 +140,38 @@ public static IHostBuilder ConfigureWebJobs(this IHostBuilder builder, Action<Ho
...
+                services.TryAddSingleton<IPrimaryHostStateProvider>(new PrimaryHostStateProvider() { IsPrimary = true });
```
mathewc: Not clear to me you should be hardcoding IsPrimary like this, as opposed to also registering PrimaryHostCoordinator as is done normally. Then IsPrimary is based on existing logic and use of IDistributedLockManager. The main point is that we need to think about Functions Host usages and host builder ordering to ensure we're using the right services.

src/Microsoft.Azure.WebJobs.Host/Scale/IScalerProvider.cs
```diff
@@ -0,0 +1,27 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.WebJobs.Host.Scale
+{
+    /// <summary>
+    /// Interface for providing scalers.
+    /// </summary>
+    public interface IScalerProvider
```
mathewc: We might consider splitting this interface apart into two separate interfaces. I need to think about it more. The main reason would be single responsibility. In my prototype I shoved them together like this for simplicity, and also because the name IScaleMonitorProvider was already taken.

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleMonitorService.cs
```diff
@@ -0,0 +1,149 @@
...
+using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
+using System.Linq;
+using Microsoft.Extensions.Options;
+using Microsoft.Azure.WebJobs.Hosting;
+
+namespace Microsoft.Azure.WebJobs.Host.Scale
+{
+    /// <summary>
+    /// Class to store scale metrics.
```
mathewc: This type is being moved from Functions Host. I don't know why you randomly changed the class xml doc to something less correct from what it already was https://github.com/Azure/azure-functions-host/blob/8c8d2df01b6a3d5dbfbf5e268241eef8d4f4eb83/src/WebJobs.Script/Scale/FunctionsScaleMonitorService.cs#L20 which is more correct.

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleMonitorService.cs
```diff
@@ -0,0 +1,149 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Hosting;
+using System;
```
mathewc, line 4: Same here - you're messing up the using orderings randomly in files you're moving over. Update your VS settings to order system usings first

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleOptions.cs
```diff
@@ -0,0 +1,95 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
```
mathewc, line 7: In all files fix using ordering

src/Microsoft.Azure.WebJobs.Host/Scale/TriggerMetadataProvider.cs
mathewc: I don't think so. An implementor of this interface has a single job - return all the trigger metadata. That metadata may be used in a variety of ways as needed by consumers. If needed, we could add some extension methods to simplify common operations. Once we have some extensions on the new APIs we can look and decide how we can simplify their logic I think.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsHostBuilderExtensions.cs
alrod: Discussed offline - `PrimaryHostStateProvider` is a minimal interface needed for `ScaleMonitorService`

src/Microsoft.Azure.WebJobs.Host/Scale/TriggerMetadataProvider.cs
alrod: Discussed offline, let's keep the interface as simple as we can

src/Microsoft.Azure.WebJobs.Host/Scale/TriggerMetadata.cs
alrod: I added the logic to get function name and type directly from metadata

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
alrod: This is solved by `IAzureComponentFactoryProvider` in `WebJobs.Host.Storage`

src/Microsoft.Azure.WebJobs.Host/Scale/IScalerProvider.cs
alrod: what if we will use the names?
```
IScaleMonitorsProvider
ITargetScalersProvider
```

alrod: We decided to IScaleMonitorProvider

sample/SampleHost/SampleHost.csproj
```diff
@@ -15,7 +15,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.EventHubs" Version="3.0.6" />
     <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.ServiceBus" Version="3.1.1" />
-    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Storage" Version="5.0.0-beta.5" />
+    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Storage" Version="5.0.0-beta.3" />
```
mathewc: What's going on here - why are we decrementing this?

src/Microsoft.Azure.WebJobs.Host.Storage/StorageServiceCollectionExtensions.cs
```diff
@@ -24,15 +24,18 @@ public static void AddAzureStorageCoreServices(this IServiceCollection services)
...
 
-            services.TryAddEnumerable(ServiceDescriptor.Transient<IConfigureOptions<JobHostInternalStorageOptions>, CoreWebJobsOptionsSetup<JobHostInternalStorageOptions>>());
-
             services.TryAddSingleton<IDelegatingHandlerProvider, DefaultDelegatingHandlerProvider>();
 
             // Adds necessary Azure services to create clients
             services.AddAzureClientsCore();
 
-            services.TryAddSingleton<IAzureBlobStorageProvider, AzureStorageProvider>();
+            services.AddAzureStorageScaleServices();
```
mathewc: I think it'd be safer to just duplicate the registrations in the new method rather than call the new API here. In the future we might add unrelated services to the new builder method that we DONT want to include in core scenarios. This configuration makes it easy to make a mistake.

src/Microsoft.Azure.WebJobs.Host/Constants.cs
```diff
@@ -15,5 +15,6 @@ internal static class Constants
         public const string AzureWebsiteInstanceId = "WEBSITE_INSTANCE_ID";
         public const string AzureWebsiteContainerName = "CONTAINER_NAME";
         public const string DateTimeFormatString = "yyyy'-'MM'-'dd'T'HH':'mm':'ss'.'fffK";
+        public const string FunctionsHostingConfigSectionName = "FunctionsHostingConfig";
```
mathewc: We don't want to have "Functions" specific naming/apis in the SDK. Wondering if there's a simple way that the config can be transformed for scaler hosts such that the SDK code is just looking for SDK scoped app setting names. Can you provide an example of the current name/value for this config so I can suggest a translation?

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleManager.cs
```diff
@@ -0,0 +1,299 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Host.Scale
+{
+    /// <summary>
+    /// Class to get scaling votes for all the triggers or for an individual trigger.
```
mathewc: Any reason you changed the previous text "Manages scale monitoring operations."?

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleManager.cs
```diff
@@ -0,0 +1,299 @@
...
+            _targetScalersInError = new HashSet<string>();
+            _scaleOptions = scaleConfiguration;
+            _configuration = configuration;
+        }
+
+        // for mock testing only
+        internal ScaleManager()
+        {
+        }
+
```
mathewc, line 52: Remove extra spaces

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleMonitorService.cs
mathewc: I guess to avoid the cast we can just define a static helper ScalerManager.GetScalersToSample that takes in the required parameters

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleMonitorService.cs
```diff
@@ -0,0 +1,150 @@
...
+            _timer = new Timer(OnTimer, null, Timeout.Infinite, Timeout.Infinite);
+            _logger = loggerFactory.CreateLogger<ScaleMonitorService>();
+            _scaleOptions = scaleOptions;
+            _primaryHostStateProvider = primaryHostStateProvider;
+        }
+
+        public virtual Task StartAsync(CancellationToken cancellationToken)
+        {
+            // start the timer by setting the due time
+            string tbsState = _scaleOptions.Value.IsTargetScalingEnabled ? "enabled" : "disabled";
```
mathewc: I don't understand the changes made to this service when it was migrated over. The existing code checks if RTS is enabled and only starts if it is. You've removed that logic why?

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleMonitorService.cs
```diff
@@ -0,0 +1,150 @@
...
+            _logger.LogInformation($"Scale monitor service is started. Target base scaling is {tbsState}.");
```
mathewc: Remove this logging. We already log the options on startup right?

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleOptions.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// <summary>
+        /// Gets or sets a value indicating whether old metrics data
+        /// will be auto purged.
+        /// </summary>
+        public bool MetricsPurgeEnabled { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether target base scaling is enabled at the host level.
+        /// </summary>
+        public bool IsTargetScalingEnabled { get; set; }
```
mathewc, line 80: To confirm, in Functions this is currently controlled by app setting TARGET_BASED_SCALING_ENABLED. Remind me again why we're translating this to an options? Just to allow additional configuration capabilities?

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleStatus.cs
```diff
@@ -12,5 +14,15 @@ public class ScaleStatus
...
+
+        /// <summary>
+        /// Gets or sets the current target worker count.
+        /// </summary>
+        public int? TargetWorkerCount { get; set; }
+
+        /// <summary>
+        /// Gets the individual scale statuses for all monitored functions.
+        /// </summary>
+        public IDictionary<string, ScaleStatus> FunctionScaleStatuses { get; set; }
```
mathewc: The changes to this type don't look right. Note that in this member we're recursively referencing this type which isn't correct.

This type ScaleStatus as per its class doc is just the status type returned from IScaleMonitor. If you need a new type that represents aggregated status across monitors/scalers, we need a new type. In Functions [here](https://github.com/Azure/azure-functions-host/blob/379e98010c1d789369df6516b0eb5cc60556cdc7/src/WebJobs.Script/Scale/ScaleStatusResult.cs#L16) we had an aggregate type defined.

We could call it HostScaleStatus, AggregateScaleStatus or something to indicate that its an aggregate status across all functions in the host.

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleMonitorManager.cs
```diff
@@ -14,12 +15,15 @@ public ScaleMonitorManager()
         {
         }
 
-        public ScaleMonitorManager(IEnumerable<IScaleMonitor> monitors)
+        public ScaleMonitorManager(IEnumerable<IScaleMonitor> monitors, IEnumerable<IScaleMonitorProvider> scalerProviders)
```
mathewc: nit: monitorProviders

src/Microsoft.Azure.WebJobs.Host/Scale/TriggerMetadata.cs
```diff
@@ -0,0 +1,47 @@
...
+
+namespace Microsoft.Azure.WebJobs.Host.Scale
+{
+    /// <summary>
+    /// Represents trigger metadata.
+    /// </summary>
+    public class TriggerMetadata
+    {
+        public TriggerMetadata(JObject metadata)
+            : this(metadata, new Dictionary<string, object>())
```
mathewc: define as case insensitive

src/Microsoft.Azure.WebJobs.Host/Scale/TriggerMetadata.cs
```diff
@@ -0,0 +1,47 @@
...
+        public TriggerMetadata(JObject metadata, IDictionary<string, object> properties)
+        {
+            Metadata = metadata;
+            Properties = properties;
+        }
+
+        /// <summary>
+        /// Gets the name of the Function this trigger belongs to.
+        /// </summary>
+        public string FunctionName => Metadata.GetValue("functionName", StringComparison.OrdinalIgnoreCase)?.Value<string>();
```
mathewc: For these properties that are derived from the metadata payload, shouldn't we just initialize these values once in the ctor?

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsHostBuilderExtensions.cs
mathewc: How does this layer with Azure Functions usage though? In Functions we're calling ConfigureWebJobsScale (I see in your PR) so it'll register this dummy impl. when we want an actual impl.

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleMonitorService.cs
alrod: Making the method static would be problematic as it uses to much stuff from the instance:
```
_targetScalersInError
_monitorManager
_targetScalerManager
_configuration
_scaleOptions
```

My vote is to leave the cast t here.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsHostBuilderExtensions.cs
alrod: I went with default registration and staring the timer only RTS is enabled
https://github.com/Azure/azure-functions-host/pull/9131#discussion_r1127055448

sample/SampleHost/SampleHost.csproj
alrod: I am fixing the versions in the tests/sample as current version is [5.0.0-beta.2](https://github.com/Azure/azure-webjobs-sdk/blob/dev/build/common.props#L5)

src/Microsoft.Azure.WebJobs.Host/Constants.cs
alrod: I just reused the name for hosting configurations in current [function host](https://github.com/Azure/azure-functions-host/blob/5a70c857d1575060d39f0af874d88f5572a29a61/src/WebJobs.Script/Config/FunctionsHostingConfigOptionsSetup.cs#L25). So basically the pairs are { "[feature_name]", "[feature_value]" } from hosting config db.
For functions host scenario the configuration will be built in function host and the section name must be the same in function host and in web jobs sdk. We can come up with different name, what about "HostingConfig"?

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleManager.cs
alrod: Reverted back.

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleMonitorService.cs
alrod: I went with default registration and staring the timer only RTS is enabled
https://github.com/Azure/azure-functions-host/pull/9131#discussion_r1127055448

Now the code is the same as it was in function host

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleOptions.cs
alrod: Yes, there are 2 levels on enabling TBS:
1. App setting.
2. Specific trigger level.

BTW we made TBS default strategy recently (not released yet):
https://github.com/Azure/azure-functions-host/commit/e65a0a179fe957f084652b7a8e0b29a74253c4dc

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleStatus.cs
alrod: You proposed the FunctionScaleStatuses in the commit:
https://github.com/Azure/azure-webjobs-sdk/pull/2954/commits/79a0486f4f7d18a902da55921771b9bb2b84f071#diff-bd390eccc6a95518edc67c13fef7a99bf7bfae7933b0a8ed4b6eb0180d5ae4b0R66

I do not think we want to introduce a new type, my previous proposal was to have 2 methods (see the commit diff):
```
public async Task<ScaleStatus> GetScaleStatusAsync(ScaleStatusContext context)
public async Task<IDictionary<string, ScaleStatus>> GetScaleStatusesAsync(ScaleStatusContext context)
```
What do you think?

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleMonitorService.cs
alrod: I do not see the log with the options , not sure how we are adding this.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
```diff
@@ -0,0 +1,379 @@
...
+            var section = scaleHost.Services.GetService<IConfiguration>().GetSection(Constants.HostingConfigSectionName);
+            Assert.False(section.GetValue<string>("sovemalue") == "1");
+            Assert.True(section.GetValue<string>("Microsoft.Azure.WebJobs.Host.EndToEndTests") == "1");
+            Assert.True(section.GetValue<string>("microsoft.azure.webJobs.host.endtoendtests") == "1");
+
+            await TestHelpers.Await(async () =>
+            {
+                IScaleManager scaleManager = scaleHost.Services.GetService<IScaleManager>();
+
+                var scaleStatus = await scaleManager.GetAggrigatedScaleStatusAsync(new ScaleStatusContext());
```
chiangvincent: nit: typo GetAggregatedScaleStatusAsync

src/Microsoft.Azure.WebJobs.Host/Scale/IScaleManager.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// Provides scale status.
+    /// </summary>
+    public interface IScaleManager
+    {
+        /// <summary>
+        /// Gets the current scale status <see cref="AggregatedScaleStatus">.
+        /// </summary>
+        /// <param name="context">The <see cref="ScaleStatusContext"/>.</param>
+        /// <returns>The current <see cref="ScaleStatus"/>.</returns>
+        Task<AggregatedScaleStatus> GetAggrigatedScaleStatusAsync(ScaleStatusContext context);
```
chiangvincent: Nit: typo GetAggregatedScaleStatusAsync

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsServiceCollectionExtensions.cs
```diff
@@ -89,6 +89,9 @@ public static IWebJobsBuilder AddWebJobs(this IServiceCollection services, Actio
             services.TryAddSingleton<IDistributedLockManager, InMemoryDistributedLockManager>();
             services.TryAddSingleton<IScaleMonitorManager, ScaleMonitorManager>();
             services.TryAddSingleton<ITargetScalerManager, TargetScalerManager>();
+            services.TryAddSingleton<IScaleManager, ScaleManager>();
+            services.AddHostedService<ScaleMonitorService>();
+            services.AddSingleton<IScaleMetricsRepository, NullScaleMetricsRepository>();
```
chiangvincent: What is the purpose of adding a null scale metric repository here over an `InMemoryScaleMetricsRepository`?

alrod: we need `NullScaleMetricsRepository` to resolve `IScaleMetricsRepository` in `ScaleMonitorService` as it's added by default in webjobs host.
`IScaleMetricsRepository` is implemented in function host

src/Microsoft.Azure.WebJobs.Host/Scale/TriggerMetadata.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        /// <summary>
+        /// Gets the name of the Function this trigger belongs to.
+        /// </summary>
+        public string FunctionName => _name;
+
+        /// <summary>
+        /// Gets the type of the trigger.
+        /// </summary>
+        public string Type => _type;
```
chiangvincent, line 40: For reference, [here ](https://msazure.visualstudio.com/One/_git/AAPT-Antares-ScaleController?path=/src/ScaleController/Models/TriggerType.cs&version=GBv3.x&line=1&lineEnd=20&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents) are the available trigger types in scale controller v3. It could be good to define this enum in webjobs sdk instead of scale controller as well

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
```diff
@@ -0,0 +1,379 @@
...
+        internal abstract class TestExtensionScalerProvider : IScaleMonitorProvider, ITargetScalerProvider
+        {
+            private IOptions<ScaleOptions> _scaleOptions;
+            private IScaleMonitor _scaleMonitor;
+            private ITargetScaler _targetScaler;
+            private TriggerMetadata _triggerMetadata;
+
+            public TestExtensionScalerProvider(IConfiguration config, IOptions<ScaleOptions> scaleOptions, TriggerMetadata triggerMetadata)
+            {
+                Assert.Equal(scaleOptions.Value.ScaleMetricsMaxAge, TimeSpan.FromMinutes(4));
```
chiangvincent, line 237: We should enforce a schema of required values in the TriggerMetadata for each trigger type. For example, for ServiceBus triggers, enforce that TriggerMetadata contains queueName, etc. before instantiating this class. 

Is  webjobs sdk is the best place to put this schema, or in each of the individual extensions?

src/Microsoft.Azure.WebJobs.Host.Storage/StorageServiceCollectionExtensions.cs
```diff
@@ -36,6 +36,13 @@ public static void AddAzureStorageCoreServices(this IServiceCollection services)
             services.AddSingleton<IConcurrencyStatusRepository, BlobStorageConcurrencyStatusRepository>();
         }
 
+        public static void AddAzureStorageScaleServices(this IServiceCollection services)
+        {
+            services.TryAddEnumerable(ServiceDescriptor.Transient<IConfigureOptions<JobHostInternalStorageOptions>, CoreWebJobsOptionsSetup<JobHostInternalStorageOptions>>());
+            services.TryAddSingleton<IAzureBlobStorageProvider, AzureStorageProvider>();
+            services.AddSingleton<IConcurrencyStatusRepository, BlobStorageConcurrencyStatusRepository>();
```
chiangvincent, line 43: How do we ensure that the right storage account is selected for the AzureStorageProvider and for the BlobStorageConcurrencyStatusRepository?

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsServiceCollectionExtensions.cs
chiangvincent: Got it, so scale controller will have to implement the IScaleMetricsRepository as well

src/Microsoft.Azure.WebJobs.Host/Scale/AggregatedScaleStatus.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Azure.WebJobs.Host.Scale
+{
+    /// <summary>
+    /// Represents aggrigated scale status.
```
chiangvincent: nit: typo aggregated

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleStatus.cs
```diff
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using System.Collections.Generic;
```
chiangvincent: Remove if not needed?

src/Microsoft.Azure.WebJobs.Host/Scale/TriggerMetadata.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        /// <summary>
+        /// Gets all the properties tagged to this instance.
+        /// </summary>
+        public IDictionary<string, object> Properties { get; }
```
chiangvincent, line 45: Added another comment on this somewhere else, but this could also be a good place to add validations for attributes needed for each trigger type

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/WebJobs.Host.FunctionalTests.csproj
```diff
@@ -31,7 +31,7 @@
     <PackageReference Include="Moq" Version="4.16.0" />
     <PackageReference Include="xunit" Version="2.3.1" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.3.1" />
-    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Storage" Version="5.0.0-beta.5" />
+    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Storage" Version="5.0.0-beta.3" />
```
chiangvincent: This will ultimately be updated right? Which prod version will this PR have dependency on for `Microsoft.Azure.WebJobs.Extensions.Storage` ?

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
```diff
@@ -0,0 +1,379 @@
...
+            {
+                // Adding host.json here
+                config.AddJsonStream(new MemoryStream(Encoding.UTF8.GetBytes(hostJson)));
+
+                // Adding app settings
+                config.AddInMemoryCollection(new Dictionary<string, string>
+                {
+                    { "app_setting1", "value1" },
+                    { "app_setting2", "value2"},
+                    { $"{Constants.HostingConfigSectionName}:Microsoft.Azure.WebJobs.Host.EndToEndTests", "1" }
```
chiangvincent, line 78: We should also test the case where the app setting is defined with double underscore __ instead of colon : (hierarchically defined app settings general case)

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
```diff
@@ -0,0 +1,379 @@
...
+            int targetWorkerCount = (int)Char.GetNumericValue(TargetScalerDescriptor.FunctionId[TargetScalerDescriptor.FunctionId.Length - 1]) + 1;
+
+            return Task.FromResult(new TargetScalerResult()
+            {
+                TargetWorkerCount = targetWorkerCount
+            });
+        }
+    }
+
+    internal class FakeAzureComponentFactory
```
chiangvincent: Remove if not needed

chiangvincent: on a related note, how is Managed Identity supported in this new model? I remember in the old version of the API the AzureComponentFactory was injected

chiangvincent (dismissed):

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleManager.cs
```diff
@@ -0,0 +1,300 @@
...
+                scaleMonitorsToSample.AddRange(scaleMonitors);
+            }
+
+            return (scaleMonitorsToSample, targetScalersToSample);
+        }
+
+        internal static ScaleVote GetAggregateScaleVote(IEnumerable<ScaleVote> votes, ScaleStatusContext context, ILogger logger)
+        {
+            ScaleVote vote = ScaleVote.None;
+            if (votes.Any())
```
chiangvincent, line 254: A logical simplification: just take the max scale vote of all the votes. I believe this logically equivalent to scale out if any workers are voting to scale out, and scale in iff all votes are scale in.

src/Microsoft.Azure.WebJobs.Host/Scale/TriggerMetadata.cs
alrod: WebJobs should be agnostic and do not know about trigger types.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
alrod: This should be in the extension, see SB PR:
https://github.com/Azure/azure-sdk-for-net/pull/34557

src/Microsoft.Azure.WebJobs.Host.Storage/StorageServiceCollectionExtensions.cs
alrod: We will set AzureWebbJobsStorage setting for the ScaleHost

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/WebJobs.Host.FunctionalTests.csproj
alrod: PROD version will `Microsoft.Azure.WebJobs.Extensions.Storage.5.0.0-beta.3`. This change is just points the tests dll to the future prod version. Referencing to `Microsoft.Azure.WebJobs.Extensions.Storage.5.0.0-beta.5`  does not make change as current version is `Microsoft.Azure.WebJobs.Extensions.Storage.5.0.0-beta.2`

src/Microsoft.Azure.WebJobs.Host/Scale/TriggerMetadata.cs
alrod: WebJobs should be agnostic to an extension

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
alrod: we are using `TriggerData.Properties` to pass `AzureComponentFactory`:
https://github.com/Azure/azure-sdk-for-net/pull/34557

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleManager.cs
alrod: This logic was migrated from RDS and it looks right to me as it also rely one worker count and if there is any votes. I do not see why we should change it.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
alrod: I think  "__"  shoud be normalized to ":" by ScalerHost creator (SC).

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
alrod: we are using `TriggerData.Properties` to pass `AzureComponentFactory`:
https://github.com/Azure/azure-sdk-for-net/pull/34557

mathewc (approved):

src/Microsoft.Azure.WebJobs.Host/Scale/TriggerMetadata.cs
```diff
@@ -0,0 +1,52 @@
...
+        private readonly string _type;
+
+        public TriggerMetadata(JObject metadata)
+            : this(metadata, new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase))
+        {
+        }
+
+        public TriggerMetadata(JObject metadata, IDictionary<string, object> properties)
+        {
+            Metadata = metadata;
```
mathewc: Should add null checks

src/Microsoft.Azure.WebJobs.Host.Storage/StorageServiceCollectionExtensions.cs
chiangvincent: Can the cx define a custom connection for storage account (i.e. not AzureWebJobsStorage)? If so, how would we handle that case?

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsServiceCollectionExtensions.cs
```diff
@@ -87,8 +88,9 @@ public static IWebJobsBuilder AddWebJobs(this IServiceCollection services, Actio
             services.TryAddSingleton<IFunctionInstanceLogger, FunctionInstanceLogger>();
             services.TryAddSingleton<IHostInstanceLogger, NullHostInstanceLogger>();
             services.TryAddSingleton<IDistributedLockManager, InMemoryDistributedLockManager>();
-            services.TryAddSingleton<IScaleMonitorManager, ScaleMonitorManager>();
-            services.TryAddSingleton<ITargetScalerManager, TargetScalerManager>();
+
+            services.AddCommonScaleServices();
+            services.AddSingleton<IScaleMetricsRepository, NullScaleMetricsRepository>();
```
fabiocav: Consider using `TryAddSingleton` here instead.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsServiceCollectionExtensions.cs
```diff
@@ -159,5 +161,54 @@ public static IWebJobsBuilder AddWebJobs(this IServiceCollection services, Actio
...
+            services.AddCommonScaleServices();
+            services.TryAddSingleton<IPrimaryHostStateProvider>(new PrimaryHostStateProvider() { IsPrimary = true });
+
+            if (configure != null)
+            {
+                services.Configure(configure);
+            }
+
+            // Options logging
+            services.AddTransient(typeof(OptionsFactory<>));
```
fabiocav: Why are we adding these options services as part of the scale logic? In the context of Functions, this is already handled, but for WebJobs, this shouldn't be tied to this logic.

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleManager.cs
```diff
@@ -0,0 +1,297 @@
...
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Host.Scale
+{
+    /// <summary>
+    /// Manages scale monitoring operations.
+    /// </summary>
+    public class ScaleManager
```
fabiocav: Does this type need to be public? 

Should we define (and register) an interface for this?

src/Microsoft.Azure.WebJobs.Host/Scale/TriggerMetadata.cs
fabiocav: Address this?

fabiocav (commented):
Quick review on some of the DI changes. Sharing this, but will continue to iterate.

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleManager.cs
mathewc: We had defined an interface, but I just made it a public service type, similar to the decision I made for ConcurrencyManager. The rational is that it makes it easier to add new operations in the future w/o breaking an interface.

If we went the interface route, we wouldn't have an IScaleManager interface because that implies a larger capability set than just GetStatus. We'd end up defining just an IHostScaleStatusProvider or IScaleStatusProvider or some such interface for just that operation. I considered this.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsServiceCollectionExtensions.cs
mathewc: Yeah, it's not really. AddWebJobsScale configures a host as a ScaleHost, and we want the options logging stuff as well to log ScaleOptions. So it's common registrations that we want to do for normal WebJobs hosts as well as scaler hosts. Could be factored into a shared method like I did for AddCommonScaleServices.

src/Microsoft.Azure.WebJobs.Host.Storage/StorageServiceCollectionExtensions.cs
```diff
@@ -36,6 +36,13 @@ public static void AddAzureStorageCoreServices(this IServiceCollection services)
             services.AddSingleton<IConcurrencyStatusRepository, BlobStorageConcurrencyStatusRepository>();
         }
 
+        public static void AddAzureStorageScaleServices(this IServiceCollection services)
```
brettsam, line 39: Is this meant to ever be called by a customer?

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsHostBuilderExtensions.cs
```diff
@@ -135,6 +136,26 @@ public static IHostBuilder ConfigureWebJobs(this IHostBuilder builder, Action<Ho
             return builder;
         }
 
+        /// <summary>
+        /// Configures the specified <see cref="IHostBuilder"/> as a scale manager host.
```
brettsam: Is that something a customer will do? Or do we need to mention this is only for internal infrastructure or something like that?

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsServiceCollectionExtensions.cs
brettsam: can you call `services.AddOptions()` instead? That conditionally registers everything you'd need for Options -- and if they change something in future versions, it'd continue to work.

https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Options/src/OptionsServiceCollectionExtensions.cs#L22

src/Microsoft.Azure.WebJobs.Host.Storage/StorageServiceCollectionExtensions.cs
alrod: Yes, this will be called from ScaleController, see example here:
https://github.com/Azure/azure-webjobs-sdk/blob/scaler-apis/test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs#L84

Function host is not supposed to call this.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsHostBuilderExtensions.cs
alrod: yes, this will be called from `ScaleController and a webjobs customer is not suppose to create a "scale manager host". I I extended the description.

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsServiceCollectionExtensions.cs
alrod: This was copy pasted from [here](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsServiceCollectionExtensions.cs#L119). Do we want to change this in the both places: `services.AddTransient(typeof(OptionsFactory<>));` -> `services.AddOptions()` ?

alrod: Bunch of the tests are broken after the change:
https://github.com/Azure/azure-webjobs-sdk/pull/2954/checks?check_run_id=11964665844

Do we want to fix them all revert `services.AddOptions()`?

src/Microsoft.Azure.WebJobs.Host/Scale/IScaleStatusProvider.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Host.Scale
+{
+    /// <summary>
+    /// Interface for providing a scale status
```
mathewc: "Interface for providing aggregate scale status across all functions being monitored by the host."

brettsam (approved):

build/common.props
```diff
@@ -1,8 +1,8 @@
 <Project>
   <PropertyGroup>
     <!-- Packages can have independent versions and only increment when released -->
-    <Version>3.0.36$(VersionSuffix)</Version>
-    <HostStorageVersion>5.0.0-beta.2$(VersionSuffix)</HostStorageVersion>
+    <Version>3.0.37$(VersionSuffix)</Version>
+    <HostStorageVersion>5.0.0-beta.3$(VersionSuffix)</HostStorageVersion>
```
fabiocav: This package is about to be published as a GA (non-beta) release, OK to add the changes here?

fabiocav (approved):

pragnagopa:
@alrod - Please update the PR description and would be great if you can add PRs links to extension SDK updates and  Functions Host

build/common.props
alrod: Updated to `<HostStorageVersion>5.0.0$(VersionSuffix)</HostStorageVersion>`

PR #2958, opened 2023-02-16
Scaler API updates

mathewc:


src/Microsoft.Azure.WebJobs.Host.Storage/StorageServiceCollectionExtensions.cs
```diff
@@ -31,17 +31,12 @@ public static void AddAzureStorageCoreServices(this IServiceCollection services)
             // Adds necessary Azure services to create clients
             services.AddAzureClientsCore();
 
-            services.AddAzureStorageScaleServices(null);
+            services.AddAzureStorageScaleServices();
         }
 
-        public static void AddAzureStorageScaleServices(this IServiceCollection services, Action<JobHostInternalStorageOptions> configure)
```
mathewc, line 37: Not sure why scaler hosts would need to configure this?

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsHostBuilderExtensions.cs
```diff
@@ -160,6 +161,7 @@ public static IHostBuilder ConfigureWebJobs(this IHostBuilder builder, Action<Ho
                     services.Configure(configureScaleOptions);
                 }
 
+                services.AddFeatureManagement();
```
mathewc, line 164: Suggest we use the feature APIs built into .NET core https://learn.microsoft.com/en-us/azure/azure-app-configuration/use-feature-flags-dotnet-core?tabs=core5x. It's a clean model that composes with the configuration system. Heck, we might even be able to replace our custom stuff in the functions host with this.

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleManager.cs
```diff
@@ -224,8 +223,8 @@ public async Task<ScaleStatus> GetScaleStatusAsync(ScaleStatusContext context)
                     if (!_targetScalersInError.Contains(scalerUniqueId))
                     {
                         string assemblyName = GetAssemblyName(scaler.GetType());
-                        string flag = _hostingConfig.Value.GetFeature(assemblyName);
-                        if (flag == "1")
+                        bool featureEnabled = await _featureManager.IsEnabledAsync(assemblyName);
```
mathewc, line 226: We should find a way to fix the feature names here, so for the code checked into the SDK meaningful feature names are used, rather than assembly name. Perhaps the scaler host can do that modification when it adds the feature flags to config?

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleMonitorService.cs
```diff
@@ -73,7 +73,8 @@ private async Task TakeMetricsSamplesAsync()
         {
             try
             {
-                (_scaleManager as ScaleManager).GetScalersToSample(out List<IScaleMonitor> scaleMonitorsToProcess, out List<ITargetScaler> targetScalersToProcess);
+                // TODO: Need to fix this cast to ScaleManager
+                var (scaleMonitorsToProcess, targetScalersToSample) = await (_scaleManager as ScaleManager).GetScalersToSampleAsync();
```
mathewc, line 77: We need to find a way to remove this cast ideally

src/Microsoft.Azure.WebJobs.Host/Scale/TriggerMetadata.cs
```diff
@@ -15,30 +11,31 @@ namespace Microsoft.Azure.WebJobs.Host.Scale
     /// </summary>
     public class TriggerMetadata
     {
-        private readonly List<object> _properties;
+        public TriggerMetadata(string functionName, JObject metadata)
```
mathewc, line 14: I revamped the trigger metadata / provider APIs to make them cleaner. We may need to iterate more.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
```diff
@@ -11,17 +17,11 @@
...
-using System.IO;
-using System.Linq;
-using System.Text;
-using System.Threading;
-using System.Threading.Tasks;
 using Xunit;
 
 namespace Microsoft.Azure.WebJobs.Host.EndToEndTests.Scale
 {
+    [Trait(TestTraits.CategoryTraitName, TestTraits.ScaleMonitoring)]
```
mathewc, line 24: Using traits like this make it easy to run just the tests you care about by feature in VS :)

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
```diff
@@ -33,6 +33,13 @@ public class ScaleHostEndToEndTests
         [InlineData(true)]
         public async Task ScaleManager_GetScaleStatusAsync_ReturnsExpected(bool tbsEnabled)
         {
+            var triggerMetadata = new List<TriggerMetadata>()
+            {
+                new TriggerMetadata(Function1Name, new JObject { { "type", "testExtensionATrigger" } }),
+                new TriggerMetadata(Function2Name, new JObject { { "type", "testExtensionATrigger" } }),
+                new TriggerMetadata(Function3Name, new JObject { { "type", "testExtensionBTrigger" } }, new Dictionary<string, object> { { "foo", "bar" } }),
```
mathewc, line 40: I'd like to discuss more why this Properties bag is needed for triggers, but for now assuming it is, I retain it in the API.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
```diff
@@ -88,58 +83,35 @@ public async Task ScaleManager_GetScaleStatusAsync_ReturnsExpected(bool tbsEnabl
...
 
                 // Adding app settings
-                config.AddInMemoryCollection(new List<KeyValuePair<string, string>>()
+                config.AddInMemoryCollection(new Dictionary<string, string>
                 {
-                    new KeyValuePair<string, string>("app_setting1", "value1"),
-                    new KeyValuePair<string, string>("app_setting2", "value2")
+                    { "app_setting1", "value1" },
+                    { "app_setting2", "value2"},
+                    { "FeatureManagement:Microsoft.Azure.WebJobs.Host.EndToEndTests", "true" }
```
mathewc, line 93: Here's how a feature flag can be added. We can also control the root "FeatureManagement" name - we could have it be "WebJobsFeatures" or something.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
```diff
@@ -88,58 +83,35 @@ public async Task ScaleManager_GetScaleStatusAsync_ReturnsExpected(bool tbsEnabl
...
-                // replace IConcurrencyStatusRepository for the test
-                services.AddSingleton<IConcurrencyStatusRepository, TestConcurrencyStatusRepository>();
-                // Adding services for scale host.
-                services.AddSingleton(triggerMetadataProvider);
-                services.AddSingleton<IScaleMetricsRepository, TestScaleMetricsRepository>();
+                services.AddAzureStorageScaleServices();
 
-                // Specific services registrations for scale host
-                services.AddSingleton(triggerMetadataProvider);
+                services.AddSingleton<IConcurrencyStatusRepository, TestConcurrencyStatusRepository>();
```
mathewc, line 101: Let's strive to keep the builder usage clean here - I keep cleaning it up, then it gets bloated again.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
alrod: This is needed to pass `AzureComponentFactory `for example (maybe some other properties in the future). We can not define `AzureComponentFactory` in `TriggerMetadata` as it would bring another nuget package reference and we do no want that.

test/Microsoft.Azure.WebJobs.Host.UnitTests/PublicSurfaceTests.cs
```diff
@@ -16,6 +14,7 @@ namespace Microsoft.Azure.WebJobs.Host.UnitTests
     /// fail any time new dependencies or public surface area are added, ensuring
     /// we review such additions carefully.
     /// </summary>
+    [Trait(TestTraits.CategoryTraitName, TestTraits.ScaleMonitoring)]
```
alrod, line 17: Remove or add another category name.

src/Microsoft.Azure.WebJobs.Host.Storage/StorageServiceCollectionExtensions.cs
alrod: This is used in https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host.Storage/AzureStorageProvider.cs#L41. as alternative to `AzureWebJobsStorage`

But I have never seen it was used, is it some legacy thing?

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleManager.cs
alrod: TBS is enabling right now, SC also relies on the same flag name(assembly name).

Changing the flag name will cause the TBS delay (we did commit ETA as mid Feb):
1. SC needs to be updated to use the new flag name.
2. Stamps configs needs to be updated - some stamps already have TBS.
3. Function host with new web jobs sdk reference needs to be releases for TBS+RTS.

src/Microsoft.Azure.WebJobs.Host/Scale/TriggerMetadataProvider.cs
```diff
@@ -1,54 +1,24 @@
...
+        private readonly IEnumerable<TriggerMetadata> _triggerMetadata;
 
-        public TriggerMetadataProvider(JArray fucntionMetadata, Dictionary<string, IEnumerable<object>> metadataProperties)
+        public TriggerMetadataProvider(IEnumerable<TriggerMetadata> triggerMetadata)
         {
-            _fucntionMetadata = fucntionMetadata;
-            _metadataProperties = metadataProperties;
-        }
-
-        public IEnumerable<TriggerMetadata> GetTriggersMetadata(string triggerType)
```
alrod, line 23: Do we want to filer by trigger type?

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
alrod: The root name should be "HostingConfig" as function host already use it.

mathewc: We don't want a name like "HostingConfig" in the SDK layer. We should use a name that makes sense from the WebJobs SDK point of view. Functions host can override the source as needed when registering its provider.

src/Microsoft.Azure.WebJobs.Host/Scale/TriggerMetadataProvider.cs
mathewc: No, we really want the metadata provider interface to do a single job - return the metadata. Consumers can filter however they need.

test/Microsoft.Azure.WebJobs.Host.UnitTests/PublicSurfaceTests.cs
mathewc: I added this here because when doing scaler work, we're changing the surface area and want to run these tests. Note that tests can have multiple traits so this isn't a problem.

PR #2962, opened 2023-02-28
Update ReleaseAsync call to not pass the function cancellation token when there is an exception

### LINKED ISSUE #2961: Singleton lock is not released when function cancellation token has been cancelled
When drain mode is called or we shutdown the host and stop all the listeners, the function invocation cancellation token gets cancelled. As a result, in the scenario where a function is using a `[Singleton]` attribute we end up trying to release the singleton lock twice. This leads to several exceptions:

1. `System.Threading.Tasks.TaskCanceledException : The operation was canceled.
    at Azure.Core.CancellationHelper.ThrowOperationCanceledException(Exception innerException, CancellationToken cancellationToken)`

2. `System.InvalidOperationException : The timer has already been stopped`

Example of what is happening:

- Function is invoked
- /drain is called -> “DrainMode mode enabled”
- “Calling StopAsync on all registered listeners”
- “Stopping the listener…”
- After the function finishes executing:
    - Because we are stopping the listeners/drain mode is enabled, we cancel the `functionCancellationTokenSource`
    - We eventually call `ReleaseLockAsync` to release the singleton lock and pass in the `functionCancellationToken`
        - This will stop the lease renewal timer and try to release the blob lock
    - ReleaseLockAsync throws an exception in Azure.Core (Azure.Storage.Blobs.BlobRestClient.ReleaseLeaseAsync) because the cancellation token has been cancelled (`TaskCanceledException`)
    - We catch that exception in `ExecuteWithLoggingAsync()`, and because there is an exception we have logic in place to release the singleton lock here as well (if the singleton is still held)
    - We do this via the same `ReleaseLockAsync` method that a) stops the lease renewal timer and b) releases the blob lock
    - At this point the renewal timer has already been stopped and so we fail with exception “The timer has already been stopped”
- “Stopped the listener…”
- “Call to StopAsync complete…”
-  << we never end up releasing the singleton lock >>

This leads to a) the singleton lock not being released and b) an exception that surfaces to the customer function, failing their function execution

liliankasem:
resolves #2961

In the scenario where a function is using a `[Singleton]` attribute and the function cancellation token source is cancelled, we get an exception from Azure.Core when trying to release the singleton lock blob via `ReleaseLockAsync`. This is because we are passing in the function cancellation token through and it has already been cancelled. As a result, we end up calling the `ReleaseLockAsync` again for a second time which leads to another exception as we try to stop the lease renewal timer again.

I don't see any benefit to us passing our function cancellation token to the blob clients `ReleaseLockAsync` function. I feel in any scenario, we will want to release that singleton lock. I have updated the code to not pass in a cancellation token here. 

Open to better suggestions on how to address this.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonManager.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonManager.cs
@@ -166,7 +166,7 @@ namespace Microsoft.Azure.WebJobs.Host
                 await handle.LeaseRenewalTimer.StopAsync(cancellationToken);
             }
 
-            await _lockManager.ReleaseLockAsync(handle.InnerLock, CancellationToken.None);
+            await _lockManager.ReleaseLockAsync(handle.InnerLock, cancellationToken);
 
             string msg = string.Format(CultureInfo.InvariantCulture, "Singleton lock released ({0})", handle.InnerLock.LockId);
             _logger?.LogDebug(msg);
--- a/src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonManager.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonManager.cs
@@ -166,7 +166,7 @@ namespace Microsoft.Azure.WebJobs.Host
                 await handle.LeaseRenewalTimer.StopAsync(cancellationToken);
             }
 
-            await _lockManager.ReleaseLockAsync(handle.InnerLock, cancellationToken);
+            await _lockManager.ReleaseLockAsync(handle.InnerLock, CancellationToken.None);
 
             string msg = string.Format(CultureInfo.InvariantCulture, "Singleton lock released ({0})", handle.InnerLock.LockId);
             _logger?.LogDebug(msg);
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
@@ -345,7 +345,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                             SingletonLock singleton = await parameterHelper.GetSingletonLockAsync();
                             if (singleton != null && singleton.IsHeld)
                             {
-                                await singleton.ReleaseAsync(cancellationToken);
+                                await singleton.ReleaseAsync(CancellationToken.None);
                             }
                         }
 
--- a/src/Microsoft.Azure.WebJobs.Host/Timers/TaskSeriesTimer.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Timers/TaskSeriesTimer.cs
@@ -72,11 +72,7 @@ namespace Microsoft.Azure.WebJobs.Host.Timers
                 throw new InvalidOperationException("The timer has already been stopped.");
             }
 
-            if (!_cancellationTokenSource.IsCancellationRequested)
-            {
-                _cancellationTokenSource.Cancel();
-            }
-
+            Cancel();
             return StopAsyncCore(cancellationToken);
         }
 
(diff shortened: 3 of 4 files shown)


src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonManager.cs
```diff
@@ -166,7 +166,7 @@ public async virtual Task ReleaseLockAsync(RenewableLockHandle handle, Cancellat
                 await handle.LeaseRenewalTimer.StopAsync(cancellationToken);
             }
 
-            await _lockManager.ReleaseLockAsync(handle.InnerLock, cancellationToken);
+            await _lockManager.ReleaseLockAsync(handle.InnerLock, CancellationToken.None);
```
mathewc: I'm wondering if this is the right layer to make this fix. From the standpoint of the isolated SingletonManager class, it makes sense that its operations take a token and allow the caller to cancel those operations. It's up to the caller to decide whether they want to support cancellation, or to pass CancellationToken.None in on the call.

So I think we should look up the call stack here - perhaps it's at the point where we're calling this method that we need to pass None?

liliankasem: @mathewc that's a good point. A layer up from here is [SingletonLock](https://github.com/Azure/azure-webjobs-sdk/blob/aeb7c0c6a4e3dc72d0528b75fa5e797e453cc6dd/src/Microsoft.Azure.WebJobs.Host/Singleton/SingletonLock.cs#L100) which also doesn't feel like the right place for it; but a layer up from that is where we have an exception and decide to release the lock in [FunctionExecutor](https://github.com/Azure/azure-webjobs-sdk/blob/aeb7c0c6a4e3dc72d0528b75fa5e797e453cc6dd/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs#L348) and I think that feels like the right point to pass in `CancellationToken.None`. I've updated the code to have that change - let me know what you think!

mathewc (approved):

cyptus:
@liliankasem is there any workaround we can implement until this patch is released and updated in our codebase? could we catch this kind of exception anywhere?

liliankasem:
> @liliankasem is there any workaround we can implement until this patch is released and updated in our codebase? could we catch this kind of exception anywhere?

Unfortunately no, there isn't a workaround. The exception will occur in the host and there isn't a way for customer code to handle that.

cyptus:
@liliankasem thank you. actually i managed to catch this exception by registering a custom implementation of `IQueueProcessorFactory`. An own `QueueProcessor` can now overwrite `CompleteProcessingMessageAsync`, which is called with this particular exception and handle it.

is this a safe approach?

```c#
protected override async Task CompleteProcessingMessageAsync(
    QueueMessage message,
    FunctionResult result,
{
    if (IsIrrelevantTimerException(result))
    {
        await DeleteMessageAsync(message, cancellationToken).ConfigureAwait(false);
    }
    else
    {
        await base.CompleteProcessingMessageAsync(message, result, cancellationToken);
    }
}

private static bool IsIrrelevantTimerException(FunctionResult result)
{
    // waiting for https://github.com/Azure/azure-webjobs-sdk/pull/2962
    
    return result.Succeeded == false &&
           result.Exception is InvalidOperationException operationException &&
           operationException.Message == "The timer has already been stopped." &&
           operationException.StackTrace?.Contains(
               "Microsoft.Azure.WebJobs.Host.Timers.TaskSeriesTimer.StopAsync"
           ) == true;
}
```

lazaro-ansaldi:
Hi @liliankasem ,

In which framework version was this fix included?
I'm using 4.19.0.20414 with Net6 and having the same issue with a [Singleton] queue-triggered function.

I really appreciate any help you can provide.

image

liliankasem:
> Hi @liliankasem ,
> 
> In which framework version was this fix included? I'm using 4.19.0.20414 with Net6 and having the same issue with a [Singleton] queue-triggered function.
> 
> I really appreciate any help you can provide.
> 
> image

Should be host version v4.21+

kylewarren:
Hello @liliankasem, are you all confident this is only with singletons? We are getting a similar error running our service bus triggers, all started in the past month, and seeing it on v3.0.37 as well.

PR #2966, opened 2023-03-10
Scaler API review updates

mathewc:


src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsHostBuilderExtensions.cs
```diff
@@ -152,26 +149,8 @@ public static IHostBuilder ConfigureWebJobs(this IHostBuilder builder, Action<Ho
         {
             builder.ConfigureServices((context, services) =>
             {
-                WebJobsBuilder webJobsBuilder = new WebJobsBuilder(services);
+                IWebJobsBuilder webJobsBuilder = services.AddWebJobsScale(configureScaleOptions);
```
mathewc, line 152: Moving this to follow the existing pattern

src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsServiceCollectionExtensions.cs
```diff
@@ -162,5 +161,54 @@ public static IWebJobsBuilder AddWebJobs(this IServiceCollection services, Actio
...
+            return builder;
+        }
+
+        /// <summary>
+        /// Adds scale services common to both normal WebJobs and ScaleHost scenarios.
+        /// </summary>
+        /// <param name="services"></param>
+        private static void AddCommonScaleServices(this IServiceCollection services)
+        {
+            services.AddOptions<ScaleOptions>()
```
mathewc, line 202: Centralizing common logic so we don't get copy/paste errors in the future

src/Microsoft.Azure.WebJobs.Host/Scale/AggregateScaleStatus.cs
```diff
@@ -0,0 +1,33 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Azure.WebJobs.Host.Scale
+{
+    /// <summary>
+    /// Represents the aggregate scale status across all functions being monitored by the host.
+    /// </summary>
+    public class AggregateScaleStatus
```
mathewc, line 11: Renamed this type and some of its members

src/Microsoft.Azure.WebJobs.Host/Scale/IScaleManager.cs
```diff
@@ -1,20 +0,0 @@
...
-// Licensed under the MIT License. See License.txt in the project root for license information.
-
-using System.Threading.Tasks;
-
-namespace Microsoft.Azure.WebJobs.Host.Scale
-{
-    /// <summary>
-    /// Provides scale status.
-    /// </summary>
-    public interface IScaleManager
```
mathewc, line 11: Issue was that this interface had a general name about scale management, but only exposed a single status API. Would have renamed the interface to reduce its scope to something like IHostScaleMonitor, or IHostScaleStatusProvider. However, I don't think we need an interface at all. Like we do for ConcurrencyManager, we can expose the service type directly which allows us to add things in the future.

src/Microsoft.Azure.WebJobs.Host/Scale/NullScaleMetricsRepository.cs
```diff
@@ -8,15 +8,16 @@ namespace Microsoft.Azure.WebJobs.Host.Scale
...
         public Task WriteMetricsAsync(IDictionary<IScaleMonitor, ScaleMetrics> monitorMetrics)
         {
             return Task.CompletedTask;
         }
 
         public Task<IDictionary<IScaleMonitor, IList<ScaleMetrics>>> ReadMetricsAsync(IEnumerable<IScaleMonitor> monitors)
         {
-            var dictionary = new Dictionary<IScaleManager, IList<ScaleMetrics>>();
-            return Task.FromResult((IDictionary<IScaleMonitor, IList<ScaleMetrics>>)dictionary);
+            return Task.FromResult((IDictionary<IScaleMonitor, IList<ScaleMetrics>>)_emptyMetrics);
```
mathewc, line 20: This was returning the wrong type (IScaleManager).

test/Microsoft.Azure.WebJobs.Host.UnitTests/PublicSurfaceTests.cs
```diff
@@ -14,6 +14,7 @@ namespace Microsoft.Azure.WebJobs.Host.UnitTests
     /// fail any time new dependencies or public surface area are added, ensuring
     /// we review such additions carefully.
     /// </summary>
+    [Trait(TestTraits.CategoryTraitName, TestTraits.ScaleMonitoring)]
```
mathewc, line 17: Keep this here so we can run these along with scale tests

PR #2968, opened 2023-03-15
Do not treat warnings as errors for WebJobs.Host.TestCommon.csproj

michaelpeng36:
Allows releases builds to complete even though `WebJobs.Host.TestCommon.csproj` has a reference to `Microsoft.Azure.WebJobs.Host.Storage 5.0.0-beta.2`


kashimiz (approved):
:shipit:

PR #2970, opened 2023-03-29
Adding InMemoryScaleMetricsRepository

alrod:
The implementation will be consumed in SC v3 and the extensions tests.
FunctionHost has the implementation based on azure storage blobs.


test/Microsoft.Azure.WebJobs.Host.UnitTests/Scale/InMemoryScaleMetricsRepositoryTest.cs
```diff
@@ -0,0 +1,145 @@
...
+            {
+                MetricsPurgeEnabled = purgeEnabled
+            };
+            FieldInfo fieldInfo = typeof(ScaleOptions).GetField("_scaleMetricsMaxAge", BindingFlags.NonPublic | BindingFlags.Instance);
+            fieldInfo.SetValue(options, TimeSpan.FromMilliseconds(maxAgeInMs));
+            var repository = new InMemoryScaleMetricsRepository(Options.Create(options), _loggerFactory);
+            var monitor1 = new TestScaleMonitor1();
+
+            // write metrics
+            Dictionary<IScaleMonitor, ScaleMetrics> metricsMap = new Dictionary<IScaleMonitor, ScaleMetrics>()
```
chiangvincent, line 133: Can you add a test case where there are multiple samples, and some are removed while some remain? Mainly just to test that we can iterate through a list of multiple metrics samples and remove at the same time, since that was the error last time:

```
foreach (ScaleMetrics metric in pair.Value)
{
    if (metric.Timestamp < threshold)
    {
        bool removed = _monitorMetrics.TryRemove(pair.Key, out _);
    }
}
```

chiangvincent (changes requested):

chiangvincent (approved):

PR #2978, opened 2023-05-09
Introduce RPC packages and extension points.

### LINKED ISSUE #2985: Add RPC extension packages and abstractions
https://github.com/Azure/azure-functions-host/issues/9235

This issue tracks the work to add the abstraction layers for enabling WebJob extensions to add their own Host <--> Worker RPC services. We will start small exposing only the ability to map gRPC services.

Due to the way WebJobs extensions are loaded into an isolated assembly load context, we cannot expose gRPC service mapping in a direct fashion. Instead, we need to split this into two assemblies:

1. `Microsoft.Azure.WebJobs.Host.Rpc`
    - referenced by: webjobs extensions (transitively) and Azure Functions host.
    - purpose: this assembly contains only the necessary types to bridge AspNetCore endpoint extensibility from registered webjobs extensions to the functions host.
    - reason: we cannot perform the `Grpc.AspNetCore.Server`'s `.MapGrpcService<T>()` call from within the Functions host default assembly load context. It **must** be performed within the webjobs extension load context, thus an extension point from an already unified assembly must be chosen. In this case, we will use `EndpointDataSource` from AspNetCore.
2. `Microsoft.Azure.WebJobs.Extensions.Rpc`
    - reference by: webjobs extensions (and **not** Azure functions host).
    - purpose: expose APIs to perform the actual extension RPC service mapping.
    - reason: we do not want to expose the `EndpointDataSource` extensibility pattern directly. It is too much boilerplate, and we instead want something that maps onto the existing `IWebJobsExtensionBuilder` experience.

jviau:
Part of https://github.com/Azure/azure-functions-host/issues/9235

resolves #2985

## Overview

This PR introduces extension points to enable the extension RPC scenario described in the issue above. The changes are as follows:

First, we introduce a package `Microsoft.Azure.WebJobs.Host.Rpc`. This a separate package is it brings in the AspNetCore framework, which I did not want to add to `Microsoft.Azure.WebJobs.Host` package. The goal of this package is to add a unified extension point for the Functions host and WebJobs extensions to use. However, WebJobs extensions are not to reference this package or use the types directly. This will be explained later. This package introduces a single type `WebJobsRpcEndpointDataSource`, which is the type the functions host will look for in the webjobs service container to discover extensions-registered RPC endpoints.

Next, we introduce another package `Microsoft.Azure.WebJobs.Extensions.Rpc`. This is not an independent extension unto itself, rather this is the package _other_ webjobs extensions will reference (and **not** `Microsoft.Azure.WebJobs.Host.Rpc`). While `Microsoft.Azure.WebJobs.Host.Rpc` will be referenced ultimately by both the functions gRPC host (directly) and WebJobs extensions (transitively through this package), this package will **only** be referenced by WebJobs extensions. The functions host must never have a reference to `Microsoft.Azure.WebJobs.Extensions.Rpc`. The reason for this is due to how WebJobs extensions are loaded in their own assembly load context and that gRPC assemblies are private. This means gRPC types in the Functions host do **not equal** the gRPC types a WebJobs extension loads - even if they reference the **exact same gRPC nuget package**. When we ultimately call `.MapGrpcServive<TService>` from `Grpc.AspnetCore.Server`, this call must be made completely in the WebJobs extension assembly load context. This also is the reason the `EndpointDataSource` was chosen as the common extension point between host and extensions - that type only relies on already unified assemblies.

### Package naming
I am open to changing the name of the two packages, or even moving the extension-specific package to https://github.com/Azure/azure-webjobs-sdk-extensions.

Some alternative naming:

`Microsoft.Azure.WebJobs.Host.Rpc` -> `Microsoft.Azure.WebJobs.Rpc.Core`
`Microsoft.Azure.WebJobs.Extensions.Rpc` -> `Microsoft.Azure.WebJobs.Rpc`

## TODO

- [x] Unit tests
- [x] Fix sln folder structure <-- noticed `WebJobs-WithAnalyzers.sln` is out of date already, so did not add anything to it. It can be addressed in another PR.


src/Microsoft.Azure.WebJobs.Host.Rpc/WebJobs.Host.Rpc.csproj
```diff
@@ -0,0 +1,21 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\..\build\common.props" />
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
```
jviau, line 5: This is intended only for out-of-proc worker scenarios. The functions host is `net6.0` in these cases, but do I want to multi-target this also for extensions not yet moved to `net6.0`? For example, Durable Functions host extension only targets `netcoreapp3.1` at the latest.

Should this target `netcoreapp3.1` also, or should extensions be required to upgrade considering `net6.0` is the oldest still in supported TFM (excluding netfx)?

src/Microsoft.Azure.WebJobs.Host.Rpc/WebJobs.Host.Rpc.csproj
```diff
@@ -0,0 +1,21 @@
...
+    <LangVersion>latest</LangVersion>
+    <PackageId>Microsoft.Azure.WebJobs.Host.Rpc</PackageId>
+    <AssemblyName>Microsoft.Azure.WebJobs.Host.Rpc</AssemblyName>
+    <RootNamespace>Microsoft.Azure.WebJobs.Host.Rpc</RootNamespace>
+    <Description>This package contains the runtime assemblies for Microsoft.Azure.WebJobs.Host.Rpc. For more information, please visit http://go.microsoft.com/fwlink/?LinkID=320971</Description>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.Azure.WebJobs.Host\WebJobs.Host.csproj" />
```
jviau, line 14: This p2p reference is actually not needed/used - but figured I would set the dependency chain in place now rather than later.

src/Microsoft.Azure.WebJobs.Extensions.Rpc/WebJobsExtensionBuilderRpcExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        /// <summary>
+        /// Maps a gRPC service as a host/worker RPC extension.
+        /// </summary>
+        /// <typeparam name="T">The type of gRPC service to register.</typeparam>
+        /// <param name="builder">The <see cref="IWebJobsExtensionBuilder" /> to add the gRPC service to.</param>
+        /// <param name="configure">Action to configure the gRPC extension.</param>
+        /// <returns>The <paramref name="builder" /> with gRPC service added.</returns>
+        /// <exception cref="ArgumentNullException">If <paramref name="builder" /> is null.</exception>
+        public static IWebJobsExtensionBuilder MapGrpcService<T>(
```
jviau: There is potential for confusion with this name - will extenders mistakenly assume this is for registering an externally callable gRPC service on the functions web host?

Is there anything that can be done to clarify this is for host <--> worker RPC only?

src/Microsoft.Azure.WebJobs.Extensions.Rpc/WebJobsExtensionBuilderRpcExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+            this IWebJobsExtensionBuilder builder, Action<GrpcServiceEndpointConventionBuilder> configure = null)
```
mattchenderson: Is the callback taking a GrpcServiceEndpointConventionBuilder really necessary? Wondering if we can trim down the API surface here.

src/Microsoft.Azure.WebJobs.Extensions.Rpc/Implementation/GrpcExtension.cs
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        /// <inheritdoc />
+        public void Apply(IEndpointRouteBuilder builder, ILogger logger)
+        {
+            ArgumentNullException.ThrowIfNull(builder);
+
+            logger?.GrpcServiceApplied(_extension.Name, typeof(TService).Name);
+            GrpcServiceEndpointConventionBuilder b = builder.MapGrpcService<TService>();
+            _configure?.Invoke(b);
```
mattchenderson: Do we need this? It seems like we could just have the endpoint conventions remain owned by us.

mattchenderson (commented):
Just some questions about the exposure of `GrpcServiceEndpointConventionBuilder`

src/Microsoft.Azure.WebJobs.Extensions.Rpc/WebJobsExtensionBuilderRpcExtensions.cs
mattchenderson: Would `MapWorkerGrpcService` or something like that make sense?

src/Microsoft.Azure.WebJobs.Extensions.Rpc/WebJobsExtensionBuilderRpcExtensions.cs
jviau: We can leave it off and only add it back if requested.

src/Microsoft.Azure.WebJobs.Host.Rpc/WebJobs.Host.Rpc.csproj
fabiocav: We can start here with `net6.0`. We'd need to make sure bundles are frozen on V3 to make sure we don't pick up those updates there, but unless we have a very good reason to expand, we should keep this scoped to `net6.0`

src/Microsoft.Azure.WebJobs.Extensions.Rpc/WebJobsExtensionBuilderRpcExtensions.cs
jviau: That will work great for now :). Thanks!

jviau:
Re-opening this PR as a branch off this repo instead of a fork.

PR #2986, opened 2023-05-25
Roranjan/ai managed identity

RohitRanjanMS:
Adding support for AAD Auth in the ApplicationInsights package. 

System Assigned Identity
`APPLICATIONINSIGHTS_AUTHENTICATION_STRING` = `Authorization=AAD`

User Assigned Identity
`APPLICATIONINSIGHTS_AUTHENTICATION_STRING` =
`Authorization=AAD;ClientId={​​Client ID of the user-assigned managed identity}​​`

**Steps to enable ManagedIdentity based ingestion**
- On the Functions App
>1. Enable `System assigned` identity or configure `User assigned` identity.
>2. Set `APPLICATIONINSIGHTS_AUTHENTICATION_STRING` environment variable.
- On the AI Resource
>1. Assign "Monitoring Metrics Publisher" role to Functions App identity.
>2. Disable "Local Authentication".

This will enable AAD auth for AI instrumentation in the Host, snapshot debugger, profiler and the language specific agents(e.g., [Java)](https://learn.microsoft.com/en-us/azure/azure-monitor/app/monitor-functions)


test/Microsoft.Azure.WebJobs.Host.UnitTests/Loggers/ApplicationInsightsConfigurationTests.cs
```diff
@@ -58,7 +60,7 @@ public void DependencyInjectionConfiguration_Configures_With_InstrumentationKey(
             using (var host = builder.Build())
             {
                 var config = host.Services.GetService<TelemetryConfiguration>();
-
+
```
jviau: nit: whitespace 
```suggestion

```

jviau:
To put the topics we discussed in the previous meeting regarding this more concretely:

I believe what @mattchenderson was suggesting is for us to add a package ref to `Microsoft.Extensions.Azure` and leverage `AzureComponentFactory.CreateTokenCredential` to get the token credential. This way you don't need to manage our own options structure or parsing. The primary unique thing we would need to add is a way to _turn off_ adding a credential at all. Maybe we could make it a subsection of the app insights option setting? We would look for a section `ApplicationInsights:Credential`, if it exists this means customer wants to use token credential. We expect the shape of that section to align with what `AzureComponentFactory.CreateTokenCredential` wants. If `ApplicationInsights:Credential` is not defined, then this means no `TokenCredential` will be used for app insights.

malsey:
We also need this feature..

RohitRanjanMS:
I will prioritize this feature and will have an update this week.

GABRIELNGBTUC:
Is there any progress on this PR?

andystaples (approved):
LGTM

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TokenCredentialOptions.cs
```diff
@@ -0,0 +1,123 @@
...
+                case AuthToken when string.IsNullOrWhiteSpace(ClientId):
+                    return new ManagedIdentityCredential();
+                case AuthToken when !string.IsNullOrWhiteSpace(ClientId):
+                    return new ManagedIdentityCredential(ClientId);
+                default:
+                    return null;
+            }
+        }
+
+        private IReadOnlyDictionary<string, string> ParseAuthenticationString(string value)
```
jviau: Instead of returning a dictionary, can this method directly assign the properties as it parses? Also recommend this does some validation to ensure a good connection string was provided (all the necessary values are present in the connection string).

If you refactor the constructor and static `Parse` method in my other suggestion, then the parsing can explicitly store the expected values in local variables, and then construct the options object with all the properties set (as in, don't use a dictionary - use local variables per value extracted from connection string).

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TokenCredentialOptions.cs
```diff
@@ -0,0 +1,123 @@
...
+
+namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
+{
+    public class TokenCredentialOptions
+    {
+        private const string AuthToken = "AAD";
+        private const string AuthAuthorizationKey = "Authorization";
+        private const string AuthClientIdKey = "ClientId";
+
+        public TokenCredentialOptions(string authenticationString)
```
jviau: nit: having a constructor which takes the connection string _and_ settable properties is not ideal (the property setters are never very useful). So either do one of two things:

1. Remove property setters
**OR**
2. Switch to a parameterless constructor and add a `public static TokenCredentialOptions Parse(string authenticationString)`
    - this way the auth string can be skipped entirely and a user directly constructs this options object (a bit more useful).

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
```diff
@@ -194,6 +200,16 @@ public string Format()
                 };
             }
 
+            JObject tokenCredentialOptions = null;
+            if (TokenCredentialOptions != null)
+            {
+                tokenCredentialOptions = new JObject
```
jviau, line 206: Do you think it's worth running this through sanitization incase a customer puts a full connection string in one of these properties?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
```diff
@@ -305,6 +306,13 @@ internal static LoggerFilterOptions CreateFilterOptions(LoggerFilterOptions regi
                 configuration.InstrumentationKey = options.InstrumentationKey;
             }
 
+            // Default is connection string based ingestion
+            var tokenCredential = options.TokenCredentialOptions?.CreateTokenCredential();
+            if (tokenCredential != null)
```
jviau: Can make this slightly more concise:
``` CSHarp
if (options.TokenCredentialOptions?.CreateTokenCredential() is TokenCredential credential)
{
    configuration.SetAzureTokenCredential(credential);
}
```

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TokenCredentialOptions.cs
```diff
@@ -0,0 +1,123 @@
...
+        /// Create an <see cref="TokenCredential"/> from an authentication string.
+        /// </summary>
+        /// <param name="value">
+        /// The authentication string containing semi-colon separated key=value tokens.</param>
+        /// <returns>New <see cref="ManagedIdentityCredential"/> from the authentication string.</returns>
+        public TokenCredential CreateTokenCredential()
+        {
+            switch (Authorization)
+            {
+                case AuthToken when string.IsNullOrWhiteSpace(ClientId):
```
jviau: this switch statement ultimately calls the same constructor in both cases

``` CSharp
public ManagedIdentityCredential(string clientId = null, TokenCredentialOptions options = null)
```

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TokenCredentialOptions.cs
```diff
@@ -0,0 +1,123 @@
...
+                    return new ManagedIdentityCredential();
+                case AuthToken when !string.IsNullOrWhiteSpace(ClientId):
+                    return new ManagedIdentityCredential(ClientId);
+                default:
+                    return null;
```
jviau: Is failing silently here a good idea? Might be better to throw.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TokenCredentialOptions.cs
```diff
@@ -0,0 +1,123 @@
...
+                if (tokens.ContainsKey(AuthClientIdKey))
+                {
+                    ClientId = tokens[AuthClientIdKey];
+                }
+            }
+        }
+        /// <summary>
+        /// The authentication kind.
+        /// </summary>
+        public string Authorization { get; set; }
```
jviau: This might not be necessary to store (for now). We only support a value of `AAD` correct? We can validate that during connection string parsing and throw if it is not `AAD`. Failing fast here will help expose bad configuration earlier to the developer.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
```diff
@@ -23,6 +23,12 @@ public class ApplicationInsightsLoggerOptions : IOptionsFormatter
         /// </summary>
         public string ConnectionString { get; set; }
 
+        /// <summary>
+        /// Gets or sets Application Insights Authentication Mode. If set, this will
+        /// take precedence over the default connection string based ingestion.
+        /// </summary>
+        public TokenCredentialOptions TokenCredentialOptions { get; set; }
```
jviau, line 30: How is this being set? If you are planning on reading the env variable for this, I imagine you will need to read the env variable, build, and set this as part of a `IConfigureOptions`

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TokenCredentialOptions.cs
```diff
@@ -0,0 +1,123 @@
...
+        /// </remarks>
+        public string ClientId { get; set; }
+
+        /// <summary>
+        /// Create an <see cref="TokenCredential"/> from an authentication string.
+        /// </summary>
+        /// <param name="value">
+        /// The authentication string containing semi-colon separated key=value tokens.</param>
+        /// <returns>New <see cref="ManagedIdentityCredential"/> from the authentication string.</returns>
+        public TokenCredential CreateTokenCredential()
```
jviau: nit: this can be `internal`. No need to expose this to public API just yet

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/WebJobs.Logging.ApplicationInsights.csproj
```diff
@@ -27,6 +27,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Azure.Identity" Version="1.4.0" />
```
jviau: Need to use `1.10.2`. All versions earlier have a severe vulnerability, and `1.10.3` has an incompatible dependency with functions host (why they rev'd a dependencies major version in a patch change is beyond me).

jviau (changes requested):

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TokenCredentialOptions.cs
RohitRanjanMS: Fixed.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
RohitRanjanMS: I have removed the Authorization property. ClientId is masked, I just need this for troubleshooting.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/TokenCredentialOptions.cs
RohitRanjanMS: Removed switch, now this will always return a TokenCredential.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
RohitRanjanMS: Setting this in the ScriptHostBuilderExtensions, along with the connectionstring and InstrumentationKey. This will be another PR.

jviau (approved):

fernando-almeida-sprinklr:
@RohitRanjanMS is there any reason why this wouldn't work for AKS? If it's meant to work can you explain what could have been missed that's not specified on the steps above. I've unsuccessfully tried to use both the clientId for a user-assigned identity and the identity assigned to the AKS cluster so far.

fernando-almeida-sprinklr:
in between I was able to get it to work. For AKS this seems to work only with one of the user assigned identities assigned to agentpool VMSS. Wouldn't it be possible to use a workload identity for this?

PR #2991, opened 2023-06-07
 Introduce RPC packages and extension points. 

### LINKED ISSUE #2985: Add RPC extension packages and abstractions

jviau:
Part of https://github.com/Azure/azure-functions-host/issues/9235

resolves #2985

## Overview

This PR introduces extension points to enable the extension RPC scenario described in the issue above. The changes are as follows:

First, we introduce a package `Microsoft.Azure.WebJobs.Host.Rpc`. This a separate package is it brings in the AspNetCore framework, which I did not want to add to `Microsoft.Azure.WebJobs.Host` package. The goal of this package is to add a unified extension point for the Functions host and WebJobs extensions to use. However, WebJobs extensions are not to reference this package or use the types directly. This will be explained later. This package introduces a single type `WebJobsRpcEndpointDataSource`, which is the type the functions host will look for in the webjobs service container to discover extensions-registered RPC endpoints.

Next, we introduce another package `Microsoft.Azure.WebJobs.Extensions.Rpc`. This is not an independent extension unto itself, rather this is the package _other_ webjobs extensions will reference (and **not** `Microsoft.Azure.WebJobs.Host.Rpc`). While `Microsoft.Azure.WebJobs.Host.Rpc` will be referenced ultimately by both the functions gRPC host (directly) and WebJobs extensions (transitively through this package), this package will **only** be referenced by WebJobs extensions. The functions host must never have a reference to `Microsoft.Azure.WebJobs.Extensions.Rpc`. The reason for this is due to how WebJobs extensions are loaded in their own assembly load context and that gRPC assemblies are private. This means gRPC types in the Functions host do **not equal** the gRPC types a WebJobs extension loads - even if they reference the **exact same gRPC nuget package**. When we ultimately call `.MapGrpcServive<TService>` from `Grpc.AspnetCore.Server`, this call must be made completely in the WebJobs extension assembly load context. This also is the reason the `EndpointDataSource` was chosen as the common extension point between host and extensions - that type only relies on already unified assemblies.

### Package naming
I am open to changing the name of the two packages, or even moving the extension-specific package to https://github.com/Azure/azure-webjobs-sdk-extensions.

Some alternative naming:

`Microsoft.Azure.WebJobs.Host.Rpc` -> `Microsoft.Azure.WebJobs.Rpc.Core`
`Microsoft.Azure.WebJobs.Extensions.Rpc` -> `Microsoft.Azure.WebJobs.Rpc`

## TODO

- [x] Unit tests
- [x] Fix sln folder structure <-- noticed `WebJobs-WithAnalyzers.sln` is out of date already, so did not add anything to it. It can be addressed in another PR.


build/common.props
```diff
@@ -5,7 +5,7 @@
     <HostStorageVersion>5.0.0$(VersionSuffix)</HostStorageVersion>
     <LoggingVersion>4.0.3$(VersionSuffix)</LoggingVersion>
     
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <!-- <TargetFramework>netstandard2.0</TargetFramework> -->
```
jviau: I pushed new commits to undo this, but GitHub is not updating this PR - very weird.
 
I'll do another PR immediately after this to undo these changes.

src/Microsoft.Azure.WebJobs.Extensions.Rpc/WebJobsExtensionBuilderRpcExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// <summary>
+        /// Maps a gRPC service as a host/worker RPC extension.
+        /// </summary>
+        /// <typeparam name="T">The type of gRPC service to register.</typeparam>
+        /// <param name="builder">The <see cref="IWebJobsExtensionBuilder" /> to add the gRPC service to.</param>
+        /// <returns>The <paramref name="builder" /> with gRPC service added.</returns>
+        /// <exception cref="ArgumentNullException">If <paramref name="builder" /> is null.</exception>
+        /// <remarks>
+        /// This gRPC service is <b>not</b> for external gRPC communication. It is only for RPC between the
+        /// out-of-prov worker and the host.
```
liliankasem:
```suggestion
        /// out-of-proc worker and the host.
```

src/Microsoft.Azure.WebJobs.Extensions.Rpc/Implementation/Logs.cs
```diff
@@ -0,0 +1,27 @@
...
+        [LoggerMessage(EventId = 1400, Level = LogLevel.Trace, Message = "Begin applying RPC host/worker extensions.")]
+        public static partial void ApplyRpcExtensionsBegin(this ILogger logger);
+
+        [LoggerMessage(EventId = 1401, Level = LogLevel.Trace, Message = "Done applying RPC host/worker extensions. {count} endpoints added.")]
+        public static partial void ApplyRpcExtensionsEnd(this ILogger logger, int count);
+
+        [LoggerMessage(EventId = 1402, Level = LogLevel.Error, Message = "Error applying RPC host/worker extensions.")]
+        public static partial void ApplyRpcExtensionsError(this ILogger logger, Exception ex);
+
+
```
liliankasem: nit: extra line

liliankasem (approved):
This all looks great to me, it's very clean - great work!! 

Regarding naming: I feel `WebJobs.Rpc.Core` is a better name. As for the other package, it will only be used by extensions right? So WebJobs.Extensions.Rpc feels right but WebJobs.Rpc is fine too

satvu:
For the second one I agree that `WebJobs.Extensions.Rpc` sounds right and this name seems more immediately obvious/relevant. For `Microsoft.Azure.WebJobs.Host.Rpc` I feel like I can be convinced either way.

jviau:
> Regarding naming: I feel WebJobs.Rpc.Core is a better name.

👍 switched to this name

PR #3001, opened 2023-07-17
Adding ability to enable filters for live metrics in ApplicationInsights logging options

surgupta-msft:
Issue related to PR - https://github.com/Azure/azure-functions-host/issues/8973

Writeup for the changes - https://github.com/Azure/azure-functions-host/issues/8973#issuecomment-1636192111

Adding a setting `EnableLiveMetricsFilters` that filters logs before they are sent to Live Metrics.
- When false, all logs are sent to Live Metrics, regardless of any log filter configuration. 
- When true, the logs are filtered before they are sent. Setting to true may yield better performance.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
@@ -112,6 +112,13 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
         public DependencyTrackingOptions DependencyTrackingOptions { get; set; }
 
         /// <summary>
+        /// Gets or sets a value that filters logs before they are sent to Live Metrics. False by default.
+        /// When false, all logs are sent to Live Metrics, regardless of any log filter configuration. When true,
+        /// the logs are filtered before they are sent. Setting to true may yield better performance.
+        /// </summary>
+        public bool EnableLiveMetricsFilters { get; set; } = false;
+
+        /// <summary>
         /// Gets or sets HTTP request collection options. 
         /// </summary>
         public HttpAutoCollectionOptions HttpAutoCollectionOptions { get; set; } = new HttpAutoCollectionOptions();
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
@@ -114,7 +114,7 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
         /// <summary>
         /// Gets or sets a value that filters logs before they are sent to Live Metrics. False by default.
         /// When false, all logs are sent to Live Metrics, regardless of any log filter configuration. When true,
-        /// the logs are filtered before they are sent. Setting to true may yield better performance.
+        /// the logs are filtered before they are sent.
         /// </summary>
         public bool EnableLiveMetricsFilters { get; set; } = false;
 
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsLoggingBuilderExtensions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsLoggingBuilderExtensions.cs
@@ -61,8 +61,17 @@ namespace Microsoft.Extensions.Logging
             builder.AddConfiguration();
             builder.Services.AddApplicationInsights(loggerOptionsConfiguration, additionalTelemetryConfiguration);
 
-            builder.Services.PostConfigure<LoggerFilterOptions>(o =>
+            builder.Services.AddOptions<LoggerFilterOptions>()
+                .PostConfigure<ApplicationInsightsLoggerOptions>((o, appInsightsOptions) =>
             {
+                // The custom filtering below is only needed if we are sending all logs to Live Metrics.
+                // If we are filtering (or not using Live Metrics) we don't need to do this.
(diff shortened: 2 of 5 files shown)


src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
```diff
@@ -340,8 +346,13 @@ internal static LoggerFilterOptions CreateFilterOptions(LoggerFilterOptions regi
                 {
                     quickPulseProcessor = new QuickPulseTelemetryProcessor(next);
```
brettsam, line 350: While we're in here... if LiveMetrics is disabled there's also no need to add this Processor. Can we exclude that as well? I suspect there'll be a slight perf bump from that too.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsLoggingBuilderExtensions.cs
```diff
@@ -61,8 +62,20 @@ public static class ApplicationInsightsLoggingBuilderExtensions
...
-            {
+            builder.Services.AddOptions<LoggerFilterOptions>()
+                .PostConfigure<IOptions<ApplicationInsightsLoggerOptions>>((o, appInsightsOptions) =>
+                {
+                // The custom filtering below is only needed if we are sending all logs to Live Metrics.
+                // If we are filtering (or not using Live Metrics) we don't need to do this.
+                if (appInsightsOptions != null)
+                {
+                    if (!appInsightsOptions.Value.EnableLiveMetrics ||
+                        appInsightsOptions.Value.EnableLiveMetricsFilters)
```
brettsam, line 73: I think we need to tweak this... if `EnableLiveMetrics` is false, we should just immediately return, no matter what `EnableLiveMetricsFilters` is. I *think* in this case (you should add a configuration test) if `EnableLiveMetrics` is false and `EnableLiveMetricsFilters` is false... we'd still create this custom filter, when we really shouldn't.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
```diff
@@ -111,6 +111,13 @@ public TimeSpan QuickPulseInitializationDelay
         /// </summary>
         public DependencyTrackingOptions DependencyTrackingOptions { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value that filters logs before they are sent to Live Metrics. False by default.
+        /// When false, all logs are sent to Live Metrics, regardless of any log filter configuration. When true,
+        /// the logs are filtered before they are sent. Setting to true may yield better performance.
```
brettsam: On second thought, let's remove the last sentence about performance b/c we've not actually tested it. I don't want us to promise something that we haven't tested.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsLoggingBuilderExtensions.cs
surgupta-msft: I feel we need the entire condition as it is.

_if EnableLiveMetrics is false, we should just immediately return, no matter what EnableLiveMetricsFilters is._

^ this is happening with current condition. If EnableLiveMetrics is false, first part of `||` will be always true and we will always return.

Second part of `||` is still needed for the scenario where EnableLiveMetrics is true but we want live metrics to be filtered (i.e. EnableLiveMetricsFilters  = true)

Let me know if I am missing anything.

brettsam: you're right :-) I was misreading!

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
surgupta-msft: Done

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
```diff
@@ -338,10 +344,18 @@ internal static LoggerFilterOptions CreateFilterOptions(LoggerFilterOptions regi
                 .Use((next) => new OperationFilteringTelemetryProcessor(next))
                 .Use((next) =>
                 {
-                    quickPulseProcessor = new QuickPulseTelemetryProcessor(next);
+                    if (options.EnableLiveMetrics)
```
surgupta-msft: @brettsam  Here I added the condition.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Loggers/ApplicationInsightsConfigurationTests.cs
```diff
@@ -679,6 +700,46 @@ public void DependencyInjectionConfiguration_ConfiguresActiveOnlyOnce()
...
+                    {
+                        o.InstrumentationKey = "some key";
+                        o.EnableLiveMetricsFilters = true;
+                    });
+                })
+                .Build())
+            {
+                var filterOptions = host.Services.GetService<IOptions<LoggerFilterOptions>>().Value;
+                var rule = SelectAppInsightsRule(filterOptions, "Category");
+                Assert.Equal(null, rule.Filter);
```
brettsam, line 719: nit: you can use `Assert.IsNull()` for these. No need to update this, but just for future reference :-)

brettsam (approved):

PR #3006, opened 2023-07-27
Enum TryParse speedup

andystaples:
Small improvement to a call to Enum.TryParse for LogLevel in a telemetry filter. Should make this call ~20x more performant and reduce memory use to 0b

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/LogLevelExtension.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/LogLevelExtension.cs
@@ -34,5 +34,41 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights.Extensions
                     return logLevel.ToString(CultureInfo.InvariantCulture);
             }
         }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/LogLevelExtension.cs
```diff
@@ -34,5 +34,41 @@ internal static string ToStringOptimized(this LogLevel logLevel)
                     return logLevel.ToString(CultureInfo.InvariantCulture);
             }
         }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal static bool TryParseOptimized(this string logLevel, out LogLevel level)
+        {
+            switch (logLevel)
```
RohitRanjanMS, line 41: Switch is case sensitive by default. Do we have to worry about case?

andystaples: Actually, I researched this. Enum.TryParse which we are using now is also case sensitive by default. I looked into some true case insensitive options, but they all 10x'd the time cost of the method, even some really cool dictionary methods. I could just add the all lower equivalents to the case statement, but I figure since this behavior is the same as the existing method, no need.

RohitRanjanMS (approved):

PR #3011, opened 2023-08-15
Adding IListenerDecorator support

mathewc:
Exposing a new decorator feature for layering of listener behaviors. I've refactored a couple of our existing platform listener behaviors (singleton and function listener error handling) on top of this new feature to prove it out.

The Logic Apps team will be using this feature to implement some new advanced features in their extension, requiring dynamic listener behaviors. The decorators feature will allow them to add that.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListenerDecorator.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListenerDecorator.cs
@@ -1,0 +1,27 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Host.Listeners
+{
+    internal class FunctionListenerDecorator : IListenerDecorator
+    {
+        private readonly ILoggerFactory _loggerFactory;
+        private readonly IOptions<JobHostOptions> _jobHostOptions;
+
+        public FunctionListenerDecorator(ILoggerFactory loggerFactory, IOptions<JobHostOptions> jobHostOptions)
+        {
+            _loggerFactory = loggerFactory;
+            _jobHostOptions = jobHostOptions;
+        }
+
+        public IListener Decorate(ListenerDecoratorContext context)
+        {
+            // wrap the listener with a function listener to handle exceptions
+            bool allowPartialHostStartup = _jobHostOptions.Value.AllowPartialHostStartup;
+            return new FunctionListener(context.Listener, context.FunctionDefinition.Descriptor, _loggerFactory, allowPartialHostStartup);
+        }
+    }
+}
--- a/src/Microsoft.Azure.WebJobs.Host/Listeners/HostListenerFactory.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Listeners/HostListenerFactory.cs
@@ -14,7 +14,6 @@ using Microsoft.Azure.WebJobs.Host.Scale;
 using Microsoft.Azure.WebJobs.Logging;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Options;
 
 namespace Microsoft.Azure.WebJobs.Host.Listeners
 {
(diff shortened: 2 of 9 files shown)


src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListenerDecorator.cs
```diff
@@ -0,0 +1,27 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.WebJobs.Host.Listeners
+{
+    internal class FunctionListenerDecorator : IListenerDecorator
```
mathewc, line 9: Previously this "decorator" and Singleton were both listener wrappers that were hardcoded in the listener pipeline. Refactoring these out into decorators.

test/Microsoft.Azure.WebJobs.Host.FunctionalTests/HostListenerFactoryTests.cs
```diff
@@ -25,17 +26,45 @@ namespace Microsoft.Azure.WebJobs.Host.UnitTests.Listeners
     public class HostListenerFactoryTests
     {
         private readonly IJobActivator _jobActivator;
-        private IConfiguration _configuration = new ConfigurationBuilder().Build();
+        private readonly IHost _testHost;
+        private IConfiguration _configuration;
 
         public HostListenerFactoryTests()
         {
-            var hostBuilder = new HostBuilder();
```
mathewc, line 32: I refactored these tests to use a proper builder because I need to test the ordering of the default decorators added by our services DI. The test below ensures that when a user adds their own decorators, they're ordered correctly with built-ins.

src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListenerDecorator.cs
```diff
@@ -0,0 +1,27 @@
...
+        private readonly IOptions<JobHostOptions> _jobHostOptions;
+
+        public FunctionListenerDecorator(ILoggerFactory loggerFactory, IOptions<JobHostOptions> jobHostOptions)
+        {
+            _loggerFactory = loggerFactory;
+            _jobHostOptions = jobHostOptions;
+        }
+
+        public IListener Decorate(ListenerDecoratorContext context)
+        {
```
apranaseth, line 21: Can it return a task? So any async operation we can do before decorating the listener.
From Logic Apps perspective, whether there is a need of dynamic listener gets decided at the host startup. For making this decision to create the dynamic listener, the workflow definition is loaded from the file system and checked if it has concurrency configuration. There is some I/O needed to evaluate the need of dynamic listener, so having an asynchronous method will help in this case.

mathewc: Hmmm. I think a better design would be for your workflow definition to be loaded earlier in the startup path and be made available via DI through a WorkflowOptions or other service that can be bound to. At that point it's already loaded an all your components can access the cached data w/o having to read the data again.

src/Microsoft.Azure.WebJobs.Host/Listeners/HostListenerFactory.cs
```diff
@@ -229,5 +219,37 @@ internal static bool IsDisabledByProvider(Type providerType, MethodInfo jobFunct
...
+
+            return listener;
+        }
+
+        private IListener ApplyDecorators(IEnumerable<IListenerDecorator> decorators, IListener listener, IFunctionDefinition functionDefinition, Type rootListenerType)
+        {
+            foreach (var decorator in decorators)
+            {
+                var context = new ListenerDecoratorContext(functionDefinition, rootListenerType, listener);
+                listener = decorator.Decorate(context);
```
brettsam, line 249: This all looks good to me. The only real thought I have is "how do we know if decorators are being applied?" -- from a debugging standpoint.

Could/should we log here that we're applying a decorator? Or add logs in each decorator to say that they're doing something?

I know we only have 2 and they've been built-in for years, but technically this opens some things up...

brettsam (approved):

src/Microsoft.Azure.WebJobs.Host/Listeners/HostListenerFactory.cs
mathewc: Great suggestion. I'll add a log statement

src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListenerDecorator.cs
fabiocav: 👍🏼

src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListenerDecorator.cs
apranaseth: I think we should be able to handle that another way. Fine with this being sync call.

apranaseth (approved):

src/Microsoft.Azure.WebJobs.Host/Listeners/HostListenerFactory.cs
fabiocav: +1 to this.

fabiocav (approved):

PR #3013, opened 2023-08-22
Add API for differentiating user and system scopes

jviau:
This new API is "internal" and only meant to be used by the functions host and extensions (in limited cases). They are marked `[Obsolete]` and in an `.Internal` namespace with doc to advise end users against using it.

The purpose of this API is to give us a way to identify if the current execution context is in a "system" or "user" scope. A system scope being it is executing system code, and user scope being user code. This is to improve our ability to distinguish the two and change logic based on that. One such example is to alter our logging behaviors, performing system logging only when in a system scope.


.github/policies/resourceManagement.yml
jviau, line 1: I think this is line endings? I kept undoing the change and it came back immediately.

brettsam (approved):

RohitRanjanMS (approved):

PR #3014, opened 2023-08-24
Functions - BlobTrigger container scan should not skip past changes made during the scan

AartBluestoke:
When there are continuations to the container scan, the later pages should not advance the timestamp past when the first page of the scan occurs.

Without this change an update to a file on the first page that occurs while the later API calls are occurring might be missed.

old process:
0. old threshold
1. write file v1. 
2. begin scan 
3. read file v1. 
4. write file v2, 
5. write occurs to another, 
6. fetch next batch of results from the container scan
7. find 'another, t5'
8. highlighted line of code sets the 'high water mark' to t5, skipping over , file v2 which wrote at t4.
9. container scan ends
10. next scan begins at t5, missing the update at t4

the correct 'high water mark' is the minimum of t5 and t2, when the scan began

new process:
0. old threshold
1. write file v1. 
2. begin scan 
3. find 'file v1, t1', set high water mark to v1
4. write file v2, 
5. write occurs to another, 
6. fetch next batch of results from the container scan
7. find 'another, t5'
8. the high water mark IS NOT updated
9. container scan ends
10. next scan begins at t1, finding the file written at t4.
11. the scan finds the file written at t5 again, but rejects it via ETAG duplication.


AartBluestoke:
@microsoft-github-policy-service agree

AartBluestoke:
PR raised in response to bug discussed in azure support ticket TrackingID#2308210030000318

src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/ScanBlobScanLogHybridPollingStrategy.cs
```diff
@@ -214,6 +214,7 @@ private void ThrowIfDisposed()
             // if starting the cycle, reset the sweep time
             if (continuationToken == null)
             {
+                containerScanInfo.CurrentScanBeginTime = DateTime.UtcNow;
```
AartBluestoke: is the best name here "Scan" or "Sweep"? (both words are used interchangeably in this code)

fabiocav:
@AartBluestoke apologies for the state of this PR. This extension has since moved to https://github.com/Azure/azure-sdk-for-net and a lot of changes have happened in this area. 

I'm closing this as a stale PR, but thank you for the contribution.

PR #3016, opened 2023-09-11
Add message id to the log in QueueProcessor.cs

yuanweilei:
Add message id to the log, when the message is moving to poison queue.
Since the message id will be changed after moved to poison queue.


mathewc:
What version of the storage extension are you using? Note that this code has moved to the Azure SDK repo here https://github.com/Azure/azure-sdk-for-net/blob/fd461551f8423a2e1dafc902baa9767755ae380d/sdk/storage/Microsoft.Azure.WebJobs.Extensions.Storage.Common/src/Shared/Queues/QueueProcessor.cs#L131 and all updates are shipping from that repo.

I agree with you that this would be a nice improvement, but this PR should be closed and the work done in the Azure SDK repo.

komayama:
Hi @mathewc ,

I have create new PR on `azure-sdk-for-net`.
https://github.com/Azure/azure-sdk-for-net/pull/39062

jviau:
@mathewc should we delete that extension then if it has moved?

PR #3018, opened 2023-09-15
Scaler API discovery from extension

alrod:


test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
```diff
@@ -5,21 +5,28 @@
...
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
+using Microsoft.VisualStudio.TestPlatform.ObjectModel.Utilities;
 using Newtonsoft.Json.Linq;
 using Xunit;
 using static Microsoft.Azure.WebJobs.Host.EndToEndTests.Scale.ScaleHostEndToEndTests;
 
+[assembly: WebJobsStartup(typeof(TestExtensionA), "testExtensionATrigger")]
```
alrod, line 27: `Name` property is used to store trigger type to call `IWebJobsTriggerStartup.Configure` with appropriate `TriggerMetadata`

PR #3022, opened 2023-09-21
Log "Canceled" function execution when there is a FunctionInvocationCanceledException

liliankasem:
//draft - I think this is breaking change but putting up a PR for design discussion with Host team.


liliankasem:
Revisit for v5+

PR #3026, opened 2023-10-12
Add option to log the URL query string for HTTP functions to LoggerOptions

andystaples:
Currently, parameters on queries to HTTP triggered functions are stripped from the request URL before the URL is logged in the Request Telemetry. This PR adds an option to host.json logging settings which may be set to true to disable this behavior, logging the entire request URL along with the parameters.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
@@ -124,9 +124,9 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
         public HttpAutoCollectionOptions HttpAutoCollectionOptions { get; set; } = new HttpAutoCollectionOptions();
 
         /// <summary>
-        /// Gets or sets a value that removes the query parameters from functions URLs. True by default.
-        /// When true, the query parameters are stripped off of the functions URL for logging HTTP trigger calls.
-        /// When false, query parameters are logged.
+        /// Gets or sets a value that removes the query parameters from functions URLs. False by default.
+        /// When false (default), the query parameters are stripped off of the functions URL for logging HTTP trigger calls.
+        /// When true, query parameters are logged.
         /// </summary>
         public bool EnableQueryStringTracing { get; set; } = false;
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
```diff
@@ -19,8 +19,9 @@ internal class WebJobsTelemetryInitializer : ITelemetryInitializer
         private static readonly string _currentProcessId = Process.GetCurrentProcess().Id.ToString();
         private readonly string _sdkVersion;
         private readonly string _roleInstanceName;
+        private readonly ApplicationInsightsLoggerOptions _options;
 
-        public WebJobsTelemetryInitializer(ISdkVersionProvider versionProvider, IRoleInstanceProvider roleInstanceProvider)
+        public WebJobsTelemetryInitializer(ISdkVersionProvider versionProvider, IRoleInstanceProvider roleInstanceProvider, ApplicationInsightsLoggerOptions options)
```
brettsam: This should use `IOptions<ApplicationInsightsLoggerOptions>` so it gets auto-injected. You won't need the factory above then.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
```diff
@@ -123,6 +123,13 @@ public TimeSpan QuickPulseInitializationDelay
         /// </summary>
         public HttpAutoCollectionOptions HttpAutoCollectionOptions { get; set; } = new HttpAutoCollectionOptions();
 
+        /// <summary>
+        /// Gets or sets a value that removes the query parameters from functions URLs. True by default.
+        /// When true, the query parameters are stripped off of the functions URL for logging HTTP trigger calls.
+        /// When false, query parameters are logged.
+        /// </summary>
+        public bool SanitizeRequestURL { get; set; } = true;
```
brettsam: I don't love this name... can it be something specific about the QueryString? Like `EnableQueryStringTracing` or something? (that's not great either... but `Sanitize` makes me think we're doing something to obscure the whole url rather than just touching the QueryString)

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Initializers/WebJobsTelemetryInitializer.cs
andystaples: Good call, fixed

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
andystaples: Agreed, having the option be a new opt-in feature rather than a new opt-out seems prudent. Renamed to EnableQueryStringTracing for lack of a better descriptor, set the default value to true, and flipped all the logic

RohitRanjanMS (approved):

PR #3027, opened 2023-10-16
Proposal for Scaler APIs discovery in extension

alrod:


test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
```diff
@@ -86,28 +93,38 @@ public async Task ScaleManager_GetScaleStatusAsync_ReturnsExpected(bool tbsEnabl
...
             {
-                if (t.Type == ATriggerType)
-                {
-                    hostBuilder.ConfigureTestExtensionAScale(t);
-                }
-                else
+                builder.UseHostId(hostId);
+
+                // Get the interface type you want to find
+                Type interfaceType = typeof(IWebJobsStartup2);
```
mathewc: Would be good to clean this code up into a pattern we can point people to to follow for actual ScaleHost applications. I think you can simply use the existing [UseExternalStartup](https://github.com/Azure/azure-webjobs-sdk/blob/8960a7294b854b65c2b416b23dfc2fcc80574725/src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsBuilderExtensions.cs#L206) method, passing in your own discoverer and correct context type? That way we leverage all the existing activation logic.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
```diff
@@ -20,6 +24,9 @@
 using Xunit;
 using static Microsoft.Azure.WebJobs.Host.EndToEndTests.Scale.ScaleHostEndToEndTests;
 
+[assembly: WebJobsStartup(typeof(TestExtensionA), "testExtensionATrigger")]
```
mathewc: Are you proposing we'd update all the existing extensions this way with an Name ctor param? I think we should use all the existing assembly attributes declared by extensions as is. If not, recommend using in this test what we expect extensions to use. I.e. I think all extensions are currently accepting the default value for Name.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
```diff
@@ -86,28 +93,38 @@ public async Task ScaleManager_GetScaleStatusAsync_ReturnsExpected(bool tbsEnabl
...
                 {
-                    hostBuilder.ConfigureTestExtensionBScale(t);
+                    var attributes = assembly.GetCustomAttributes<WebJobsStartupAttribute>();
+
+                    foreach (var attribute in attributes)
+                    {
+                        if (interfaceType.IsAssignableFrom(attribute.WebJobsStartupType))
+                        {
+                            // Filter out triggers needed to be be configured
+                            var triggersToConfigure = triggerMetadata.Where(x => x.Type == attribute.Name);
```
mathewc: If we accept the default Name from existing attribute definitions, are we sure these lookups will always match? I.e. we need to verify that the source of the TriggerMetadata.Type is the same as we're using here.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
```diff
@@ -233,27 +250,31 @@ public ITargetScaler GetTargetScaler()
...
     {
-        public static IHostBuilder ConfigureTestExtensionAScale(this IHostBuilder builder, TriggerMetadata triggerMetadata)
+        public void Configure(IWebJobsBuilder builder)
+        {
+        }
+
+        public void Configure(WebJobsBuilderContext context, IWebJobsBuilder builder)
         {
-            builder.ConfigureServices((context, services) =>
+            var scaleHostBuilderContext = context as ScaleHostBuilderContext;
```
mathewc: recommend doing a null check here which is the pattern extensions will follow, since their configure method can be called in 2 different ways.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
alrod: I add `ScaleHostBuilderContext.GetTriggersMetadata(string type)` instead.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
alrod: Yes, using [UseExternalStartup](https://github.com/Azure/azure-webjobs-sdk/blob/8960a7294b854b65c2b416b23dfc2fcc80574725/src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsBuilderExtensions.cs#L206) makes sense.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
alrod: I add ScaleHostBuilderContext.GetTriggersMetadata(string type) instead.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
```diff
@@ -233,27 +235,38 @@ public ITargetScaler GetTargetScaler()
...
-                    return new TestExtensionAScalerProvider(config, scaleOptions, triggerMetadata);
-                });
-            });
-
-            return builder;
+                    builder.Services.AddSingleton<IScaleMonitorProvider>(serviceProvider =>
+                    {
+                        IConfiguration config = serviceProvider.GetService<IConfiguration>();
+                        IOptions<ScaleOptions> scaleOptions = serviceProvider.GetService<IOptions<ScaleOptions>>();
+                        return new TestExtensionAScalerProvider(config, scaleOptions, triggerMetadata);
```
mathewc: I think we're really close to being able to clean all this up so scaler providers can be registered as a Type, rather than having to construct explicitly this way.

If we introduce a new `ITriggerMetadataProvider` interface, the host can register that with DI to provide the metadata, then this extension logic simply becomes the following, allowing DI to be fully used for construction:

```csharp
if (scaleHostBuilderContext != null)
{
    builder.Services.AddSingleton<IScaleMonitorProvider, TestExtensionAScalerProvider>();
    builder.Services.AddSingleton<ITargetScalerProvider, TestExtensionAScalerProvider>();
}
```

We'll then have to decide on ScaleHostBuilderContext since this means it wouldn't be used for trigger metadata. Need to see if there are any other properties that context type needs to convey.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
```diff
@@ -86,28 +94,22 @@ public async Task ScaleManager_GetScaleStatusAsync_ReturnsExpected(bool tbsEnabl
...
-                    scaleOptions.IsRuntimeScalingEnabled = true;
-                });
+            {
+                builder.UseHostId(hostId);
 
-            foreach (TriggerMetadata t in triggerMetadata)
+                Assembly assembly = Assembly.GetExecutingAssembly();
+
+                var scaleHostBuilderContext = new ScaleHostBuilderContext(triggerMetadata);
+                new DefaultStartupTypeLocator(assembly);
```
mathewc: ? you're newing up an unused type here

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
```diff
@@ -86,28 +94,22 @@ public async Task ScaleManager_GetScaleStatusAsync_ReturnsExpected(bool tbsEnabl
...
-                    scaleOptions.IsTargetScalingEnabled = tbsEnabled;
-                    scaleOptions.MetricsPurgeEnabled = false;
-                    scaleOptions.ScaleMetricsMaxAge = TimeSpan.FromMinutes(4);
-                    scaleOptions.IsRuntimeScalingEnabled = true;
-                });
+            {
+                builder.UseHostId(hostId);
 
-            foreach (TriggerMetadata t in triggerMetadata)
+                Assembly assembly = Assembly.GetExecutingAssembly();
```
mathewc: This isn't needed - DefaultStartupTypeLocator has a default ctor you can use

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
alrod: `IScaleMonitorProvider/ITargetScalerProvider` return singular `IScaleMonitor `and `ITargetScaler`. It looks like we need a some new interface to return collection of scalers in this case.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
alrod: The DefaultStartupTypeLocator  ctor returns `testhost` assembly but we need `Microsoft.Azure.WebJobs.Host.EndToEndTests`:
image

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
mathewc: My latest commit addresses this

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
```diff
@@ -84,34 +91,37 @@ public async Task ScaleManager_GetScaleStatusAsync_ReturnsExpected(bool tbsEnabl
...
+
+                var webJobsBuilderContext = new WebJobsBuilderContext
                 {
-                    hostBuilder.ConfigureTestExtensionBScale(t);
-                }
-            }
+                    Configuration = context.Configuration,
+                    EnvironmentName = context.HostingEnvironment.EnvironmentName
+                };
+                webJobsBuilderContext.Properties["IsAzureWebJobsScaleHost"] = true;
```
mathewc, line 105: Still considering the best way to flow this property. An alternative might be to add the "context" HostBuilderContext in this scope as a property on WebJobsBuilderContext. We could then leverage the Property collection of that type, and could also set this flag automatically in ConfigureWebJobsScale.

test/Microsoft.Azure.WebJobs.Host.EndToEndTests/Scale/ScaleHostEndToEndTests.cs
```diff
@@ -198,108 +211,185 @@ public Task<HostConcurrencySnapshot> ReadAsync(CancellationToken cancellationTok
...
-            private IOptions<ScaleOptions> _scaleOptions;
-            private IScaleMonitor _scaleMonitor;
-            private ITargetScaler _targetScaler;
-            private TriggerMetadata _triggerMetadata;
+        }
 
-            public TestExtensionScalerProvider(IConfiguration config, IOptions<ScaleOptions> scaleOptions, TriggerMetadata triggerMetadata)
+        public void Configure(WebJobsBuilderContext context, IWebJobsBuilder builder)
+        {
+            if (context.Properties.ContainsKey("IsAzureWebJobsScaleHost"))
```
mathewc, line 224: Still thinking what helper we could add to simplify this Properties check - e.g. an IsScaleHost extension method somewhere

mathewc:
@alrod Assuming this 1 year old draft PR is obsolete? Any work we need to do here?

PR #3032, opened 2023-10-24
Aligned OriginalFormatString with properties.

### LINKED ISSUE #2722: null logging entry when using SEQ
When using an azure function with Seq.Extensions.Logging, some logging done by Host.Aggregator or Host.Results doesn't contain any message.

#### Repro steps

1. Create a new empty azure function project with an empty http trigger.
2. Install Seq.Extensions.Logging
3. Configure SEQ in startup.cs
 ```csharp
    {
        {
            builder.Services.AddLogging(logging => logging.AddSeq("http://localhost:5341/"));
        }
    }
```
4. Trigger a function so that `Host.Results` can log a result.

#### Expected behavior

All logging in SEQ from azure function sdk should contains message.
image

#### Actual behavior

Some entry by Host.Results and Host.Aggregator are logged with `null` as message
image
image

#### Known workarounds

Supress Host.Results and Host.Aggregator logging to prevent null entry.

#### Related information 

If I understood the problem correctly, some logging are done without an EventId which cause logging without a message
* https://github.com/Azure/azure-webjobs-sdk/blob/804820cdce76b12f2e36ac12eb632beceba3e7da/src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/LoggerExtensions.cs#L59

RohitRanjanMS:
Updated the OriginalFormatString to align with the properties.
Fixes #2722 

image

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/States/FunctionResultAggregateState.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Loggers/Logger/States/FunctionResultAggregateState.cs
@@ -9,7 +9,7 @@ namespace Microsoft.Azure.WebJobs.Logging
 {
     internal readonly struct FunctionResultAggregateState : IReadOnlyList<KeyValuePair<string, object>>
     {
-        public const string OriginalFormatString = "Aggregate result (function={Name}, count={Count}, time={Timestamp}, duration={AverageDuration}ms, maxDuration={MaxDuration}ms, minDuration={MinDuration}ms, success={Successes}, failure={Failures}, rate={SuccessRate})";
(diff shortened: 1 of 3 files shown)


andystaples (approved):
LGTM

matt-richardson:
Hey @RohitRanjanMS :wave:

I just started digging into this - I'm still seeing this error in my logs.

It appears that we need a new release of `Microsoft.Azure.WebJobs` to get this fix.
[According to nuget](https://www.nuget.org/packages/Microsoft.Azure.WebJobs#versions-body-tab), the last release was a bout a week before this PR merged.

Do you have the superpowers to make this happen? If not, what's the best path forward?

matt-richardson:
> It appears that we need a new release of `Microsoft.Azure.WebJobs` to get this fix. [According to nuget](https://www.nuget.org/packages/Microsoft.Azure.WebJobs#versions-body-tab), the last release was a bout a week before this PR merged.

Success - we've now got a new release - Microsoft.Azure.WebJobs [3.0.41](https://www.nuget.org/packages/Microsoft.Azure.WebJobs/3.0.41) has been released which has this fix.

Thanks!

PR #3042, opened 2023-11-08
updating ci build and signing behavior

brettsam:


azure-pipelines.yml
```diff
@@ -48,32 +38,29 @@ jobs:
 
   steps:
   - template: build/install-dotnet.yml
-  
-  - task: 1ESHostedPoolValidation@1
-  - pwsh: |
-      $packageSuffix = 0
-      if (-not $$(hasTag)) {
-        $packageSuffix = $(buildNumber)
-        Write-Host "No git tag found. Setting packages suffix to '$packageSuffix'"
```
soninaren, line 57: I am guessing we don't use package suffix anymore.

soninaren (approved):

azure-pipelines.yml
soninaren: nevermind, i see we are directly using buildNumber

brettsam: we do -- it's just added to the 'pack' step (`--version-suffix`)

```
      custom: 'pack'
      ${{ if eq(variables.hasTag, true) }}:
        arguments: '--no-build -c Release -o $(buildOutputDirectory)'
      ${{ else }}:
        arguments: '--no-build -c Release -o $(buildOutputDirectory) --version-suffix -$(buildNumber)'
```

soninaren: yeah, i just saw that a little later.

PR #3045, opened 2023-11-15
Adding support for Application Insights Diagnostics event logging

RohitRanjanMS:
Implementation of EventListener that listens to events produced by ApplicationInsights SDK.
Events are batched to reduce noise.

Summary
________________________________________________________________________________________________________________________________
Diagnostic source 'Microsoft.Azure.WebJobs.Logging.ApplicationInsights.ApplicationInsightsEventListener' emitted event 'ApplicationInsightsEventListener': 
Ping sent. Outgoing etag: ''. Incoming etag: ''. Response: 'False'
Service endpoint updated.
State timer tick finished: 3 ms
BlobBaseClient.GetProperties.Start id = '00-d3046018d250cd9324ac637180997926-da26604af207dfea-00'
Azure.Core.Http.Request.Start id = '00-d3046018d250cd9324ac637180997926-b505f7c6898aa277-00'
Http request is already instrumented.
Http request is already instrumented.
Azure.Core.Http.Request.Stop id = '00-d3046018d250cd9324ac637180997926-b505f7c6898aa277-00'
TelemetryInitializerBase.Initialize - httpContextAccessor.HttpContext is null, returning.
BlobBaseClient.GetProperties.Stop id = '00-d3046018d250cd9324ac637180997926-da26604af207dfea-00'
TelemetryInitializerBase.Initialize - httpContextAccessor.HttpContext is null, returning.
Ping sent. Outgoing etag: ''. Incoming etag: ''. Response: 'False'
Service endpoint updated.
State timer tick finished: 4 ms
BlobBaseClient.GetProperties.Start id = '00-621c13030668fbec2ed123e3301d6432-be4097fcaf08d2b5-00'
Azure.Core.Http.Request.Start id = '00-621c13030668fbec2ed123e3301d6432-19548d99b30f85b9-00'
Http request is already instrumented.
Http request is already instrumented.
Azure.Core.Http.Request.Stop id = '00-621c13030668fbec2ed123e3301d6432-19548d99b30f85b9-00'
TelemetryInitializerBase.Initialize - httpContextAccessor.HttpContext is null, returning.
BlobBaseClient.GetProperties.Stop id = '00-621c13030668fbec2ed123e3301d6432-be4097fcaf08d2b5-00'
TelemetryInitializerBase.Initialize - httpContextAccessor.HttpContext is null, returning.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
@@ -142,6 +142,7 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
         /// Disabled by default.
         /// </summary>
         public EventLevel? DiagnosticsEventListenerLogLevel { get; set; }
+        
         public string Format()
         {
             JObject sampling = null;
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
@@ -231,7 +231,7 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
                 { nameof(EnableDependencyTracking), EnableDependencyTracking },
                 { nameof(DependencyTrackingOptions), dependencyTrackingOptions },
                 { nameof(TokenCredentialOptions), tokenCredentialOptions },
-                { nameof(DiagnosticsEventListenerLogLevel), DiagnosticsEventListenerLogLevel?.ToString() }
+                { nameof(DiagnosticsEventListenerLogLevel), DiagnosticsEventListenerLogLevel?.ToString() },
             };
 
             return options.ToString(Formatting.Indented);
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Diagnostics.Tracing;
 using Microsoft.ApplicationInsights.SnapshotCollector;
 using Microsoft.ApplicationInsights.WindowsServer.Channel.Implementation;
 using Microsoft.Azure.WebJobs.Hosting;
@@ -136,6 +137,11 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
         /// </summary>
         public bool EnableQueryStringTracing { get; set; } = false;
 
+        /// <summary>
+        /// Gets or sets the event level that enables diagnostic logging by event listener.
+        /// Disabled by default.
+        /// </summary>
+        public EventLevel? DiagnosticsEventListenerLogLevel { get; set; }
         public string Format()
         {
             JObject sampling = null;
@@ -223,7 +229,8 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
                 { nameof(EnableQueryStringTracing), EnableQueryStringTracing },
                 { nameof(EnableDependencyTracking), EnableDependencyTracking },
                 { nameof(DependencyTrackingOptions), dependencyTrackingOptions },
-                { nameof(TokenCredentialOptions), tokenCredentialOptions }
+                { nameof(TokenCredentialOptions), tokenCredentialOptions },
+                { nameof(DiagnosticsEventListenerLogLevel), DiagnosticsEventListenerLogLevel?.ToString() }
             };
 
             return options.ToString(Formatting.Indented);
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Constants/ApplicationInsightsDiagnosticConstants.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Constants/ApplicationInsightsDiagnosticConstants.cs
@@ -5,6 +5,6 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
 {
     public static class ApplicationInsightsDiagnosticConstants
     {
-        public const string DiagnosticSourcePrefix = "Microsoft.Azure.WebJobs.Logging.ApplicationInsights.";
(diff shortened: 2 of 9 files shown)


src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/SelfDiagnostics/ApplicationInsightsEventListener.cs
```diff
@@ -0,0 +1,123 @@
...
+            {
+                return;
+            }
+            lock (_syncLock)
+            {
+                // batch up to 30 events in one log
+                StringBuilder sb = new StringBuilder();
+                // start with a new line
+                sb.AppendLine(string.Empty);
+                while (_logBuffer.TryDequeue(out string line))
```
brettsam: There's not anything stopping more items from coming into this `_logBuffer` while you're building up the string, is there? Seems like you could get more than 30 lines showing up here?

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/SelfDiagnostics/ApplicationInsightsEventListener.cs
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        public void Flush()
+        {
+            if (_logBuffer.Count == 0)
+            {
+                return;
+            }
+            lock (_syncLock)
+            {
```
brettsam, line 79: Check for `_logBuffer.Count == 0` again once you have the lock? Another thread could have potentially flushed everything already...

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/SelfDiagnostics/ApplicationInsightsEventListener.cs
brettsam: As an example of a tweak you can make -- snapshot the `_logBuffer` and re-initialze it as a new queue. We do this in here, for example: https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Diagnostics/FileWriter.cs#L57-L67. 

Still may end up with more than 30 entries as it's a race, but at least you know it won't log uncontrollably.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
```diff
@@ -136,6 +137,11 @@ public TimeSpan QuickPulseInitializationDelay
         /// </summary>
         public bool EnableQueryStringTracing { get; set; } = false;
 
+        /// <summary>
+        /// Gets or sets the event level that enables diagnostic logging by event listener.
+        /// Disabled by default.
+        /// </summary>
+        public EventLevel? DiagnosticsEventListenerLogLevel { get; set; }
```
jviau, line 144: nit: newline after
```suggestion
        public EventLevel? DiagnosticsEventListenerLogLevel { get; set; }
        
```

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/SelfDiagnostics/ApplicationInsightsEventListener.cs
RohitRanjanMS: yes, I found this during testing at high volume. Thought it's ok if the count is more than 30, there's no hard limit but it can potentially grow really big. 
Made the change.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/ApplicationInsightsLoggerOptions.cs
```diff
@@ -223,7 +229,8 @@ public string Format()
                 { nameof(EnableQueryStringTracing), EnableQueryStringTracing },
                 { nameof(EnableDependencyTracking), EnableDependencyTracking },
                 { nameof(DependencyTrackingOptions), dependencyTrackingOptions },
-                { nameof(TokenCredentialOptions), tokenCredentialOptions }
+                { nameof(TokenCredentialOptions), tokenCredentialOptions },
+                { nameof(DiagnosticsEventListenerLogLevel), DiagnosticsEventListenerLogLevel?.ToString() }
```
jviau: nit, end with `,` so diffs are smaller:
```suggestion
                { nameof(DiagnosticsEventListenerLogLevel), DiagnosticsEventListenerLogLevel?.ToString() },
```

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Constants/ApplicationInsightsDiagnosticConstants.cs
```diff
@@ -5,6 +5,6 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
 {
     public static class ApplicationInsightsDiagnosticConstants
     {
-        public const string ApplicationInsightsDiagnosticSourcePrefix = "Microsoft.Azure.WebJobs.Logging.ApplicationInsights.";
+        public const string DiagnosticSourcePrefix = "Microsoft.Azure.WebJobs.Logging.ApplicationInsights.";
```
jviau: Was this shipped before? If so, this is public and cannot be changed.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
```diff
@@ -297,6 +298,10 @@ internal static LoggerFilterOptions CreateFilterOptions(LoggerFilterOptions regi
             QuickPulseInitializationScheduler delayer,
             Action<TelemetryConfiguration> additionalTelemetryConfig)
         {
+            if (options.DiagnosticsEventListenerLogLevel != null)
+            {
+                _ = new ApplicationInsightsEventListener((EventLevel)options.DiagnosticsEventListenerLogLevel);
```
jviau: I think this might be optimized away or GC'd - there is no persistent reference to it. Look into using a `ITelemetryModule`, which is an AppInsights SDK hook to supply a generate piece of code to start up with the appinsights SDK

brettsam (approved):

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Constants/ApplicationInsightsDiagnosticConstants.cs
RohitRanjanMS: reverted it.

src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Extensions/ApplicationInsightsServiceCollectionExtensions.cs
RohitRanjanMS: Added TelemetryModule

PR #3047, opened 2023-11-22
Adding support for pre-aggregated (standard) metrics.

RohitRanjanMS:
Enabling support for AutocollectedMetricsExtractor.

> Extracts auto-collected, pre-aggregated (aka. "standard") metrics from telemetry. Metric Extractors participate in the telemetry pipeline as telemetry processors. They examine telemetry items going through the pipeline and create pre-aggregated metrics based on the encountered items.
https://learn.microsoft.com/en-us/dotnet/api/microsoft.applicationinsights.extensibility.autocollectedmetricsextractor?view=azure-dotnet

Without AutocollectedMetricsExtractor . Log based and standard metrics shows the exact number. 
image

With AutocollectedMetricsExtractor enabled. Standard metrics are _almost_ accurate while log based metrics are off. This will be more evident if the dropped %tage is very high due to sampling.
image

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/WebJobs.Logging.ApplicationInsights.csproj
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/WebJobs.Logging.ApplicationInsights.csproj
@@ -27,7 +27,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Azure.Identity" Version="1.10.4" />
+    <PackageReference Include="Azure.Identity" Version="1.10.2" />
     <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.21.0" />
     <PackageReference Include="Microsoft.ApplicationInsights.PerfCounterCollector" Version="2.21.0" />
     <PackageReference Include="Microsoft.ApplicationInsights.SnapshotCollector" Version="1.4.3" />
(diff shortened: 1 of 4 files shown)


src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/WebJobs.Logging.ApplicationInsights.csproj
```diff
@@ -27,7 +27,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Azure.Identity" Version="1.10.4" />
+    <PackageReference Include="Azure.Identity" Version="1.10.2" />
```
brettsam: This shouldn't be getting rolled back?

RohitRanjanMS: I am aligning this with what we have in the host https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/WebJobs.Script.csproj#L43

brettsam: I think this was explicitly moved forward a couple weeks ago -- https://github.com/Azure/azure-webjobs-sdk/pull/3046/files#diff-97ff29e7c186ba87bf63628248cdec6f393d6c38410d71f3f68232554c126db6.

When we move this to the host, we'll also move that forward. Roll this back and we'll move the host forward when we pull this in.

RohitRanjanMS: Done.

brettsam (approved):

PR #3049, opened 2023-12-08
Ensure SingletonListener is applied before custom Listener decorators

mathewc:
In recent PR https://github.com/Azure/azure-webjobs-sdk/pull/3011, SingletonListener wrapping was included as a platform decorator ordered AFTER any custom user decorators. The problem with that is that it prevents custom decorators from completely controlling or short-circuiting listener startup if SingletonAttribute is used.

For example, assume a custom decorator is added that needs to conditionally disable listener startup for a particular function. The way things are now, this custom decorator is applied AFTER SingletonListener, meaning the singleton would be acquired and held even though the inner listener is never started. SingletonListener and the inner listener it wraps should always be started/stopped as a unit.

The fix is to push Singleton wrapping down into the core ListenerFactory so it happens at the right time any time the target listener is created. Not only is this more correct, but it also has the intended side effect of ordering it before custom decorators. Logic Apps is the only user of this new feature currently, and this will actually fix a bug they could run into in their "dynamic listeners" feature. For example, if the target of their dynamic listener is also Singleton, the SingletonListener currently will start and will stay running (i.e. will hold the lease) even when the target listener is stopped. That isn't correct. They probably haven't run into this (and might never have) because they currently only support ServiceBusTrigger in these dynamic scenarios, and that is never a singleton.

The following example shows the the wrapped listener order before and after, in a scenario where there's a Singleton QueueTrigger function, and a custom Decorator.

Before:
====
FunctionListener
SingletonListener
**CustomDecoratorListener**
QueueListener

After:
====
FunctionListener
**CustomDecoratorListener**
SingletonListener
QueueListener


src/Microsoft.Azure.WebJobs.Host/Hosting/WebJobsServiceCollectionExtensions.cs
```diff
@@ -217,7 +217,6 @@ private static void AddListenerDecorators(IServiceCollection services)
         {
             // Order is important for these platform decorator registrations!
             // They will be applied in this order after any user registered decorators.
-            services.TryAddEnumerable(ServiceDescriptor.Singleton<IListenerDecorator, SingletonListenerDecorator>());
             services.TryAddEnumerable(ServiceDescriptor.Singleton<IListenerDecorator, FunctionListenerDecorator>());
```
mathewc, line 220: The remaining built in decorator applies a final outer error handling listener. I think it makes sense for us to retain this as the final outer listener (i.e. it can't be overridden by custom decorators). Though an argument could also be made for pushing this down as well.

jviau (approved):
Looks good.

Do you think it is worth keeping `SingletonListenerDecorator` and manually instantiating/calling it within `ListenerFactory`?

mathewc:
> Looks good.
> 
> Do you think it is worth keeping `SingletonListenerDecorator` and manually instantiating/calling it within `ListenerFactory`?

I'd prefer to leave decorators to being those that are registered and generically applied by the HostListenerFactory.

PR #3056, opened 2024-01-24
AI Package Update

RohitRanjanMS:
Downgrading AI related packages (except the base Microsoft.ApplicationInsights package) to avoid major version bump in the dependent packages.

Diff with new set of packages.
image

Diff with only Microsoft.ApplicationInsights upgraded to 2.22
image


jviau (commented):
nuget says 2.21.0 is deprecated - do you know why?

jviau (approved):

PR #3058, opened 2024-02-06
Exposing the ability to override the job instance service provider

fabiocav:
The changes in this PR enable consumers to override the service provider used by job instances, enabling scenarios like externally provided scopes.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInstance.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInstance.cs
@@ -3,37 +3,38 @@
 
 using System;
 using System.Collections.Generic;
+using Microsoft.Azure.WebJobs.Host.Bindings;
 using Microsoft.Azure.WebJobs.Host.Protocols;
-using Microsoft.Extensions.DependencyInjection;
 
 namespace Microsoft.Azure.WebJobs.Host.Executors
 {
     internal class FunctionInstance : IFunctionInstanceEx, IDisposable
     {
-        private IServiceScope _instanceServicesScope;
-        private readonly IServiceScopeFactory _serviceScopeFactory;
-        private IServiceProvider _instanceServices;
+        private IInstanceServicesProviderFactory _instanceServicesProviderFactory;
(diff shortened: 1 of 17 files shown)


kshyju (approved):

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionInstance.cs
```diff
@@ -4,18 +4,17 @@
...
-        private IServiceScope _instanceServicesScope;
-        private readonly IServiceScopeFactory _serviceScopeFactory;
-        private IServiceProvider _instanceServices;
+        
+        private readonly IInstanceServicesFeatureFactory _instanceServicesFeatureFactory;
+        private IInstanceServicesFeature _instanceServicesFeature;
 
         public FunctionInstance(Guid id, IDictionary<string, string> triggerDetails, Guid? parentId, ExecutionReason reason, IBindingSource bindingSource,
-            IFunctionInvoker invoker, FunctionDescriptor functionDescriptor, IServiceScopeFactory serviceScopeFactory)
+            IFunctionInvoker invoker, FunctionDescriptor functionDescriptor, IInstanceServicesFeatureFactory instanceServiceFeatureFactory)
```
mathewc: I think the whole "Features" concept and naming is an Http specific thing in ASP.NET Core and might not make conceptual sense here. We don't use "Features" anywhere else in this SDK.

Instead, why isn't this just an IFunctionInstanceServiceProviderFactory interface or something? Seems like you can retain the pluggability you seek with a single factory interface.

fabiocav: There isn't anything HTTP specific about this. 

Though we're not fully implementing the pattern here, this aligns with it and is also something we use in the worker (https://github.com/Azure/azure-functions-dotnet-worker/blob/6647c454b1820711125a2639b0a3fee47dffc088/src/DotNetWorker.Core/Context/FunctionContext.cs#L62)

As discussed, this still requires both interfaces, but I'm changing the name here.

mathewc (approved):

PR #3063, opened 2024-03-04
Updating version to 3.0.41

fabiocav:


kshyju (approved):
Unrelated to your PR change. But, Is the whitespace at line 7 intentional ?

fabiocav:
> Unrelated to your PR change. But, Is the whitespace at line 7 intentional?

I think so. Seems to be an attempt to keep the version property block separate from the others.

PR #3082, opened 2024-07-01
Improving sanitization of error messages in StorageClientProvider.

mathewc:


src/Microsoft.Azure.WebJobs.Host.Storage/WebJobs.Host.Storage.csproj
```diff
@@ -1,6 +1,4 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-
-  <Import Project="..\Microsoft.Azure.WebJobs.Shared\WebJobs.Shared.projitems" Label="Shared" />
```
mathewc, line 3: The Sanitizer was getting compiled into both WebJobs.Host and WebJobs.Host.Storage, and that created a compilation issue because the latter also has internals visible to the former. This shared reference isn't needed for now, so just removing. Could have also resolved by aliasing the assembly reference.

src/Microsoft.Azure.WebJobs.Shared/Sanitizer.cs
```diff
@@ -2,6 +2,8 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Linq;
```
mathewc, line 5: As part of these changes, I copied over the latest from https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Sanitizer.cs.

Had to make a couple minor changes due to the language version used in these projects (e.g. no static local functions).

src/Microsoft.Azure.WebJobs.Host.Storage/WebJobs.Host.Storage.csproj
fabiocav: Haven't looked, but were there any public types in the shared project? Just want to make sure we don't end up with a breaking change as a result.

fabiocav (approved):
Approving, with a pending question about the shared project removal.

mathewc:
> Approving, with a pending question about the shared project removal.

No public types - the shared project just has 2 internal types.

jviau (approved):

PR #3094, opened 2024-09-05
Sanitizing FunctionInvocationException and its inner exception

surgupta-msft:
Resolves - [issue](https://github.com/orgs/Azure/projects/473/views/53?pane=issue&itemId=75688560)

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/FunctionInvocationException.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/FunctionInvocationException.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
+using Microsoft.Azure.WebJobs.Logging;
 using System;
 using System.Runtime.Serialization;
 
@@ -18,12 +19,13 @@ namespace Microsoft.Azure.WebJobs.Host
         }
 
         /// <inheritdoc/>
-        public FunctionInvocationException(string message) : base(message)
+        public FunctionInvocationException(string message) : base(Sanitizer.Sanitize(message))
         {
         }
 
         /// <inheritdoc/>
-        public FunctionInvocationException(string message, Exception innerException) : base(message, innerException)
+        public FunctionInvocationException(string message, Exception innerException) 
+            : base(Sanitizer.Sanitize(message), new Exception(Sanitizer.Sanitize(innerException.Message)))
         {
         }
 
@@ -50,7 +52,7 @@ namespace Microsoft.Azure.WebJobs.Host
         /// <param name="methodName">The fully qualified method name.</param>
         /// <param name="innerException">The exception that is the cause of the current exception (or null).</param>
         public FunctionInvocationException(string message, Guid instanceId, string methodName, Exception innerException)
-            : base(message, methodName, innerException)
+            : base(Sanitizer.Sanitize(message), methodName, new Exception(Sanitizer.Sanitize(innerException.Message)))
         {
             InstanceId = instanceId;
         }
--- a/src/Microsoft.Azure.WebJobs.Host/FunctionInvocationException.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/FunctionInvocationException.cs
@@ -79,7 +79,7 @@ namespace Microsoft.Azure.WebJobs.Host
 
         private static Exception UpdateMessage(Exception exception, string message)
         {
-            if (string.IsNullOrEmpty(message))
+            if (string.IsNullOrEmpty(message) || exception is null)
             {
                 return exception;
             }
--- a/src/Microsoft.Azure.WebJobs.Host/FunctionInvocationException.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/FunctionInvocationException.cs
@@ -3,6 +3,7 @@
 
 using Microsoft.Azure.WebJobs.Logging;
 using System;
+using System.Reflection;
 using System.Runtime.Serialization;
 
 namespace Microsoft.Azure.WebJobs.Host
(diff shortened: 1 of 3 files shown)


src/Microsoft.Azure.WebJobs.Shared/Sanitizer.cs
```diff
@@ -18,7 +18,7 @@ internal static class Sanitizer
 
         // List of keywords that should not be replaced with [Hidden Credential]
```
surgupta-msft, line 19: This is a shared file but there seems to be some difference in code between this file and the one in Host - [Sanitizer.cs](https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Sanitizer.cs). What's the guidance for this? Just copy paste and always keep the code exactly same at both the places?

test/Microsoft.Azure.WebJobs.Host.UnitTests/Loggers/ApplicationInsightsLoggerTests.cs
```diff
@@ -158,6 +158,52 @@ public void LogFunctionResult_Failed_SendsCorrectTelemetry()
             // TODO: Beef up validation to include properties
         }
 
+        [Fact]
+        public void LogFunctionResult_Sanitized_SendsCorrectTelemetry()
+        {
+            FunctionInvocationException fex = new FunctionInvocationException("Failed AccountKey=", new Exception("Text 'key':'secret'"));
```
surgupta-msft, line 164: Exception message has AccountKey and Inner exception message has "secret" here.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Loggers/ApplicationInsightsLoggerTests.cs
```diff
@@ -158,6 +158,52 @@ public void LogFunctionResult_Failed_SendsCorrectTelemetry()
...
+            Assert.Equal(LogLevel.Error.ToString(), requestTelemetry.Properties[LogConstants.LogLevelKey]);
+            // TODO: Beef up validation to include properties
+
+            // Exception needs to have associated id
+            Assert.Equal(expectedOperationId, exceptionTelemetry.Context.Operation.Id);
+            Assert.Equal(expectedRequestId, exceptionTelemetry.Context.Operation.ParentId);
+            Assert.Equal(_functionShortName, exceptionTelemetry.Context.Operation.Name);
+            Assert.Same(fex, exceptionTelemetry.Exception);
+            Assert.Equal("Failed [Hidden Credential]", exceptionTelemetry.Exception.Message);
+            Assert.Equal("Text '[Hidden Credential]'", exceptionTelemetry.Exception.InnerException.Message);
```
surgupta-msft, line 201: Validating that secret and account key has been replaced

src/Microsoft.Azure.WebJobs.Shared/Sanitizer.cs
```diff
@@ -66,8 +66,10 @@ internal static string Sanitize(string input)
                         break;
                     }
 
+                    int startIndexAfterToken = startIndex + token.Length;
+
                     // Find the end of the secret. It most likely ends with either a double quota " or tag opening <
-                    int credentialEnd = t.IndexOfAny(ValueTerminators, startIndex);
+                    int credentialEnd = t.IndexOfAny(ValueTerminators, startIndexAfterToken);
```
surgupta-msft, line 72: The current sanitizer code does not catch the scenarios such as `"key": "value"`. Since it looks for ValueTerminator `"` and ends up giving output `[Hidden Credential]: "value"`. Hence, looking for value with start index after token `key`.

surgupta-msft: We have a followup item to improve this class overall - https://github.com/Azure/azure-functions-host/issues/10383

src/Microsoft.Azure.WebJobs.Host/FunctionInvocationException.cs
```diff
@@ -18,12 +20,13 @@ public FunctionInvocationException() : base()
...
         /// <inheritdoc/>
-        public FunctionInvocationException(string message) : base(message)
+        public FunctionInvocationException(string message) : base(Sanitizer.Sanitize(message))
         {
         }
 
         /// <inheritdoc/>
-        public FunctionInvocationException(string message, Exception innerException) : base(message, innerException)
+        public FunctionInvocationException(string message, Exception innerException)
+            : base(Sanitizer.Sanitize(message), UpdateMessage(innerException, Sanitizer.Sanitize(innerException?.Message)))
```
mathewc, line 29: I think that we should be doing the Sanitization where the log is actually written, not when the exception is constructed. For example, we want to retain the ability of .NET Isolated custom middleware to receive the full exception info as thrown by the customer code.

So we should trace the handling/logging of this exception up to where it's actually written. Basically anywhere where we're handling and logging exception details for exceptions that come from user code, we need to sanitize. That's beyond just function invocations, but also custom startup code as well, which you can verify by throwing a similar exception from custom startup logic.

In the Azure Functions Host, we actually attempt to sanitize exceptions [here](https://github.com/Azure/azure-functions-host/blob/aa332f6c72c0b87ec83dec5bb7b9ac24d9947229/src/WebJobs.Script.WebHost/Diagnostics/SystemLogger.cs#L181) where they're logged. Why isn't that logic handling this?

src/Microsoft.Azure.WebJobs.Shared/Sanitizer.cs
mathewc: Yes - the goal is to keep making the same identified improvements to each over time. The class comments on this Type indicate that "Note: This file is shared between the WebJobs SDK and Script repos. Update both if changes are needed."

Look at the file history of each - I think we revved them both recently.

src/Microsoft.Azure.WebJobs.Shared/Sanitizer.cs
mathewc: I'm thinking that for dealing with customer thrown exceptions, it might be better for us to have a more comprehensive approach, since attempting to identify all the possible signatures customers may log is impossible. For example, you as a customer might log an exception with a message "Account Info {userName}, {pwd}" which token scanning won't identify.

Instead, I think we're going to need to identify all the places where we're invoking user code and handling exceptions, and wrap or mark those exceptions somehow such that later in the logging pipeline when they're logged we can completely avoid logging the exception details (perhaps we can log exception Type, etc.).

src/Microsoft.Azure.WebJobs.Host/FunctionInvocationException.cs
surgupta-msft: I will verify why that logic is not handling this and get back on the PR.

surgupta-msft:
I agree with the PR feedback here - closing this for now. Will discuss with the team and come up with a better plan to tackle these scenarios.

PR #3097, opened 2024-09-24
Delay Adaptive Sampling

### LINKED ISSUE #3100: Missing initialization logs
The Functions runtime generates important telemetry during startup, but some logs are sampled out due to Adaptive sampling being enabled by default with a MaxTelemetryItemsPerSecond of 20. This results in some startup logs being lost when the sampling processor makes its decision.

RohitRanjanMS:
Resolves #3100

The Functions runtime generates important telemetry during startup, but some logs are sampled out due to Adaptive sampling being enabled by default with a `MaxTelemetryItemsPerSecond` of 20. This results in some startup logs being lost when the sampling processor makes its decision.

The proposed change introduces a 15-second delay before Adaptive sampling begins, allowing initial telemetry to be more reliably ingested into AppInsights. The delay duration can be configured, and the feature can be disabled via host.json or appsettings override.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Processors/DelayedSamplingProcessor.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Processors/DelayedSamplingProcessor.cs
@@ -11,14 +11,13 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
     internal class DelayedSamplingProcessor : ITelemetryProcessor
     {
         private readonly AdaptiveSamplingTelemetryProcessor _samplingProcessor;
-        private ITelemetryProcessor next;
-        private bool isSamplingEnabled = false;        
+        private ITelemetryProcessor _next;
+        private bool _isSamplingEnabled = false;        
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Processors/DelayedSamplingProcessor.cs
```diff
@@ -0,0 +1,54 @@
...
+    internal class DelayedSamplingProcessor : ITelemetryProcessor
+    {
+        private readonly AdaptiveSamplingTelemetryProcessor _samplingProcessor;
+        private ITelemetryProcessor _next;
+        private bool _isSamplingEnabled = false;        
+
+        public DelayedSamplingProcessor(ITelemetryProcessor next, ApplicationInsightsLoggerOptions options)
+        {
+            _next = next;
+            _samplingProcessor = new AdaptiveSamplingTelemetryProcessor(options.SamplingSettings, null, next);
```
jviau: Consider extracting construction of `AdaptiveSamplingTelemetryProcessor` and setting of its excluded/included types into a helper method. Can use it here and in `ApplicationInsightsServiceCollectionExtensions`.

RohitRanjanMS: Moved to helper method.

jviau (approved):

PR #3103, opened 2024-10-16
Remove storage extensions project

jviau:
This project has moved to be owned by the Azure SDK team directly (a long time ago).


satvu (approved):

surgupta-msft:
How can we make sure that `Microsoft.Azure.WebJobs.Shared.Storage` is not shared with any other repo? Since it's a shared project, I expected that its removal will have impact on few projects instead of only one - `Microsoft.Azure.WebJobs.Host.Storage`?

I have one example, but I am not sure if it is 100% valid. May be we can just check if there is anything to validate here. The class [Sanitizer.cs](https://github.com/Azure/azure-functions-host/blob/49325e6a5bc896977527c273ef6d70f6a813187d/src/WebJobs.Script/Sanitizer.cs#L13) is shared between [host](https://github.com/Azure/azure-functions-host/blob/49325e6a5bc896977527c273ef6d70f6a813187d/src/WebJobs.Script/Sanitizer.cs#L13) and [webjobs](https://github.com/Azure/azure-webjobs-sdk/blob/bc9290c5b3590759c523800eec1302e37f829ebb/src/Microsoft.Azure.WebJobs.Shared/Sanitizer.cs#L13).

PR #3125, opened 2025-04-09
Fix flaky unit test relying on timing

jviau:
This test was relying on timing control by comparing two durations. This was not always accurate. This PR refactors to use a gate system to ensure the calls are sequential. The test controls the 1st gate, and the system under test controls the 2nd. The test asserts the 2nd gate does not open until _after_ it has opened the 1st gate.


surgupta-msft (approved):

surgupta-msft:
If needed, we can also request @brettsam to take a look to the PR as he has made changes to these tests before.

PR #3128, opened 2025-04-23
Use common artifacts

jviau:
Adopt common artifacts feature of dotnet SDK. CI yaml files are not updated as they provide explicit nuget package output paths via `-o`, meaning this change does not affect them.

### DIFF
--- a/eng/build/Engineering.props
+++ b/eng/build/Engineering.props
@@ -1,5 +1,12 @@
 <Project>
 
+  <!-- artifacts -->
+  <PropertyGroup>
+    <ArtifactsPath>$(RepoRoot)out</ArtifactsPath>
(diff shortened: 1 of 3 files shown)


eng/build/Engineering.props
```diff
@@ -1,5 +1,12 @@
 <Project>
 
+  <!-- artifacts -->
+  <PropertyGroup>
+    <ArtifactsPath>$(RepoRoot)out</ArtifactsPath>
```
surgupta-msft, line 5: Should it be `$(RepoRoot)/out` here? May be I am missing some context here, just wanted to double check

jviau: `RepoRoot` will have the trailing slash as part of it already (this is a common pattern in msbuild, not sure why they do that, but they do)

surgupta-msft (approved):

PR #3134, opened 2025-06-03
Introducing ScaleLoggerExtension.

alrod:
Introduce new ILogger extension methods. New events will be caught in ScaleController and logged to customer's App Insights if needed. There are 2 new methods in public surface area to allow extensions to report triggers votes including trigger name, queue length, concurrency and optimal reason.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Scale/ScaleLoggerExtension.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Scale/ScaleLoggerExtension.cs
@@ -1,4 +1,7 @@
-﻿using System;
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
 using System.Collections.Generic;
 using Microsoft.Extensions.Logging;
 
--- a/src/Microsoft.Azure.WebJobs.Host/Scale/ScaleLoggerExtension.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Scale/ScaleLoggerExtension.cs
@@ -1,0 +1,93 @@
+﻿using System;
+using System.Collections.Generic;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Host.Scale
+{
+    public static class ScaleLoggerExtension
+    {
+        private static readonly EventId FunctionScaleErrorEventId = new EventId(8001, "FunctionScaleError");
+        private static readonly EventId LogFunctionScaleVoteEventId = new EventId(8002, "LogFunctionScaleVote");
+        private static readonly string FunctionNameKey = "functionName";
+        private static readonly string TargetWorkerCountKey = "targetWorkerCount";
+        private static readonly string QueueLengthKey = "queueLength";
+        private static readonly string ConcurrencyKey = "concurrency";
+        private static readonly string VoteKey = "vote";
+        private static readonly string ReasonKey = "reason"; // Renamed from 'Reason'
+
+        // <summary>
+        // Logs a scale vote for a function with the specified name, vote, and optional reason.
+        // </summary>
+        public static void LogFunctionScaleVote(this ILogger logger, string functionName, string vote, string reason = null)
+        {
+            using (logger.BeginScope(new Dictionary<string, object>
+            {
+                [FunctionNameKey] = functionName,
+                [VoteKey] = vote,
+                [ReasonKey] = reason
+            }))
+            {
+                logger.LogDebug(
+                    LogFunctionScaleVoteEventId,
+                    "Function '{functionName}' vote: '{vote}'.", // Fixed typo
+                    functionName,
+                    vote);
+            }
+        }
+
+        // <summary>
+        // Logs a scale vote for a function with the specified name, target worker count, queue length, concurrency, and optional reason.
+        // </summary>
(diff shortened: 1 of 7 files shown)


src/Microsoft.Azure.WebJobs.Host/Scale/ScaleLoggerExtension.cs
```diff
@@ -0,0 +1,96 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.Host.Scale
+{
+    public static class ScaleLoggerExtension
```
jviau: Consider using high-performance logging which will pre-compile the template.
https://learn.microsoft.com/en-us/dotnet/core/extensions/high-performance-logging

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleLoggerExtension.cs
```diff
@@ -0,0 +1,96 @@
...
+        private static readonly string ConcurrencyKey = "concurrency";
+        private static readonly string VoteKey = "vote";
+        private static readonly string ReasonKey = "reason"; // Renamed from 'Reason'
+
+        // <summary>
+        // Logs a scale vote for a function with the specified name, vote, and optional reason.
+        // </summary>
+        public static void LogFunctionScaleVote(this ILogger logger, string functionName, string vote, string reason = null)
+        {
+            using (logger.BeginScope(new Dictionary<string, object>
```
jviau: If you include all the parameters in the template, the scope will not be necessary. In your message below, `functionName` and `vote` will already become log properties. All you would need is to add `reason` to it.

And then if you _really_ don't want to add `reason` to the message, you can define your own struct which implements `IEnumerable<KeyValuePair<string, object>>`. The contents of enumeration are your properties, and then the formatter you pass in to the `Log` call is your message.

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleLoggerExtension.cs
```diff
@@ -0,0 +1,96 @@
...
+        private static readonly string TargetWorkerCountKey = "targetWorkerCount";
+        private static readonly string QueueLengthKey = "queueLength";
+        private static readonly string ConcurrencyKey = "concurrency";
+        private static readonly string VoteKey = "vote";
+        private static readonly string ReasonKey = "reason"; // Renamed from 'Reason'
+
+        // <summary>
+        // Logs a scale vote for a function with the specified name, vote, and optional reason.
+        // </summary>
+        public static void LogFunctionScaleVote(this ILogger logger, string functionName, string vote, string reason = null)
```
jviau, line 55: Why is this added to public API? I am not familiar with scale controller APIs, but if this is expected for webjobs extension packages to call this, could we instead have them return these values as part of the API and then scale controller crafts the log from that?

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleLoggerExtension.cs
alrod: We want to add all values to the scope as those values need to be extracted in SC and logged as separate columns.

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleLoggerExtension.cs
alrod: It would be relay hard to pass those parameters using the API as WebJobs returns final scaling vote for whole function app combining all the triggers votes. Instead, we want webjob extensions/wejobs SDK log special events so ScaleController can catch these events and log them.

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleLoggerExtension.cs
alrod: included reason + high-performance logging

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleLoggerExtension.cs
jviau: Not a fan of this approach, as it is a bit obscure for extensions to understand how to use this log API independently vs it being a natural part of existing API.

But won't hold up this PR on that. We can revisit this approach later.

jviau (approved):

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleManager.cs
```diff
@@ -169,26 +169,33 @@ private async Task<IDictionary<string, TargetScalerResult>> GetTargetScalersResu
                         catch (NotSupportedException ex)
                         {
                             string targetScalerUniqueId = GetTargetScalerFunctionUniqueId(targetScaler);
-                            _logger.LogWarning($"Unable to use target based scaling for Function '{targetScaler.TargetScalerDescriptor.FunctionId}'. Metrics monitoring will be used.", ex);
+
+                            _logger.LogFunctionScaleError(
+                                "Unable to use target based scaling, switching to metrics monitor.",
```
alrod, line 174: Change from warning to error level

PR #3139, opened 2025-07-23
Add FunctionAbortedException and FunctionCanceledException

### LINKED ISSUE #3143: Add new exception type for retry-able error
Add new exception types related to https://github.com/Azure/azure-functions-host/issues/10936 - this will allow extensions to react to this specific type of error and share it between consumers like webjobs extensions and the azure functions host.

satvu:
### Issue describing the changes in this PR

resolves #3143

Pre-requisite for https://github.com/Azure/azure-functions-host/pull/11159 and will also be available for extensions to consume.

### Pull request checklist

**IMPORTANT**: Currently, changes must be backported to the `in-proc` branch to be included in Core Tools and non-Flex deployments.

* [x] My changes **do not** require documentation changes
    * [ ] Otherwise: Documentation issue linked to PR
* [ ] My changes **should not** be added to the release notes for the next release
    * [ ] Otherwise: I've added my notes to `release_notes.md`
* [x] My changes **do not** need to be backported to a previous version
    * [ ] Otherwise: Backport tracked by issue/PR #issue_or_pr
* [x] My changes **do not** require diagnostic events changes
    * Otherwise: I have added/updated all related diagnostic events and their documentation (Documentation issue linked to PR)
* [x] I have added all required tests (Unit tests, E2E tests)


src/Microsoft.Azure.WebJobs.Host/Exceptions/FunctionAbortedException.cs
```diff
@@ -0,0 +1,44 @@
...
+using System;
+using System.Runtime.Serialization;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// Exception thrown when a job function invocation is aborted.
+    /// </summary>
+    [Serializable]
```
jviau: `[Serializable]` and the protected ctor below are deprecated starting in .net8. Considering that is the lowest in-support .NET version, maybe we should stop including this?

https://learn.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib0051

jviau (approved):

liliankasem:
Actually we have been meaning to move https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Description/Workers/FunctionInvocationCanceledException.cs to the WebJobs SDK, do you think we can include that in this PR since we're touching exceptions already?

src/Microsoft.Azure.WebJobs.Host/Exceptions/FunctionAbortedException.cs
Frulfump: The .NET Team has stopped including that attribute for any new Exception created since .NET 8 so I think it makes sense for second party (and third-party) to also not include this attribute.

https://github.com/dotnet/designs/blob/ece1f853d2ba4c174151291056ff29d9101d0091/accepted/2020/better-obsoletion/binaryformatter-obsoletion.md?plain=1#L54

liliankasem (approved):

src/Microsoft.Azure.WebJobs.Host/Exceptions/FunctionAbortedException.cs
```diff
@@ -0,0 +1,28 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// Exception thrown when a job function invocation is aborted.
+    /// </summary>
+    public class FunctionAbortedException : FunctionTimeoutException
```
mathewc: What does "abort" mean? From this hierarchy, it looks like you're saying all aborts are due to timeouts? Then why do we need a new exception type?

src/Microsoft.Azure.WebJobs.Host/Exceptions/FunctionInvocationCanceledException.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    public class FunctionInvocationCanceledException : FunctionInvocationException
+    {
+        public FunctionInvocationCanceledException(string invocationId, Exception innerException)
+            : base($"The invocation request with id '{invocationId}' was canceled before the request was sent to the worker.", innerException) { }
```
mathewc: I get that you're wanting to move https://github.com/Azure/azure-functions-host/blob/dev/src/WebJobs.Script/Description/Workers/FunctionInvocationCanceledException.cs from Functions to the SDK.

Generally, we strive to keep the SDK layer general, and about WebJobs, not Functions specific. In this case, the SDK doesn't know anything about "workers". I wonder if this shouldn't just stay in Functions.

satvu: @liliankasem Thoughts? Does the log message need to include anything about a "worker"? If it would be useful for extensions or other components to know if an invocation was canceled (would they retry or behave in a different way) we can update this. If not, we can keep this in the host.

src/Microsoft.Azure.WebJobs.Host/Exceptions/FunctionAbortedException.cs
jviau: "Abort" means the platform elected to terminate the execution, but it is not a direct fault of the function itself.

This type inherits from `FunctionTimeoutException`, as odd as it is, to avoid a breaking change. We will be switching some function failure cases from the timeout exception to the abort. Any extension which conditions off the exception failure type would then have their logic break if that type changes. Durable is one such extension which does exactly that: they check for timeout exception and special case that failure. By inheriting from that exception their logic will continue to work.

src/Microsoft.Azure.WebJobs.Host/Exceptions/FunctionInvocationCanceledException.cs
liliankasem: I think we wanted to have it here originally and just put it in Functions because it was faster, but given we have other FunctionException here it made sense to me to have it here. But I agree with your message about the SDK not knowing anything about Workers - we can update the message to remove reference to the worker.

Either I don't feel super strongly about this, it was just a lingering work item for a couple years and we can just leave it as is.

satvu: @mathewc I think this could be useful information for extensions - if we updated the log message to something like "Invocation request with id `{invocationId}` was canceled" (no mention of a worker), would that be agreeable?

src/Microsoft.Azure.WebJobs.Host/Exceptions/FunctionAbortedException.cs
mathewc: So you're saying in this repo we want to bake in an undesirable class hierarchy to simplify external consumer code? Instead, why don't you just declare your new exception not as a subclass, and have your consumer code deal with handling both types of exceptions properly during the transition?

src/Microsoft.Azure.WebJobs.Host/Exceptions/FunctionInvocationCanceledException.cs
```diff
@@ -0,0 +1,13 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    public class FunctionInvocationCanceledException : FunctionInvocationException
+    {
+        public FunctionInvocationCanceledException(string invocationId, Exception innerException)
+            : base($"The invocation request with id '{invocationId}' was canceled.", innerException) { }
```
mathewc: nit: The convention we use in this repo when logging IDs is to use "Id" casing, e.g. https://github.com/Azure/azure-webjobs-sdk/blob/bf1d2f79119994cecc206c3eb18e4bdbeb026ccf/src/Microsoft.Azure.WebJobs.Host/Loggers/FunctionInstanceLogger.cs#L86

src/Microsoft.Azure.WebJobs.Host/Exceptions/FunctionAbortedException.cs
jviau: @mathewc because that is a breaking behavior change. Today we are throwing a `FunctionTimeoutException` in a place that we really should never have. But now customers have taken a dependency on that behavior. To have a middle ground of not breaking customers but also differentiating this exception type we opted for class hierarchy.

mathewc: Well, it's also breaking for anyone catching a FunctionTimeoutException currently if you start throwing a new derived type that doesn't actually represent a timeout, right? I think this "is-a" relationship is wrong.

satvu: Updated the naming and definition to:

FunctionAbortTimeoutException - Exception thrown for function invocations that are being aborted due to another concurrently running function timing out.
 
The hierarchy makes much more sense with this naming and logic.

mathewc (approved):

src/Microsoft.Azure.WebJobs.Host/Exceptions/FunctionAbortTimeoutException.cs
```diff
@@ -0,0 +1,28 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs.Host
+{
+    /// <summary>
+    /// Exception thrown when a function invocation is aborted due to another concurrently running function timing out.
+    /// </summary>
+    public class FunctionAbortTimeoutException : FunctionTimeoutException
```
fabiocav: Thought we had settled on `FunctionTimeoutAbortException` :)

fabiocav (changes requested):
Can you please avoid the file moves with this change (avoid the move to the Exceptions folder). Keep the exceptions aligned with areas (similar to what you'd do with other types)

satvu:
> Can you please avoid the file moves with this change (avoid the move to the Exceptions folder). Keep the exceptions aligned with areas (similar to what you'd do with other types)

Both of the ones I moved were in the root of the project (`src/Microsoft.Azure.Wejobs.Host`) and I didn't feel like they were any more "general use" than the exceptions that were in the `/Exceptions` folder (in fact, both are in a hierarchy that implements [FunctionException](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Exceptions/FunctionException.cs), which is stored in `src/Microsoft.Azure.Wejobs.Host/Exceptions`).

fabiocav:
> > Can you please avoid the file moves with this change (avoid the move to the Exceptions folder). Keep the exceptions aligned with areas (similar to what you'd do with other types)
> 
> Both of the ones I moved were in the root of the project (`src/Microsoft.Azure.Wejobs.Host`) and I didn't feel like they were any more "general use" than the exceptions that were in the `/Exceptions` folder (in fact, both are in a hierarchy that implements [FunctionException](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Exceptions/FunctionException.cs), which is stored in `src/Microsoft.Azure.Wejobs.Host/Exceptions`).

Let's not add this change as part of this PR. We can move the exceptions defined int he root (and in `Exceptions`) into the appropriate areas, but for the new ones you're defining, just ensure they're in an area/structure that aligns with their functionality (and avoids general buckets). The ones defined in `Exceptions` folder shouldn't be there, so let's avoid expanding on that pattern.

jviau (approved):

fabiocav (approved):

PR #3146, opened 2025-08-25
Create a function-level request telemetry when none exists.

RohitRanjanMS:
Most instrumented triggers (like HTTP, Service Bus, or Event Hub) create their own activities. In OTel mode, uninstrumented triggers leave Activity.Current as null, so we create a root activity for the function run. This ensures complete telemetry correlation and distributed tracing. Behavior is unchanged for instrumented triggers or when using the Application Insights SDK.


src/Microsoft.Azure.WebJobs.Host/WebJobs.Host.csproj
```diff
@@ -20,6 +20,7 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="2.1.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.1.1" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+    <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="9.0.6" />
```
liliankasem: Any concerns about bring this dep in? Any impact to the host?

RohitRanjanMS: I don't anticipate any issue. We have the exact same version in the host as well.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -97,43 +98,55 @@ public async Task<IDelayedException> TryExecuteAsync(IFunctionInstance functionI
...
 
-                    if (loggedStartedEvent)
-                    {
-                        _functionInstanceLogger.DeleteLogFunctionStarted(functionStartedMessageId);
+                        if (loggedStartedEvent)
+                        {
+                            _functionInstanceLogger.DeleteLogFunctionStarted(functionStartedMessageId);
+                        }
                     }
+                    activity?.Stop();
```
jviau: Should this be in a finally block? Consider calling `Dispose` instead also (so any analyzers we add won't complain dispose isn't being called)

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -97,43 +98,55 @@ public async Task<IDelayedException> TryExecuteAsync(IFunctionInstance functionI
...
+                    // If no current activity exists, create one for the entire function run.
+                    // HTTP, Service Bus, Event Hub, and other instrumented triggers will have their own activities.
+                    // BeginFunctionScope creates a function activity when AppInsights SDK is enabled.
+                    // In OTel mode, Activity.Current will be null unless the trigger is instrumented.
+                    if (Activity.Current is null)
                     {
-                        functionStartedMessage.Failure = FunctionFailure.FromException(exception);
-                        exceptionInfo = ExceptionDispatchInfo.Capture(exception);
-                        exceptionInfo = await InvokeExceptionFiltersAsync(parameterHelper.JobInstance, exceptionInfo, functionInstanceEx, parameterHelper.FilterContextProperties, logger, cancellationToken);
+                       activity = activitySource.StartActivity(functionInstanceEx.FunctionDescriptor.LogName, ActivityKind.Server);
```
jviau: Could this go before line 100? That then avoids the extra indentation level for the `using (parameterHelper)` block.

You could also move the activity creation to a helper (which returns `null` if one already exists), so you can also have it in a `using` block.

src/Microsoft.Azure.WebJobs.Host/WebJobs.Host.csproj
jviau: More than functions use this package though, it is public. We will need to evaluate if there are any breaking changes here.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
RohitRanjanMS: Changed to `using`.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
RohitRanjanMS: Moving it before line 100 i.e., `BeginFunctionScope` will break the existing AppInsights SDK integration.
Moved it to a helper and using `using` block now.

src/Microsoft.Azure.WebJobs.Host/WebJobs.Host.csproj
RohitRanjanMS: Removed the dependency on System.Diagnostics.DiagnosticSource

src/Microsoft.Azure.WebJobs.Host/Abstractions/IActivitySourceAbstraction.cs
```diff
@@ -0,0 +1,15 @@
...
+#nullable enable
+
+namespace Microsoft.Azure.WebJobs.Host.Abstractions
+{
+    /// <summary>
+    /// Abstraction for ActivitySource to avoid dependency on System.Diagnostics.DiagnosticSource
+    /// </summary>
+    public interface IActivitySourceAbstraction
+    {
+        IActivityAbstraction? StartActivity(string name, ActivityKindAbstraction kind = ActivityKindAbstraction.Server);
```
jviau: Do we need this to be nullable? Being explicit would have the benefit of callers clearly expressing what kind they are.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -57,7 +60,9 @@ public FunctionExecutor(
                 ConcurrencyManager concurrencyManager,
                 ILoggerFactory loggerFactory = null,
                 IEnumerable<IFunctionFilter> globalFunctionFilters = null,
-                IDrainModeManager drainModeManager = null)
+                IDrainModeManager drainModeManager = null,
+                IActivitySourceAbstraction activitySource = null,
```
jviau: Do we need `IActivityContextProvider`? If we had just `IActivitySourceAbstraction`, we could have the implementation on the host side determine to no-op if an activity source already exists.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -701,6 +709,32 @@ private FunctionCompletedMessage CreateStartedMessage(IFunctionInstance instance
...
+            // HTTP, Service Bus, Event Hub, and other instrumented triggers will have their own activities.
+            // BeginFunctionScope creates a function activity when AppInsights SDK is enabled.
+            // In OTel mode, Activity.Current will be null unless the trigger is instrumented.
+            if (_activityContextProvider.HasCurrentActivity)
+            {
+                return null;
+            }
+
+            // Start new activity for the function execution
+            return _activitySource.StartActivity(
```
jviau: I wonder if the `IActivitySourceAbstraction.StartActivity` should take in `IFunctionInstanceEx` and let the implementation decide what properties to use.

src/Microsoft.Azure.WebJobs.Host/Abstractions/IActivityAbstraction.cs
```diff
@@ -0,0 +1,15 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Azure.WebJobs.Host.Abstractions
+{
+    /// <summary>
+    /// Abstraction for Activity to avoid dependency on System.Diagnostics.DiagnosticSource
+    /// </summary>
+    public interface IActivityAbstraction : IDisposable
```
jviau: Do we need this abstraction? `System.Diagnostics.Activity` should still be available without the package reference, right? It will just be an older version of it.

jviau (commented):
Can we reduce the abstractions down to just one?

``` CSharp
public interface IFunctionsExecutorActivitySource
{
    System.Diagnostics.Activity? StartActivity(IFunctionInstanceEx instance);
}
```

src/Microsoft.Azure.WebJobs.Host/Abstractions/IActivityAbstraction.cs
jviau: Ah it isn't available in netstandard2.0.

Options:

1. Would bringing in a low version of the package be safer?
2. Can we use `IDisposable` instead? I don't see much usage of the `IActivityAbstraction` beside a `using` block.

src/Microsoft.Azure.WebJobs.Host/Abstractions/IActivitySourceAbstraction.cs
RohitRanjanMS: `StartActivity` may return `null` in case of sampling or if there is no activity source listener. 
Removed Kind.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
RohitRanjanMS: Removed `IActivityContextProvider`

src/Microsoft.Azure.WebJobs.Host/Abstractions/IActivityAbstraction.cs
RohitRanjanMS: Removed `IActivityAbstraction`, the method is now returning `IDisposable`.

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -701,6 +706,23 @@ private FunctionCompletedMessage CreateStartedMessage(IFunctionInstance instance
             return message;
         }
 
+#nullable enable
+        private IDisposable? TryCreateFunctionActivity(IFunctionInstanceEx functionInstanceEx)
+        {
+            // Return null if no activity source is provided
+            if (_activitySource is null)
```
jviau: small nit: can reduce to:

``` CSharp
return _activitySource?.StartActivity(functionInstanceEx);
```

Maybe can even get rid of the function entirely and in-line it.

jviau (approved):

satvu (approved):

PR #3148, opened 2025-08-26
Add logging to investigate pipeline failures

satvu:
Adding logging to investigate why tests using this value will fail on some pipeline runs but not others.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/DefaultHostIdProvider.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/DefaultHostIdProvider.cs
@@ -43,6 +43,34 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
             return Task.FromResult(_hostId);
         }
 
+        public string GetFirstJobMethod()
+        {
+            // Search through all types for the first job method.
+            // The reason we have to do this rather than attempt to use the entry assembly
+            // (Assembly.GetEntryAssembly) is because that doesn't work for WebApps, and the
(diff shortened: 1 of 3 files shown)


src/Microsoft.Azure.WebJobs.Host/Executors/DefaultHostIdProvider.cs
```diff
@@ -30,7 +32,13 @@ public Task<string> GetHostIdAsync(CancellationToken cancellationToken)
             if (_hostId == null)
             {
                 _hostId = ComputeHostId();
```
jviau, line 35: Recommend adding some logging within `ComputeHostId()`. Looking at that implementation, it is a hash of an assembly name. So I would assume if this is failing, that assembly name is changing. But from what and to what?

src/Microsoft.Azure.WebJobs.Host/Executors/DefaultHostIdProvider.cs
```diff
@@ -18,6 +19,7 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
     internal class DefaultHostIdProvider : IHostIdProvider
     {
         private readonly ITypeLocator _typeLocator;
+        private readonly ILogger<DefaultHostIdProvider> _logger;
```
jviau, line 22: This is never assigned to, will always be null.

jviau (commented):
You don't need to actually check this in right? You just need to run CI with it, correct? You can push these changes to ADO and run the CI there.

satvu:
REsolved

Issue #3165, opened 2025-11-14
Low level logging bugs can result in orphaned invocations and incorrect accounting


mathewc:
In FunctionExecutor [here](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs#L83) as part of invocation processing we have code that attempts to ensure that for any invocation that starts, we also complete that invocation. This is done via logging to the `IAsyncCollector<FunctionInstanceLogEntry>`. So there's a call [here](https://github.com/Azure/azure-webjobs-sdk/blob/7a710cbc5bd729392b74e85785deb03f38d850b4/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs#L111) when the invocation starts, and a call [here](https://github.com/Azure/azure-webjobs-sdk/blob/7a710cbc5bd729392b74e85785deb03f38d850b4/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs#L134) when the invocation completes. There's exception handling to ensure that we always complete each invocation.

However, if an exception happens when we're logging [here](https://github.com/Azure/azure-webjobs-sdk/blob/7a710cbc5bd729392b74e85785deb03f38d850b4/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs#L129) the invocation won't be completed. Because the [LogFunctionCompleted](https://github.com/Azure/azure-webjobs-sdk/blob/7a710cbc5bd729392b74e85785deb03f38d850b4/src/Microsoft.Azure.WebJobs.Host/Loggers/FunctionInstanceLogger.cs#L59) event goes through the ILogger chain, any any low level exceptions can blow this up.

PR #3166, opened 2025-11-18
Improve FunctionExecutor error handling

mathewc:
Addresses https://github.com/Azure/azure-webjobs-sdk/issues/3165.

For context, we had an Azure Functions bug where one of the registered loggers was throwing an exception. See https://github.com/Azure/azure-functions-host/pull/11473. The test added in this PR simulates this.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
@@ -124,20 +124,24 @@ namespace Microsoft.Azure.WebJobs.Host.Executors
                         functionStartedMessage.EndTime = DateTimeOffset.UtcNow;
                     }
 
-                    // If function started was logged, don't cancel calls to log function completed.
-                    bool loggedStartedEvent = functionStartedMessageId != null;
-                    _functionInstanceLogger.LogFunctionCompleted(functionStartedMessage);
-
-                    if (instanceLogEntry != null)
+                    try
                     {
-                        CompleteInstanceLogEntry(instanceLogEntry, functionStartedMessage.Arguments, exceptionInfo);
-                        await _functionEventCollector.AddAsync(instanceLogEntry);
-                        _resultsLogger?.LogFunctionResult(instanceLogEntry);
+                        _functionInstanceLogger.LogFunctionCompleted(functionStartedMessage);
                     }
-
-                    if (loggedStartedEvent)
+                    finally
                     {
-                        _functionInstanceLogger.DeleteLogFunctionStarted(functionStartedMessageId);
+                        if (instanceLogEntry != null)
+                        {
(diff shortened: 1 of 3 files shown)


src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -124,20 +124,25 @@ public async Task<IDelayedException> TryExecuteAsync(IFunctionInstance functionI
...
+                        if (instanceLogEntry != null)
+                        {
+                            // it's important that we always complete the log entry even in the event of an exceptions above
+                            CompleteInstanceLogEntry(instanceLogEntry, functionStartedMessage.Arguments, exceptionInfo);
+                            await _functionEventCollector.AddAsync(instanceLogEntry);
+                            _resultsLogger?.LogFunctionResult(instanceLogEntry);
+                        }
+
+                        bool loggedStartedEvent = functionStartedMessageId != null;
+                        if (loggedStartedEvent)
```
alrod:
```suggestion
                        if (functionStartedMessageId != null)
```

src/Microsoft.Azure.WebJobs.Host/Executors/FunctionExecutor.cs
```diff
@@ -124,20 +124,24 @@ public async Task<IDelayedException> TryExecuteAsync(IFunctionInstance functionI
...
-                    if (loggedStartedEvent)
+                    finally
                     {
-                        _functionInstanceLogger.DeleteLogFunctionStarted(functionStartedMessageId);
+                        if (instanceLogEntry != null)
+                        {
+                            // it's important that we always complete the log entry even in the event of an exceptions above
+                            CompleteInstanceLogEntry(instanceLogEntry, functionStartedMessage.Arguments, exceptionInfo);
+                            await _functionEventCollector.AddAsync(instanceLogEntry);
+                            _resultsLogger?.LogFunctionResult(instanceLogEntry);
```
RohitRanjanMS, line 138: `LogFunctionResult` may also fail — are there any implications if it throws an exception?

mathewc: Yes, but I think we're fine here. If that happened, the below DeleteLogFunctionStarted call would be skipped which is a noop relic these days and not crucial anyways in such an error condition. The HandleExceptionAsync call would also be skipped, but that's also OK because right now it only handles Timeout/Fatal exceptions, which in the case of broking logging won't be the case.

In the end, pretty much any line of code that is calling out to service implementations like these can throw, and most of them are also doing logging. Instead of having deep nested try/finally blocks for each line of code, I think we need to be pragmatic and just ensure that the critical completions are performed as I've done.

RohitRanjanMS (approved):

PR #3167, opened 2025-11-19
fixing race in Binder that can cause NullReferenceExceptions

brettsam:
For a long time, we've seen a decent number of errors coming from this code with the stack:

```
System.InvalidOperationException : Error while handling parameter _binder after function returned: ---> System.NullReferenceException : Object reference not set to an instance of an object.
   at async Microsoft.Azure.WebJobs.Binder.Complete(CancellationToken cancellationToken) at D:\a\_work\1\s\src\Microsoft.Azure.WebJobs.Host\Bindings\Runtime\Binder.cs : 156
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.Azure.WebJobs.Host.Bindings.Runtime.RuntimeValueProvider.SetValueAsync(Object value,CancellationToken cancellationToken) at D:\a\_work\1\s\src\Microsoft.Azure.WebJobs.Host\Bindings\Runtime\RuntimeValueProvider.cs : 34
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at async Microsoft.Azure.WebJobs.Host.Executors.FunctionExecutor.ParameterHelper.ProcessOutputParameters(CancellationToken cancellationToken) at D:\a\_work\1\s\src\Microsoft.Azure.WebJobs.Host\Executors\FunctionExecutor.cs : 959
```

~~It's unclear how this is even possible given that the code here can seemingly never be null. But we're going to do some null-checks here to see if we can prevent these exceptions while we continue to investigate.~~

Edit:
We figured out the underlying issue. In Functions if you have multiple input bindings [we call BindAsync() all in parallel](https://github.com/Azure/azure-functions-host/blob/d8ffdbc293d8a424de7f0b89dce2a7a4e56c8c12/src/WebJobs.Script/Description/Workers/WorkerFunctionInvoker.cs?plain=1#L144C13-L158C20). This adds them to the underlying `_binders` list.
 
But `list.Add(item)` isn't thread-safe -- it can cause internal resizing, etc, which means that internally things can be in a bad state. Then later enumerating it can see null items when there should be none. If you use one of these null items, you get a NullRef.

I was confused why the typical "add while enumerating" exceptions weren't being thrown, and it seemed like we would never even be in this state. But now it all makes sense with how Functions calls `BindAsync()` in parallel.

So the issue isn't "adding in one thread while enumerating in another" but it's "adding from multiple threads and then later enumerating". That explains why it never throws "collection was modified".

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/Runtime/Binder.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/Runtime/Binder.cs
@@ -153,18 +153,8 @@ namespace Microsoft.Azure.WebJobs
         /// <returns></returns>
         internal async Task Complete(CancellationToken cancellationToken)
         {
-            if (_binders == null)
-            {
-                return;
-            }
-
             foreach (IValueBinder binder in _binders)
             {
-                if (binder == null)
-                {
-                    continue;
-                }
-
                 // Binding can only be uses for non-Out parameters, and their binders ignore this argument.
                 await binder.SetValueAsync(value: null, cancellationToken: cancellationToken);
             }
--- a/src/Microsoft.Azure.WebJobs.Host/Bindings/Runtime/Binder.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Bindings/Runtime/Binder.cs
@@ -153,8 +153,18 @@ namespace Microsoft.Azure.WebJobs
         /// <returns></returns>
         internal async Task Complete(CancellationToken cancellationToken)
         {
+            if (_binders == null)
+            {
+                return;
+            }
+
             foreach (IValueBinder binder in _binders)
             {
+                if (binder == null)
+                {
+                    continue;
+                }
+
                 // Binding can only be uses for non-Out parameters, and their binders ignore this argument.
                 await binder.SetValueAsync(value: null, cancellationToken: cancellationToken);
             }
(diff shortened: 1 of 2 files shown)


RohitRanjanMS (approved):

src/Microsoft.Azure.WebJobs.Host/Bindings/Runtime/Binder.cs
```diff
@@ -153,8 +153,18 @@ IWatcher IWatchable.Watcher
         /// <returns></returns>
         internal async Task Complete(CancellationToken cancellationToken)
         {
+            if (_binders == null)
```
mathewc: I think the best theory is that the _binders List (which isn't thread safe) is being modified by multiple threads and intermediate state in the list is leading to null refs. I bet in about 30 min a multi-threaded unit test could be written to prove this out. Then we just need to add synchronization to fix this.

mathewc: Also, _binders itself can never be null, what's happening is the enumeration below is returning a null value from the list. So I think we'd just need the check you added on 163 if we go this route. But I'm hoping the synchronization fix works.

jviau: @mathewc if your theory of a thread safety issue here is correct, should we address that? Switching `_binders` to be an `ImmutableList` would work (and re-assign on add), or we could use locking.

brettsam: new fix coming in...

mathewc (approved):

RohitRanjanMS (approved):

PR #3190, opened 2026-04-02
Fix target scaler fallback race condition on multi-worker plans

alrod:
## Problem

On Elastic Premium plans, when a target scaler throws `NotSupportedException` (e.g. ServiceBus without Manage claim), the fallback to incremental scale monitoring fails if the Scale Controller and primary host run on different workers.

`_targetScalersInError` is a static in-process field. Only the `admin/host/scale/status` code path (called by the Scale Controller) invokes target scalers and populates this set. The `ScaleMonitorService.TakeMetricsSamplesAsync()` path (running on the primary host) reads the set but never populates it. When these run on different workers, the primary host's copy stays empty — it never discovers the failure, excludes the incremental monitor, collects no metrics, and produces no scale-in votes. The app stays stuck at max workers indefinitely.

## Fix

- **Extract `TryGetScaleResultAsync`** — single method that calls `GetScaleResultAsync`, catches `NotSupportedException`, and adds the scaler to `_targetScalersInError`. Eliminates code duplication between `GetTargetScalersResultAsync` and the new probe in `ScaleMonitorService`.

- **Probe target scalers in `ScaleMonitorService`** — `TakeMetricsSamplesAsync` now periodically calls `TryGetScaleResultAsync` on target scalers so the primary host independently discovers failures. Runs on first tick, then every 10 minutes to balance detection speed vs overhead.

- **Thread safety** — replaced `HashSet<string>` with `ConcurrentDictionary<string, byte>` for `_targetScalersInError`. The read path (`ContainsKey` in `GetScalersToSample`) and write path (`TryAdd` in `TryGetScaleResultAsync`) can now execute concurrently without locks.

- **Cleanup** — removed redundant re-initialization of `_targetScalersInError` in the constructor (field is already initialized at declaration; `ScaleManager` is a singleton so the constructor only runs once, but the assignment was unnecessary).

## Testing

- Updated existing `GetScalersToSample_FallsBackToMonitor_OnTargetScalerError` assertion to match updated log message format.
- Added `OnTimer_ProbesTargetScalers_FallsBackToMonitor` — verifies that `ScaleMonitorService` independently detects a faulty target scaler and falls back to incremental monitoring.
- All 27 scale tests pass (21 ScaleManager + 6 ScaleMonitorService).

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Scale/ScaleMonitorService.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Scale/ScaleMonitorService.cs
@@ -30,6 +30,8 @@ namespace Microsoft.Azure.WebJobs.Host.Scale
         private readonly ITargetScalerManager _targetScalerManager;
         private readonly IConfiguration _configuration;
         private bool _disposed;
+        private static DateTime _nextTargetScalerValidationTime = DateTime.MinValue;
+        private static readonly TimeSpan _targetScalerValidationInterval = TimeSpan.FromMinutes(10);
 
         public ScaleMonitorService(
             IScaleStatusProvider scaleStausProvider,
@@ -93,6 +95,29 @@ namespace Microsoft.Azure.WebJobs.Host.Scale
             try
             {
                 var (scaleMonitorsToProcess, targetScalersToSample) = ScaleManager.GetScalersToSample(_monitorManager, _targetScalerManager, _scaleOptions, _configuration);
+
+                // Periodically probe target scalers to discover any that throw NotSupportedException.
+                // This ensures the primary host independently detects target scaler failures,
+                // even when the Scale Controller runs on a different worker.
+                // Runs on first tick, then every 10 minutes to avoid per-tick overhead
+                // while still catching runtime permission changes (e.g. managed identity revocation).
+                if (targetScalersToSample.Any() && DateTime.UtcNow >= _nextTargetScalerValidationTime)
+                {
+                    foreach (var targetScaler in targetScalersToSample)
+                    {
+                        try
+                        {
+                            await ScaleManager.TryGetScaleResultAsync(targetScaler, new TargetScalerContext(), _logger, "ScaleMonitorService").ConfigureAwait(false);
+                        }
+                        catch (Exception)
+                        {
+                            // Ignore transient errors — only NotSupportedException triggers fallback
+                            // and that is already handled inside TryGetScaleResultAsync.
+                        }
+                    }
+
+                    _nextTargetScalerValidationTime = DateTime.UtcNow + _targetScalerValidationInterval;
+                }
 
                 if (scaleMonitorsToProcess.Any())
                 {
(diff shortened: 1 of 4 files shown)


src/Microsoft.Azure.WebJobs.Host/Scale/ScaleMonitorService.cs
```diff
@@ -94,6 +96,29 @@ private async Task TakeMetricsSamplesAsync()
             {
                 var (scaleMonitorsToProcess, targetScalersToSample) = ScaleManager.GetScalersToSample(_monitorManager, _targetScalerManager, _scaleOptions, _configuration);
 
+                // Periodically probe target scalers to discover any that throw NotSupportedException.
+                // This ensures the primary host independently detects target scaler failures,
+                // even when the Scale Controller runs on a different worker.
+                // Runs on first tick, then every 10 minutes to avoid per-tick overhead
+                // while still catching runtime permission changes (e.g. managed identity revocation).
+                if (targetScalersToSample.Any() && DateTime.UtcNow >= _nextTargetScalerValidationTime)
```
mathewc, line 104: This method is all about taking metrics samples for incremental scaling. It doesn't make any sense to be making TBS calls here. Also, this approach makes a bunch of service calls whose results are just thrown away.

Instead, why don't we address the fundamental issue which is the _targetScalersInError collection as you're using it is supposed to be shared state across N workers, similar to metrics samples and snapshots. So one thing you could do is store that info in storage in the same way so its accessible by all instances.

alrod: Thanks @mathewc. A couple of points on the probe approach:

1. **Service call frequency** — the probe wasn't per-tick. It ran on first tick, then cached for 10 minutes (`_nextTargetScalerValidationTime`). So it was one TBS call per scaler per 10 minutes, not "a bunch." For comparison, the concurrency snapshot ([`_concurrencyStatusRepository.ReadAsync`](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Scale/ScaleManager.cs#L148)) does a blob storage read on every SC call (~10s) with no caching.

2. **Storage-backed approach** — I agree it's architecturally cleaner. I've implemented it in https://github.com/Azure/azure-webjobs-sdk/pull/3191 following the [`BlobStorageConcurrencyStatusRepository`](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host.Storage/BlobStorageConcurrencyStatusRepository.cs) / [`NullConcurrencyStatusRepository`](https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Scale/NullConcurrencyStatusRepository.cs) pattern. It is a significantly bigger change (~700 lines vs ~150), but it fixes the root cause and keeps TBS calls out of the metrics sampling path.

mathewc: I do think I like the storage option better. Seems less hacky. One question I do have on this whole "fallback" approach - this ScaleMonitorService won't even be polling metrics if IsRuntimeScalingEnabled=false which is the normal case, right? Only if RTS is being used will IsRuntimeScalingEnabled=true. So usually TakeMetricsSamplesAsync won't even be running, so wondering how the fallback works in the normal case.

alrod:
closing in favor - https://github.com/Azure/azure-webjobs-sdk/pull/3191

PR #3191, opened 2026-04-08
Fix target scaler fallback race condition via shared storage

alrod:
## Problem

On Elastic Premium plans, when a target scaler throws `NotSupportedException` (e.g., ServiceBus without Manage claim), the fallback to incremental scale monitoring fails if the Scale Controller and primary host run on different workers.

`_targetScalersInError` was a static in-process `HashSet<string>`. Only the `admin/host/scale/status` code path (called by the Scale Controller) invokes target scalers and populates this set. The `ScaleMonitorService.TakeMetricsSamplesAsync()` path (running on the primary host) reads the set but never populates it. When these run on different workers, the primary host's `_targetScalersInError` stays empty — it never discovers the failure, excludes the incremental monitor, collects no metrics, and produces no scale-in votes. The app stays stuck at max workers indefinitely.

## Root Cause

The fundamental issue is that `_targetScalersInError` is process-local shared state that needs to be visible across multiple workers — the same class of problem solved by `IScaleMetricsRepository` (for metrics) and `IConcurrencyStatusRepository` (for concurrency snapshots), both of which use blob storage.

## Fix

Introduce a new internal `ITargetScalerErrorRepository` abstraction backed by blob storage, following the established `BlobStorageConcurrencyStatusRepository` pattern.

### New Files

| File | Description |
|------|-------------|
| `ITargetScalerErrorRepository` | Internal interface: `AddAsync`, `GetAsync` |
| `NullTargetScalerErrorRepository` | No-op default (used when storage isn't configured) |
| `BlobStorageTargetScalerErrorRepository` | Blob-backed impl persisting to `scale/{hostId}/targetScalersInError.json` |

### Modified Files

| File | Description |
|------|-------------|
| `ScaleManager.cs` | Removed static `_targetScalersInError`. Injected `ITargetScalerErrorRepository`. Writes errors via `AddAsync` on `NotSupportedException`. `GetScalersToSample` is now async and accepts the repository directly (reads inside `IsTargetScalingEnabled` block). |
| `ScaleMonitorService.cs` | Injected `ITargetScalerErrorRepository`. Passes repository to `GetScalersToSample`. |
| `WebJobsServiceCollectionExtensions.cs` | Registers `NullTargetScalerErrorRepository` as default |
| `StorageServiceCollectionExtensions.cs` | Registers `BlobStorageTargetScalerErrorRepository` when Azure Storage is configured |

### Key Design Decisions

**ETag-based optimistic concurrency:** `AddAsync` uses a read-modify-write loop with ETag conditions to prevent lost updates when multiple instances write concurrently. Uses `IfMatch` for existing blobs and `IfNoneMatch = *` for new blobs. Retries up to 3 times on 412/409.

**TTL-based expiry (no ClearAsync):** Instead of explicitly clearing errors on startup (which races with other instances), the blob stores a `lastUpdated` timestamp alongside the scaler set. `GetAsync` returns empty if the data is older than 10 minutes. This makes recovery automatic — once the error stops occurring (e.g., customer grants Manage claim and restarts), no new `AddAsync` calls refresh the timestamp, and the data ages out.

**Internal interface:** `ITargetScalerErrorRepository` is `internal` — it's an implementation detail, not public API surface.

### How It Fixes the Race Condition

1. Worker A (running Scale Controller) calls `GetScaleStatusAsync` → catches `NotSupportedException` → writes scaler ID to blob via `AddAsync`
2. Worker B (primary host) calls `TakeMetricsSamplesAsync` → `GetScalersToSample` reads blob via repository → sees the error → falls back to incremental monitor → collects metrics → produces scale-in votes

### Recovery Flow

1. TBS call fails → `NotSupportedException` → `AddAsync()` → blob created/updated with current timestamp
2. As long as the error persists, each Scale Controller poll refreshes the timestamp
3. Customer grants Manage claim → restarts app → TBS succeeds → no `AddAsync` calls → blob ages out after 10 minutes → TBS used again

## Testing

- Updated existing `GetScalersToSample_FallsBackToMonitor_OnTargetScalerError` test to use repository
- Updated all `ScaleManager` constructor calls for new parameter
- Added `GetScalersToSample_WithErrorSet_FiltersCorrectly` test
- Added `GetScalersToSample_WithNullErrorSet_NoFiltering` test (uses `NullTargetScalerErrorRepository`)
- Added `GetScaleStatusAsync_SecondCall_ReadsPersistedError` test
- Added `OnTimer_CrossWorkerFallback_SharedRepository` test
- Added `GetAsync_StaleData_ReturnsEmpty` TTL expiry test
- Added `GetAsync_FreshData_ReturnsScalers` TTL expiry test
- Updated E2E blob tests to verify new `TargetScalerErrorState` format
- All 162 scale + public surface tests pass (0 errors, 0 warnings)


src/Microsoft.Azure.WebJobs.Host/Scale/ScaleManager.cs
```diff
@@ -60,7 +61,8 @@ internal ScaleManager()
         /// <returns>A task that returns the <see cref="AggregateScaleStatus"/>.</returns>
         public async Task<AggregateScaleStatus> GetScaleStatusAsync(ScaleStatusContext context)
         {
-            var (scaleMonitorsToProcess, targetScalersToProcess) = GetScalersToSample(_monitorManager, _targetScalerManager, _scaleOptions, _configuration);
+            var scalersInError = await _targetScalerErrorRepository.GetAsync(CancellationToken.None);
```
mathewc: This load should only happen if TBS is enabled, right? Shouldn't this be pushed inside of GetScalersToSample, inside the IsTargetScalingEnabled block? That also addresses the duplication of this line that you have now in ScaleMontorService as well.

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleManager.cs
```diff
@@ -228,10 +228,11 @@ internal static (List<IScaleMonitor>, List<ITargetScaler>) GetScalersToSample(
             if (scaleOptions.Value.IsTargetScalingEnabled)
             {
                 HashSet<string> targetScalerFunctions = new HashSet<string>();
+                var errored = targetScalersInError ?? new HashSet<string>();
```
mathewc: Rather than new up a set here, can't you just do `!targetScalersInError?.Contains(scalerUniqueId)`? However, why even accept null in this method? Looks like your repository doesn't even return null?

src/Microsoft.Azure.WebJobs.Host.Storage/BlobStorageTargetScalerErrorRepository.cs
```diff
@@ -0,0 +1,138 @@
...
+        private BlobContainerClient? _blobContainerClient;
+
+        public BlobStorageTargetScalerErrorRepository(IHostIdProvider hostIdProvider, ILoggerFactory loggerFactory, IAzureBlobStorageProvider azureStorageProvider)
+        {
+            _hostIdProvider = hostIdProvider;
+            _logger = loggerFactory.CreateLogger(LogCategories.Scale);
+            _blobStorageProvider = azureStorageProvider;
+        }
+
+        public async Task AddAsync(string scalerUniqueId, CancellationToken cancellationToken)
```
mathewc, line 41: A possibility for lost updates exist here if two instances concurrently attempt to add a scaler. So you'll either want to do an Etag retry loop, or move to a one blob per scaler id design. Probably the former.

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleMonitorService.cs
```diff
@@ -50,19 +52,25 @@ public ScaleMonitorService(
...
+        public async Task StartAsync(CancellationToken cancellationToken)
         {
             if (_scaleOptions.Value.IsRuntimeScalingEnabled)
             {
                 _logger.LogInformation("Runtime scale monitoring is enabled.");
 
+                // Clear any persisted target scaler errors from previous runs.
+                // This allows TBS to be re-evaluated after an app restart,
+                // e.g. after the customer grants the Manage claim.
+                await _targetScalerErrorRepository.ClearAsync(cancellationToken);
```
mathewc: Clearing like this can delete newly reported errors from other instances (race condition). Will also be unnecessary deletions, and open the window for periods of time where actual errors aren't reported.

Instead, why don't you consider tagging the blob with a last updated timestamp, and in the repository return empty if its older than a TTL. This way it clears itself automatically.

src/Microsoft.Azure.WebJobs.Host/Scale/ITargetScalerErrorRepository.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.Host.Scale
+{
+    /// <summary>
+    /// Provides functionality for persisting target scaler errors across multiple host instances.
+    /// When a target scaler throws <see cref="System.NotSupportedException"/>, the scaler identifier
+    /// is recorded so all instances can fall back to incremental scale monitoring.
+    /// </summary>
+    public interface ITargetScalerErrorRepository
```
mathewc: These new interfaces/types are just an internal implementation detail so we should make them internal so as not to increase our public surface area

test/Microsoft.Azure.WebJobs.Host.UnitTests/PublicSurfaceTests.cs
```diff
@@ -281,6 +281,7 @@ public void WebJobs_Host_VerifyPublicSurfaceArea()
                 "ITargetScaler",
                 "ITargetScalerManager",
                 "ITargetScalerProvider",
+                "ITargetScalerErrorRepository",
```
mathewc: We don't want this new public surface

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleManager.cs
alrod: Good catch -- moved the `GetAsync` call inside `GetScalersToSample`, within the `IsTargetScalingEnabled` block. The method now accepts `ITargetScalerErrorRepository` instead of a pre-loaded `ISet<string>`, and is async. This eliminates the duplicated load in both `ScaleManager.GetScaleStatusAsync` and `ScaleMonitorService.TakeMetricsSamplesAsync`, and avoids the unnecessary blob read when TBS is disabled.

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleManager.cs
alrod: Agreed -- removed null support from the parameter. `GetScalersToSample` now requires a non-null `ITargetScalerErrorRepository` (no default value). DI always provides one (`NullTargetScalerErrorRepository` when storage isn't configured), so null is never the real case. Updated the test that was passing null to use `NullTargetScalerErrorRepository` instead.

src/Microsoft.Azure.WebJobs.Host.Storage/BlobStorageTargetScalerErrorRepository.cs
alrod: Good point -- added ETag-based optimistic concurrency with a retry loop (max 3 attempts). `ReadBlobWithETagAsync` now returns the content + ETag, and `WriteBlobAsync` passes `IfMatch = etag` for existing blobs or `IfNoneMatch = ETag.All` for new blobs. On 412/409, the loop re-reads and retries.

src/Microsoft.Azure.WebJobs.Host/Scale/ScaleMonitorService.cs
alrod: Great suggestion -- replaced `ClearAsync` with TTL-based expiry. The blob format now includes a `lastUpdated` timestamp alongside the scaler set. `GetAsync` returns empty if the data is older than 10 minutes (aligns with Scale Controller poll interval). `AddAsync` refreshes the timestamp on every write, keeping it alive while the error persists. Removed `ClearAsync` from the interface, all implementations, and `ScaleMonitorService.StartAsync`. Recovery is automatic: once the customer fixes the issue and restarts, no new `AddAsync` calls occur, the blob ages out, and TBS is retried.

src/Microsoft.Azure.WebJobs.Host/Scale/ITargetScalerErrorRepository.cs
alrod: Done -- changed `ITargetScalerErrorRepository` to `internal`. `Microsoft.Azure.WebJobs.Host.Storage` already has `InternalsVisibleTo` for the Host assembly, so the blob implementation compiles fine.

test/Microsoft.Azure.WebJobs.Host.UnitTests/PublicSurfaceTests.cs
alrod: Removed -- `ITargetScalerErrorRepository` is now internal, so it no longer appears in the public surface area test.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Scale/ScaleMonitorServiceTests.cs
```diff
@@ -213,6 +213,80 @@ await TestHelpers.Await(() =>
             var metricsWritten = _metricsRepository.Metrics[monitor2].Take(5);
             Assert.Equal(testMetrics2, metricsWritten);
         }
+
+        [Fact]
+        public async Task OnTimer_CrossWorkerFallback_SharedRepository()
```
mathewc: This test doesn't exercise any timers or any ScaleMonitorService APIs. I think you should move it to ScaleManagerTests and name it something like GetScalersToSample_FiltersScalersInError or something like that

test/Microsoft.Azure.WebJobs.Host.UnitTests/Scale/ScaleManagerTests.cs
```diff
@@ -471,12 +475,164 @@ public async Task GetTargetScalersResult_ReadsConcurrencyStatus_OnlyWhenDCAndPer
...
+            ScaleManager scaleManager = new ScaleManager(_monitorManagerMock.Object, _targetScalerManagerMock.Object, _metricsRepositoryMock.Object, concurrencyStatusRepositoryMock.Object, _targetScalerErrorRepository, options, _loggerFactory, _configuration, concurrencyOptions);
 
             await scaleManager.GetScaleStatusAsync(context);
 
             var expectedCalls = expectReadCalled ? Times.Once() : Times.Never();
             concurrencyStatusRepositoryMock.Verify(p => p.ReadAsync(It.IsAny<CancellationToken>()), expectedCalls);
         }
+
+        [Fact]
+        public async Task InMemoryTargetScalerErrorRepository_Lifecycle()
```
mathewc: This and the next tests are test cases for a test helper class? We don't usually do that - likely not necessary

src/Microsoft.Azure.WebJobs.Host.Storage/BlobStorageTargetScalerErrorRepository.cs
```diff
@@ -0,0 +1,184 @@
...
+        {
+            try
+            {
+                for (int attempt = 0; attempt < MaxRetries; attempt++)
+                {
+                    // Read current state with ETag
+                    var (state, etag) = await ReadBlobWithETagAsync(cancellationToken);
+                    var set = state?.Scalers ?? new HashSet<string>();
+                    if (!set.Add(scalerUniqueId))
```
mathewc: This block reads strange - recommend just dropping the if statement and always doing the add.

There is an optimization you could make along these lines to avoid unnecessary blob updates, but I don't think you need to go that far:

```
bool added = set.Add(scalerUniqueId);
bool ttlNeedsRefresh = state?.LastUpdated == null || (DateTime.UtcNow - state.LastUpdated.Value) > (DefaultTtl / 2);
if (!added && !ttlNeedsRefresh)
{
    return; // nothing to persist
}
```

test/Microsoft.Azure.WebJobs.Host.UnitTests/Scale/ScaleMonitorServiceTests.cs
alrod: Good call -- this test doesn't exercise any ScaleMonitorService APIs. Looking closer, the scenario it covers (write error via `GetScaleStatusAsync` then read via `GetScalersToSample`) is already fully covered by `GetScalersToSample_FallsBackToMonitor_OnTargetScalerError` in ScaleManagerTests, and the filtering logic is covered by `GetScalersToSample_WithErrorSet_FiltersCorrectly`. Deleted it since it's redundant.

test/Microsoft.Azure.WebJobs.Host.UnitTests/Scale/ScaleManagerTests.cs
alrod: Agreed -- removed both. The helper is trivially simple and already exercised indirectly by `GetScalersToSample_FallsBackToMonitor_OnTargetScalerError` and `GetScaleStatusAsync_SecondCall_ReadsPersistedError`.

src/Microsoft.Azure.WebJobs.Host.Storage/BlobStorageTargetScalerErrorRepository.cs
alrod: Done -- dropped the if statement and always calling `set.Add()`. Keeping it simple without the TTL refresh optimization since the write is infrequent.

mathewc (approved):

PR #3200, opened 2026-04-30
Remove duplicate target scaler vote log from ScaleManager

alrod:
### Problem

The ScaleManager logs 'Function X vote: TargetWorkerCount=N.' (EventId 8002) after each target scaler call. However, the individual extension target scalers already log the same information with more context:

- **EventHubs** [EventHubsTargetScaler.cs#L111](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/eventhub/Microsoft.Azure.WebJobs.Extensions.EventHubs/src/Listeners/EventHubsTargetScaler.cs#L111): Target worker count for function X is N (EventHubName, EventCount, Concurrency, PartitionCount)
- **ServiceBus** [ServiceBusTargetScaler.cs#L111](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusTargetScaler.cs#L111): Target worker count for function X is N (EntityPath, MessageCount, Concurrency)

This produces duplicate log entries every polling cycle per function, adding noise to customer App Insights and platform Kusto telemetry.

### Changes

- **ScaleManager.cs**: Removed the redundant LogFunctionScaleVote call for target scalers (was at line 193). The extension-level logs provide strictly more information.
- **ScaleManagerTests.cs**: Updated test assertions that checked for the now-removed log line, adjusting log array indices.

### Testing

- All 24 ScaleManagerTests pass
- All 10 ScaleLoggerExtensionTests pass

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Host/Scale/ScaleManager.cs
+++ b/src/Microsoft.Azure.WebJobs.Host/Scale/ScaleManager.cs
@@ -190,7 +190,6 @@ namespace Microsoft.Azure.WebJobs.Host.Scale
                                 TargetWorkerCount = context.WorkerCount
                             };
                         }
-                        _logger.LogFunctionScaleVote(targetScaler.TargetScalerDescriptor.FunctionId, result.TargetWorkerCount);
                         targetScaleVotes.Add(targetScaler.TargetScalerDescriptor.FunctionId, result);
                     }
                     catch (Exception exc)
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.WebJobs.Host/Scale/ScaleManager.cs
```diff
@@ -190,7 +190,6 @@ private async Task<IDictionary<string, TargetScalerResult>> GetTargetScalersResu
                                 TargetWorkerCount = context.WorkerCount
                             };
                         }
-                        _logger.LogFunctionScaleVote(targetScaler.TargetScalerDescriptor.FunctionId, result.TargetWorkerCount);
```
mathewc, line 193: Hmmm, I get what you're saying that most (if not all) scalers are already doing their own logging. However, it might be beneficial to retain these central logs to ensure this info is always logged. Otherwise you're relying on the individual scalers which might not log. So this is more of an assurance.

How big of an issue is the extra log really?

PR #3205, opened 2026-05-21
Strip _MS.AggregationIntervalMs from custom metrics when EnableMetricsCustomDimensionOptimization is enabled

pmalmsten:
## Summary

When `ApplicationInsightsLoggerOptions.EnableMetricsCustomDimensionOptimization` is `true`, the `WebJobsTelemetryInitializer` now also strips the `_MS.AggregationIntervalMs` custom dimension from `MetricTelemetry` items, in addition to the existing `HostInstanceId` removal.

`_MS.AggregationIntervalMs` is added by the Application Insights classic SDK's metric-aggregation pipeline (`MetricAggregateToApplicationInsightsPipelineConverterBase.ConvertAggregateToTelemetry` sets it before calling `TelemetryClient.Track`). The OpenTelemetry Azure Monitor exporter does not emit this dimension on custom metrics, so this change brings the classic SDK output into parity with the OTel exporter when the optimization flag is enabled (default-off).

## Why

`_MS.AggregationIntervalMs` provides no user-facing diagnostic value but is projected as a custom dimension onto every pre-aggregated custom metric, where it acts as a high-cardinality multiplier on every unique timeseries. Production telemetry from a Logic Apps Standard / Application Insights deployment shows the impact:

- The dimension takes **66 distinct values** per metric (varying aggregation intervals in ms), exploding timeseries counts via the cross-product.
- Across all classic-SDK workloads on a single Application Insights resource (24h sample): **64,559 unique timeseries** today vs. an estimated **9,122** without this dimension — an **85.9%** reduction.
- For Logic Apps Standard specifically: **30,752 → 1,738** unique timeseries (**94.3%** reduction); 96.5% of timeseries today carry this dimension.
- The resulting **6x–13x per-region multiplier** drives breaches of the Azure Monitor 50,000-active-timeseries-per-subscription-region custom metrics quota.

Independent confirmation that the dimension is purely a classic-SDK artifact:

- Inspection of the Azure Monitor OTel exporter source (`MetricHelper.cs`) shows it only copies `metricPoint.Tags` into properties and never sets `_MS.AggregationIntervalMs`.
- Workloads migrated from the classic SDK to the OTel exporter (AKS pods, Container Apps) showed **0 records** with `_MS.AggregationIntervalMs` across 100M+ datapoints/day; rolling OTel back caused the dimension to reappear at 42–48% of metrics within hours, and re-deploying OTel dropped it back to 0.

This change brings the classic SDK to parity for workloads that opt in via `EnableMetricsCustomDimensionOptimization`, dropping their active timeseries well within quota without losing any meaningful diagnostic information.

## Pipeline ordering

The Application Insights SDK pipeline runs `ITelemetryInitializer.Initialize` inside `TelemetryClient.Track`, before the processor chain and the channel:

```
SDK aggregator sets _MS.AggregationIntervalMs on MetricTelemetry
  → TelemetryClient.Track(item)
  → ITelemetryInitializer.Initialize  ← WebJobsTelemetryInitializer removes the dimension here
  → ITelemetryProcessor chain
  → ITelemetryChannel.Send             (serialized envelope, no _MS.AggregationIntervalMs)
```

## Changes

- `Constants/LoggingConstants.cs` — add `AggregationIntervalMsKey = "_MS.AggregationIntervalMs"`, matching the style of the existing `HostInstanceIdKey` constant.
- `Initializers/WebJobsTelemetryInitializer.cs` — when `EnableMetricsCustomDimensionOptimization` is enabled and the telemetry item is a `MetricTelemetry`, also remove `LoggingConstants.AggregationIntervalMsKey` from `Properties`.
- `test/.../Loggers/WebJobsTelemetryInitializerTests.cs` — three new unit tests covering: removal when optimization is enabled, preservation when disabled, and that non-metric telemetry (`TraceTelemetry`) is unaffected.

## Behavior

- Default (`EnableMetricsCustomDimensionOptimization = false`): unchanged.
- When opted in: `MetricTelemetry` items leaving the channel will not carry `_MS.AggregationIntervalMs` regardless of whether the dimension was added by `TelemetryClient.GetMetric(...).TrackValue(...)` aggregation or by `AutocollectedMetricsExtractor`. Other telemetry types are unaffected.

## Tests

- New unit tests pass; all 35 tests in `WebJobsTelemetryInitializerTests` and `ApplicationInsightsLoggerTests` pass.
- Build succeeds with no new warnings.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Constants/LoggingConstants.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Constants/LoggingConstants.cs
@@ -9,12 +9,6 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
         public const string Unknown = "[Unknown]";
         public const string ClientIpKey = "ClientIp";
         public const string HostInstanceIdKey = "HostInstanceId";
-
-        // Property added by the Application Insights SDK metric aggregation pipeline
-        // (e.g. when metrics are emitted via TelemetryClient.GetMetric(...).TrackValue(...)).
-        // This dimension is no longer published on custom metrics once migrated to the
-        // OpenTelemetry exporter, so it is stripped when EnableMetricsCustomDimensionOptimization
-        // is enabled to keep behavior consistent across exporters.
         public const string AggregationIntervalMsKey = "_MS.AggregationIntervalMs";
     }
 }
--- a/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Constants/LoggingConstants.cs
+++ b/src/Microsoft.Azure.WebJobs.Logging.ApplicationInsights/Constants/LoggingConstants.cs
@@ -9,5 +9,12 @@ namespace Microsoft.Azure.WebJobs.Logging.ApplicationInsights
         public const string Unknown = "[Unknown]";
         public const string ClientIpKey = "ClientIp";
         public const string HostInstanceIdKey = "HostInstanceId";
+
+        // Property added by the Application Insights SDK metric aggregation pipeline
+        // (e.g. when metrics are emitted via TelemetryClient.GetMetric(...).TrackValue(...)).
+        // This dimension is no longer published on custom metrics once migrated to the
+        // OpenTelemetry exporter, so it is stripped when EnableMetricsCustomDimensionOptimization
+        // is enabled to keep behavior consistent across exporters.
+        public const string AggregationIntervalMsKey = "_MS.AggregationIntervalMs";
     }
 }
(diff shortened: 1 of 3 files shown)


RohitRanjanMS:
Hi @pmalmsten, thanks for the thorough write-up and the strong data behind it — the intent absolutely makes sense. That said, stripping  _MS.AggregationIntervalMs  would be a breaking change for customers who have already opted into this optimization: on upgrade, the dimension would silently disappear from their custom metrics, potentially affecting existing alerts, dashboards, and queries that reference it. A change like this can only ship in a major release, and unfortunately there's no major release planned for the classic Application Insights SDK.
Since the classic SDK is on a deprecation path, our recommendation is to migrate to OpenTelemetry, which already produces the desired output (no  _MS.AggregationIntervalMs  on custom metrics) without needing this change.
Really appreciate the contribution — we just don't have a safe vehicle to take it in the classic SDK.

